PR #12, opened 2020-01-13
CLI help output formatting

lrlna:
Hello! 

I am thinking of changing up the way help output is getting generated. I'd like to input some additional information about each command and have links to docs, if applicable. 

I am thinking shell-api should output an object that then can be formatted by the CLI or the browser packages. Since colouring and spacing will have to be done differently for the browser and terminal, it also doesn't make a lot of sense to do it directly in shell packages. What do y'all think?

here is the object I am thinking of outputting:
_Edit: simplified the arguments object to not have the top level array_
```js
var arguments ={
      help: "Top Level Commands", // what's currently listed under 'help' in yaml file
      docs: "https://docs.mongodb.com/manual/reference/mongo-shell/#command-helpers",
      attr: [
        { "use": "Set current database" },
        { "it": "Result of the last line evaluated; use to further iterate" },
        { ".exit": "Quit the mongo shell" },
        { "db.help()": "Help on db methods" },
        { "db.mycoll.help()": "Help on collection methods" },
        { "sh.help()": "Sharding helpers" },
        { "rs.help()": "Replica set helpers" },
        { "help admin": "Administrative help" }
      ]   
}
```

I've hard coded something for the top-level `help()` to play around with formatting. Here is what ^ that info (+ more) would look like: 
Untitled

@durran @aherlihy what do you think?

### DIFF
--- a/packages/cli-repl/lib/cli-repl.js
+++ b/packages/cli-repl/lib/cli-repl.js
@@ -4,10 +4,8 @@ const { CliServiceProvider } = require('mongosh-service-provider-server');
 const Mapper = require('mongosh-mapper');
 const { compile } = require('mongosh-mapper');
 const ShellApi = require('mongosh-shell-api');
+const ansi = require('ansi-escape-sequences');
 // const _ = require('lodash');
-
-const COLORS = { RED: "31", GREEN: "32", YELLOW: "33", BLUE: "34", MAGENTA: "35" };
-const colorize = (color, s) => `\x1b[${color}m${s}\x1b[0m`;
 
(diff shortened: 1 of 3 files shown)


lrlna:
I think we are pausing on this a little bit until Durran sorts out i18n. Once that gets sorted it'll be easier to think about the `help` formatting!

^ this is more of an FYI for Anna, since we talked about all of this whilst you've been away.

aherlihy:
Thanks Irina! This looks like it's going in a great direction and when it's ready, should be continued.

PR #14, opened 2020-01-14
MONGOSH-53: pass completer function to repl.start

lrlna:
## Description
This PR takes advantage of node repl's built in `completer` param by passing in our own completer function. 

Our completer function then completes appropriate `ShellApi` attributes based on the input. For example:
```shell
$ u # completes with use
$ db.shipwrecks. # provides a range of collection level completions
$ db.shipwrecks.findAndM # completes with db.shipwrecks.findAndModify
$ db.shipwrecks.find({feature_type: "Wrecks - Visible"}). # completes with collection cursor completions
$ db.shipwrecks.aggregate([{$sort: {feature_type: 1}}]). #completes with aggregation cursor completions
```

## Open Questions
1. Is this code typescript-y enough?
2. Something that will need to be solved in the future is what should happen when the query looks like this: 

    ```shell
    $ db.shipwrecks.aggregate([{$sort: {field.name: 1}}]).hasN`
    ```
    Currently the completer function breaks of portions of the user typed query by `.` In this example, then, it will break off at `field.name`, instead of on `.hasN` and will try to suggest aggregation cursor commands too early \o/ This is likely to be solved when implementing query operators with `ace-autocompleter`, but I did want to point out that this isn't working quite yet and we should be keeping it in mind for later additions.

## Up Next
1. Query and Aggregation operators
2. Checking for version numbers
3. Suggesting collection names based on current topology. (Coincidently, should we be suggesting anything other than collection names too?)

Thanks!


packages/cli-repl/package.json
```diff
@@ -22,5 +22,8 @@
     "mongosh-service-provider-server": "file:../service-provider-server",
     "mongosh-shell-api": "file:../shell-api",
     "read": "^1.0.7"
+  },
+  "devDependencies": {
+    "mocha": "^7.0.0"
```
mcasimir: This should not be needed

packages/cli-repl/src/completer.spec.ts
```diff
@@ -0,0 +1,125 @@
+import write from './completer';
```
mcasimir: is `write` as name intended, isn't something like `completer` or `complete` more clear?

packages/cli-repl/src/completer.spec.ts
```diff
@@ -0,0 +1,125 @@
...
+    });
+
+    it('is an exact match to one of shell completions', () => {
+      const i = 'use';
+      expect(write(i)).to.deep.equal([[i], i]);
+    });
+  });
+
+  context('when context is top level db', () => {
+    // this should eventually encompass tests for DATABSE commands and
```
mcasimir: There is a typo in DATABSE and inpput :). Sure you want to leave this comment here, do we have something that our process could capture?

packages/cli-repl/src/completer.spec.ts
```diff
@@ -0,0 +1,125 @@
...
+
+    it('returns all suggestions', () => {
+      const i = 'db.shipwrecks.find({feature_type: "Wrecks - Visible"}).';
+      const collCursorComplete = Object.keys(shellTypes.Cursor.attributes);
+      const adjusted = collCursorComplete.map(c => `${i}${c}`)
+
+      expect(write(i)).to.deep.equal([adjusted, i]);
+    });
+
+    it('does not have a match', () => {
```
mcasimir: I struggled a bit to understand the test case is it more 'returns empty suggestions when it does not have a match'? Maybe i still misunderstood it :).

packages/cli-repl/src/completer.ts
```diff
@@ -0,0 +1,76 @@
+import { types as shellTypes } from 'mongosh-shell-api';
+/**
+ * Return complete suggestions given currently typed line
+ *
+ * @param {any} output - The output.
+ *
+ * @returns {string} The output.
+ */
+function completer(line: string): any {
```
mcasimir: Isn't return type 'string' according to jsdoc?

packages/cli-repl/src/completer.ts
```diff
@@ -0,0 +1,76 @@
...
+  const rsComplete = Object.keys(shellTypes.ReplicaSet.attributes);
+  const shardComplete = Object.keys(shellTypes.Shard.attributes);
+
+  // suggest SHELLAPI commands
+  if (splitLine.length <= 1) {
+    // TODO: this should also explicitly suggest 'sh', 'rs', and 'db' strings
+    const hits = filterComplete(shellComplete, elToComplete);
+    return [hits.length ? hits : [], line];
+  } else if (firstLineEl === 'db' && splitLine.length === 2) {
+    // TODO: @lrlna suggest DATABASE commands (currently not available in
```
mcasimir: Should we create a task for this?

packages/cli-repl/src/completer.ts
```diff
@@ -0,0 +1,76 @@
...
+ * Return complete suggestions given currently typed line
+ *
+ * @param {any} output - The output.
+ *
+ * @returns {string} The output.
+ */
+function completer(line: string): any {
+  // keep initial line param intact to always return in return statement
+  // check for contents of line with:
+  const splitLine = line.split('.');
```
mcasimir: Is line the whole input? I guess for now it really does not matter, and hopefully in general, but do we need to support autocomplete in case of 'var x = db.coll.find'?

packages/cli-repl/src/completer.spec.ts
lrlna: ha! i copied this over from another test and forgot to rename

packages/cli-repl/src/completer.ts
lrlna: i think there is an epic Max created. Shell API only has crud so far, and this is technically not part of crud.

packages/cli-repl/src/completer.ts
lrlna: line would be the whole input, yea! and it should work with the `var x = db.coll.findA`. I've just added a test for this case so we can keep track of it.

packages/cli-repl/package.json
lrlna: I feel like I should be able to run tests from this directory too, no? there are scripts defined above in package.json so I thought they should probably run.

packages/cli-repl/src/completer.spec.ts
lrlna: it's more of "does not return completions if there is no match" can rename it to that <- if it makes more sense.

packages/cli-repl/src/completer.ts
lrlna: i couldn't get the compiler to compile, and `any` worked. I changed the `typescript` definition to `[string[], string]` now as it's more specific.

PR #15, opened 2020-01-14
MONGOSH-58: Browser repl package setup + initial presentation ui

mcasimir:
This PR adds a package for browser repl with setup for Webpack, Karma etc ..

I'm assuming that by merging this PR the package will not be actually released as except for the setup it only contains placeholder code.


.gitignore
```diff
@@ -1,5 +1,10 @@
 .DS_Store
 .lock-wscript
+<<<<<<< HEAD
```
lrlna: rebase conflict

.editorconfig
```diff
@@ -0,0 +1,8 @@
+root = true
```
lrlna: i am wondering if this file should be specific to your own system? since not everyone is using the same editor, this won't be triggered for everyone.

packages/browser-repl/config/webpack.config.prod.js
```diff
@@ -0,0 +1,36 @@
+const path = require('path');
+const WebpackVisualizerPlugin = require('webpack-visualizer-plugin');
```
lrlna: did you want this one in `prod` config too, or just `dev` ?

packages/browser-repl/config/karma.conf.js
```diff
@@ -0,0 +1,35 @@
+process.env.CHROME_BIN = require('puppeteer').executablePath();
```
lrlna: ooooo whatchu thinking of doing with puppeteer? :D

.editorconfig
mcasimir: That is a good question editorconfig is not editor specific. I've seen it added somewhere else in mongo-js and I did the same, not changing much for my editor.

packages/browser-repl/config/webpack.config.prod.js
mcasimir: only prod

packages/browser-repl/config/karma.conf.js
mcasimir: run tests I hope :D

packages/browser-repl/config/webpack.config.prod.js
mcasimir: But should be excluded from the npm packaging

.editorconfig
lrlna: ah ok gotcha

PR #21, opened 2020-01-28
COMPASS-52: use repl.setupHistory given a mongosh_repl_history file

lrlna:
## Description
Adds a repl function to add a history file to save command history. **Does not** include omitting sensitive info just yet -- stay tuned for next PR.

XBGLLndZsv

### DIFF
--- a/packages/cli-repl/src/cli-repl.ts
+++ b/packages/cli-repl/src/cli-repl.ts
@@ -1,16 +1,19 @@
-import repl, { REPLServer } from 'repl';
-import util from 'util';
-import read from 'read';
-import { Transform } from 'stream';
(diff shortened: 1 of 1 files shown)


packages/cli-repl/src/cli-repl.ts
```diff
@@ -1,16 +1,19 @@
...
+import Mapper from 'mongosh-mapper';
 import completer from './completer';
+import { Transform } from 'stream';
+import readline from 'readline';
 import write from './writer';
+import util from 'util';
+import path from 'path';
+import read from 'read';
+import os from 'os';
+import fs from 'fs';
```
lrlna: no _you_ organised imports in descending length order 😅 😅 😅

PR #22, opened 2020-01-28
MONGOSH-72: implement commands

mcasimir:
Allows to run commands like 'use', 'it', 'show' and 'help'.

NOTES:
- the `show` command does not seem to be available in shell api.
- Would be nice to have a more robust parsing for commands.
- The output of the commands is currently rendered as "stringified" string with quotes. We need the object model for commands in order to fix that.


mcasimir:
NOTES:
- the `show` command does not seem to be available in shell api.
- Would be nice to have a more robust parsing for commands.
- The output of the commands is currently rendered as "stringified" string with quotes. We need the object model for commands in order to fix that.

PR #23, opened 2020-01-30
[java-shell] Basic functionality and tests

kornilova-l:
Main class is `MongoShell`.
It has method `eval(String) -> CompletableFuture<MongoShellResult>` that accepts shell command(s) and returns instance of `MongoShellResult` subclass (e.g. `BooleanResult`, `IntResult`, `CursorResult`, `InsertOneResult`)

In this version database operations are performed synchronously even though the API is async (`CompletableFuture<MongoShellResult>`).
I'm not sure if it's possible to execute database operations on different thread because GraalJS `Context` is not thread-safe. So different threads cannot access the same GraalJS objects. This problem will require some research.

Usage example and instruction on how to run tests is in [README.md](https://github.com/kornilova-l/mongosh/blob/java-shell/java-shell/README.md)


lrlna:
just wanted to pop in and say _i am very excited_ about this!!

PR #24, opened 2020-01-30
shell can redact and omit sensitive info, as a treat

lrlna:
## Description
This PR adds a flag to redact information in user's command history using [mongodb-redact](https://github.com/mongodb-js/redact) module. This is based on a flag(`--redactInfo`) when user starts the shell. The default for this flag is `false` to keep it compatible with the current shell. 

This PR also automatically removes user management commands from history (entirely). 

Example of starting shell without `--redactInfo`:
lF30gBfAuY

Example of starting shell with `--redactInfo`:
0jpSixzEN9

## Open Questions
Any other commands I should be omitting? I just copied the ones from [this line in current mongodb shell](https://github.com/mongodb/mongo/blob/master/src/mongo/shell/dbshell.cpp#L308).

Thanks! ✨

### DIFF
--- a/packages/cli-repl/src/history.ts
+++ b/packages/cli-repl/src/history.ts
@@ -11,9 +11,9 @@ import redactInfo from 'mongodb-redact';
  */
 function changeHistory(history: string[], redact: boolean = false) {
   const hiddenCommands =
-    RegExp('createUser|auth|updateUser|changeUserPassword', 'ig');
+    RegExp('createUser|auth|updateUser|changeUserPassword', 'g');
 
-  if (hiddenCommands.test(history[0])) history.shift();
+  if (hiddenCommands.test(history[0])) return history.shift();
   if (redact) history[0] = redactInfo(history[0]);
 }
 
--- a/packages/cli-repl/src/history.ts
+++ b/packages/cli-repl/src/history.ts
@@ -1,0 +1,19 @@
+/**
+ * Modifies command history array based on sensitive information.
+ * If redact option is passed, also redacts sensitive info.
+ *
+ * @param {array} History - Array of commands, where the first command is the
+ * most recent.
+ *
+ * @param {boolean} Redact - Option to redact sensitive info.
+ */
+function changeHistory(history: string[], redact: boolean = false) {
+  const hiddenCommands =
+    RegExp('createUser|auth|updateUser|changeUserPassword', 'ig');
+
+  if (hiddenCommands.test(history[0])) {
+    history.shift();
+  }
+}
+
+export default changeHistory;
--- a/packages/cli-repl/src/history.ts
+++ b/packages/cli-repl/src/history.ts
@@ -1,3 +1,5 @@
+import redactInfo from 'mongodb-redact';
+
 /**
  * Modifies command history array based on sensitive information.
  * If redact option is passed, also redacts sensitive info.
@@ -11,9 +13,8 @@ function changeHistory(history: string[], redact: boolean = false) {
   const hiddenCommands =
     RegExp('createUser|auth|updateUser|changeUserPassword', 'ig');
 
-  if (hiddenCommands.test(history[0])) {
-    history.shift();
-  }
+  if (hiddenCommands.test(history[0])) history.shift();
+  if (redact) history[0] = redactInfo(history[0]);
 }
 
 export default changeHistory;
--- a/packages/cli-repl/package.json
+++ b/packages/cli-repl/package.json
@@ -18,6 +18,7 @@
   "dependencies": {
     "lodash.set": "^4.3.2",
     "minimist": "^1.2.0",
+    "mongodb-redact": "^0.2.0",
     "mongosh-mapper": "file:../mapper",
     "mongosh-service-provider-server": "file:../service-provider-server",
     "mongosh-shell-api": "file:../shell-api",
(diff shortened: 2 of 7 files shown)


packages/cli-repl/src/history.ts
```diff
@@ -0,0 +1,20 @@
...
+ * Modifies command history array based on sensitive information.
+ * If redact option is passed, also redacts sensitive info.
+ *
+ * @param {array} History - Array of commands, where the first command is the
+ * most recent.
+ *
+ * @param {boolean} Redact - Option to redact sensitive info.
+ */
+function changeHistory(history: string[], redact: boolean = false) {
+  const hiddenCommands =
```
mcasimir: Is case insensitive what we want? Also that could be a constant in the outer scope.

packages/cli-repl/src/history.ts
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * Modifies command history array based on sensitive information.
+ * If redact option is passed, also redacts sensitive info.
+ *
+ * @param {array} History - Array of commands, where the first command is the
+ * most recent.
+ *
+ * @param {boolean} Redact - Option to redact sensitive info.
+ */
+function changeHistory(history: string[], redact: boolean = false) {
```
mcasimir: Following up on what we discussed: what about exporting a function that takes a history entry and returns either a new history entry or falsy? ie.

``` js
function filterHistoryEntry(entry) {
  const hiddenCommands = RegExp('createUser|auth|updateUser|changeUserPassword', 'ig');

  if (hiddenCommands.test(entry)) return;
  return redactInfo(entry);
}
```

NOTE: If it is ok for you I can extract it after we merge and i rebase from master.

packages/cli-repl/src/history.ts
lrlna: hmmmm i am not sure. if the command is `Createuser` technically it's not a valid command and will error, but if someone goes in and types their password along with the wrong command I don't want it to be showing up in history either. so idk! what do you think?

packages/cli-repl/src/history.ts
lrlna: i made this function exportable so you can just require it from cli-repl, so there is no rush to extract it if you want to use it right away.

for it to work in node's cli, i have to modify the history object and I think the cleanest way is just modifying then having a new history entry or falsy. This function will return what you passed it in if there are no modifications necessary. Does that make sense? There are examples in the test file of what I mean.

packages/cli-repl/src/history.ts
mcasimir: Don't know :). Not sure if something bad can happen if we get too liberal with that regex and we start to filter out safe input. For me is ok as is, but maybe max would have a better opinion.

packages/cli-repl/src/history.ts
mcasimir: Not sure I understood what you mean there :). 

I would not use that function as is now, I just need the core feature, as that would result just in plain duplicated code.

Why do you think is wrong to extract the function and using it like this:

``` js
import historyFilter from 'whatever shared place';

function changeHistory(history: string[], redact: boolean = false) {
   const filteredEntry = historyFilter(history[0], redact);
   if (!filteredEntry) {
      history.shift();
      return;
   }

   history[0] = filteredEntry;
}
```

packages/cli-repl/src/history.ts
```diff
@@ -0,0 +1,20 @@
...
+  const hiddenCommands =
+    RegExp('createUser|auth|updateUser|changeUserPassword', 'ig');
+
+  if (hiddenCommands.test(history[0])) history.shift();
```
mcasimir: Just found out a small one: you may want to return here. It does not cause direct issues, but in case of redact you are re-redacting the previous entry (if one). May throw if `history[0]` is `null` after shift and `redactInfo` is not happy about it.

lrlna: ah! good point, let me do a check if history actually has a thing.

in node repl this function only get triggered when the first item comes in, so it's always there, but might not be a case in other uses.

packages/cli-repl/src/history.ts
lrlna: hmmmm so for me history is an array of strings, where the first item is the latest. The way I use this function now is quite clean:

```js
      repl.on('flushHistory', function() {
        changeHistory(repl.history, redactInfo);
      })
```
So using your proposed API is a lot less clean for my use case.

I also find it weird if when an API endpoint returns a false _or_ returns a value. Imo APIs should in generally return same structure, or to be more specific the same data type, of a thing, since it makes them more predictable to use.

mcasimir: Ok not holding further on that opinion then, i'll work with that. Although I still disagree (especially on the sort order) :).

packages/cli-repl/src/history.ts
mcasimir: The correct behaviour is to return, otherwise the previous element may be redacted twice (not sure that is a problem), maybe we need a specific test case for that if you don't have one.

lrlna: check the latest release! it's just returning ^__^

packages/cli-repl/src/history.ts
lrlna: yea maybe let's talk about history a bit more irl next week! Maybe I am completely misunderstanding this \o/

I think the func is so very smol, we can change it anytime if it turns out to be a bad idea™

PR #25, opened 2020-01-31
Add i18n module

durran:


packages/i18n/src/translator.ts
```diff
@@ -0,0 +1,104 @@
...
+   * @param {string} key - The key.
+   *
+   * @returns {string} The translated string.
+   */
+  translateApiHelp(key: string): string {
+    const value = this.find(key);
+    if (typeof value === 'string') {
+      return value;
+    }
+    return Mustache.render(TEMPLATE, value);
```
lrlna: I am wondering if we should let the implementations handle the formatting that this `TEMPLATE` is doing. Instead have `Translator` return just the translation object.

I am thinking we will run into trouble, for example, when we want to pad things in the browser vs. the shell, or add new extra new lines, or change colour of items.

PR #28, opened 2020-02-05
RFC: API for mongosh logger

lrlna:
## Description
This pr is an RFC on how I am thinking about implementing the logger from an API perspective. This PR is better viewed under the [View File](https://github.com/mongodb-js/mongosh/blob/6d608c3352b222543a9f23f5e680c1856952d199/packages/logger/README.md) option in the diff.

The tl;dr: should live on the Mapper level, but should be it's own package that could be used without having to be hooked into the mapper/shell api.

I am thinking from an implementation perspective, I will mirror it to pino.js, but the API endpoints will handle both logging with pino to file, and logging to segment.

## Open Questions
Something we talked about previously in regards to logging is that folks should be able to pass in a flag when using the CLI to send or not to send information to segment. This begs the question: where should we store this file, and how should we be accessing information from it. Are we looking into something like `.mongoshrc`?

_Edit: sorry for the old school javascript examples! Will fix once we agree on the design_


packages/logger/README.md
```diff
@@ -0,0 +1,99 @@
...
+```js
+var CliServiceProvider = require('mongosh-service-provider-server').CliServiceProvider
+var ShellApi = require('mongosh-shell-api')
+var Mapper = require('mongosh-mapper')
+
+var loggerPath = path.join(os.homedir(), '.mongosh_log')
+
+CliServiceProvider.connect(driverUri, driverOptions)
+  .then(function (serviceProvider) {
+    var mapper = new Mapper(serviceProvider, loggerPath)
```
mcasimir: What if mapper would take a logger instance instead?

lrlna: that's also a possibility!

Do you feel like that'd be too many configurable things to do on the level of browser/Compass/CLI if now we have to configure the logger too? That was my biggest dilemma when thinking this through

lrlna: I guess I was mainly thinking, if someone was doing their own implementation of the whole thing, they might forget to setup a logger entirely and so no logging will ever happen. If that's something we want, then passing a Logger instance is a really good approach.

packages/logger/README.md
```diff
@@ -0,0 +1,99 @@
...
+    var shellApi = new ShellApi(this.mapper)
+  })
+```
+
+## API
+### `new Logger([destination])`
```
mcasimir: If we are going to enforce the constructor interface i'd prefer `new Logger(fileTransport)`, `new Logger(consoleTransport)`. 

We may also think to avoid that and just establish a contract between mapper and logger, then we can just instantiate `pino` or anything else that is appropriate, without over-constraining.

lrlna: i don't want to enforce constructor interface. These should work as well:

```js
var logger = require('@mongosh/logger')()
```
or:
```js
var logger = require('@mongosh/logger')(/path/to/dest/)
```
or:
```js
var logger = require('@mongosh/logger)({browser: {toObject: true}})
```
or:
```js
var logger = require('@mongosh/logger')
var log = logger(/path/to/file)
```
I'd prefer to not have to deal with any sort of transport layers, but instead provide defaults. `pino` handles both browser and node environments so we can piggy back on that.

What do you mean by "establish a contract"?

packages/logger/README.md
lrlna: I guess if we also let the implementer pass in the logger instance they can also decide on what logger to use, as long as it follows the `log.info`/ `log.error` pattern (most JS ones do). I guess it all depends on do we want to give implementers the choice?

packages/logger/README.md
```diff
@@ -0,0 +1,99 @@
...
+Create a new Logger instance. Provide destination path to file if you want to save
+information locally. Otherwise data will only be sent to
```
mcasimir: Would never expect the logger to report info to DWH. 

I like that the telemetry / reporting implementation is passed as a dependency like the logger is. But by any mean i would segregate the two interfaces.

```
interface TelemetryReporter {
   reportEvent(type: string, payload: any): Promise<void>
}

interface Logger {
   info(...args: any): void;
   error(...args: any): void;
   warn(...args: any): void;
}

const segmentTelemetryReporter = new SegmentTelemetryReporter(segmentApiKey);
const fileLogger = require('pino')( pino.destination('/log/path'));

const mapper = new Mapper(shellApi, {logger: fileLogger, telemetryReporter: segmentTelemetryReporter});
```


```

Telemetry.reportEvent

lrlna: This RFC is for a logger that is a wrapper around both a logger and telemetry. Otherwise I'd straight up use `pino`. 

I don't think I understand this sentence:
> But by any mean i would segregate the two interfaces.

lrlna: Maybe a drawing will help on what this RFC is trying to encapsulate:

0B1F9CE1-E965-425E-BF81-2647449A40A3

mcasimir: Sorry for the confusion, I was referring to the interface segregation in the SOLID principles.

packages/logger/README.md
mcasimir: I prefer that form of dependency inversion as it makes it more testable (we can just do `new Mapper(serviceProviderMock, loggerMock))`.

It will take the creational responsibility out of the Mapper layer (the mapper won't need to know how to create a logger) allowing for independent evolution of the two implementations (as long as the interface is honored)

packages/logger/README.md
mcasimir: Here I was still thinking we were trying to do only a Logger. In which case, as you mentioned somewhere else, I had the evil idea of just use `pino` and add  an interface (as in typescript Interface) in the mapper.

packages/logger/README.md
lrlna: yea i think what you're saying makes sense. let me have a think about it some more!

lrlna:
I have realised that there is one thing I wanted to do in the future, which is to get some Performance API statistics to a local log file. And I don't think it'd work that well with what I have proposed here, since the Mapper will then have a responsibility of starting a performance observer and ending it, which is probably not the responsibility it should be caring for entirely. 

tl;dr I think what I proposed doesn't necessarily work, and I'll try to think of a different approach.

PR #29, opened 2020-02-06
MONGOSH-78: Extract presentation responsibility from shell-api / mapper

mcasimir:
This is a proposal to take away presentational concerns from the logic layer (shell-api / mapper) and move it to the layer that deals with the presentation itself.

## Motivation

Browser and cli applies different representation to the output.
  
# Description

The presentation of evaluated results coming from shell api happens now in two places:

- In the `cli-repl` the `write` function (renamed to `formatOutput` to highlight that that's where the presentation occurs).
- In the `browser-repl` in the `shell-output-line` component.

That is done by creating a plain object model for the shell api (each shell-api object type exposes a method to get the type itself). The `toReplString` (name that does not fit much here) returns a plain (materialised) object representation of the api object (that is necessary for things like Cursors which wont  hold the result set otherwise).

**Note**: this is mostly necessary to avoid using `instanceOf` which fails to work with `Proxy` and potential bridges (ie. iframe / java).

The object model can be used to tell which api object has to be presented and apply the proper format to it.

### Some examples:

In the cli a list of document can be presented as the output of `util.inspect` whereas in the browser a table or a more interactive presentation can be used.

## Alternatives

The solution in this PR is the most flexible (it allows alternate presentation for everything, should work in every context), but we may not need it. Some simplifications could be:

1. `toReplString` always returns `Promise<string>` (now occasionally return arrays), and is going to be the final representation (no further manipulation of the output should be needed it such case).

In the browser we could use something like `ansi-to-html` to deal with coloring / color codes and some url parser for links. No interaction with documents.

2. Like n. 1 but we hack something to recognise documents and we have a special presentation for those only.

I'm not aware about the status of the `idea` integration but we may run this by them too.

## Changes

- Adds a `shellApiType` function to the return types of shell api.
- Help is now an object and not a string anymore
- The result of `it` is now of the type `CursorIterationResult`
- Adds an example of usage in the browser-repl
- Adds an example of usage in the browser-cli

## Notes

- maybe `shellApiType` can be a property directly rather than a function.
- The documents returned by `findOne` are still plain objects (may or may be not an issue)
- @lrlna the help implementation is based on your structure `{help, docs, attr}` hope that will work with your PR.
- the implementation should be polished: probably we could merge `toReplString` and `shellApiType` in one method `toReplOutput` or `toReplValue` returning: `{type, value, pretty}`.

Screenshot 2020-02-06 at 11 49 18


packages/shell-api/bin/compile-shell-api.js
```diff
@@ -129,11 +136,12 @@ const loadAll = () => {
   const mainLib = yaml.load(main);
   const FILES = fs.readdirSync(yamlDir).filter((s) => (/[A-Z]/.test( s[0])));
   let exports = '\nexport default ShellApi;\n';
-  let types = [];
-  let typeConsts = [];
+  const types = [];
+  const typeConsts = [];
```
lrlna: not a review comment but a legit question: does typescript not complain that this is a const and you push into it? o.O o.O o.O

mcasimir: For `let/const` the behaviour is like javascript: does not matter for `push` as we are not reassigning the variable. 

It would instead complain if the type is readonly:

```
const types: readonly string[] = []; // or Readonly<Array<string>>
types.push('something');
```

packages/shell-api/src/help.js
```diff
@@ -0,0 +1,28 @@
+import i18n from 'mongosh-i18n';
+
+// TODO: This probably should be in i18n itself
```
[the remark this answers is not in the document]
mcasimir: Yep that would be better

packages/cli-repl/src/cli-repl.ts
```diff
@@ -62,16 +62,43 @@ class CliRepl {
...
+    const evaluationResult = await this.evaluate(originalEval, input, context, filename);
+
+    if (this.isShellApiType(evaluationResult)) {
+      return formatOutput(
+        await evaluationResult.toReplString(),
+        evaluationResult.shellApiType()
+      );
+    }
+
+    return formatOutput(evaluationResult);
```
lrlna: i think the most proper way of doing formatting in the cli is through repl's `writer` function. It's one the [options](https://nodejs.org/api/repl.html#repl_repl_start_options) we pass in to repl.start.

My reasoning there is that I want to use as much of the formatting as node repl provides us and as much of the standard API as possible.

packages/cli-repl/src/format-output.ts
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * Return the pretty string for the output.
+ *
+ * @param {any} value - The output value.
+ * @param {string} type - The shell api type if known or undefined.
+ *
+ * @returns {string} The output.
+ */
+export default function formatOutput(output: any, type?: string): string {
+  if (type === 'Cursor') {
```
lrlna: I like the idea of doing this style of comparison! I think if we ever want to do anything interactive, it'll be more or less straightforward this way.

packages/cli-repl/src/cli-repl.ts
```diff
@@ -62,16 +62,43 @@ class CliRepl {
       this.connect(driverUri, driverOptions);
     }
   }
+  /**
+   * Returns true if a value is a shell api type
+   *
+   * @param {any} evaluationResult - The result of evaluation
+   */
+  private isShellApiType(evaluationResult: any): boolean {
```
lrlna: question about the shell type: does it also give types to errors?

packages/shell-api/bin/compile-shell-api.js
```diff
@@ -28,8 +29,7 @@ const attrTemplate = (attrName, lib, base = '') => {
   const lhs = `    this${base}.${attrName}`;
 
   if (attrName === 'help') {
-    return `${lhs} = () => (i18n.translateApiHelp('${attr}'));
-${lhs}.toReplString = () => (i18n.translateApiHelp('${attr}'));`
+    return `${lhs} = () => new Help(${JSON.stringify(attr)});`;
```
lrlna: this is good!

packages/cli-repl/src/cli-repl.ts
mcasimir: I think you're right. Don't know why i ended up keeping the one in evaluate instead of putting it there.

packages/cli-repl/src/cli-repl.ts
mcasimir: Is a good question, for now it does not, but i have a feeling it will get more clear what does that job as soon as we figure a way to deal with evaluation of code in a unified way. We should give that a stab next.

Also even in the PR that's already duplicated code. Definitely something to look into.

lrlna: Definitely something we can start looking into later. I haven't started on error formatting yet anyways.

PR #30, opened 2020-02-06
MONGOSH-53: add topology-awareness to cli's completer

lrlna:
## Description
Adds a version filtering to completer's completer function. Currently used
mdbVersion is the now the first param in that function.

Filtering happens for both minimum and maximum semver version provided by
shell-types.js file in shell-api.

`shell-api` was adjusted to have valid semver versions for method following the
`major.minor.patch` format. They have also been stringified, since `4.4.0` is
not a valid number format.

## Open Question
I am doing an incredibly hacky version of extracting the current system version
of mongod. Any ideas on getting it another way?

Still looking into a way of comparing a version that's a local developer version that doesn't have a specific semver attached to it.


packages/cli-repl/src/cli-repl.ts
```diff
@@ -45,6 +46,12 @@ class CliRepl {
       this.serviceProvider = serviceProvider;
       this.mapper = new Mapper(this.serviceProvider);
       this.shellApi = new ShellApi(this.mapper);
+      // .toString() output since child_process execution returns a Buffer
+      // grab version info from the first line from output
+      this.mdbVersion = execSync('mongod --version')
+        .toString()
+        .split('\n')[0]
+        .replace('db version ', '');
```
lrlna: this is hyper hacky ^ 
i am thinking there maybe should be a way for `shell-api` to provide us with a current mongod version.

mcasimir: Technically we have to implement it anyway to do: `db.version()`

lrlna: ah good! this hack can then quickly be rewritten once we get that. 

i think with the node driver, we just need to get [serverInfo](https://mongodb.github.io/node-mongodb-native/3.5/api/Admin.html#serverInfo) and that will return all the info we will need

mcasimir: Actually, didn't @durran implement the version command already? (the one you are calling).

lrlna: hmmmm not that i know! there is [this version](https://github.com/mongodb-js/mongosh/blob/master/packages/cli-repl/bin/mongosh.js#L12), but this looks at the version of `mongosh` package, rather than `mongod`

mcasimir: Ohh `mongo version` returns the "shell version". Ok never mind. I think we can switch the hack to an hardcoded string.

lrlna: > I think we can switch the hack to an hardcoded string.

I don't understand what you mean 😅 `mongod version` returns the DB version for me:
Capture d’écran, le 2020-02-06 à 16 03 39

PR #31, opened 2020-02-12
Fix java-shell after recent changed to js codebase

kornilova203:
`Cursor` member functions are now invoked using `apply` e.g. `return (_a = this._cursor).hasNext.apply(_a, __spread(arguments));`

This did not work with first implementation of `com.mongodb.mongosh.service.Cursor` class because Java functions annotated as `@HostAccess.Export` don't have `apply` method even though they can be invoked from js code.

To fix the problem I create a "real" js functions that wraps calls to Java code (see [Cursor.functionProducer](https://github.com/mongodb-js/mongosh/pull/31/files#diff-ca51e378bcf32fca779df8a4df6f42c8R15))


durran:
@kornilova203 I apologise I missed this! Can you rebase against master and then I'll pull in. Thank you so much!

kornilova203:
@durran,
Yes, I'll rebase and continue with the development
No problem :)

kornilova203:
Hi @durran,
I rebased onto master

Though I'm not sure if I should keep devDependencies and scripts in my `package.json` or in global one

To run tests:
1. compile mongosh project `npm run bootstrap` etc
2. set mongodb uri to packages/java-shell/src/test/resources/URI.txt
3. ```
   cd java-shell
   ./gradlew test
    ```

PR #33, opened 2020-02-13
MONGOSH-79: shell-browser: topology aware autocomplete

mcasimir:
Features:
- Implements "topology aware" autocompletion
- add `getServerVersion` to the service provider layer

Chores:
- `npm run start:compass` now runs the shell embedded in compass master
- add `eslint-plugin-mocha` to fail in case focused tests are committed (ie. `it.only(...`)

NOTES:

@lrlna I'm currently passing only the last word to the cli-repl autocomplete, is that what it expects? or would it be better to pass the whole text instead?  Would it work with newline characters?

Feb-14-2020 17-14-15


packages/browser-repl/src/lib/autocompleter/autocompleter.ts
```diff
@@ -0,0 +1,24 @@
+/**
```
lrlna: It's a bit weird for me to have a `./lib` directory within a `./src`, since lib is usually where our compiled modules go. Maybe adding this to `./modules` might work better?

packages/browser-repl/src/lib/autocompleter/shell-api-autocompleter.ts
```diff
@@ -0,0 +1,30 @@
...
+import { Autocompleter, Completion } from './autocompleter';
+
+export class ShellApiAutocompleter implements Autocompleter {
+  private serverVersion: string;
+
+  constructor(serverVersion) {
+    this.serverVersion = serverVersion;
+  }
+
+  async getCompletions(code: string): Promise<Completion[]> {
```
lrlna: I am not sure I understand why this needs to be async just yet? `cliReplCompleter` does not return anything async just yet, and might not do that for a quite a bit still.

packages/browser-repl/src/components/ace-autocompleter-adapter.ts
```diff
@@ -0,0 +1,47 @@
+import { Autocompleter, Completion } from '../lib/autocompleter/autocompleter';
```
lrlna: there is an [ace-autocomplete](https://github.com/mongodb-js/ace-autocompleter) which may or may not be helpful here to get your line positions and other ace-y things

packages/browser-repl/src/lib/runtime-helpers/open-context-runtime.ts
```diff
@@ -0,0 +1,65 @@
...
+
+    this.setupEvaluationContext(
+      interpreterEnvironment.getContextObject(),
+      serviceProvider
+    );
+
+    this.interpreter = new Interpreter(interpreterEnvironment);
+  }
+
+  async getCompletions(code: string): Promise<Completion[]> {
```
lrlna: I don't think I quite understand why the completion function has to be in this open context runtime as well? is it because you get the server version first and that can only happen from the service provider?

what if the server version gets added to the `ns`? or just even the redux state so you don't have to set up completions in `runtime`?

packages/browser-repl/src/lib/autocompleter/autocompleter.ts
mcasimir: You mean rename `lib`? Sure, i think another option would be to drop `lib` and put everything as peer of components.

lrlna: > i think another option would be to drop lib and put everything as siblings of components.

that's also a good idea!

packages/browser-repl/src/lib/runtime-helpers/open-context-runtime.ts
mcasimir: The whole idea of the `Runtime` is for the top level/react-specific stuff to not have any direct dependency on shell-api / mapper and service provider, since that would not work with WebWorkers or more isolated environments.

I agree 100% we should do differently otherwise. I hope we can sort out If we need that or not, we may drop the runtime object entirely in that case.

packages/browser-repl/src/lib/autocompleter/shell-api-autocompleter.ts
mcasimir: We can make it sync, i stared with one autocompleter / without the adapter and ACE had the callback for some reason

PR #34, opened 2020-02-17
MONGOSH-51: add usage menu to mongosh --help

lrlna:
## Description
Uses i18n module to localise help menu for `mongosh --help`. Additionally, outputs help menu when an invalid option is provided to instantiate mongosh. For example:

Y39WwcWpLs

Available options were just copied over from `mongo --help`, so we can adjust them in the future to what we actually support.

### Help Menu Structure
I'd like to keep this structure with all the cli help commands for consistency: 

- usage: **bold**
- section headings: standard yellow colour, **bold**, and separated by new lines above and below for clarity
- examples: description in plain text, actual example in standard green. when more than one example is provided, separate each section by new line. (right now there is only one example, but we should add a few more in the future)
- docs information: **bold**
- docs link: green

If user's terminal is not TTY, we will keep text as is without colours/bolds/underlines/etc.


durran:
LGTM

PR #35, opened 2020-02-18
Add show databases

aherlihy:
Going to go through the `show` command first. This is just `show dbs` and `show databases` but I wanted to make sure there wasn't anything else I should be updating for the whole path.


lrlna:
it looks good to me, anna!

packages/mapper/lib/mapper.js
```diff
@@ -19,6 +19,7 @@ class Mapper {
     this.awaitLoc = []; // track locations where await is needed
     this.checkAwait = false;
     this.cursorAssigned = false;
+    this.databases = { test: new Database(this, 'test') };
```
mcasimir: Do we have to memoize the database object?

packages/mapper/lib/mapper.js
```diff
@@ -48,15 +49,33 @@ class Mapper {
...
+  async show(_, arg) {
+    switch (arg) {
+      case 'databases':
+      case 'dbs':
+        const result = await this._serviceProvider.listDatabases('admin');
+        if (!('databases' in result)) {
+          throw new Error('Error: invalid result from listDatabases');
+        }
+        return result.databases.reduce((str, db) => {
+          return `${str}\n${db.name}\t${db.sizeOnDisk}B`;
```
mmarcon: Minor thing: `mongo` uses GB instead of B

packages/mapper/lib/mapper.js
aherlihy: I'm not sure what specifically you mean by memoize. Previously we weren't saving db objects, but now we are so that when you switch back to a db the collection objects that get created through proxy usage stick around for autocomplete/etc.

packages/mapper/lib/mapper.js
aherlihy: Yeah I noticed that, but I'm not convinced its better. The same `show dbs` command on the old shell just returns 0GB, 0GB, 0GB which i think is less informative than B if they're small. Should we update it to do what `ls` does, i.e. "human readable format" (https://web.archive.org/web/20120507054320/http://codeaid.net/javascript/convert-size-in-bytes-to-human-readable-format-(javascript))

mmarcon: I was thinking the same. One downside is the results won't be nicely aligned as they are in `mongo`:

```
> show dbs
100YWeatherSmall   0.108GB
admin              0.000GB
agenda             0.000GB
agendaDb           0.000GB
city               0.002GB
citybike           0.005GB
compass           12.454GB
config             0.000GB
debugging          0.000GB
foo                0.000GB
graphql_guide      0.000GB
gridfs             0.009GB
las_vegas          0.000GB
local              0.000GB
metrics            0.000GB
ships              0.001GB
test               3.084GB
video              0.256GB
```
`mongo` uses decimals, maybe that's good enough?

aherlihy: I think they'll be aligned if we use the tab character right? As long as there aren't any exceedingly long db names. We can do some math to make sure the size is rounded to 4-5 digits?

packages/mapper/lib/mapper.js
mcasimir: Yep i meant that (store it so we don't have to refetch later). Makes sense!

PR #36, opened 2020-02-18
MONGOSH-81: shell-compass: persisted history

mcasimir:
Saves history between sessions


packages/compass-shell/src/components/compass-shell/compass-shell.jsx
```diff
@@ -7,29 +7,74 @@ import { connect } from 'react-redux';
...
+
+    try {
+      await this.props.historyStorage.save(history);
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.error(error);
+    }
+  }
+
+  loadHistory = async() => {
```
lrlna: what if you load up history when this plugin is loaded? (`onActivated`) that way your props always have history and then you add and remove it as you go.

lrlna: ^ reason I am saying this is it might be nice to have this logic out of a view component and in a redux state management part of the plugin.

mcasimir: That's good too, had a chat with rhys and decided to avoid redux as there is nothing else interested in that state, but i'm open to changing it.

Where the save history goes in that case? We could also set two properties: `initialHistory` and `saveHistory`.

lrlna: yea using redux state will definitely require you to change a few things so up to you!

remind me of the difference between saveHistory and initialHistory?

mcasimir: `saveHistory` would be the function called to save the history when it changes `initialHistory` would be the array of history entries you suggested

mcasimir: I think i'll merge this as is, and revisit it now that we integrate the shell properly, as i believe that is going to involve more redux and it will make sense to have it consistent

PR #37, opened 2020-02-19
fix multiline input

lrlna:
## Description
Callback on `this.evaluate` was catching and printing recoverable errors that happen when you type in multiline input. This PR removes the catch, and lets original node eval deal with multiline, and just pass on the result over to `writer` and then `formatOutput`.


mcasimir:
Looks good

PR #39, opened 2020-02-27
MONGOSH-80: shell-browser: styling

mcasimir:
- Add styles as per: https://mongodb.invisionapp.com/share/8ZVWIV1SYWH#/screens/404221930
- Fixes some failing tests
- Use i18n in the browser repl too

NOTE: object inspection is only a placeholder, will be implemented later.

Feb-27-2020 14-45-52


packages/browser-repl/src/components/types/help-output.tsx
```diff
@@ -51,8 +57,9 @@ export class HelpOutput extends Component<HelpOutputProps> {
   render(): JSX.Element {
     const help = this.props.value;
 
+    const className = classnames(styles['help-output']);
```
lrlna: i can't figure out what exactly in your css adds the extra line break between various help items when you do `db.coll1.help()`, but imo, it be nicer if they are a bit more squished together without that extra `\n`

packages/browser-repl/src/components/utils/inspect.ts
```diff
@@ -0,0 +1,61 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
+const utilInspect = require('util').inspect;
+const utilInspectCustom = utilInspect.custom || 'inspect';
+
+const formatBsonType = (value): any => ({
+  [utilInspectCustom](): string {
+    return `${value._bsontype}(${(JSON.stringify(value))})`;
```
lrlna: something to consider here for formatting bson is `[EJSON.stringify](https://github.com/mongodb/js-bson#ejsonstringifyvalue-replacer-space-options)` function. If `JSON.stringify()` is satisfactory for you, then ignore!

packages/browser-repl/src/components/utils/inspect.ts
```diff
@@ -0,0 +1,61 @@
...
+function isBsonType(value): any {
+  return value && value._bsontype;
+}
+
+function isObject(value): any {
+  return value && typeof value === 'object' && !Array.isArray(value);
+}
+
+function formatProperty(value): any {
+  if (isObject(value) && isBsonType(value)) {
```
lrlna: so i think our [js-bson](https://github.com/mongodb/js-bson) package will format the value if it's a bson type, and otherwise leave it as is. and if you end up using it, you might not need this check.

packages/browser-repl/src/components/utils/inspect.ts
```diff
@@ -0,0 +1,61 @@
...
+function formatObject(object): any {
+  const viewObject = {};
+  for (const key of Object.keys(object)) {
+    viewObject[key] = formatProperty(object[key]);
+  }
+  return viewObject;
+}
+
+function toViewValue(value): any {
+  if (isBsonType(value)) {
```
lrlna: similar comment to the above: js-bson package _might_ be helpful here to not have these extra checks.

packages/browser-repl/src/components/ace-theme.js
```diff
@@ -0,0 +1,130 @@
...
+
+const keywordsColor = '#FF7DC3';
+const stringsColor = '#35DE7B';
+const literalsColor = '#2DC4FF';
+const supportColor = '#a5e3ff';
+const classesColor = '#EDB210';
+const commentsColor = uiColors.gray.light1;
+const variablesColor = '#FF6F44';
+
+const layoutCss = `
```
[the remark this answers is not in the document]
mcasimir: That would be just amazing, I did not try to style it with just css, i only copy-pasted and changed the code from `mongo-ace-theme`. I'll give that a try.

packages/browser-repl/src/components/ace-theme.js
```diff
@@ -0,0 +1,130 @@
+import {uiColors} from '@leafygreen-ui/palette';
```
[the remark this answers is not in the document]
mcasimir: I agree, possibly i would also add it to the eslint, do you know if that would annoy someone?

packages/browser-repl/src/components/shell-output-line.tsx
```diff
@@ -23,68 +32,80 @@ export class ShellOutputLine extends Component<ShellOutputLineProps> {
...
     }
 
-    return inspected;
+    return <ObjectOutput value={value} />;
   }
 
-  renderValue(): JSX.Element {
-    const {shellApiType} = this.props.entry;
+  private isError(value: any): boolean {
+    return typeof value.message === 'string' && typeof value.stack === 'string';
```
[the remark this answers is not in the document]
mcasimir: That case should have been dealt before in the caller. There is a test that would catch it in case we move that check around.

packages/browser-repl/src/components/types/help-output.tsx
mcasimir: I think you'll have to show me what you mean :D

packages/browser-repl/src/components/utils/inspect.ts
mcasimir: That should be a temporary solution, we should revisit that after we implement the supplementary types for the shell: https://docs.mongodb.com/manual/core/shell-types/

PR #50, opened 2020-03-06
feat(shell-api): allow fluent methods

mcasimir:
Implements: `db.coll1.skip(..).limit(..)`.

Allows to mark methods as fluent in shell api `yaml` by adding `__fluent: true` to the method spec.

In order to support that change also moves methods definitions outside the constructor and method attributes (ie. `find.help`) as prototype properties (would not compile otherwise).

Also reverts mongodb driver version to avoid 'cursor closed' issue.


packages/mapper/src/mapper.integration.spec.ts
```diff
@@ -0,0 +1,86 @@
+import { expect } from 'chai';
+import { CliServiceProvider } from 'mongosh-service-provider-server';
```
lrlna: ooooooohhh having integration tests is very nice!

PR #52, opened 2020-03-09
MONGOSH-105: add autocompletions for find and aggregation queries

lrlna:
## Description
Preloads autocompletions from `[ace-autocompleter](https://github.com/mongodb-js/ace-autocompleter)` for all collection and aggergation queries. 

## Open Questions
I am feeling this module is getting a bit unruly, and getting quite hacky with all the `.pop()`s, `.split()`s and `.trim()`s.

I think in the future (maybe after mongodb world), we can consider writing a parser to have a PT with a bit of context to give the necessary information for more accurate suggestions. This would also allow for queries that look like this, that are currently not possible to autocomplete:

```js
var cat = "Nori"; db.cats.aggregate([ {$mat
```

### DIFF
--- a/packages/cli-repl/src/completer.ts
+++ b/packages/cli-repl/src/completer.ts
@@ -78,20 +78,25 @@ function completer(mdbVersion: string, line: string): [string[], string] {
         mdbVersion, COLL_CURSOR_COMPLETIONS, elToComplete, splitLine);
       return [hits.length ? hits : [], line];
     }
-    // complete aggregation queries/stages
-    if (splitLine[2].includes('aggregate') && splitLine[2].includes('([')) {
-      const expressions = BASE_COMPLETIONS.concat(accumulators(
-        elToComplete, mdbVersion));
+
+    // complete aggregation and collection  queries/stages
+    if (splitLine[2].includes('([') || splitLine[2].includes('({')) {
+      let expressions;
+      if (splitLine[2].includes('aggregate')) {
+        // aggregation needs extra accumulators to autocomplete properly
+        expressions = BASE_COMPLETIONS.concat(accumulators(
+          elToComplete, mdbVersion));
+      } else {
+        // collection quering just needs MATCH COMPLETIONS
+        expressions = MATCH_COMPLETIONS;
+      }
       // split on {, as a stage/query will always follow an open curly brace
-      const preAggQuery = line.split("{");
-      const prefix = preAggQuery.pop().trim();
+      const splitQuery = line.split("{");
+      const prefix = splitQuery.pop().trim();
       const command = line.split(prefix).shift();
       const hits = filterQueries(mdbVersion, expressions, prefix, command);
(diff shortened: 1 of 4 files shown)


packages/cli-repl/src/completer.ts
```diff
@@ -34,49 +67,89 @@ function completer(mdbVersion: string, line: string): [string[], string] {
...
+				);
+			});
+		} else if (stage.includes(GROUP)) {
+			return ACCUMULATORS;
+		}
+	}
+	return [];
+}
+
+function filterQueries(mdbVersion: string, completions: any, prefix: string, split: string) {
```
lrlna: i have this in another function entirely, since I felt that putting `filterShellAPI` and `filterQueries` together in one although created less duplicate code, was also a bit less readable.

packages/cli-repl/src/completer.ts
```diff
@@ -34,49 +67,89 @@ function completer(mdbVersion: string, line: string): [string[], string] {
...
 }
 
-// TODO: @lrlna this should also take this.shellApi.version of sorts and also
-// match on version:
-// semver.gte(version, .version)
-function filterComplete(mdbVersion: string, completions: object, toMatchTo: string, split?:
-                        string[]) {
+// stage completions based on current stage string.
+function accumulators(stage: string, mdbVersion: string) {
+  if (stage !== '') {
```
mcasimir: Some times when ifs starts to nest I find the early-return pattern useful to flatten it back, this is from a random google search: https://www.freecodecamp.org/forum/t/the-return-early-pattern-explained-with-javascript-examples/19364.

Absolutely fine as is, just want to share with you.

packages/cli-repl/src/completer.ts
mcasimir: I like it

packages/cli-repl/src/completer.ts
```diff
@@ -34,49 +67,89 @@ function completer(mdbVersion: string, line: string): [string[], string] {
...
   return [[line], line];
 }
 
-// TODO: @lrlna this should also take this.shellApi.version of sorts and also
-// match on version:
-// semver.gte(version, .version)
-function filterComplete(mdbVersion: string, completions: object, toMatchTo: string, split?:
-                        string[]) {
+// stage completions based on current stage string.
+function accumulators(stage: string, mdbVersion: string) {
```
mcasimir: what about a name that is more inline with the comment? like `getStageCompletions` or `stageCompletionsAccumulators`.

packages/cli-repl/src/completer.ts
lrlna: ah that's a good catch!

lrlna:
@mcasimir would you happen to know why the AceAutocompleterAdapter test fails on windows? Does it fail for you locally too? (I, once again :((((, can't run anything browser related locally)

PR #53, opened 2020-03-09
Async Rewriter

aherlihy:
The async rewriter is ready for testing integrating with browser (it can already be run from the CLI) with `npm run start-async`. Most common use cases should be covered, but there is still work to do with classes and scope.

Since this is behind a feature flag and contains only a few lines of changes in the code outside the new package, I will like merge it ASAP so that I don't have to keep rebasing.


packages/async-rewriter/lib/antlr/ECMAScriptLexer.d.ts
```diff
@@ -0,0 +1,141 @@
+export var ECMAScriptLexer: typeof ECMAScriptLexer;
```
mcasimir: you may want to add `lib/` to `.gitignore`

packages/cli-repl/src/cli-repl.ts
```diff
@@ -195,20 +196,8 @@ class CliRepl {
...
-
-          // Pass the locations to a parser so that it can add 'await' if any function calls contain 'await' locations
-          const syncStr = compile(input, this.mapper.awaitLoc);
-          if (syncStr.trim() !== input.trim()) {
-            console.log(`DEBUG: rewrote input "${input.trim()}" to "${syncStr.trim()}"`);
-          }
-
-          // Eval the rewritten string, this time for real
-          this.mapper.checkAwait = false;
+          const syncStr = compile(input, types, this.mapper.scope);
```
mcasimir: Nice!

PR #54, opened 2020-03-10
Align service provider interface with implementations

mcasimir:
Some preliminary work for db and collection commands:

- Fixes some typings issues between `service-provider-core` and `service-provider-server`.
- Extract duplicated tests to functions in `service-provider-server` so we don't have to maintain duplicated files for `CompassServiceProvider` and `CliServiceProvider`.


durran:
LGTM

PR #58, opened 2020-03-10
feat(service-provider): add dropDatabase

mcasimir:
- Add `dropDatabase` to `ServiceProvider`
- Add eslint to `service-provider-server`


packages/service-provider-server/src/cli-service-provider.ts
```diff
@@ -576,6 +555,33 @@ class CliServiceProvider implements ServiceProvider {
...
+    writeConcern?: WriteConcernDoc
+  ): Promise<Result> {
+    // Defaults are based on old shell implementation.
+    // See: https://github.com/mongodb/mongo/blob/d2b75b4e2a6d1e9db7cbb6120c34b0b44476828e/src/mongo/shell/db.js#L7
+    const defaultWriteConcern = {
+      w: 'majority',
+      wtimeout: 10 * 60 * 1000
+    };
+
+    return await this.nodeTransport.runCommand(db, {
```
durran: Instead of running the `dropDatabase` command directly, I would call `db.dropDatabase` on the Node driver directly and add that to the Node transport. This is to protect us from driver specification changes that would cause the commands being built to change in some manner. If the spec changes and we're calling the driver's helper method, we get the changes for free - this should apply to all commands that the driver has helpers for.

durran: (It also gives us TXN support for free)

mcasimir: Looked into that, did not find a way to pass the write concerns: https://mongodb.github.io/node-mongodb-native/3.1/api/Db.html#dropDatabase

PR #59, opened 2020-03-10
MONGOSH-53: emit log events in mapper

lrlna:
## Description
Uses [`nanobus`](github.com/choojs/nanobus) to emit and listen to log events. Most log events are emitted in the mapper, with the exception of errors we receive and handle directly in the CLI and the connect event.

Each action is listened to separately (this is why it's a bit of a larger PR) since eventually most of these events will only be logged to user's internal file _with a few exceptions_ of things actually going off to telemetry. Listening to them individually helps us sort out what information can go where.

Mapper now takes an instance of `EventEmitter`, which will default to an EventEmitter if none are provided. This will allow other implementers (i.e. browser, jet brains) to have their own listeners that are best suited for their implementations.

I know it's a bit long, but most of it is just the listener definitions \o/

_edit: will squash everything before I merge in!_


packages/cli-repl/src/cli-repl.ts
```diff
@@ -39,9 +42,14 @@ class CliRepl {
    */
   async connect(driverUri: string, driverOptions: NodeOptions): Promise<void> {
     console.log(i18n.__(CONNECTING), driverUri);
+    // @ts-ignore
```
mcasimir: What was the problem here? 'ts-ignore' is dangerous, there are many workarounds that are safer, maybe i can help you to find one.

mcasimir: maybe this didn't work: https://github.com/choojs/nanobus/blob/master/index.d.ts, :/ .. or it wants to be used as class like: `new Nanobus('mongosh')`.

packages/cli-repl/src/cli-repl.ts
```diff
@@ -9,6 +9,8 @@ import Mapper from 'mongosh-mapper';
 import completer from './completer';
 import i18n from 'mongosh-i18n';
 import formatOutput from './format-output';
+import nanobus from 'nanobus';
```
mcasimir: just curious, what is the difference with EventEmitter?

packages/cli-repl/src/cli-repl.ts
mcasimir: Another a neat way is to use an interface in place of the actual type:

```
interface Bus {
  emit(string, ...any) => void; 
}
```

An then just

```
  private bus: Bus
```

That would work in typescript since classes does not have to implement interfaces explicitly. This is also a sort of pattern.

packages/cli-repl/src/logger.ts
```diff
@@ -0,0 +1,162 @@
+import pino from 'pino'
+import path from 'path'
+import os from 'os'
+
+function logger (bus) {
+  const dest = path.join(os.homedir(), './.mongosh_log')
+  const log = pino({ name: 'monogsh' }, pino.destination(dest))
+
+  bus.on('*', function() {
```
mcasimir: fancy!

packages/mapper/src/mapper.ts
```diff
@@ -81,9 +89,12 @@ export default class Mapper {
       case 'dbs':
         const result = await this.serviceProvider.listDatabases('admin');
         if (!('databases' in result)) {
-          throw new Error('Error: invalid result from listDatabases');
+          const err = new Error('Error: invalid result from listDatabases');
+          this.messageBus.emit('error', err);
```
mcasimir: Would it be an option to handle the error logging only once? Dunno, maybe with a `Proxy`, so we avoid to test that we emit the error each time there is an exception.

packages/cli-repl/src/logger.ts
mcasimir: Maybe this is the difference :)

mcasimir: Maybe is a stupid idea, but would it be possible to use this for everything? emitting all the infos as an object:

```
bus.emit('method:aggregate', { collection, pipeline });
```

lrlna: Yep, this is the difference! Nanobus has a star event that we can append stuff too that all events might get. 

I don't think it's good to use this for everything since we are going to be sending different information to telemetry and logging and it will be on case by case bases. For example, collection information should probably not be sent to telemetry, since that might conflict with GDPR.

packages/mapper/src/mapper.ts
lrlna: Yea, I think that'd be possible. 

I already do this in the CLI since that's where we catch all the errors, so I think this might be already redundant call. Good catch!

packages/cli-repl/src/logger.ts
mcasimir: Makes sense!

packages/cli-repl/src/cli-repl.ts
mcasimir: Forgot to mention the quickest fix of all: `private bus: any`

lrlna: Probably the cleanest way is just doing `new Nanobus('mongosh')`. I'll do that!

PR #60, opened 2020-03-11
Add eslint to transport server

mcasimir:


packages/transport-server/src/node-cursor.spec.ts
```diff
@@ -140,25 +140,6 @@ describe('NodeCursor', () => {
     });
   });
 
-  describe('#explain', () => {
```
mcasimir: Explain has to be implemented and can't return a cursor since the driver returns results

packages/transport-server/src/node-transport.integration.spec.ts
```diff
@@ -48,7 +48,7 @@ describe('NodeTransport [ integration ]', function() {
 
       beforeEach(async() => {
         result = await nodeTransport.
-          aggregate('music', 'bands', [{ $match: { name: 'Aphex Twin' }}]);
+          aggregate('music', 'bands', [{ $match: { name: 'Aphex Twin' } }]);
```
lrlna: you know Durran wrote these tests when .... :joy:

packages/transport-server/src/node-transport.ts
```diff
@@ -32,7 +32,8 @@ class NodeTransport implements Transport {
    * @returns {NodeTransport} The Node transport.
    */
   static async fromURI(uri: string, options: NodeOptions = {}): Promise<NodeTransport> {
-    const mongoClient = new MongoClient(uri, { ...DEFAULT_OPTIONS, ...options });
+    const clientOptions: any = { ...DEFAULT_OPTIONS, ...options };
```
lrlna: ahhh i see what you meant about type definition when declaring a variable

PR #62, opened 2020-03-15
fix: shell api commands not working in browser

mcasimir:
Creates context with shell api attributes so that it will contain attributes of shell api that are non own properties.

Fixes an issue where commands like `show` and `use` where not accessible from the browser repl.


mcasimir:
Fix moved to https://github.com/mongodb-js/mongosh/pull/64, will move tests for electron runtime too

PR #64, opened 2020-03-16
refactor: unify set ctx code and move to mapper

mcasimir:
Moves the code that assign the ShellApi methods and attributes to the global context to the `setCtx` method of `Mapper`.

The code was present in 2 forms in both `browser-repl` and `cli-repl`.

Is that the right place to factor it?

NOTES:

1. `new ShellApi()` is moved in the `setCtx` of mapper too.
2. We run the commands directly from the context, is that ok?
3. If we keep this do we have to expose shell api? like `mapper.shellApi`?


aherlihy:
Let me know when it's not WIP anymore and I'll review!

PR #67, opened 2020-03-16
Fix up async-writer api

aherlihy:
NOTE: this doesn't include any implementation changes, only changing the packge to return a class instead of a function.

This is my suggestion for what the async writer API should look like. The major change is that the mapper doesn't have to construct the SymbolTable separately, it's initialized and stored within the AsyncWriter class. So to use from the mapper:

```
import { types } from 'shell-api'; // this will be an argument when the context object PR is merged
const globalScope = { db: types.Database };

const writer = new AsyncWriter(globalScope, types); // initialize
const rewrittenCode = writer.compile('code'); // use
```
The `globalScope` argument is the set of global variables (like `db`) and their types. The `types` argument is the ShellApi types from `shell-api-types` and is passed as an argument in order to not add an extra dependency to async-rewriter package (also because in theory this package can be used independently of the shell itself).

There is a little bit of weirdness in the async-writer class because of how babel's plugins work. The `InferTypeVisitor` has to be it's own class, yet it requires access to the symbol table, but I cannot modify the constructor or pass it any arguments/options. I don't like defining the visitor within the constructor but that's the only way I could think to do it. Open to suggestions if you think it can be improved!

### DIFF
--- a/packages/async-rewriter/src/async-writer-babel.ts
+++ b/packages/async-rewriter/src/async-writer-babel.ts
@@ -1,167 +1,179 @@
 /* eslint no-console:0 */
 import * as babel from '@babel/core';
 // import printAST from 'ast-pretty-print';
+import SymbolTable from './symbol-table';
 
-export default (symbols): any => {
-  const debug = (str, type?): void => {
-    str = `  ${str}${type === undefined ? '' : ` ==> ${type}`}`;
-    // console.log(str);
-  };
+const debug = (str, type?): void => {
+  str = `  ${str}${type === undefined ? '' : ` ==> ${type}`}`;
+  // console.log(str);
+};
 
-  const inferTypeVisitor = {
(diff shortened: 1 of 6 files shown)


packages/async-rewriter/src/async-writer-babel.ts
```diff
@@ -1,167 +1,179 @@
...
 
-  const inferTypeVisitor = {
-    Identifier: {
-      exit(path): void {
-        const id = path.node.name;
-        let sType = symbols.lookup(id);
-        if (typeof sType === 'string') {
-          sType = symbols.types[sType];
+export default class AsyncWriter {
+  public symbols: SymbolTable;
```
aherlihy: This is public in case we need to access the symbol table from elsewhere, potentially for debugging or logging.

PR #74, opened 2020-03-19
feat: add createIndex and createIndexes

mcasimir:
Implements `db.coll.createIndex` and `db.coll.createIndexes`.

Changes:
- service provider: Implement `createIndexes` and `getIndexes` (used in mapper for tests)
- mapper: Implement `createIndexes` and `createIndex` as wrapper of `createIndexes` (as is currenly).  
- shell api: Add  `createIndexes` and `createIndex` specs and help

Chores:
- Use `ts-sinon` in mapper tests to mock the `ServiceProvider` interface.

### DIFF
--- a/packages/service-provider-server/src/cli-service-provider.ts
+++ b/packages/service-provider-server/src/cli-service-provider.ts
@@ -448,7 +448,8 @@ class CliServiceProvider implements ServiceProvider {
   }
 
   /**
-   * Insert many documents into the colleciton.
+   * Insert many documents into the collection. {
+   * db('db1').collection('coll1').listIndexes().toArray()}
    *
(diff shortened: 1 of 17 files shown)


packages/service-provider-server/src/cli-service-provider.ts
```diff
@@ -448,7 +448,8 @@ class CliServiceProvider implements ServiceProvider {
   }
 
   /**
-   * Insert many documents into the colleciton.
+   * Insert many documents into the collection. {
+   * db('db1').collection('coll1').listIndexes().toArray()}
```
aherlihy: I think this got put in the wrong place?

PR #85, opened 2020-03-25
formatting 'show dbs' in implemenetors instead of mapper

lrlna:
## Description

I noticed we were formatting db size and creating a table in the mapper and
figured that should be in cli/browser/other implementors instead.

Result is the same as before, in cli:
Capture d’écran, le 2020-03-25 à 13 15 54

in browser:
Capture d’écran, le 2020-03-25 à 13 17 17


packages/browser-repl/src/components/shell-output-line.tsx
```diff
@@ -48,6 +49,10 @@ export class ShellOutputLine extends Component<ShellOutputLineProps> {
       return <HelpOutput value={value} />;
     }
 
+    if (shellApiType === 'CommandResult') {
```
mcasimir: I think the idea of `CommandResult` was a bit different. @aherlihy suggested to use it for all the commands that has pre-formatted output. So that would not work really well, if we want to do what you suggested we need a specific type that identifies the `ShowDbsOutput` something like `ShowDbsResult`.

lrlna: ah! i was not aware that this type of result was reserved for something specific. I'll change this particular one to `ShowDbsResult` then.

PR #93, opened 2020-03-31
move write dir before creating logger instance

lrlna:


packages/cli-repl/src/cli-repl.spec.ts
```diff
@@ -0,0 +1,17 @@
...
+import parseCliArgs from './arg-parser';
+import CliRepl from './cli-repl';
+import { expect } from 'chai';
+
+describe('cli-repl', () => {
+  context('constructor', () => {
+    it('starts without errors', (done) => {
+      const options = parseCliArgs([]);
+      const driverOptions = mapCliToDriver(options);
+      const cli = new CliRepl('mongodb://127.0.0.1:27017', driverOptions, options)
```
mcasimir: This will try to connect, we may need to start a server

packages/cli-repl/src/cli-repl.spec.ts
```diff
@@ -0,0 +1,17 @@
...
+      cli.close()
+      expect(cli).to.not.throw();
```
mcasimir: `expect(...).to.not.throw();` wants a function, it will try to run `cli()`.

packages/cli-repl/src/cli-repl.spec.ts
```diff
@@ -0,0 +1,17 @@
...
+import CliRepl from './cli-repl';
+import { expect } from 'chai';
+
+describe('cli-repl', () => {
+  context('constructor', () => {
+    it('starts without errors', (done) => {
+      const options = parseCliArgs([]);
+      const driverOptions = mapCliToDriver(options);
+      const cli = new CliRepl('mongodb://127.0.0.1:27017', driverOptions, options)
+      cli.close()
```
mcasimir: Since `new CliRepl` runs async code there is no guarantee this would ever run after the repl started.

packages/cli-repl/src/cli-repl.spec.ts
```diff
@@ -0,0 +1,17 @@
...
+      expect(cli).to.not.throw();
+      done();
```
mcasimir: This has to be called in a callback.

packages/cli-repl/src/cli-repl.spec.ts
mcasimir: since `new CliRepl()` starts node repl the stdin and stdout are messed up entirely.

An option there is to pass mocks of those streams.

lrlna: for other readers: this was resolved outside of this PR. We will think of an easier way to test `CliRepl` instances in a different PR.

packages/cli-repl/src/cli-repl.spec.ts
```diff
@@ -0,0 +1,10 @@
+import mapCliToDriver from './arg-mapper';
```
mcasimir: If this file does nothing i'd remove it, would only make eslint unhappy once we add it

PR #96, opened 2020-04-01
`db.coll.stats()` and related helpers

mcasimir:
Implements `db.coll.stats()`, `db.coll.storageSize()`,  `db.coll.dataSize()` and `db.coll.totalSize()`.

```
$ mongosh > db.coll1.stats().then(s => [s.storageSize, s.totalIndexSize, s.size])
[ 20480, 20480, 165 ]
$ mongosh > db.coll1.storageSize()
20480
$ mongosh > db.coll1.totalIndexSize()
20480
$ mongosh > db.coll1.dataSize()
165
$ mongosh > db.coll1.totalSize()
40960
```


packages/cli-repl/src/logger.ts
```diff
@@ -230,6 +230,26 @@ function logger(bus: any, logDir: string) {
...
+    const params = { collection };
+    log.info('method:dataSize', params);
+  });
+
+  bus.on('method:storageSize', function(collection) {
+    const params = { collection };
+    log.info('method:storageSize', params);
+  });
+
+  bus.on('method:totalSize', function(collection) {
```
lrlna, line 249: I am wondering if we ought to namespace is a bit more in the future, something along the lines of: `method:db.coll.storageSize`, so it's extra extra clear. Definitely not a `now` fix though!

mcasimir: Makes sense

PR #97, opened 2020-04-01
MONGOSH-106, MONGOSH-107: add telemetry opt-in command, add userid

lrlna:
## Description
This PR creates a config file where currently two things are stored:
- userId
- telemetry opt-in.

For example:
Capture d’écran, le 2020-04-01 à 17 18 35

When user initial logs in, we write the file, setup `userId`, and toggle `enableTelemetry` to false. User also receives a message to opt in. They can subsequently opt in with `enableTelemetry` command or opt out with `disableTelemetry`.

For example: 
Capture d’écran, le 2020-04-01 à 17 15 21

## Next Steps
This is not yet hooked up to Segment, so next thing would be to do that, as well as hook up the logger to be notified when user opts in and opts out. That way it can take into consideration when to write to Segment.

### DIFF
--- a/packages/cli-repl/src/cli-repl.ts
+++ b/packages/cli-repl/src/cli-repl.ts
@@ -1,10 +1,10 @@
 import { CliServiceProvider, NodeOptions } from '@mongosh/service-provider-server';
 import { changeHistory } from '@mongosh/history';
 import formatOutput from './format-output';
+import { TELEMETRY } from './constants';
 import repl, { REPLServer } from 'repl';
 import CliOptions from './cli-options';
 import Mapper from '@mongosh/mapper';
-import { TELEMETRY } from './constants';
 import completer from './completer';
 import i18n from '@mongosh/i18n';
 import { ObjectId } from 'bson';
@@ -106,6 +106,7 @@ class CliRepl {
           this.enableTelemetry = config.enableTelemetry;
           return;
         }
+        this.bus.emit('error', e)
         throw err;
       }
 
--- a/packages/cli-repl/src/cli-repl.ts
+++ b/packages/cli-repl/src/cli-repl.ts
@@ -98,24 +98,64 @@ class CliRepl {
   generateOrReadTelemetryConfig(): void {
     const configPath = path.join(this.mongoshDir, 'config');
 
-    fs.open(configPath, 'wx', (err, fd) => {
-      if (err) {
-        if (err.code === 'EEXIST') {
-          const config = JSON.parse(fs.readFileSync(configPath, 'utf8'));
-          this.userId = config.userId;
-          this.enableTelemetry = config.enableTelemetry;
-          return;
-        }
-        this.bus.emit('error', e)
-        throw err;
-      }
+    let fd;
 
+    try {
+      fd = fs.openSync(configPath, 'wx');
       this.userId = new ObjectId(Date.now());
       this.enableTelemetry = false;
-      const config = { userId: this.userId, enableTelemetry: this.enableTelemetry };
-      fs.writeFileSync(fd, JSON.stringify(config));
-    });
+      this.writeConfigFileSync(configPath);
+    } catch (err) {
+      if (err.code === 'EEXIST') {
+        const config = JSON.parse(fs.readFileSync(configPath, 'utf8'));
+        this.userId = config.userId;
+        this.enableTelemetry = config.enableTelemetry;
+        return;
+      }
+      this.bus.emit('error', err)
+      throw err;
+
+    } finally {
+      if (fd !== undefined) fs.closeSync(fd);
+    }
   }
+
+  /**
+  * sets CliRepl.enableTelemetry based on a bool, and writes the selection to
+  * config file.
+  *
+  * @param {boolean} enabled - enabled or disabled status
+  *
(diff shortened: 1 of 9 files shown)


packages/cli-repl/src/cli-repl.ts
```diff
@@ -68,6 +75,48 @@ class CliRepl {
...
+  }
+
+  /**
+  * Checks if config file exists.
+  *
+  * If exists: sets userId and enabledTelemetry to this.
+  * If does not exist: writes a new file with a newly generated ObjectID for
+  * userid and enableTelemetry set to false.
+  */
+  generateOrReadTelemetryConfig(): void {
```
mcasimir, line 98: How do we sync `fs.open` here?

packages/cli-repl/src/cli-repl.ts
```diff
@@ -197,6 +239,28 @@ class CliRepl {
...
+  *
+  * @returns {string} Status of telemetry logging: disabled/enabled
+  */
+  toggleTelemetry(enabled: boolean): string {
+    this.enableTelemetry = enabled;
+
+    const configPath = path.join(this.mongoshDir, 'config');
+    const config = { userId: this.userId, enableTelemetry: this.enableTelemetry };
+    fs.writeFileSync(configPath, JSON.stringify(config));
+
```
mcasimir, line 256: These line are also in `generateOrReadTelemetryConfig` would it be an idea to put them together in a function? Dunno writeConfig, readConfig :/

packages/cli-repl/src/cli-repl.ts
mcasimir: like if `fs.open` would take long enough (or it would hang)

mcasimir:
Noticed now bootstrap fails too

packages/cli-repl/src/constants.ts
```diff
@@ -1,6 +1,12 @@
 import i18n from '@mongosh/i18n';
 import clr from './clr';
 
+export const TELEMETRY = `
+${i18n.__('cli-repl.cli-repl.telemetry')}
+${i18n.__('cli-repl.cli-repl.enableTelemetry')}${clr('enableTelemtry()', 'bold')}.
```
mcasimir, line 6: I think your worst nightmare is here again :D

packages/cli-repl/src/cli-repl.ts
lrlna: Both `fs.open` and `fs.openSync` could be blocking or take long enough, the only difference (iirc) is that `fs.open` is backed by a thread pool. If you are worried about it taking too long, I can look into how to add a timeout. 

I did look at having it done synchronously, and here is what that looked like:

```js
    let fd;

    try {
      fd = fs.openSync(configPath, 'wx');

      this.userId = new ObjectId(Date.now());
      this.enableTelemetry = false;
      const config = { userId: this.userId, enableTelemetry: this.enableTelemetry };
      fs.writeFileSync(fd, JSON.stringify(config));
    } catch (err) {
      if (err.code === 'EEXIST') {
        const config = JSON.parse(fs.readFileSync(configPath, 'utf8'));
        this.userId = config.userId;
        this.enableTelemetry = config.enableTelemetry;
        return;
      }
      this.bus.emit('error', e)
      throw err;

    } finally {
      if (fd !== undefined) fs.closeSync(fd);
    }
```

packages/cli-repl/src/cli-repl.ts
lrlna: Good catch! I didn't even notice 😅 Will write a lil abstracted function

packages/cli-repl/src/constants.ts
lrlna: it took me literally 20 min to understand what you meant :D

packages/cli-repl/src/cli-repl.ts
mcasimir: I meant that with the async version we could have race conditions. Like `this.userId not being set when we try to read it.

lrlna: a race condition is a horse of a different colour! i can make it sync.

lrlna:
@mcasimir I think I addressed everything you've mentioned now.

In terms of testing -- it'll have to be under end to end tests. I don't think there is a way to do this otherwise :/. I have so far thought of possibly moving all the reads and writes over to another file, but I feel like that will give us only the ability to test actual `fs` functions (which is already done in node itself), and creates a bit more work around managing and passing around information such as `this.userId`, or `this.mongoshDir` or `this.enableTelemetry`.

mcasimir:
Yeah i think we are good with e2e tests, let's add some close after then.

> I have so far thought of possibly moving all the reads and writes over to another file, but I feel like that will give us only the ability to test actual `fs` functions.

Was not suggesting that .. but it doesn't sound like a bad idea to me. If we could do constructor injection (`new CliRepl(dep1, dep2, dep3)`) it would be just great and improve the testability, but, even without we could get a better separation of concerns: `CliRepl` would know to use the configuration but not how to store, retrieve, serialize and deserialize it, so it would have one responsiblity less.

mcasimir:
also LGTM :)

PR #99, opened 2020-04-02
Move evaluator to Shell API

aherlihy:


packages/shell-api/src/top-level-api.ts
```diff
@@ -0,0 +1,72 @@
...
+      case 'use':
+        return this.use(argv[0]);
+      case 'show':
+        return this.show(argv[0]);
+      case 'it':
+        return this.it();
+      case 'help':
+        return this.help();
+      case 'var':
+        this._mapper.cursorAssigned = true;
```
mcasimir, line 37: always wanted to ask this one, do we have to do something more here? like handle also `let`, `const`, direct assignment? Could we just reset the cursor?

aherlihy: I wish we could remove that case but not iterating the cursor when it's not returned to the evaluator is expected current shell behavior that we're not going to be able to remove anytime soon I think. I do think you're right that we should do some more checking to see if cursor is assigned or not instead of just parsing for `var`. So get rid of `cursorAssigned` entirely, assume that all cursors are not iterated. It would probably happen as you said in the print statement, where we check if the value returned to the evaulator is of type cursor and if so then iterate it.

Otherwise we would have to pass some state into the async rewriter so that we could record when parsing if the find command result is assigned to anything. I made a ticket: https://jira.mongodb.org/browse/MONGOSH-161

aherlihy:
Superseded by https://github.com/mongodb-js/mongosh/pull/103

PR #103, opened 2020-04-07
Refactor evaluation into ShellEvaluator

aherlihy:
Summary of changes:

- This `shell-api` package is now 100% YAML generated and has no internal dependencies other than i18n. This was done because that package is required by many internal packages and it makes it easier to avoid circular dependencies if we move the non-generated JS out of the generated library. This will also make it easier if we decide to remove generation entirely.
    1. I also renamed the generated `types` object to `signatures` because types is too generic a name.
- Define a new `ShellEvaluator` class in JavaScript that will serve as the entry point to the backend. 
    1. Contains the `Mapper` and `AsyncRewriter` instances.
    2. Provides a `customEval` function that takes in a JS evaluator and parses for unix-style commands, which are called directly on the `Mapper`. If the input is regular JS, it runs the input through the async rewriter and invokes the evaluator argument on the result.
  3. `setCtx` moved to this package from the `Mapper` so that the context and the symbol table could be updated in the same place. Additionally the `cli-repl` and `browser` packages will only require this `ShellEvaluator` package and will not need to require or know about the `Mapper`. There's some slight weirdness here where we are setting the `context` attribute of the Mapper here as well, but I anticipate that will go away when we remove/thin out the `Mapper` class entirely so that state is tracked within the `ShellEvaluator` class.

Two side effect changes:
- Remove the `cursorAssigned` variable because we will need to track that better, see https://jira.mongodb.org/browse/MONGOSH-161
- Remove async-rewrite feature flag so input gets rewritten every time.


packages/async-rewriter/src/symbol-table.ts
```diff
@@ -20,22 +20,25 @@ export default class SymbolTable {
       this.add(s, { type: 'classdef', returnType: this.types[s], lib: true });
     });
   }
-  deepCopy(): SymbolTable {
```
aherlihy, line 23: Can ignore these changes, that just allows us to instantiate the Symbol Table and add global objects in two separate steps instead of one.

packages/browser-runtime-core/src/open-context-runtime.ts
```diff
@@ -45,6 +45,6 @@ export class OpenContextRuntime implements Runtime {
 
   private setupEvaluationContext(context: object, serviceProvider: object): void {
     const mapper = new Mapper(serviceProvider);
-    mapper.setCtx(context);
```
aherlihy, line 48: TODO for browser support

packages/browser-runtime-electron/src/electron-runtime.spec.ts
```diff
@@ -4,7 +4,7 @@ import sinon from 'sinon';
 import { CliServiceProvider } from '@mongosh/service-provider-server';
 import { ElectronRuntime } from './electron-runtime';
 
-describe('Mapper (integration)', function() {
+xdescribe('Mapper (integration)', function() { // TODO
```
aherlihy, line 7: TODO for browser support

packages/cli-repl/src/completer.spec.ts
```diff
@@ -1,5 +1,5 @@
 import completer from './completer';
-import { types as shellTypes } from '@mongosh/shell-api';
```
aherlihy, line 2: I think my editor made this change automatically, can revert if anyone sees a problem with it.

packages/shell-api-types/bin/compile-shell-api-types.js
```diff
@@ -190,9 +190,20 @@ const loadAll = () => {
...
-  let exports = '\nexport default ShellApi;\n';
-  const types = ['const unknown = {\n  type: \'unknown\',\n  attributes: {}\n}'];
-  const typeConsts = ['unknown'];
+  let exports = '';
+  const types = [
+    'const unknown = {\n  type: \'unknown\',\n  attributes: {}\n}',
+    `const ShellApi = {
+    type: 'ShellApi',
+    hasAsyncChild: false,
+    attributes: {
```
aherlihy, line 199: This irritates me slightly because we are defining these types in a separate place from where they are implemented (i.e. they exist on the ShellAPI object but are not getting generated from code. So if we update the top-level commands we have to do it in two places)

packages/shell-api/src/shell-api.ts
```diff
@@ -0,0 +1,133 @@
+import AsyncWriter from '@mongosh/async-rewriter';
+import Mapper from '@mongosh/mapper';
+import { types, Help } from '@mongosh/shell-api-types';
+
+class ShellApi {
+  public mapper: any;
+  private asyncWriter: AsyncWriter;
+  private bus: any;
+  private container: any;
+  constructor(serviceProvider, bus, container) {
```
aherlihy, line 10: Passing a reference to the container so that we can start/stop telemetry: yes or no?

packages/shell-api/src/shell-api.ts
```diff
@@ -0,0 +1,133 @@
...
+    this.mapper = new Mapper(serviceProvider, bus);
+    this.asyncWriter = new AsyncWriter({}, types);
+    this.bus = bus;
+    this.container = container;
```
lrlna, line 14: So the container is only used for telemetry right now, eh? I wonder if we could default it to noop, so you don't have to do [those checks below](https://github.com/mongodb-js/mongosh/pull/103/files#diff-3c80e6a0146b17d9c12433c0f55c1147R62)?

aherlihy: Yeah, I thought there might be chance there will be other things we need to do beyond telemetry. Can you think of any?

packages/shell-api-types/bin/compile-shell-api-types.js
mcasimir: Can't read the code anymore, you mean command parsing vs invoking?

aherlihy: Ah sorry, went away after rebasing, I will make a new comment.

packages/shell-api-types/bin/compile-shell-api-types.js
```diff
@@ -190,9 +190,20 @@ const loadAll = () => {
   const main = fs.readFileSync(path.join(yamlDir, 'main.yaml'));
   const mainLib = yaml.load(main);
   const FILES = fs.readdirSync(yamlDir).filter((s) => (/[A-Z]/.test( s[0])));
-  let exports = '\nexport default ShellApi;\n';
-  const types = ['const unknown = {\n  type: \'unknown\',\n  attributes: {}\n}'];
-  const typeConsts = ['unknown'];
+  let exports = '';
+  const types = [
```
aherlihy, line 194: (@mcasimir this is where I was commenting.)

This irritates me slightly - because the ShellAPI class is no longer generated from YAML, but it still needs to be present in the `shell-types-signatures` file, we have to define the attributes in two separate places (once in `shell-types-signatures` and once in `shell-api/src/shell-api.ts`. This is because the signatures file is generated also from YAML but ShellApi is no longer in YAML. Perhaps we can get rid of the need to have ShellAPI in the `shell-types-signatures` file entirely to fix the issue.

aherlihy: Actually - nevermind, I just moved it out of this file so the issue should be fixed. It was only being used in the cli-repl completer module I believe.

The change is here: https://github.com/mongodb-js/mongosh/commit/d7c84b4675df0784083cce17465d7cecd687b215

packages/shell-api/src/shell-api.ts
lrlna: I missed this! Sorry! 

Nothing comes to mind right now aside from telemetry.

packages/shell-evaluator/src/shell-evaluator.ts
```diff
@@ -0,0 +1,133 @@
+import AsyncWriter from '@mongosh/async-rewriter';
+import Mapper from '@mongosh/mapper';
+import { signatures, Help } from '@mongosh/shell-api';
+
+class ShellEvaluator {
+  public mapper: any;
+  private asyncWriter: AsyncWriter;
+  private bus: any;
+  private container: any;
```
mcasimir, line 9: Let's add a type there.

packages/shell-evaluator/src/shell-evaluator.ts
```diff
@@ -0,0 +1,133 @@
+import AsyncWriter from '@mongosh/async-rewriter';
+import Mapper from '@mongosh/mapper';
+import { signatures, Help } from '@mongosh/shell-api';
+
+class ShellEvaluator {
+  public mapper: any;
+  private asyncWriter: AsyncWriter;
+  private bus: any;
```
mcasimir, line 8: Let's add a type there.

packages/shell-evaluator/src/shell-evaluator.ts
```diff
@@ -0,0 +1,133 @@
...
+  private container: any;
+  constructor(serviceProvider, bus, container?) {
+    this.mapper = new Mapper(serviceProvider, bus);
+    this.asyncWriter = new AsyncWriter(signatures);
+    this.bus = bus;
+    this.container = container;
```
mcasimir, line 14: What is `container`? That makes me think of dependency injection but is not so clear since the other deps are passed as normal parameters. Wouldn't you rather have something that tells what's the purpose there? If is only for telemetry something telemetry related like `TelemetryConfig` perhaps.

By the way, typescript has duck-typing everywhere and types are matched by shape, so interfaces does not have to be implemented explicitly and we can have an object implementing multiple interfaces.

That means that unlike java we can define interfaces for what we expect from dependencies we have as parameters.

For example if you expect `container` to be an object having `toggleTelemetry(boolean)` you can enforce that rule with an interface or type there:

```
interface Container {
   toggleTelemetry(boolean): void
}
```

The implementor of `Container` does not need to know about that at all, it means that in the caller we can pass any object of any class that has a `toggleTelemetry(boolean)` method.

packages/shell-evaluator/src/shell-evaluator.ts
```diff
@@ -0,0 +1,133 @@
+import AsyncWriter from '@mongosh/async-rewriter';
+import Mapper from '@mongosh/mapper';
+import { signatures, Help } from '@mongosh/shell-api';
+
+class ShellEvaluator {
+  public mapper: any;
```
mcasimir, line 6: Don't we want to make that `private`?

Also i'd add a type there.

packages/shell-evaluator/src/shell-evaluator.ts
mcasimir, line 8: I would add a type here too.

packages/shell-evaluator/src/shell-evaluator.ts
```diff
@@ -0,0 +1,133 @@
...
+  private bus: any;
+  private container: any;
+  constructor(serviceProvider, bus, container?) {
+    this.mapper = new Mapper(serviceProvider, bus);
+    this.asyncWriter = new AsyncWriter(signatures);
+    this.bus = bus;
+    this.container = container;
+  }
+
+  public toReplString(): string {
```
mcasimir, line 17: I think `toReplString` and `shellApiType` are not needed here. An instance of this object will never be evaluated, nor constructed inside the shell context. For what i get there is nothing that could call these two methods.

packages/shell-evaluator/src/shell-evaluator.ts
```diff
@@ -0,0 +1,133 @@
...
+        }
+        return;
+      case 'disableTelemetry':
+        if (this.container) {
+          this.container.toggleTelemetry(false);
+        }
+        return;
+      default:
+        const rewrittenInput = this.asyncWriter.compile(input);
+        console.log('\x1b[36m%s\x1b[0m', `rewrote "${input.trim()}" to "${rewrittenInput.trim()}"`);
```
mcasimir, line 73: Should we use `debug` here too like we do in compass?

packages/shell-evaluator/src/shell-evaluator.ts
```diff
@@ -0,0 +1,133 @@
...
+
+  /**
+   * Evaluates the input code and wraps the result with the type
+   *
+   * @param {function} originalEval - the javascript evaluator.
+   * @param {String} input - user input.
+   * @param {Context} context - the execution context.
+   * @param {String} filename
+   */
+  public async customEval(originalEval, input, context, filename) {
```
mcasimir, line 86: Would be good to have a return type here: `Promise<{type: string, value: any}>`

packages/shell-evaluator/src/shell-evaluator.ts
```diff
@@ -0,0 +1,133 @@
...
+   * The `contextObject` is prepared so that it can be used as global object
+   * for the repl evaluationi.
+   *
+   * @note The `contextObject` is mutated, it will retain all of its existing
+   * properties but also have the global shell api objects and functions.
+   *
+   * @param {Object} - contextObject an object used as global context.
+   */
+  setCtx(contextObject: any): void {
+    // Add API methods for VSCode
```
mcasimir, line 117: I think those are also for scripts in general

packages/shell-evaluator/src/shell-evaluator.ts
mcasimir: Typescript has duck-typing everywhere and types are matched by shape, so interfaces does not have to be implemented explicitly and we can have an object implementing multiple interfaces.

That means that unlike java we can define interfaces for what we expect from dependencies we have as parameters without affecting the 
implementation.

For example if you expect `container` to be an object having `toggleTelemetry(boolean)` you can enforce that rule with an interface or type there:

```
interface Container {
   toggleTelemetry(boolean): void
}
```

The implementor of `Container` does not need to know about that at all, it means that in the caller we can pass any object of any class that has a `toggleTelemetry(boolean)` method and could have any other methods too.

PR #108, opened 2020-04-15
MONGOSH-103: redact information from logs and connection dialogue

lrlna:
## Description
Removes sensitive information from logs if exists. Uses `mongodb-redact` module just like in `@mongosh/history`. To remove credentials in `driverURI` currently a separate regex to replace information with `<credentials>`.

Examples:
Capture d’écran, le 2020-04-15 à 15 07 06

And in logs:
Capture d’écran, le 2020-04-15 à 15 06 41

### DIFF
--- a/packages/cli-repl/src/index.ts
+++ b/packages/cli-repl/src/index.ts
@@ -1,4 +1,5 @@
 import CliRepl from './cli-repl';
+import redactPwd from './redact-pwd';
 import parseCliArgs from './arg-parser';
 import mapCliToDriver from './arg-mapper';
 import generateUri from './uri-generator';
@@ -13,6 +14,7 @@ export {
   USAGE,
   TELEMETRY,
   CliRepl,
+  redactPwd,
   completer,
   generateUri,
   parseCliArgs,
--- a/packages/cli-repl/src/logger.ts
+++ b/packages/cli-repl/src/logger.ts
@@ -1,3 +1,5 @@
+import redactInfo from 'mongodb-redact';
+import redactPwd from './redact-pwd';
 import { uuid } from 'uuidv4';
 import pino from 'pino';
 import path from 'path';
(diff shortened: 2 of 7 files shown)


packages/cli-repl/src/logger.ts
```diff
@@ -38,217 +40,217 @@ function logger(bus: any, logDir: string) {
 
   bus.on('method:aggregate', function(collection, pipeline) {
     const params = { collection, pipeline };
-    log.info('method:aggregate', params);
+    log.info('method:aggregate', redactInfo(params));
```
mcasimir, line 43: We do `redactInfo(params)` everywhere, maybe we can extract this in a function:

```
function logMethodCall(method, params) {
   log.info(method, redactInfo(params));
}
```

This way next time we have to do a change in format is going to be just one place.

packages/cli-repl/src/logger.ts
```diff
@@ -12,7 +14,7 @@ function logger(bus: any, logDir: string) {
   bus.on('*', function() {});
 
   bus.on('connect', function(info) {
-    const params = { sessionID, info };
+    const params = { sessionID, info: redactPwd(info) };
```
mcasimir, line 17: Why not `redactPassword`, may be just me being exceptionally confused by abbreviations though.

packages/cli-repl/src/index.ts
```diff
@@ -13,6 +14,7 @@ export {
   USAGE,
   TELEMETRY,
   CliRepl,
+  redactPwd,
```
mcasimir, line 17: I think i added this in some PR, not sure what you think about it, but i would not export anything as public unless is necessary. 

If others start to use what we expose we will be limited in the possibility to change things, Is kinda the same thing like marking something that is private as public.

packages/cli-repl/src/redact-pwd.ts
```diff
@@ -0,0 +1,4 @@
+export default function retractPassword(uri: string): string {
```
mcasimir, line 1: I think we got a typo here: `retractPassword => redactPassword`

PR #109, opened 2020-04-16
Add return type signatures to Cursor

mcasimir:
- Adds `returnType: 'Cursor'`, `__fluent: true` and `returnsPromise: true` where appropriate in `Cursor.yaml`.
- Fixes the implementation of `forEach`.
- Commented out few not implemented methods in Cursor.yaml so we don't forget about it.


mcasimir:
I read your comment, just removed the lines that didn't match, we can use `contain` or `have.members` don't remember what is the `chai` version.

PR #110, opened 2020-04-16
Define CommandResult/ShowDbsResult in YAML

aherlihy:


packages/shell-api/yaml/CommandResult.yaml
```diff
@@ -0,0 +1,8 @@
+class:
+    <<: *__defaultClass
+    help:
+        help: 'The CommandResult class.'
```
lrlna, line 4: help is generated from `i18n` module, and should be read from a set of keys from there. For example [DatabaseClass](https://github.com/mongodb-js/mongosh/blob/master/packages/shell-api/yaml/Database.yaml#L10) has this kind of entry. the `i18n` module then has the corresponding key pairs here in [en](https://github.com/mongodb-js/mongosh/blob/master/packages/i18n/src/locales/en_US.js#L1037) and [de](https://github.com/mongodb-js/mongosh/blob/master/packages/i18n/src/locales/de_DE.js#L953).

aherlihy: Right, feel free to push to this PR to update that.

PR #114, opened 2020-04-16
MONGOSH-109: standardise log emit format

lrlna:
## Description

Use `mognosh:` namespace throughout the project for logs. Emit a single event
from all the methods, `mongosh:api-call`, that results in having a single
listener.

New format when logging:

```js
this.messageBus.emit(
  'mongosh:api-call',
  {
    method: 'count',
    class: 'Collection',
    db: collection._database._name,
    coll: collection._name,
    arguments: { query, options }
  }
);
```

Most functions currently have `Database` or `Collection` as their first
argument, which is actually the entire `Database` or `Collection` class. So, we
should watch out logging the entire `arguments` when transitioning to the more
abstracted out way to log method calls.

This PRs also (again like in a previous one) adds a little log for the
input rewrite we are doing for easier debugging with internal users.


packages/mapper/src/mapper.ts
```diff
@@ -171,15 +203,24 @@ export default class Mapper {
...
     const dbOptions: any = {};
-    this.messageBus.emit('metho:bulkWrite', collection._name, operations);
+    const db = collection._database._name;
+    const coll = collection._name;
+    this.messageBus.emit(
+      'mongosh:api-call',
+      {
+        method: 'bulkWrite',
+        class: 'Collection',
+        db, coll, arguments: { operations, options }
```
lrlna, line 213: do we want to log the whole of operations here, in fact?

mcasimir: I'd say, let's emit those so we have them in the telemetry. We could add some filter where we log if it gets messy.

Again not super against to giving names there, but i'd really start with arguments being an array and see if we need more.

We probably can't generate the names, so if we want to move that to either the `compile-shell-api.js` or to a decorator we would lose them.

packages/mapper/src/mapper.ts
```diff
@@ -463,15 +564,28 @@ export default class Mapper {
...
     }
+
+    this.messageBus.emit(
+      'mongosh:api-call',
+      {
+        method: 'findOneAndReplace',
+        class: 'Collection',
+        db,
+        coll,
+        arguments: { filter, replacement, findOneAndReplaceOptions }
```
lrlna, line 582: do we want to log the replacement document, or just the filters/options?

packages/mapper/src/mapper.ts
```diff
@@ -495,18 +609,31 @@ export default class Mapper {
...
     }
+
+    this.messageBus.emit(
+      'mongosh:api-call',
+      {
+        method: 'findOneAndUpdate',
+        class: 'Collection',
+        db,
+        coll,
+        arguments: { filter, update, findOneAndUpdateOptions }
```
lrlna, line 627: same as with `replacement` above: do we want to log the `update` doc?

packages/mapper/src/mapper.ts
```diff
@@ -526,18 +653,30 @@ export default class Mapper {
...
     if ('writeConcern' in options) {
       dbOptions.writeConcern = options.writeConcern;
     }
+
+    this.messageBus.emit(
+      'mongosh:api-call',
+      {
+        method: 'insert',
+        class: 'Collection',
+        db, coll, arguments: { d, options }
```
lrlna, line 668: same with the document here: to log or not to log? 🤔

packages/mapper/src/mapper.ts
```diff
@@ -560,19 +699,30 @@ export default class Mapper {
...
     if ('writeConcern' in options) {
       dbOptions.writeConcern = options.writeConcern;
     }
 
+    this.messageBus.emit(
+      'mongosh:api-call',
+      {
+        method: 'insertMany',
+        class: 'Collection',
+        db, coll, arguments: { docs, options }
```
lrlna, line 714: I am keen to say let's not log the very many docs people will be inserting. Keeping it here for a sec, in case someone thinks otherwise.

packages/mapper/src/mapper.ts
```diff
@@ -595,18 +745,30 @@ export default class Mapper {
...
     if ('writeConcern' in options) {
       dbOptions.writeConcern = options.writeConcern;
     }
+
+    this.messageBus.emit(
+      'mongosh:api-call',
+      {
+        method: 'insertOne',
+        class: 'Collection',
+        db, coll, arguments: { doc, options }
```
lrlna, line 760: same as before about `doc`

packages/mapper/src/mapper.ts
```diff
@@ -665,18 +842,22 @@ export default class Mapper {
...
     const dbOptions: any = {};
-    this.messageBus.emit('method:save', collection._name, doc);
+    const db = collection._database._name;
+    const coll = collection._name;
+    this.messageBus.emit(
+      'mongosh:api-call',
+      {
+        method: 'save',
+        class: 'Collection',
+        db, coll, arguments: { doc, options }
```
lrlna, line 852: same as before about `doc`

packages/mapper/src/mapper.ts
```diff
@@ -697,14 +878,24 @@ export default class Mapper {
...
     const dbOptions: any = {};
-    this.messageBus.emit('method:replaceOne', collection._name, filter);
+    const db = collection._database._name;
+    const coll = collection._name;
+    this.messageBus.emit(
+      'mongosh:api-call',
+      {
+        method: 'replaceOne',
+        class: 'Collection',
+        db, coll, arguments: { filter, replacement, options }
```
lrlna, line 888: same as before about `replacement`

packages/mapper/src/mapper.ts
```diff
@@ -728,25 +919,41 @@ export default class Mapper {
...
+
     let result;
-    this.messageBus.emit('method:update', collection._name, filter);
+
     if (options.multi) {
       result = await this.serviceProvider.updateMany(
-        collection._name,
-        collection._database._name,
+        db,
+        coll,
```
lrlna, line 948: ^ so in `serviceProvider` `update` function takes `(db, collection, filter, update, options)`, what we had here before was `(collection, db, filter, update, options)`. So this little change fixes the possible bug with updates!

packages/mapper/src/mapper.ts
```diff
@@ -728,25 +919,41 @@ export default class Mapper {
...
         filter,
         update,
         options,
       );
     } else {
       result = await this.serviceProvider.updateOne(
-        collection._name,
-        collection._database._name,
+        db,
```
lrlna, line 955: `updateOne` also takes `(db, collection, filter, update, options).

packages/mapper/src/mapper.ts
mcasimir: i think the rule should be, emit everything then we decide what to log in the logger.ts and what to send to segment in telemetry.ts or however that will be called.

packages/mapper/src/mapper.ts
mcasimir: i start to see that could happen everywhere :/, In case of some import script we would have log files bigger than the database

packages/mapper/src/mapper.ts
lrlna: I think my concern is that it can be a _very_ large document, or in case of `insertMany`, thousands of documents someone is reading from another file, and I don't want to pollute.

lrlna:
I am confused: when did we say this would be generated in YAML? I don't think I am following.

packages/mapper/src/mapper.ts
lrlna: yes, exactly! that's why I want to be careful o.O

packages/mapper/src/mapper.ts
mcasimir: nice catch

mcasimir:
> I am confused: when did we say this would be generated in YAML? I don't think I am following.

Sorry for the confusion, I'm just using a bit of caution about we being able to pull out the decorator refactoring soon. I mentioned YAML for that reason. If we can't make it we could do the same in the YAML compilation.

PR #116, opened 2020-04-17
telemetry config: enable telemetry by default

lrlna:
## Description

As discussed after a previous demo, set up mongosh CLI config to have telemetry
enabled by default (opt-out).

This PR additonal standardises sessionID to be the same as the log file for
easier access when debugging. SessionID is the logged for user's convenience.,

### DIFF
--- a/packages/cli-repl/src/cli-repl.ts
+++ b/packages/cli-repl/src/cli-repl.ts
@@ -129,7 +129,7 @@ class CliRepl {
   *
   * @returns {string} Status of telemetry logging: disabled/enabled
   */
-  toggleTelemetry(enabled: boolean): void {
+  toggleTelemetry(enabled: boolean): string {
     this.enableTelemetry = enabled;
     this.disableGreetingMessage = true;
 
@@ -137,9 +137,9 @@ class CliRepl {
     this.writeConfigFileSync(configPath);
 
     if (enabled) {
-      console.log(i18n.__('cli-repl.cli-repl.enabledTelemetry'));
+      return i18n.__('cli-repl.cli-repl.enabledTelemetry');
     } else {
-      console.log(i18n.__('cli-repl.cli-repl.disabledTelemetry'));
+      return i18n.__('cli-repl.cli-repl.disabledTelemetry');
     }
   }
 
--- a/packages/cli-repl/src/cli-repl.ts
+++ b/packages/cli-repl/src/cli-repl.ts
@@ -34,6 +34,7 @@ class CliRepl {
   private repl: REPLServer;
   private bus: Nanobus;
   private enableTelemetry: boolean;
+  private disableGreetingMessage: boolean;
   private userId: ObjectId;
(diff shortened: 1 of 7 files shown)


packages/cli-repl/src/cli-repl.ts
```diff
@@ -174,7 +182,7 @@ class CliRepl {
    */
   greet(): void {
     console.log(`Using MongoDB: ${this.mdbVersion} \n`);
-    console.log(TELEMETRY);
+    if (!this.disableGreetingMessage) console.log(TELEMETRY);
```
lrlna, line 185: @mmarcon so I was thinking the telemetry greeting was getting kind of annoying. So I am thinking of hiding once a user either runs `enableTelemetry` or `disableTelemetry`. What do you think? Or would you always prefer it there?

packages/shell-evaluator/src/shell-evaluator.ts
```diff
@@ -78,12 +78,12 @@ class ShellEvaluator {
...
       case 'enableTelemetry':
         if (this.container) {
-          this.container.toggleTelemetry(true);
+          return this.container.toggleTelemetry(true);
         }
         return;
       case 'disableTelemetry':
         if (this.container) {
-          this.container.toggleTelemetry(false);
+          return this.container.toggleTelemetry(false);
```
lrlna, line 86: this needs a return, since `this.container.toggleTelemetry()` returns a string. If we don't return, output is just `undefined`. Please let me know if not having a `return` here was intentional, and I am missing something!

packages/cli-repl/src/cli-repl.ts
mmarcon: How hard would it be to show it only the first time I launch the shell? Essentially, I'd do something like this (tweaking the copy a bit):

**1st time**
```
$ mongosh

Connecting to: mongodb://127.0.0.1:27017
Using MongoDB: 4.2.0

To help improve our products, anonymous usage data is collected and sent to MongoDB periodically (https://www.mongodb.com/legal/privacy-policy).
You can opt-out by running the disableTelemetry() command.

$ mongosh >
```

**Next time**
```
$ mongosh

Connecting to: mongodb://127.0.0.1:27017
Using MongoDB: 4.2.0

$ mongosh >
```
@fredtruman @Sgrinfy happy to get some feedback on the copy.

lrlna: not hard at all. if that's what we want, it's easily manageable.

mmarcon: @lrlna Yes, that's what I would do. If it's not something you can get done quickly before today's internal release no worries. You can merge this PR and we can ticket this.

lrlna: ok, fixed! i have another PR incoming with actual segment implementation that hopefully you can have with today's release as well.

PR #121, opened 2020-04-20
MONGOSH-110: send telemetry events to segment if opted-in

lrlna:
## Description

Sends information to segment if user opted-in. 

API_KEY from segment is to be read from `./config.js` file that evergreen writes when releasing the binary. This PR adds a task to `release.js` to populate that file. Evergreen already has the appropriate key to be written.

This PR also adds a few more events: `mongosh:new-user`, `mongosh-update-user`
and `mongosh:toggle-telemetry`.

Information sent to segment is stripped of possible sensitive information that
could come in queries via `mongodb-redact` module.

Adds documentation on events we listen to in `logger.ts`

## Open Questions

~@durran what's the best possible way for me to get this API key into an
Evergreen build?~

### DIFF
--- a/packages/cli-repl/src/logger.ts
+++ b/packages/cli-repl/src/logger.ts
@@ -61,7 +61,8 @@ export default function logger(bus: any, logDir: string) {
 
   let analytics = new NoopAnalytics();
   try {
-    analytics = new Analytics(process.env.SEGMENT_API_KEY);
+    // this file gets written as a part of a release
+    analytics = new Analytics(require('./config.js'));
   } catch (e) {
     bus.emit('mongosh:error', e)
   }
--- a/packages/cli-repl/src/logger.ts
+++ b/packages/cli-repl/src/logger.ts
@@ -62,9 +62,8 @@ export default function logger(bus: any, logDir: string) {
   let analytics = new NoopAnalytics();
   try {
     // this file gets written as a part of a release
-    console.log(require('./config.js'))
-    log.warn(require('./config.js').SEGMENT_API_KEY)
-    analytics = new Analytics(require('./config.js').SEGMENT_API_KEY);
+    log.warn(require('./analytics-config.js').SEGMENT_API_KEY)
+    analytics = new Analytics(require('./analytics-config.js').SEGMENT_API_KEY);
   } catch (e) {
     bus.emit('mongosh:error', e)
   }
@@ -100,7 +99,7 @@ export default function logger(bus: any, logDir: string) {
   bus.on('mongosh:error', function(error: ErrorEvent) {
     log.error(error);
 
-    if (telemetry) {
+    if (telemetry && error.name.includes('Mongosh')) {
       analytics.track({
         userId,
         event: 'mongosh:error',
@@ -143,17 +142,6 @@ export default function logger(bus: any, logDir: string) {
         userId,
         event: 'mongosh:show',
         properties: { method: args.method }
-      });
-    }
-  });
-
-  bus.on('mongosh:it', function() {
-    log.info('mongosh:it');
-
-    if (telemetry) {
-      analytics.track({
-        userId,
-        event: 'mongosh:it'
       });
     }
   });
--- a/packages/cli-repl/src/logger.ts
+++ b/packages/cli-repl/src/logger.ts
@@ -62,8 +62,9 @@ export default function logger(bus: any, logDir: string) {
   let analytics = new NoopAnalytics();
   try {
     // this file gets written as a part of a release
-    analytics = new Analytics(require('./config.js'));
-    log.warn(require('./config.js'))
+    console.log(require('./config.js'))
+    log.warn(require('./config.js').SEGMENT_API_KEY)
+    analytics = new Analytics(require('./config.js').SEGMENT_API_KEY);
   } catch (e) {
     bus.emit('mongosh:error', e)
   }
--- a/packages/cli-repl/src/logger.ts
+++ b/packages/cli-repl/src/logger.ts
@@ -63,6 +63,7 @@ export default function logger(bus: any, logDir: string) {
   try {
     // this file gets written as a part of a release
     analytics = new Analytics(require('./config.js'));
+    log.warn(require('./config.js'))
   } catch (e) {
     bus.emit('mongosh:error', e)
   }
--- a/packages/cli-repl/src/logger.ts
+++ b/packages/cli-repl/src/logger.ts
@@ -39,7 +39,16 @@ interface ErrorEvent {
 }
 
 interface ConnectEvent {
-  driverUri: string;
+  uri: string;
+  isAtlas: boolean;
+  isLocalhost: boolean;
+  serverVersion: string;
+  isEnterprise: boolean;
+  authType: string;
+  isDataLake: boolean;
+  dlVersion?: string;
+  isGenuine: boolean;
+  serverName: string;
 }
 
 export default function logger(bus: any, logDir: string) {
@@ -58,15 +67,16 @@ export default function logger(bus: any, logDir: string) {
   }
 
   bus.on('mongosh:connect', function(args: ConnectEvent) {
-    const connectionUri = redactPwd(args.driverUri);
-    const params = { sessionID, userId, connectionUri };
-    log.info('connect', params);
+    const connectionUri = redactPwd(args.uri);
+    delete args.uri;
+    const params = { sessionID, userId, connectionUri, ...args };
+    log.info('mongosh:connect', params);
 
     if (telemetry) {
       analytics.track({
         userId,
         event: 'mongosh:connect',
-        properties: { sessionID, connectionUri }
+        properties: { sessionID, connectionUri, ...args }
       });
     }
   });
@@ -112,7 +122,7 @@ export default function logger(bus: any, logDir: string) {
   });
 
   bus.on('mongosh:use', function(args: UseEvent) {
-    log.info(args);
+    log.info('mongosh:use', args);
 
     if (telemetry) {
       analytics.track({
@@ -123,7 +133,7 @@ export default function logger(bus: any, logDir: string) {
   });
 
   bus.on('mongosh:show', function(args: ShowEvent) {
-    log.info(args);
+    log.info('mongosh:show', args);
 
     if (telemetry) {
       analytics.track({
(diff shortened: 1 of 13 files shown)


packages/cli-repl/src/logger.ts
```diff
@@ -1,28 +1,92 @@
...
+
 function logger(bus: any, logDir: string) {
   const sessionID = new ObjectId(Date.now());
   const logDest = path.join(logDir, `${sessionID}_log`);
   const log = pino({ name: 'monogsh' }, pino.destination(logDest));
   console.log(`Current sessionID: ${sessionID}`);
+  let userId;
+  let telemetry;
+
+  let analytics = new NoopAnalytics();
```
lrlna, line 33: use a noop just in case something goes wrong with the definition below. That way the following `analytics` calls don't fail and pollute user with errors.

packages/mapper/src/mapper.ts
```diff
@@ -206,7 +206,7 @@ export default class Mapper {
       {
         method: 'bulkWrite',
         class: 'Collection',
-        db, coll, arguments: { operations, options }
```
lrlna, line 209: `operations` here could be a super large array of documents, so let's omit it entirely from logs.

packages/cli-repl/src/logger.ts
```diff
@@ -31,23 +95,68 @@ function logger(bus: any, logDir: string) {
...
+    if (args.arguments.pipeline) properties.arguments.pipeline = args.arguments.pipeline;
+    if (args.arguments.query) properties.arguments.query = args.arguments.query;
+    if (args.arguments.options) properties.arguments.options = args.arguments.options;
+    if (args.arguments.filter) properties.arguments.filter= args.arguments.filter;
+
+    if (telemetry) {
+      analytics.track({
+        userId,
+        event: 'mongosh:api-call',
+        properties: redactInfo(properties)
```
lrlna, line 150: @mmarcon so these are the telemetry tracking events for api-call.

`properties` would include for example:

```js
{
  method: 'aggregation',
  class: 'Collection',
  arguments: {
    pipeline: [ {$match: {"lat": {$gt: 45}}} ]
    options: { writeConcern: 1 }
  }
```

I am always omitting `db` and `coll` from analytics. 

Depending on the method, arguments could `filter` or `query` instead of `pipeline`. Or there won't be any, if a method doesn't support any. 

This will send all the api-calls to telemetry. If you want, we can be more specific, and send only specific ones over.

mmarcon: @lrlna are we redacting stuff? E.g. if a pipeline or a filter contains email addresses, we should not see them in analytics.

mmarcon: Oh yeah, looks like properties are being redacted. Cool.

lrlna: yep! let me know if something that should be redacted, isn't. Can patch up `mongodb-redacted`

packages/cli-repl/src/logger.ts
```diff
@@ -1,53 +1,189 @@
...
+
+
+  bus.on('mongosh:error', function(error: ErrorEvent) {
     log.error(error);
+
+    if (telemetry) {
+      analytics.track({
+        userId,
+        event: 'mongosh:error',
+        properties: { error }
```
aherlihy, line 107: One of the side effects of how the node repl handles errors weirdly, is that we can send errors that we've thrown to telemetry but skip other errors, right? Since we don't want to send every JS error ever triggered on the shell to telemetry? I'm not 100% clear on what errors are getting sent

packages/shell-evaluator/src/shell-evaluator.ts
```diff
@@ -123,7 +123,7 @@ class ShellEvaluator {
         this.saveState();
         const rewrittenInput = this.asyncWriter.compile(input);
         this.bus.emit(
-          'mongosh:rewrittenAsyncInput',
```
aherlihy, line 126: We could emit a little more intelligently here if you want - we could only emit when the rewritten input does not match the original input. (We'd have to set a flag to babel to not 'prettify' the generated code, or have the async-rewriter indicate that it  has inserted an await, or remove all whitespace before comparison).

As it is, it will emit an event for literally everything that gets run in the shell, so it might not be that useful.

packages/cli-repl/src/logger.ts
```diff
@@ -1,53 +1,189 @@
...
+        event: 'mongosh:api-call',
+        properties: redactInfo(properties)
+      });
+    }
   });
 }
 
-export default logger;
+// set up a noop, in case we are not able to connect to segment.
+function NoopAnalytics() {}
```
mcasimir, line 186: Can't `class` be used here?

packages/cli-repl/src/logger.ts
lrlna: good question! so we currently send everything, because all the errors ever make it to the writer [get logged](https://github.com/mongodb-js/mongosh/blob/master/packages/cli-repl/src/cli-repl.ts#L202). This doesn't include unhandled exceptions, since we don't currently handle those. 

Do you think we should be a bit more specific about what errors we send back? For example, only `MongoshInternalError`? I do agree, I think telemetry will be super polluted if we do everything.

packages/mapper/src/mapper.ts
```diff
@@ -114,20 +106,14 @@ export default class Mapper {
 
     for (let i = 0; i < 20; i++) { // TODO: ensure that assigning cursor doesn't iterate
       if (!await this.currentCursor.hasNext()) {
-        this.messageBus.emit(
-          'mongosh:it',
-          { method: 'it', arguments: { result: 'no cursor' } }
-        );
+        this.messageBus.emit('mongosh:it');
```
mcasimir, line 109: this is going to be emitted 20 times, is it ok?

packages/shell-evaluator/src/shell-evaluator.ts
lrlna: ^ that would be heaps more useful to us and to other users!

scripts/release.js
```diff
@@ -55,13 +56,28 @@ const archive = async() => {
   );
 };
 
+const writeSegmentFile = () => {
+  console.log(process.env.SEGMENT_API_KEY)
```
mcasimir, line 60: Would it be worth adding some prefix to this log like `with SEGMENT_API_KEY = `?

scripts/release.js
```diff
@@ -55,13 +56,28 @@ const archive = async() => {
...
+}
+
 /**
  * Creates dist/mongosh or dist/mongosh.exe and then creates a
  * tarball or zip of the executable as dist/mongosh-${version}-${os}.tgz or
  * zip.
  */
 const release = async() => {
   const artifact = getArtifact();
+  const segmentFile = await writeSegmentFile();
```
mcasimir, line 80: I think this `segmentFile` variable can be removed here, seems unused. Also the `await` is  probably unnecessary at this point.

packages/cli-repl/src/logger.ts
```diff
@@ -1,53 +1,189 @@
...
   const log = pino({ name: 'monogsh' }, pino.destination(logDest));
   console.log(`Current sessionID: ${sessionID}`);
+  let userId;
+  let telemetry;
+
+  let analytics = new NoopAnalytics();
+  try {
+    // this file gets written as a part of a release
+    console.log(require('./config.js'))
+    log.warn(require('./config.js').SEGMENT_API_KEY)
```
mcasimir, line 66: if this whole file is only used for Analytics id give it a more specific name like, `analytics-config.js`

packages/cli-repl/src/logger.ts
```diff
@@ -1,53 +1,189 @@
...
   const logDest = path.join(logDir, `${sessionID}_log`);
   const log = pino({ name: 'monogsh' }, pino.destination(logDest));
   console.log(`Current sessionID: ${sessionID}`);
+  let userId;
+  let telemetry;
+
+  let analytics = new NoopAnalytics();
+  try {
+    // this file gets written as a part of a release
+    console.log(require('./config.js'))
```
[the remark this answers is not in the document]
lrlna: totally on purpose while i am trying to debug the build (i grab the .tgz that this PR produces and run it locally). Will not merge with this in ofc.

packages/cli-repl/src/logger.ts
lrlna: what's the syntax for a noop class? I've only ever written it like this before.

packages/mapper/src/mapper.ts
lrlna: yea, i've been reading over my logs, and it's polluted with these. happy to remove!

scripts/release.js
lrlna: temporary :sob: just trying to see if the logs on evergreen ever get here (they don't)

packages/cli-repl/src/logger.ts
mcasimir: I meant this:

```
class NoopAnalytics {
   identify() {}
   track() {}
}
```

I was wondering if you had some form of typescript issue with that. Just a formality anyway.

lrlna: oh no, nothing to do with typescript! I am just old school and didn't know the new and fancy syntax

lrlna: in an unexpected turn of events, typescript does not, in fact like the class syntax.

packages/cli-repl/src/logger.ts
aherlihy: Yeah I'd say even only Mongosh errors and not generic javascript errors that get written to the logger but not thrown by us.

lrlna: ok, now telemetry only gets Mongosh errors. Local logs get everything tho, because I want to keep an eye on all the js throws in case we have missed something.

lrlna:
I am going to merge it in, and see if evergreen will properly inject this file when on master. If not, I'll debug tomorrow. If it doesn't inject the key properly, master will not be broken and everything will continue functioning as is. Segment just won't get any data, but that's not different from what's happening right now.

packages/cli-repl/src/logger.ts
aherlihy: Perfect, ty!

PR #123, opened 2020-04-20
Add babel type annotations [WIP]

aherlihy:
The type annotations that are going in after #122. TODO: right now it's not compiling because there are cases that aren't being handled, so there is some work to do to ensure that all possible AST nodes are handled.


aherlihy:
Changes included in #127

PR #124, opened 2020-04-21
Collection drop

mcasimir:
Implements `db.coll.drop()`


packages/i18n/src/locales/en_US.js
```diff
@@ -951,6 +951,12 @@ const translations = {
               example: '',
               parameters: {}
             },
+            drop: {
+              link: '',
```
lrlna, line 955: link: https://docs.mongodb.com/manual/reference/method/db.collection.drop
description: Removes a collection or view from the database.
example: db.students.drop()

packages/mapper/src/mapper.integration.spec.ts
```diff
@@ -3,11 +3,21 @@ import { CliServiceProvider } from '@mongosh/service-provider-server';
...
+const mongodbRunnerBefore = require('mongodb-runner/mocha/before');
+const mongodbRunnerAfter = require('mongodb-runner/mocha/after');
+
 describe('Mapper (integration)', function() {
-  this.timeout(10000);
+  this.timeout(60000);
+
+  before(function(done) {
+    try {
+      mongodbRunnerBefore({ port: 27018, timeout: 60000 }).call(this, done);
```
lrlna, line 14: I wonder if wrapping the mongodb-runner/mocha/before in a `try`/`catch` might hide the problem with this module on windows.

packages/mapper/src/mapper.ts
```diff
@@ -1580,4 +1580,37 @@ export default class Mapper {
...
+    );
+
+    try {
+      return await this.serviceProvider.dropCollection(
+        collection._database._name,
+        collection._name
+      );
+    } catch (error) {
+      if (error.codeName === 'NamespaceNotFound') {
+        return false;
```
lrlna, line 1610: we should emit an error for the logger here. Might be useful for people debugging:
```js
this.messageBus.emit('mongosh:error', new Error(`Namespace not found: ${collection._name}`));
```

I am also wondering if we should maybe also return something to user indicating that a collection could not be dropped.

packages/i18n/src/locales/en_US.js
mcasimir: thanks!

PR #130, opened 2020-04-22
Split mapper into multiple classes for different apis

mcasimir:
This is an attempt to prevent database methods and collection methods from clashing in the mapper.

An example of such methods are `runCommand` or `getName`.

## Changes:

1. Extract collection and databases api methods from the `Mapper` class in the respective new classes: `CollectionMapper` and `DatabaseMapper`. 
  The facade for command implementations like `use` or `show` still remains the `Mapper` class.
  The method invocation pattern is exactly the same as before with a `collection` or a `database` argument passed as first parameter by the shell api. Methods are basically moved with a cut and paste.
2. Implements the `db.getCollection()` since is now used also internally.
3. The `Database` constructor in shell-api does not call `new Collection()` directly anymore but instead `wrappee.getCollection(db, name)`.
 
TODO: The tests are converted to the change but they are still all in the same file.


packages/mapper/src/mapper.integration.spec.ts
```diff
@@ -86,7 +86,7 @@ describe('Mapper (integration)', function() {
 
       describe('when calling it after find', () => {
         it('returns next batch of docs', async() => {
-          mapper.find(collection, {}, { _id: 0 });
+          mapper.collectionMapper.find(collection, {}, { _id: 0 });
```
mcasimir, line 89: This is a TODO. I have to convert this tests to separate files and just have `collectionMapper.find`

aherlihy:
Sounds good to me - do you plan on moving tests now or waiting?

mcasimir:
Closed in favor of #133

PR #131, opened 2020-04-22
debug win timeout

mcasimir:
A playground PR to play around and try to get rid of the timeout in windows.


lrlna:
This can be boiled down to a more simplified code with with the `mongodb-runner/mocha/before`. If you up the timeout to 100000, you can get the permissions error. 
14nT6Q13

lrlna:
closing this, sine we no longer have azure pipelines.

PR #132, opened 2020-04-23
Add Shell BSON classes to context

aherlihy:


packages/shell-api/src/shell-bson.js
```diff
@@ -0,0 +1,43 @@
+/* eslint prefer-rest-params: 0 */
+import bson from 'bson';
+
+export default {
+  RegExp: RegExp,
+  DBRef: bson.DBRef,
+  Map: bson.Map,
```
mcasimir, line 7: Wouldn't this clash with the javascript `Map` as a global? I guess that's what happens now in the old shell or?

mmarcon: This is `Map` in mongo:

```
> Map
function() {
        this._data = {};
    }
> Map.prototype
{
	"put" : function(key, value) {
    var o = this._get(key);
    var old = o.value;
    o.value = value;
    return old;
},
	"get" : function(key) {
    return this._get(key).value;
},
	"_get" : function(key) {
    var h = Map.hash(key);
    var a = this._data[h];
    if (!a) {
        a = [];
        this._data[h] = a;
    }
    for (var i = 0; i < a.length; i++) {
        if (friendlyEqual(key, a[i].key)) {
            return a[i];
        }
    }
    var o = {key: key, value: null};
    a.push(o);
    return o;
},
	"values" : function() {
    var all = [];
    for (var k in this._data) {
        this._data[k].forEach(function(z) {
            all.push(z.value);
        });
    }
    return all;
}
}
```

mcasimir: doesn't seem native code, the BSON implementation is a polifill for ES6 Map so we are good:

https://github.com/mongodb/js-bson/blob/master/lib/map.js#L3

PR #135, opened 2020-04-23
Create standardised Mongosh Errors

lrlna:
## Description
Creates four new errors that would be distinctly used in Mongosh. They are an
extension of JavaScript Error object. These error types are then used inside
async-rewriter to provide clearly messaging, since Babel wraps our errors in
their own objects.

This also removes stacktrace from errors in CLI's formatter and leaves only
`error.name` and `error.message` to simplify the output we provide to the user.

We are also adding a `@mongosh/errors` package to unify error objects to be used
throughout the project in the future.

## Example
The following example includes standard JavaScript errors and our own errors in `mongosh`:
Capture d’écran, le 2020-04-23 à 20 06 45


packages/errors/lib/index.d.ts
```diff
@@ -0,0 +1,13 @@
+declare class MongoshRuntimeError extends Error {
```
mcasimir, line 1: I think we forgot to add `lib` to gitignore here?

packages/errors/index.ts
```diff
@@ -0,0 +1,37 @@
+class MongoshRuntimeError extends Error {
+  constructor(...args) {
+    super(...args);
+    this.name = 'MongoshRuntimeError';
+  }
+}
+
+class MongoshInternalError extends Error {
+  constructor(msg, opts?) {
+    super(opts);
```
mcasimir, line 10: `super` is the usual constructor of error, which takes the message. Should be like this:

```
constructor(msg) {
   super(
     msg + 
    '\nThis is an error inside Mongosh. Please file a bug report. ' + 
     'Please include a log file from this session.');
   this.name = 'MongoshInternalError'
}
```

lrlna: from the [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error), it seems to super the second argument.

Capture d’écran, le 2020-04-24 à 14 09 14

mcasimir: I think that example is confusing, but proves that super takes a message as first parameter, just in a super-weird way. If you check better: the first parameter of the CustomError constructor (foo) is not the message and is not passed to `super`. The rest of arguments is.

So when is used we get:

```
 new CustomError(
  'baz',  // this gets assigned to e.foo
  'bazMessage' // this is going to be the first argument of `...params`, 
                             // so the first argument of super() and it populates 
                             //  e.message` since is the message.  
```

An easier example would be this one:

```
class PoliteError extends Error {
  constructor(reason) {
    super(`We are really sorry but something went wrong: ${reason}.`)
  }
}

try {
  throw new PoliteError('life is hard');
} catch (e) {
  console.log(e.message);
}
  
// => We are really sorry but something went wrong: life is hard.
```

mcasimir:
Btw i think i already have a use for this one. In the mapper we have stuff like this:

```
    if (typeof options !== 'object' || Array.isArray(options)) {
      throw new Error('options must be an object');
    }
```

for which would be great to have custom errors!

PR #136, opened 2020-04-24
Rewrite forEach in async-writer [WIP]

aherlihy:
Rewrite forEach calls to wrap the array in `to_iterator`.

TODO:

- [ ] write to_iterator wrapper in shell api
- [ ] update cursor iterator methods


packages/shell-api/src/toIterator.js
```diff
@@ -0,0 +1,23 @@
+class Iterator {
```
aherlihy, line 1: Apparently extending the array class in typescript is quite complicated due to https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes#extending-built-ins-like-error-array-and-map-may-no-longer-work so I did it in javascript.

lrlna: I think this is fine \o/

packages/async-rewriter/src/async-writer-babel.spec.ts
```diff
@@ -929,25 +973,37 @@ describe('async-writer-babel', () => {
...
           writer.compile('fn(db)');
         } catch (e) {
           expect(e.name).to.be.equal('MongoshInvalidInputError');
+          done();
         }
       });
-      it('throws an error for db.coll', () => {
+      it('throws an error for db.coll', (done) => {
         try {
           writer.compile('fn(db.coll)');
```
mcasimir, line 986: forgot to mention last time, i think is nice to add something like: `done(new Error('expected to throw'))` after this line. To avoid a timeout in case the test fails.

aherlihy: Did you approve the PR otherwise?

aherlihy: Cool, I think the done() fixes can be put off for later to save painful rebasing

PR #137, opened 2020-04-24
`db.coll.renameCollection` and `db.coll.runCommand`

mcasimir:
Implements `db.coll.renameCollection` and `db.coll.runCommand`.

Also re-emit few events that were missing.


packages/i18n/src/locales/en_US.js
```diff
@@ -974,6 +974,18 @@ const translations = {
...
+            renameCollection: {
+              link: 'https://docs.mongodb.com/manual/reference/method/db.collection.renameCollection/',
+              description: 'Renames a collection.',
+              example: 'db.coll.renameCollection("newName")',
+              parameters: {}
+            },
+            runCommand: {
+              link: '',
+              description: 'Runs a db command with the given name where the first param is the collection name.',
+              example: '',
```
lrlna, line 986: `help` will look weird without an example. How about something like this: 
```
db.collection.runCommand( "text", { search: "searchKeywords"})
```

packages/mapper/src/mapper.ts
```diff
@@ -542,6 +553,40 @@ export default class Mapper {
...
+      );
+
+      return {
+        ok: 1
+      };
+    } catch (e) {
+      if (e.name === 'MongoError') {
+        return {
+          ok: 0,
+          errmsg: e.errmsg,
```
lrlna, line 580: do you think it's possible to have this be just `message` as a standard error would have? This will not pass the `error` check in CLI's `writer` function, which explicitly checks for `message`

The other question I have, is whether you expect these to be handled differently than regular errors?

packages/mapper/src/mapper.ts
```diff
@@ -1682,10 +1727,63 @@ export default class Mapper {
...
     return `${collection._name}`;
   }
+
+  async collection_runCommand(
+    collection: Collection,
+    commandName: string,
+    options?: Record<string, any>
+  ): Promise<any> {
+    if (typeof commandName !== 'string') {
+      throw new Error('commandName must be a string');
```
lrlna, line 1763: How do you feel about using `assert` here instead of using an Error. Lots of these are look like asserts to me.

packages/mapper/src/mapper.ts
```diff
@@ -1682,10 +1727,63 @@ export default class Mapper {
...
+    }
+
+    if (options && commandName in options) {
+      throw new Error('commandName cannot be passed as an option');
```
lrlna, line 1767: Same here, I feel like these are `assert` territory.

packages/mapper/src/mapper.ts
mcasimir: Here i'm just re-creating the behaviour of the old shell. It looked so weird to me, i'd expect it to throw. 

We should not introduce breaking changes now, rather in a second moment and discuss it case by case, keep it on the back of your mind for now, we'll revisit this later, there is a chance it was done by mistake in the old shell.

packages/mapper/src/mapper.ts
mcasimir: Sounds good to me!

packages/i18n/src/locales/en_US.js
mcasimir: Thanks!

packages/mapper/src/mapper.ts
mcasimir: Going to do that in a separate PR cause i have to fix it in other paces too.

packages/mapper/src/mapper.ts
lrlna: yea, that sounds good to me, we can reorganize it later.

The question that remains is how do you expect to handle `ok:1` and `ok:0` return types in the `writer`? Do we warn the user, do we tell them operation succeeded (ok: 1). Afaik, we have no other return types of this kind (and we will get more of them, I am sure). I just want to make sure we handle them properly. Happy if we do this in another PR though!

mcasimir: Probably a shell api type would do that, but yep definitely another PR, we have plenty to do first, and for now the user is as good as it was in the old shell, so we are fine. 

I'd also bring this kind of topics to @max and go through prioritisation.

PR #138, opened 2020-04-27
Always pass dbOptions

mcasimir:
Preparatory work for `getWriteConcern`, `setWriteConcern`:

1. Add `DatabaseOptions`, `CommandOptions` and related types to `service-provider-core`.
2. Make sure all the methods accepts a `dbOptions` parameter and that `dbOptions` is always passed down to `client.db`.
3. Make sure we set `returnNonCachedInstance: true` as dbOption always

### DIFF
--- a/packages/mapper/src/mapper.ts
+++ b/packages/mapper/src/mapper.ts
@@ -15,7 +15,8 @@ import {
 
 import {
   ServiceProvider,
-  Document
+  Document,
+  DatabaseOptions
 } from '@mongosh/service-provider-core';
 
 import { EventEmitter } from 'events';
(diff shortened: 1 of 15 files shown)


packages/mapper/src/mapper.ts
```diff
@@ -140,32 +141,22 @@ export default class Mapper {
...
     if ('readConcern' in options) {
       dbOptions.readConcern = options.readConcern;
     }
     if ('writeConcern' in options) {
-      dbOptions.writeConcern = options.writeConcern;
+      Object.assign(dbOptions, options.writeConcern);
     }
 
-    let cmd;
-    if (coll === null) {
```
mcasimir, line 153: This would be unreachable code, a collection always have a name and is always passed as parameter by shell api.

packages/mapper/src/mapper.ts
```diff
@@ -140,32 +141,22 @@ export default class Mapper {
...
     const coll = collection._name;
 
-    const dbOptions: any = {};
+    const dbOptions: DatabaseOptions = {};
 
     if ('readConcern' in options) {
       dbOptions.readConcern = options.readConcern;
     }
     if ('writeConcern' in options) {
-      dbOptions.writeConcern = options.writeConcern;
```
mcasimir, line 149: the `writeConcern` option gets filtered out:

https://github.com/mongodb/node-mongodb-native/blob/fa052a514be0a485831fe5e01c3c6eafe2b23140/lib/db.js#L71

we merge `w, j, ...` directly

packages/service-provider-core/src/admin.ts
```diff
@@ -0,0 +1,19 @@
+import Result from "./result";
+
+export default interface Admin {
```
mcasimir, line 3: just extracted this interface as is not related to db, or collection methods

packages/service-provider-server/src/cli-service-provider.ts
```diff
@@ -67,12 +73,12 @@ class CliServiceProvider implements ServiceProvider {
 
   async renameCollection(
     database: string,
-    collection: string,
+    oldName: string,
```
mcasimir, line 76: this is actually a database kind-of method, didn't know how to communicate that, thought that probably renaming this parameter would help.

PR #140, opened 2020-04-27
Revert symbol table state after JS error

aherlihy:


packages/shell-evaluator/src/shell-evaluator.ts
```diff
@@ -112,12 +120,19 @@ class ShellEvaluator {
...
           'mongosh:rewrittenAsyncInput',
           { original: input.trim(), rewritten: rewrittenInput.trim() }
         );
-        return originalEval(rewrittenInput, context, filename);
+        try {
+          return await originalEval(rewrittenInput, context, filename);
+        } catch (err) {
+          // This is for browser/Compass
+          this.revertState();
+          throw err;
```
mcasimir, line 134: To debug it try to throw a different error here (`new Error('revertState was called')`), should appear in the shell output

aherlihy: Ty!

PR #141, opened 2020-04-27
Error on reassigning shell api types

aherlihy:


packages/async-rewriter/src/symbol-table.ts
```diff
@@ -84,13 +85,24 @@ export default class SymbolTable {
     return this.signatures.unknown;
   }
 
+  private checkIfApi(key, item?): void {
+    if (item === undefined) {
+      item = this.lookup(key);
+    }
+    if (item.api) {
+      // waiting on new error types
+      throw new Error(`InvalidInput - cannot reassign API type ${key}`);
```
lrlna, line 94: new errors are merged in ✨

packages/async-rewriter/src/async-writer-babel.spec.ts
```diff
@@ -3137,4 +3137,65 @@ switch(TEST) {
...
+      expect(writer.symbols.scopeAt(0).db).to.deep.equal({
+        api: true, ...signatures.Database
+      });
+    });
+    it('regular add', (done) => {
+      input = 'const db = 1';
+      try {
+        writer.compile(input);
+      } catch (err) {
+        done();
```
mcasimir, line 3155: if is sync just removing done and the try catch would do the same:

```
it('regular add', (done) => {
      input = 'const db = 1';
      writer.compile(input);
 });
```

packages/async-rewriter/src/symbol-table.ts
mcasimir, line 94: FYI: this job can be done without the rewriting when we assign the global context in the shell-evaluator:

```
Object.defineProperty(contextObject, 'db', {
  get: () => mapper.currentDatabase,
  writable: false
});
```

packages/async-rewriter/src/async-writer-babel.spec.ts
aherlihy: But if the error isn't thrown then it wouldn't notice if there's no done() statement, right?  It's basically mocha expect(...).to.throw() except not buggy

packages/async-rewriter/src/symbol-table.ts
aherlihy: lol! You're so right, since we save the state before/after js errors. Thank you!

aherlihy: Hmm, actually, that won't error (it will just prevent it from being changed). I will look into how to make it error

packages/async-rewriter/src/async-writer-babel.spec.ts
mcasimir: Sorry mistyped the example:

```
it('regular add', () => {
      input = 'const db = 1';
      writer.compile(input);
 });
```

This would fail if  anything inside throws and succeed otherwise.

`done` is only needed if you are testing something that is really asynchronous and has a callback, for example:

```
function returnAfterTimeout(ms, callback) {
    setTimeout(() => callback(null), ms);
}

it('returns at some point', (done) => {
   returnAfterTimeout(done);
});
```

aherlihy: But we want the error to be thrown, since there's no "try/catch/else" statement and mocha's expect.to.be.thrown is buggy. How would you ensure that an error is thrown without using done() and without using expect.to.be.thrown?

aherlihy: Perhaps the example will make more sense now that I've updated it with the custom errors that were just merged.

packages/async-rewriter/src/symbol-table.ts
aherlihy: For the moment, until we can figure out the segfault, going to keep the readonly check here

aherlihy: Added!

packages/async-rewriter/src/async-writer-babel.spec.ts
mcasimir: Ohh now i got it, why `expect.to.be.thrown` is buggy?

PR #142, opened 2020-04-28
MONGOSH-104: log connection information

lrlna:
## Description
Adds more detailed information about the current connection:
```
{
  isAtlas: true,
  isLocalhost: false,
  serverVersion: '4.2.6',
  isEnterprise: true,
  authType: 'scram-sha-1',
  isDataLake: false,
  dlVersion: null,
  isGenuine: true,
  serverName: 'mongodb'
}
```
Uses [mongodb-build-info](https://github.com/mongodb-js/mongodb-build-info) to determine whether a connection is `DataLake` or a `GenuineMongoDB`. (there is also [a ticket](https://jira.mongodb.org/browse/COMPASS-4260) to use this module in compass)

Adds three more methods to service-provider-server:
- __buildInfo:__ replacement for `getServerVersion`. Returns the whole of `buildInfo` that we then use to extract connection information from.
- __getTopology:__ returns current topology property on CliServiceProvider's `mongoClient`.
- __getCmdLineOpts:__ returns `cmdLineOpts`.

### DIFF
--- a/packages/cli-repl/src/cli-repl.ts
+++ b/packages/cli-repl/src/cli-repl.ts
@@ -52,6 +52,8 @@ class CliRepl {
     this.serviceProvider = await CliServiceProvider.connect(driverUri, driverOptions);
     this.ShellEvaluator = new ShellEvaluator(this.serviceProvider, this.bus, this);
     this.mdbVersion = await this.serviceProvider.getServerVersion();
+    const connectionInfo = this.getConnectionInfo(driverUri, driverOptions);
+    this.bus.emit('connect', connectionInfo);
     this.start();
   }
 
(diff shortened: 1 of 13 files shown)


packages/cli-repl/src/cli-repl.ts
```diff
@@ -51,7 +53,28 @@ class CliRepl {
 
     this.serviceProvider = await CliServiceProvider.connect(driverUri, driverOptions);
     this.ShellEvaluator = new ShellEvaluator(this.serviceProvider, this.bus, this);
-    this.mdbVersion = await this.serviceProvider.getServerVersion();
+    this.buildInfo = await this.serviceProvider.buildInfo();
+
+    let cmdLineOpts = null;
```
mcasimir, line 58: I'd really extract this whole new block in a method.

lrlna: Wouldn't I have to wrap the new block in another try/catch? Or do you mean an async block that returns object or null?

lrlna: ah nvm!

PR #144, opened 2020-04-28
Add recursive tests

aherlihy:


packages/async-rewriter/src/async-writer-babel.spec.ts
```diff
@@ -3368,4 +3368,87 @@ switch(TEST) {
...
+    describe('async', () => {
+      before(() => {
+        spy = sinon.spy(new SymbolTable([{ db: signatures.Database }, {}], signatures));
+        writer = new AsyncWriter(signatures, spy);
+        output = writer.compile(`
+function f(arg) {
+  if (arg === 'basecase') {
+    return db.coll.insertOne({});
+  }
+  return f();
```
aherlihy, line 3411: Do we expect an await to be added here?

packages/async-rewriter/src/async-writer-babel.spec.ts
```diff
@@ -3368,4 +3368,87 @@ switch(TEST) {
...
+}
+  `);
+      });
+      it('compiles correctly', () => {
+        expect(output).to.equal(`async function f(arg) {
+  if (arg === 'basecase') {
+    return await db.coll.insertOne({});
+  }
+
+  return f();
```
aherlihy, line 3421: Do we expect an await to be added here?

mcasimir: I think if is not tail recursive would probably be necessary. Not sure about that case specifically too.

PR #146, opened 2020-04-28
Support for 'this'

aherlihy:


packages/async-rewriter/src/async-writer-babel.ts
```diff
@@ -106,8 +106,23 @@ var TypeInferenceVisitor: Visitor = { /* eslint no-var:0 */
             throw new MongoshInvalidInputError(`Cannot access Mongosh API types dynamically. ${help}`);
           }
           path.node['shellType'] = state.symbols.signatures.unknown;
+          debug(`MemberExpression: { object.sType: ${lhsType.type}, property.name: ${rhs} }`, path.node['shellType']);
           return;
       }
+      if (path.node.object.type === 'ThisExpression') {
+        const classPath = path.findParent((p) => p.isClassDeclaration());
+        if (!classPath) {
+          throw new MongoshUnimplementedError('Unable to handle this outside of method definition of class declaration');
```
lrlna, line 115: I'd put "this" in quotation marks. I had to read the sentence a couple of times to understand what you mean.

packages/async-rewriter/src/async-writer-babel.ts
```diff
@@ -257,6 +274,15 @@ var TypeInferenceVisitor: Visitor = { /* eslint no-var:0 */
             const tyTS = lhsNode as unknown;
             const id = tyTS as babel.types.Identifier;
             state.symbols.updateAttribute(id.name, attrs, sType);
+          } else if (lhsNode.type === 'ThisExpression') {
+            const classPath = path.findParent((p) => p.isClassDeclaration());
+            if (!classPath) {
+              throw new MongoshUnimplementedError('Unable to handle this outside of method definition of class declaration');
+            }
+            if (attrs.length > 1) {
+              throw new MongoshUnimplementedError('Unable to handle multi-layered assignment to \'this\'');
```
lrlna, line 283: ^ yea something like your above statement for all the mentions of 'this' in errors would be really good.

packages/async-rewriter/src/async-writer-babel.spec.ts
```diff
@@ -2407,34 +2407,67 @@ function f() {
...
+      try {
+        writer.compile(input);
+      } catch (e) {
+        expect(e.name).to.be.equal('MongoshUnimplementedError');
+        done();
+      }
     });
-    describe('adds methods to class', () => {
+    it('errors in regular function', (done) => {
+      input = 'function x() { this.x = 1 }';
```
lrlna, line 2424: hmmm what if people instantiate classes like so:

```
function cat (name) {
  if (!(this instanceof Cat)) return new Cat(name)
  this._name = name
}
```
are we also saying `MongodbUnimplementedError` for something like the above?

aherlihy: That function is on it's own right, not inside of a class declaration? Yeah it would be turned off since we can't do any inference on this at all. So the keyword would only be allowed within a method of a class. It's a bit strict but I think it's an edge case anyway because I don't anticipate people will be writing classes much with the shell anyway. If the error gets reported a lot via telemetry or feature requests we can try to revisit it

packages/async-rewriter/src/async-writer-babel.ts
aherlihy: Cool, I replaced each `this` with `'this' keyword` in the error messages

packages/async-rewriter/src/async-writer-babel.ts
aherlihy: Done!

packages/async-rewriter/src/async-writer-babel.spec.ts
lrlna: it is on it's own, but that's how you'd write 'Classes' pre es6. I _feel_ like people are more likely to do this (it's quicker) than the actual es6 class declaration. But I could be wrong! This would be cool to monitor and see if lots of people hit this error!

PR #147, opened 2020-04-29
MONGOSH-183: add segment api key on evergreen build

lrlna:
## Description

Adds a script to write segment's API key to `process.env` when releasing.


lrlna:
I am going to close this PR and bring it into #121 since that one is not merged in yet. It's a bit hard to test this without those changes in master.

PR #150, opened 2020-04-29
implements `explainable.find` and `.aggregate`

mcasimir:
Implements `explainable.find()` and `explainable.aggregate()`.

The way the cursor is reused is not super pretty: we just rewrite the `toReplString` and `shellApiType`, perhaps 2 types proxying the cursor methods would be better.

### DIFF
--- a/packages/mapper/src/mapper.ts
+++ b/packages/mapper/src/mapper.ts
@@ -450,7 +450,7 @@ export default class Mapper {
    *
    * @returns {Cursor} The promise of the cursor.
    */
-  collection_find(collection, query, projection): any {
+  collection_find(collection, query, projection): Cursor {
     const options: any = {};
     const db = collection._database._name;
     const coll = collection._name;
@@ -468,12 +468,14 @@ export default class Mapper {
       }
     );
 
-    this.currentCursor = new Cursor(
+    const cursor = new Cursor(
       this,
       this.serviceProvider.find(db, coll, query, options)
     );
 
-    return this.currentCursor;
+    this.currentCursor = cursor;
+
+    return cursor;
   }
 
   /**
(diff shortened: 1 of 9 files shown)


packages/mapper/src/mapper.ts
```diff
@@ -1808,19 +1810,75 @@ export default class Mapper {
...
+        class: 'Explainable',
+        db: explainable._collection._database._name,
+        coll: explainable._collection._name,
+        arguments: methodArguments
+      }
+    );
+  }
+
   explainable_getCollection(explainable: Explainable): Collection {
+    this._emitExplainableApiCall(explainable, 'getCollection');
```
mcasimir, line 1842: @lrlna i'd like to do the same for collection and db methods, is much more readable like this. What do you think? do you see any issue with this?

lrlna: as long as you don't just pass on the `...arguments` from the emitting function, I think it will be ok.

mcasimir: That should be fine, should be easier also to review.

mcasimir: I hope

packages/service-provider-server/src/node/node-cursor.ts
```diff
@@ -377,6 +377,30 @@ class NodeCursor implements Cursor {
     this.cursor.addCursorFlag(flag, true);
     return this;
   }
+
+  async explain(verbosity: string): Promise<any> {
```
lrlna, line 381: I think if you want to merge it in as is, I'd definitely add a set of comments as to why `explain` cursor needs to be done this way. I think if we were to forget in two months that explain does not get properly done in the node driver, this function will be a bit perplexing.

mcasimir: The weird part i was mentioning is in `explainable_find()` and `explainable_aggregate()`. This part has to stay even after the driver implements explain everywhere :). I can add a comment about the driver returning the full explain, which is intended though, and about the fact we probably should move the adaptations we do here in the mapper and just use the native cursor in this layer since it already implements the interface we need.

FYI. The problem i was talking about, with the node driver, is that `explain` is only implemented for cursors, while we need it for remove / update etc too. Also the output with `aggregate` is incomplete, but it makes no difference here since we have to handle all the cases anyway.

PR #153, opened 2020-04-30
Recover from multiline errors

lrlna:
## Description

When running eval in `try`/`catch`, test if the catchable error returns a recoverable error. Uses [is-recoverable-error](github.com/lrlna/is-recoverable-error) (which is a standalone version node's isRecoverableError) to do the check. 

## Open Questions
~There is something off with `pretty-repl` as it runs into a TypeError when using `repl.Recoverable()`, so this PR introduces a regression to highlighted input by temporarily removing `pretty-repl`. If we think having this is more important than highlighted input, let's get this in and I will look into `pretty-repl` implementation to fix the `TypeError` next.~  __nvm! fixed__

If y'all have more examples of what should be recoverable and what shouldn't, I can add more tests to the `is-recoverable-error` package.


packages/cli-repl/src/cli-repl.ts
```diff
@@ -262,22 +265,36 @@ class CliRepl {
...
+        if (isRecoverableError(input)) {
+          return callback(new Recoverable(err));
+        } else {
+          result = err;
+        }
       }
+      callback (null, result)
     };
 
+    // function isRecoverableError(error) {
```
aherlihy, line 283: Any reason to keep commented function here?

mcasimir:
Thanks for this! Is pretty simple to add a test for this in cli-repl e2e, i think we should at least add one:

```
it('runs multiline input', async () => {
      shell.stdio.stdin.write('"this is a multi\nline string"\n');

      await eventually(() => {
        expect(shell.stdio.stderr).to.be.empty;
        expect(shell.stdio.stdout).to.contain('this is a multiline string');
      });
    });
```

should fail without your magic.

A super small thing i'd notice, i'd expect `isRecoverableError` to check if an instance of an error is recoverable. We could rename it to something more true to what it really does, like `throwsRecoverableError(code)`.

packages/cli-repl/src/cli-repl.ts
lrlna: i forgot, thanks!

lrlna:
@aherlihy i figured we might not want to have this silly of a package under mongodb-js. Already transferred it over! I honestly tried making it part of mongosh, but it's was a pain in the ass 1) to have `compile-ts` compile a regular cjs package and 2) this package did not work with esm so I couldn't get it to work in typescript. Plus maybe someone else might use it if it's just on its own or the node org might consider exporting their own `isRecoverableError` function.

@mcasimir `isRecoverableError` very specifically doesn't throw anything because all of our errors get handled in writer. We _do not_ want to throw anything in the eval. In fact, we return the error as the result if `isRecoverableError` is false, which is what our writer expects. This is also what happens in node. 

`isRecoverableError` looks at the input, and tries running the input through to check if we might potentially get a SyntaxError that could be recoverable. It's literally just node's code that they don't export at the moment, and we can't take advantage of because we are not using their eval entirely.

packages/cli-repl/test/e2e.spec.ts
```diff
@@ -55,6 +55,41 @@ describe('e2e', function() {
       });
     });
 
+    it.only('throws multiline input with a single line string', async () => {
+      // this is an unterminated string constant and should throw, since it does
+      // not pass: https://www.ecma-international.org/ecma-262/#sec-line-terminators
+      shell.stdio.stdin.write('"this is a multi\nline string"\n');
```
lrlna, line 61: @mcasimir I added some e2e tests, but the test you suggested is actually meant to not be recoverable as per the ecma spec in the comment there. I added some other ones that are meant to be recovered.

packages/cli-repl/package.json
```diff
@@ -30,13 +30,20 @@
...
     "@mongosh/shell-api": "^0.0.1-alpha.13",
     "@mongosh/shell-evaluator": "^0.0.1-alpha.13",
+    "acorn": "^7.1.1",
+    "acorn-class-fields": "^0.3.2",
+    "acorn-numeric-separator": "^0.3.0",
+    "acorn-private-methods": "^0.3.1",
+    "acorn-static-class-features": "^0.2.1",
     "ansi-escape-sequences": "^5.1.2",
     "bson": "^4.0.4",
+    "is-recoverable-error": "^1.0.0",
```
aherlihy, line 46: this should be @mongodb-js/is-recoverable-error right? Otherwise LGTM!

lrlna: it's not scoped, it just stayed as `is-recoverable-error`, but the package is here: https://github.com/mongodb-js/is-recoverable-error. Otherwise we have to go through a very painful process of deprecating it and creating a new package. Lots of our utils in mongodb-js are not scoped on github (for example [hadron-app-registry](https://www.npmjs.com/package/hadron-app-registry), or even [bson-transpilers](https://www.npmjs.com/package/bson-transpilers)), so I think it should be ok!

mcasimir:
> `isRecoverableError` very specifically doesn't throw anything because all of our errors get handled in writer. We _do not_ want to throw anything in the eval. In fact, we return the error as the result if `isRecoverableError` is false, which is what our writer expects. This is also what happens in node.

I know what it does, i find the name confusing cause the argument of `isRecoverableError` is not an error. I would find either `isRecoverable(code)` or `wouldThrowARecoverableError(code)` less confusing, but is a minor issue.

> i figured we might not want to have this silly of a package under mongodb-js. Already transferred it over! I honestly tried making it part of mongosh

I agree quite firmly with @aherlihy here. You can create pure js packages in mongosh and even have a plain js file in the cli-repl itself, i really would like to have that code where everyone can have access and change it without doing a PR to an external repo. If a time will come where we'd need to extract it we will do.

packages/cli-repl/test/e2e.spec.ts
mcasimir: sounds good!

packages/cli-repl/test/e2e.spec.ts
```diff
@@ -55,6 +55,40 @@ describe('e2e', function() {
...
+    it('runs an unterminated function', async () => {
+      shell.stdio.stdin.write('function x () {\nconsole.log(\'y\')\n }');
+
+      await eventually(() => {
+        expect(shell.stdio.stderr).to.be.empty;
+      });
+    });
+
+    it('runs an unterminated function', async () => {
+      shell.stdio.stdin.write('function x () {');
```
mcasimir, line 85: i think you need the `\n` terminating it otherwise it won't get executed.

packages/cli-repl/package.json
aherlihy: fair enough! It's definitely not worth the pain. LGTM!

lrlna:
> I agree quite firmly with @aherlihy here. You can create pure js packages in mongosh and even have a plain js file in the cli-repl itself, i really would like to have that code where everyone can have access and change it without doing a PR to an external repo. If a time will come where we'd need to extract it we will do.

@mcasimir it's already [part of mongodb-js](github.com/mongodb-js/is-recoverable-error) organization. All of us have `owner` access and you can make changes any time you like. I hope I can just deprecate in a bit once node makes that function exportable.

packages/cli-repl/package.json
lrlna: hah all good!

yep, directly copy pasted, the only thing I added were tests! I have added the cases you mentioned, but if you have any other ones you can think of, let me know.

mcasimir:
Would you mind if we move it as an internal package in `mongosh`? Is going to be the same than having it in a different repo, still published independently, but with the added benefit of having the code close, having the integration immediately tested on changes and we can work on that without having to link.

lrlna:
@mcasimir y'all had no problem bringing in pretty-repl as a package that's not part of mongosh. I don't understand why this particular one is turning out to be such a big deal.  Also last time I've mentioned having a package that is not a typescript package in this repo, you've mentioned you expected this package to have typescript signatures when compiled. This package will not work when compiled like that. I've spent almost two days trying to get it to work like that. Which is why it's in its own repo so we can avoid the extra complications. Plus it's a nice reusable library other people can use, 'cause I am sure we are not the only people in this boat.

mcasimir:
@lrlna not a big deal at all! Is just a feedback like any other, but if is problematic to move that code in it is just fine!

PR #154, opened 2020-05-04
Add rest of BSON types

aherlihy:
Also includes fix for Date with no argument failing.


packages/shell-api/src/shell-bson.ts
```diff
@@ -0,0 +1,71 @@
+/* eslint prefer-rest-params: 0, @typescript-eslint/ban-ts-ignore: 0 */
+import bson from 'bson';
+
+function dateHelper(...args: any[]): Date {
```
mcasimir, line 4:
```
type DateConstructorArguments = [any, any, ...any[]];

function dateHelper(...args: DateConstructorArguments): Date {
  if (args.length === 0) {
    return new Date();
  }
  if (args.length === 1) {
    return new Date(args[0]);
  }

  return new Date(Date.UTC(...args));
}
```

packages/shell-api/src/shell-bson.ts
```diff
@@ -0,0 +1,71 @@
...
+  NumberInt: function(s): any {
+    return parseInt(s, 10);
+  },
+  NumberLong: function(v): bson.Long {
+    if (v === undefined) {
+      v = 0;
+    }
+    return bson.Long.fromNumber(v);
+  },
+  Date: function(): Date|string {
```
mcasimir, line 42: `arguments` is still there but is somehow deprecated: 
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Deprecated_and_obsolete_features
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments

```
  Date: function(...args: DateConstructorArguments): Date|string {
    const date = dateHelper(...args);
```

packages/shell-api/src/shell-bson.ts
```diff
@@ -0,0 +1,71 @@
...
+    const date = dateHelper(...arguments);
+    if (new.target) {
+      return date;
+    }
+    return date.toString();
+  },
+  ISODate: function(): Date|string {
```
mcasimir, line 49:
```
  ISODate: function(...args: DateConstructorArguments): Date|string {
    return dateHelper(...args);
```

packages/shell-api/src/shell-bson.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+    it('returns string without new', () => {
+      expect(shellBson.Date()).to.be.a('string');
+    });
+    it('returns now object with new', () => {
+      // @ts-ignore
+      const date = new shellBson.Date();
+      const cDate = new Date();
+      expect(typeof date).to.equal('object');
+      // @ts-ignore
+      expect(date.getYear()).to.equal(cDate.getYear());
```
mcasimir, line 27: `getYear` is deprecated, not sure if that is necessary for this test, but you can replace it with `getFullYear`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Deprecated_and_obsolete_features

packages/shell-api/src/shell-bson.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+      expect(code.scope).to.deep.equal({ k: 'v' });
+    });
+  });
+  describe('Date', () => {
+    it('returns string without new', () => {
+      expect(shellBson.Date()).to.be.a('string');
+    });
+    it('returns now object with new', () => {
+      // @ts-ignore
+      const date = new shellBson.Date();
```
mcasimir, line 23: `any` is usually more kind than ts-ignore: `new (shellBson.Date as any)`. There are more elegant ways to define `Date` as a constructor in `shell-bson`.

packages/shell-api/src/shell-bson.ts
aherlihy: Happy to make the change, but can you explain why that's better than just any[]? It is still failing the typescript compilation with that chage.
Screen Shot 2020-05-05 at 14 12 35

mcasimir: that worked for me, let me pull again. `any[]` does not say anything about the cardinality of the array, which means it would match an empty array for example.

`[any, any, ...any[]]` says at least 2 elements has to be there.

packages/shell-api/src/shell-bson.spec.ts
aherlihy: Unfortunately `as DateConstructor` does not seem to be directly castable
Screen Shot 2020-05-05 at 14 19 47

packages/shell-api/src/shell-bson.ts
aherlihy: Neat trick!

mcasimir: This works for me: https://github.com/mongodb-js/mongosh/blob/bson-fix-test-ts/packages/shell-api/src/shell-bson.ts

PR #157, opened 2020-05-05
db.aggregate and aggregate with explain

mcasimir:
- Implements `db.aggregate` and add support for `{explain: true}` in `db.coll.aggregate`.

Changes:
- Implements `db.aggregate`
- Add tests for `db.coll.aggregate`
- Refactor the common code of  `mapper.database_aggregate` and `mapper.collection_aggregate` in a private method.
- Refactor the code to make a cursor explainable into a private method.
- Add all the methods to the `Cursor` interface so it can be stubbed properly with `stubInterface`.

**NOTE**:  `Explainable.aggregate`, `Collection.aggregate([...], {explain: true})` and `Database.aggregate([...], {explain: true})` are currently returning an `AggregationCursor` that turns into an explain once evaluated. That is not what happens in the old shell, but i'd keep it for a follow up PR as this one is quite large already.

### DIFF
--- a/packages/mapper/src/mapper.ts
+++ b/packages/mapper/src/mapper.ts
@@ -17,7 +17,8 @@ import {
 import {
   ServiceProvider,
   Document,
-  DatabaseOptions
+  DatabaseOptions,
+  Cursor as ServiceProviderCursor
 } from '@mongosh/service-provider-core';
 
 import { EventEmitter } from 'events';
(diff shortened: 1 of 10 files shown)


packages/mapper/src/mapper.ts
```diff
@@ -1861,29 +1929,41 @@ export default class Mapper {
     explainable._verbosity = verbosity;
   }
 
-  explainable_find(explainable: Explainable, query?: any, projection?: any): any {
-    this._emitExplainableApiCall(explainable, 'find', { query, projection });
-
-    // TODO: turn ExplainableCursor in a proper type.
-    const cursor = explainable._collection.find(query, projection);
-    cursor.shellApiType = (): string => 'ExplainableCursor';
+  // TODO: turn this into proper types and constructors in shell api.
```
aherlihy, line 1932: This is done right? Or are there other types

mcasimir: Want to do that in a separate PR, i have to fix the output anyway. We probably need another proxy `new ExplainableCursor(cursor, verbosity)`, for which  we keep methods in sync with the original cursor, was thinking to use anchors in YAML for that.

PR #160, opened 2020-05-07
MONGOSH-186: move show collections formatting to implementors

lrlna:
## Description
`show collections` should be formatted in implementors instead of the mapper.
This PR removes `.join('\n')` out of the mapper.


lrlna:
@mcasimir ah! i didn't know that! Let me ping Lucy and double check with her.

kornilova203:
I think `show collections` should return something like `[{name: 'coll1'}, ...]` because `show databases` has a similar format. It would feel more consistent to me. But I'm not 100% sure and it's not critical

lrlna:
@kornilova203 ah ok. I'll maybe merge this in for now, and then switch it up for a different object. 

I don't know if you're using the ShellApiTypes, but this also returns a different type, `ShowCollectionsResult`, in case that's easier for you to match on

PR #161, opened 2020-05-07
touch up mongosh greeting

lrlna:
## Description

This PR adds a few touch ups to the CLI UI:
- an extra message about the wiki on startup
- extra highlighting to the currently connected url
- changes prompt to `msh >` since `$` is not usually assigned to REPLs and `mongosh` takes up a lot of valuable space.

There is also a fix for `undefined` result. If `result.value` is of type undefined, we will return a new line. 

Capture d’écran, le 2020-05-07 à 16 43 44

### DIFF
--- a/packages/cli-repl/src/cli-repl.ts
+++ b/packages/cli-repl/src/cli-repl.ts
@@ -260,8 +260,7 @@ class CliRepl {
     const version = this.buildInfo.version;
 
     this.repl = repl.start({
-      prompt: `$ mongosh > `,
-      ignoreUndefined: true,
+      prompt: `msh > `,
       writer: this.writer,
       completer: completer.bind(null, version),
     });
--- a/packages/cli-repl/src/cli-repl.ts
+++ b/packages/cli-repl/src/cli-repl.ts
@@ -5,7 +5,7 @@ import { MongoshWarning } from '@mongosh/errors';
 import { changeHistory } from '@mongosh/history';
 import getConnectInfo from './connect-info';
 import formatOutput from './format-output';
-import { TELEMETRY } from './constants';
+import { TELEMETRY, MONGOSH_WIKI } from './constants';
 import CliOptions from './cli-options';
 import completer from './completer';
 import { REPLServer, Recoverable } from 'repl';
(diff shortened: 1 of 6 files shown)


packages/cli-repl/src/cli-repl.ts
```diff
@@ -259,8 +260,7 @@ class CliRepl {
     const version = this.buildInfo.version;
 
     this.repl = repl.start({
-      prompt: `$ mongosh > `,
-      ignoreUndefined: true,
+      prompt: `msh > `,
```
lrlna, line 263: @mmarcon what do you think about this more simplified prompt?

I find that when i am in a more collapsed terminal view the full `mongosh` takes up a lot of space and I wouldn't mind having a few characters back.

mcasimir, line 263: Is even just `>` possible?

aherlihy: There as a plan to include topology information in the prompt, which would be easy to implement but we haven't gotten around to it. For example, on the old shell indicates if it's a replset/etc and what the version of the server is, like `MongoDB Enterprise >`. We should make a ticket to track that and save it for post-beta?

aherlihy: There as a plan to include topology information in the prompt, which would be easy to implement but we haven't gotten around to it. For example, on the old shell indicates if it's a replset/etc and what the version of the server is, like MongoDB Enterprise >. We should make a ticket to track that and save it for post-beta?

lrlna: i will strongly prefer `>` unless we display replica set name. So if @mmarcon agrees, I will gladly just change it to `>`

lrlna: ahhh that's a good point. We don't have that ticket yet, but that would be useful. I just felt like just having `mongosh` was giving me nothing useful.

mmarcon: Sounds good to me. `mongo` has the same.

aherlihy: I made a ticket here: https://jira.mongodb.org/browse/MONGOSH-196

For the moment just `>` seems fine to me.

PR #165, opened 2020-05-09
Made the prompt just '>'

mmarcon:
As discussed in #161, let's make the prompt just `>`.


lrlna:
Ah sorry we merged 161 before I had the chance to fix it! Thanks for this.

PR #166, opened 2020-05-11
MONGOSH-177: handle unhandled rejections

lrlna:
## Description

There are certain asynchronous functions where we don't properly handle promise rejections. These are mainly done on startup: `connect` and `buildInfo`. This PR wraps them all in a `try`/`catch` to display the error to the user properly, and terminate the process when necessary. 

__Before:__
Capture d’écran, le 2020-04-27 à 17 25 50

__After:__
Capture d’écran, le 2020-05-09 à 19 29 09

## Open Questions
I've tried to get errors with a whole bunch of our other service-provider-* methods, and it seems like everything else gets handled well. This is due to the fact that our API methods are part of the mapper, and the result from the mapper gets handled in the writer, which explicitly checks for `result.message && result.stack`. This allows us to always format the incoming error and return it gracefully. 

I might not have seen all the unhandled `await`s. If you spotted something before that also ought to be handled, please let me know. 

## Notes
I've rearranged some methods in cli-repl, which is why the PR is a tad larger.

### DIFF
--- a/packages/cli-repl/src/cli-repl.ts
+++ b/packages/cli-repl/src/cli-repl.ts
@@ -1,14 +1,14 @@
 import { CliServiceProvider, NodeOptions } from '@mongosh/service-provider-server';
+import formatOutput, { formatError } from './format-output';
 import ShellEvaluator from '@mongosh/shell-evaluator';
 import isRecoverableError from 'is-recoverable-error';
 import { MongoshWarning } from '@mongosh/errors';
 import { changeHistory } from '@mongosh/history';
+import { REPLServer, Recoverable } from 'repl';
 import getConnectInfo from './connect-info';
-import formatOutput from './format-output';
 import { TELEMETRY, MONGOSH_WIKI } from './constants';
 import CliOptions from './cli-options';
 import completer from './completer';
-import { REPLServer, Recoverable } from 'repl';
 import i18n from '@mongosh/i18n';
 import { ObjectId } from 'bson';
 import repl from 'pretty-repl';
@@ -49,10 +49,17 @@ class CliRepl {
    * @param {string} driverUrl - The driver URI.
    * @param {NodeOptions} driverOptions - The driver options.
    */
-  async connect(driverUri: string, driverOptions: NodeOptions): Promise<void> {
+  async connect(driverUri: string, driverOptions: NodeOptions): Promise<any> {
     console.log(i18n.__(CONNECTING), clr(redactPwd(driverUri), ['bold', 'green']));
 
-    this.serviceProvider = await CliServiceProvider.connect(driverUri, driverOptions);
+    // sometimes we get a timeout when connecting. Make sure promise rejection
+    // is handled and formatted properly.
+    try {
+      this.serviceProvider = await CliServiceProvider.connect(driverUri, driverOptions);
+    } catch (e) {
+      return console.log(formatError(e));
+    }
+
     this.ShellEvaluator = new ShellEvaluator(this.serviceProvider, this.bus, this);
     this.buildInfo = await this.serviceProvider.buildInfo();
     const cmdLineOpts = await this.getCmdLineOpts();
@@ -200,7 +207,6 @@ class CliRepl {
     // This checks for error instances.
     // The writer gets called immediately by the internal `this.repl.eval`
     // in case of errors.
-    // console.log('result', result)
     if (result && result.message && typeof result.stack === 'string') {
       this.bus.emit('mongosh:error', result);
       this.ShellEvaluator.revertState();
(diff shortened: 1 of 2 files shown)


packages/cli-repl/src/cli-repl.ts
```diff
@@ -44,17 +44,37 @@ class CliRepl {
...
+
+    this.bus = new Nanobus('mongosh');
+    logger(this.bus, this.mongoshDir);
+
+    this.generateOrReadTelemetryConfig();
+
+    if (this.isPasswordMissing(driverOptions)) {
+      this.requirePassword(driverUri, driverOptions);
+    } else {
+      this.setupRepl(driverUri, driverOptions);
```
mcasimir, line 63: technically this one has still the potential to generate an unhandled rejection, i see you've got a try catch around the body of the internal calls to connect, getBuild etc.., but to be really sure we should add a catch here:

```
this.setupRepl(driverUri, driverOptions).catch((error) => {
  // ...
   console.log(formatError(error))
});
```

packages/cli-repl/src/cli-repl.ts
```diff
@@ -44,17 +44,37 @@ class CliRepl {
...
    */
-  async connect(driverUri: string, driverOptions: NodeOptions): Promise<void> {
-    console.log(i18n.__(CONNECTING), clr(redactPwd(driverUri), ['bold', 'green']));
-
-    this.serviceProvider = await CliServiceProvider.connect(driverUri, driverOptions);
+  async setupRepl(driverUri: string, driverOptions: NodeOptions): Promise<void> {
+    this.serviceProvider = await this.connect(driverUri, driverOptions);
     this.ShellEvaluator = new ShellEvaluator(this.serviceProvider, this.bus, this);
-    this.buildInfo = await this.serviceProvider.buildInfo();
+    this.buildInfo = await this.getBuild();
```
mcasimir, line 77: why not `getBuildInfo`?

packages/cli-repl/src/cli-repl.ts
```diff
@@ -98,11 +177,29 @@ class CliRepl {
...
   }
 
+  /**
+   * run buildInfo() command to get mongodbversion + other build info needed for
+   * logging
+   */
+  async getBuild(): Promise<any> {
+    // this could potentially error, so wrap it in a try/catch and format the
+    // returned error
+    try {
```
mcasimir, line 192: Putting the `.catch` on `setupRepl` we can spare this try / catch, so the execution flow is interrupted where an error occur, otherwise it would continue.

packages/cli-repl/src/cli-repl.ts
```diff
@@ -44,17 +44,37 @@ class CliRepl {
...
+   * information, and finally start the repl.
    *
    * @param {string} driverUrl - The driver URI.
    * @param {NodeOptions} driverOptions - The driver options.
    */
-  async connect(driverUri: string, driverOptions: NodeOptions): Promise<void> {
-    console.log(i18n.__(CONNECTING), clr(redactPwd(driverUri), ['bold', 'green']));
-
-    this.serviceProvider = await CliServiceProvider.connect(driverUri, driverOptions);
+  async setupRepl(driverUri: string, driverOptions: NodeOptions): Promise<void> {
```
mcasimir, line 74: I like putting this here! :)

packages/cli-repl/src/cli-repl.ts
lrlna: I am not super familiar with this syntax: does doing a `.catch()` here mean we don't have to wrap internal `await` functions in a `try`/`catch`?

packages/cli-repl/src/cli-repl.ts
lrlna: i thought it'd be confusing, because there is another function (not on `this` scope though) that's already called `getBuildInfo`.

Although, I could refactor that other function to be `formatBuildInfo`, since it plucks out and returns only the necessary information from `buildInfo` and `getCmdLineOpts` functions.

lrlna: oh nvm, i thought there was a function, but there isn't one. I'll rename this to `getBuildInfo`

packages/cli-repl/src/cli-repl.ts
mcasimir: yep, it acts on the promise itself.

`catch` is a method of the promise that is used to setup a callback for when the promise is rejected;

```
async function fn() {
  throw new Error('err')
}

fn().catch((error) => {
  console.error('Got error:', error);
});
```

lrlna: cool, yea that's what I did. All the `await` methods are getting handled in the same way, so this saves me a bunch of lines \o/

PR #167, opened 2020-05-11
Implements `db. getSiblingDB()` and `db.getCollection()`

mcasimir:
Implements `db.getSiblingDB()` and `db.getCollection()`.

- For `getSiblingDB` also factor the code with `use`
- For `getCollection` also changes the shell-api compilation for `Database` putting the collections in a separate property `_collections`, in such way  using `getCollection` we can have collection names that would normally conflict with property names in the database instance.

### DIFF
--- a/packages/mapper/src/mapper.ts
+++ b/packages/mapper/src/mapper.ts
@@ -78,12 +78,26 @@ export default class Mapper {
     this.messageBus.emit('mongosh:api-call', event);
   }
 
-  use(db): any {
-    if (!(db in this.databases)) {
-      this.databases[db] = new Database(this, db);
+  private _getDatabase(name: string): Database {
+    if (typeof name !== 'string') {
+      throw new MongoshInvalidInputError(
+        `Database name must be a string. Received ${typeof name}.`);
     }
+
+    if (!name.trim()) {
+      throw new MongoshInvalidInputError('Database name cannot be empty.');
+    }
+
+    if (!(name in this.databases)) {
+      this.databases[name] = new Database(this, name);
+    }
+
+    return this.databases[name];
+  }
+
+  use(db: string): any {
     this.messageBus.emit( 'mongosh:use', { db });
-    this.context.db = this.databases[db];
+    this.context.db = this._getDatabase(db);
 
     return `switched to db ${db}`;
   }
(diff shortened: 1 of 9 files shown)


packages/mapper/src/mapper.ts
```diff
@@ -78,12 +78,26 @@ export default class Mapper {
...
+      throw new MongoshInvalidInputError(
+        `Database name must be a string. Received ${typeof name}.`);
+    }
+
+    if (!name.trim()) {
+      throw new MongoshInvalidInputError('Database name cannot be empty.');
+    }
+
+    if (!(name in this.databases)) {
+      this.databases[name] = new Database(this, name);
```
mcasimir, line 92: Database names seems to be case insensitive. Should we do something about that here (ie. access ` this.databases[name.toLowerCase()]`), would this cause unexpected behaviours?

packages/mapper/src/mapper.ts
```diff
@@ -78,12 +78,26 @@ export default class Mapper {
...
+    if (typeof name !== 'string') {
+      throw new MongoshInvalidInputError(
+        `Database name must be a string. Received ${typeof name}.`);
+    }
+
+    if (!name.trim()) {
+      throw new MongoshInvalidInputError('Database name cannot be empty.');
+    }
+
+    if (!(name in this.databases)) {
```
mcasimir, line 91: Should we do something about case sensitivity here? https://docs.mongodb.com/manual/reference/limits/#Database-Name-Case-Sensitivity

lrlna: hmmm as I was answering this, I got a bit confused by this method. The name suggests a `read` from the driver to me, but I think you're just adding a new database to `this.databases`, which is a `write`. If this is indeed a way to save a database in our local state, I think we should handle it. Really good catch finding this, I totally didn't know case sensitivity was a thing.

My original answer was:
I think this is something we should be doing for a `write` if the driver _does not_ handle it for us (I feel like it would). If it's just a `read` method, than we should have handled it before we wrote it.

mcasimir: is `_getDatabase` a confusing name maybe? What about `_getAndCacheDatabase`?

PR #170, opened 2020-05-11
Refactor

aherlihy:
**Misc**
- I went through and made sure each package's tests could be run only in that directory, because it was driving me crazy. The result is there are a few more package.json's with `mocha` or other requirements that were previously missing. There were also a few bugs picked up here and there.
- I was having install issues because the versions were out of sync, so everything got updated to `0.0.1-alpha.14`. This will get updated to whatever we're at when it's rebased.

**Shell API**
- Delete the YAML directory and replace with Typescript files for each shell API class.
  - The metadata is now set via decorators that are defined in `decorators.ts`. Adding metadata to a method is now as easy as a `@returnsPromise` above the method! Also since we can generate decorators based on class name, `help` and `shellApiType()` are now fully automatic.
  - The "top-level" shell API class is `Mongo`, which each contains a `ServiceProvider` instance since each `Mongo` connects to a different URI and needs a new MongoClient (/StitchClient/DataService, if we decide to implement for those repls). Every shell api class has a reference to it's containing `Mongo`. The `Mongo` class contains a reference to the internal state so that all shell api classes that need to change the state can do so (for example updating the current cursor).
- `InternalState` is a class that keeps the internal state for the shell. Includes:
  - current cursor
  - current db
  - message bus (for emitting events)
  - `asyncRewriter` (because it also tracks the state in the symbol table)
  - `initialServiceProvider` which is used to generate more service providers of the same type for new connections.
  - helper methods that are used by multiple shell-api classes. This could also go in the `Mongo` class, don't really mind.
- The methods of the shell api classes now call their `mongo`'s ServiceProvider directly (i.e. doing exactly was previously in the `mapper` class). The tests from the mapper have been moved over to the shell-api, but there might be some more we should add.
- No YAML = real enums! Used for things like `ServerVersions`, `Topologies`, etc. These are defined in `enums.ts`.
  - Add a new enum called `ReplPlatform`, and have each repl instance pass the correct platform enum into the `ShellInternalState` constructor. We use this enum to limit the API to platform-appropriate classes and functions (since, for example, `Mongo` should not work in the browser). It can also be used later for platform-aware autocomplete.
- Quite a few of the CRUD commands weren't marked as deprecated in the YAML, so I updated the `serverVersions` metadata to reflect that (namely `count`, `update`, `save`, and `remove`.)
- Add a `ExplainableCursor` class based on @mcasimir's comments, I'm not sure if that's exactly what we want or not.

**Repl packages**
- Each repl class now only has to instantiate one `ShellInternalState` class on startup that accepts as arguments: a platform enum; a connected service provider; and a event emitter.
- The repl package also constructs a `ShellEvaluator`, but now it's only argument is the internal state constructed above. The Repl classes pass a connected instance of ServiceProvider into the internal state but *do not save it as a class property and do not do anything with it*.
- `BuildInfo` and other internal state is now stored in `ShellInternalState`. This is so that it can be automatically updated if `db` changes.

**Shell Evaluator**
- Remove everything that isn't JavaScript evaluation specific. So this package doesn't deal with any state, it only does eval.

**Service Provider packages**
- `getConnectionInfo` is removed from `cli-repl` and added to each Service Provider as a method so they can populate the connection information with whatever is useful for the context. No connection event was getting triggered for browser previously, so that is now fixed. As a result the cli service provider saves the URI now.
- `clone` was missing from the cursor implementations, so was added.
- add a `getNewConnection` method to each service provider. Since `new Mongo()` means we need to construct a new ServiceProvider of the same type that was originally passed into `InternalState`, it was easier to add a clone method than keep track of which ServiceProvider subclass needed to be instantiated. Open to other ideas if you don't like this!

**Async Rewriter**
- Previously `async-rewriter` had `shell-api` as a dev-dependency for tests. This was causing a circular dependency (even with dev, which is annoying) so I removed it and moved a copy of the shell-signatures into the repo ito test on. It shouldn't need regular updating unless we make major changes to the format of the signatures.
- Allow `db` to be reassigned. Other shell api types (like rs and sh) are still not allowed to be reassigned.
- Allow classes to return promises, since we want `m = new Mongo(...)` to be rewritten to `m = await new Mongo(...)`.

**Mapper**
- RIP <3

**Questions/TODO**
- Make sure that the java-shell is updated as well
- Does `shell-api/bin/report-supported-apis` still need to exist?
- More tests for the new shell-api classes?
- `ExplainableCursor` the right idea?
- Rebase with master
- Add tests for `Mongo` and `ShellInternalState`


packages/shell-api/src/database.ts
```diff
@@ -0,0 +1,155 @@
...
+@hasAsyncChild
+export default class Database extends ShellApiClass {
+  mongo: Mongo;
+  name: string;
+
+  constructor(mongo, name) {
+    super();
+    this.mongo = mongo;
+    this.name = name;
+    const proxy = new Proxy(this, {
```
mcasimir, line 29: Don't have much time to go through the rest of the code, this one i noticed previously, and it changed quite a bit in master, can you double check that we did not left bits behind?

aherlihy: Yes, that's part of the rebasing in the TODO section of this PR's description

aherlihy:
Replaced by #195

packages/shell-api/src/database.ts
aherlihy: This is the rebased PR: https://github.com/mongodb-js/mongosh/pull/195

Was complicated enough to make a new PR for the rebase, so closing this one.

PR #172, opened 2020-05-12
add eslint to cli-repl

lrlna:
## Description

Because lrlna keeps forgetting semi-colons.

### DIFF
--- a/config/eslintrc.base.js
+++ b/config/eslintrc.base.js
@@ -13,8 +13,11 @@ module.exports = {
   ],
   rules: {
     'object-curly-spacing': [2, 'always'],
+    'no-empty-function': 0,
     'valid-jsdoc': 0,
     'react/sort-comp': 0, // does not seem work as expected with typescript
+    '@typescript-eslint/no-empty-function': 0,
+    '@typescript-eslint/no-use-before-define': 0,
     '@typescript-eslint/no-explicit-any': 0,
     '@typescript-eslint/no-var-requires': 0, // seems necessary to import less files
     '@typescript-eslint/no-unused-vars': 2,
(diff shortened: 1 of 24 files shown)


config/eslintrc.base.js
```diff
@@ -13,8 +13,11 @@ module.exports = {
   ],
   rules: {
     'object-curly-spacing': [2, 'always'],
+    'no-empty-function': 0,
```
mcasimir, line 16: 👍 this one was a pest! :)

packages/cli-repl/src/cli-repl.ts
```diff
@@ -103,46 +105,50 @@ class CliRepl {
...
-          result = err;
         }
+        result = err;
       }
-      callback (null, result)
+      callback(null, result);
     };
 
+    // we can assign eval, it is not a read-only property, so ignore typescript
+    // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
```
mcasimir, line 130: Try to avoid `ts-ignore`.  Rather cast to `any` since it disables only the type checking:

```
(this.repl as any).eval = customEval;
```

packages/cli-repl/src/cli-repl.ts
```diff
@@ -103,46 +105,50 @@ class CliRepl {
...
+      const warn = new MongoshWarning('Unable to set up history file. History will not be persisting in this session');
       if (err) this.writer(warn);
 
       // repl.history is an array of previous commands. We need to hijack the
       // value we just typed, and shift it off the history array if the info is
       // sensitive.
       repl.on('flushHistory', function() {
+        // property history does exist on REPLServer, so ignore typescript
+        // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
         // @ts-ignore
```
mcasimir, line 148: Same, try to avoid // @ts-ignore

packages/cli-repl/src/cli-repl.ts
lrlna: ugh typescript :/

this.repl.eval is definitely a settable property. REPL does it consistently throughout the entire codebase. bah humbug I am old and grumpy at new javascript fads

PR #173, opened 2020-05-12
feat: disable .clear command

mcasimir:


packages/cli-repl/src/cli-repl.ts
```diff
@@ -126,8 +133,7 @@ class CliRepl {
       callback (null, result)
     };
 
-    // @ts-ignore
-    this.repl.eval = customEval;
+    (this.repl as any).eval = customEval;
```
lrlna, line 136: really, this is what's going to get typescript to behave? :/ :/ :/

PR #174, opened 2020-05-12
Allow message bus in runtime

mcasimir:
Allow to pass a message bus to the runtime so we can use it for telemetry and logging in compass:

```
new ElectronRuntime(serviceProvider, messageBus)
```


Anemy:
Nice - lgtm. The windows evergreen are expected to fail for now right?

mcasimir:
> Nice - lgtm. The windows evergreen are expected to fail for now right?

So it seems

PR #175, opened 2020-05-12
MONGOSH-89: Shell as global compass role

Anemy:
https://jira.mongodb.org/browse/MONGOSH-89

Mocks for reference: https://mongodb.invisionapp.com/share/RZVV94NAY6B#/screens/403867881

This PR updates the shell plugin to register as a global component for use in `compass-home`.

It also configures the styling of the component so that it can be clicked to toggle open and clicked to toggle closed. It is resize-able by dragging the triple bar in the middle as a handler, as well as the bar that separates it from the rest of the page. It has a max height of 800px, and a min height of 24px. We use this package for resizing: https://github.com/bokuweb/re-resizable I found it did the job nicely and haven't gotten into too many issues, the dependencies on it seem slim too. One thing we could improve is when compass is zoomed in a lot (ctrl + '+') and the shell is toggled open & closed, the resumed height is not relative to the zooming. I don't think this is a breaking bug but something we could address in future.

Toggling the shell clears it (history is maintained from Maurizio's work). Should we make it persist? What would you expect?

Here's where we add in the global plugin shell role and do some styling to make sure existing plugins scroll nicely vertically: https://github.com/mongodb-js/compass-home/pull/112

I'd like to wait for the PRs on `compass-home` and `compass` to be merged in before we merge this but I'm opening it now so we can take a look. Going to try some weird things to try and break it in the meantime. Currently to manually test I'm pulling the `compass-home` changes into the `compass` node modules in this repo and running `npm run start-compass`.

The 'i' info button is not yet added. Ticket for that: https://jira.mongodb.org/browse/COMPASS-4292

__Screenshots__
May-12-2020 19-58-01

### DIFF
--- a/packages/compass-shell/package-lock.json
+++ b/packages/compass-shell/package-lock.json
@@ -1339,9 +1339,22 @@
 			}
 		},
 		"@leafygreen-ui/icon": {
-			"version": "4.3.0",
-			"resolved": "https://registry.npmjs.org/@leafygreen-ui/icon/-/icon-4.3.0.tgz",
-			"integrity": "sha512-6DOFKxj3U0hp2yQEq/Yq19qIiIIUXwRj2fH1ddOJFYLtAoeFOCSeuF6rE2dqXrXW/puQdjgpXV60w9J8gyXGog=="
+			"version": "5.1.0",
+			"resolved": "https://registry.npmjs.org/@leafygreen-ui/icon/-/icon-5.1.0.tgz",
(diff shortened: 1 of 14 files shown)


Anemy:
Chatted with Fred on how we want the close/toggle button on the top right of the shell plugin window to look. We'll do a chevron. Ticket for that: https://jira.mongodb.org/browse/COMPASS-4294 or I'll

PR #179, opened 2020-05-13
allow bson constructors without new

aherlihy:


mmarcon:
This also fixes https://github.com/mongodb-js/vscode/issues/108

PR #181, opened 2020-05-13
adjust telemetry names for connect event

lrlna:
## Description
As per @mmarcon's request, changing telemetry events to be snake_cased instead
of camelCased.

This PR also adds `server_os` and `server_arch` information to `Connect` event.

### DIFF
--- a/packages/cli-repl/src/connect-info.spec.ts
+++ b/packages/cli-repl/src/connect-info.spec.ts
@@ -1,3 +1,4 @@
+/* eslint camelcase: 0, @typescript-eslint/camelcase: 0 */
 import { expect } from 'chai';
 import getConnectInfo from './connect-info';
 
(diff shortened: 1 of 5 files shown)


packages/cli-repl/src/connect-info.spec.ts
```diff
@@ -74,15 +75,17 @@ describe('getConnectInfo', function() {
...
+      server_version: '3.2.0-rc2',
+      is_enterprise: true,
+      auth_type: 'LDAP',
+      is_data_lake: false,
+      dl_version: null,
+      is_genuine: true,
+      non_genuine_server_name: 'mongodb',
+      server_arch: 'x86_64',
+      server_os: 'osx',
       uri: ATLAS_URI
```
mmarcon, line 89: What does `URI` look like? Something like this: `mongodb+srv://<credentials>@cluster0-e06dc.azure.mongodb.net/test?authSource=admin&replicaSet=Cluster0-shard-0&readPreference=primary&ssl=true`? Might be PII, not sure if we can collect it.

packages/cli-repl/src/logger.ts
```diff
@@ -55,11 +57,11 @@ NoopAnalytics.prototype.identify = function(): void {};
 NoopAnalytics.prototype.track = function(): void {};
 
 export default function logger(bus: any, logDir: string): void {
-  const sessionID = new ObjectId(Date.now());
-  const logDest = path.join(logDir, `${sessionID}_log`);
+  const session_id = new ObjectId(Date.now());
+  const logDest = path.join(logDir, `${session_id}_log`);
   const log = pino({ name: 'monogsh' }, pino.destination(logDest));
-  console.log(`Current sessionID: ${sessionID}`);
-  let userId;
```
mmarcon, line 62: I think this should stay `userId`

packages/cli-repl/src/connect-info.spec.ts
lrlna: yes, that's what it looks like. I do not send this to telemetry though, only to user's local log file

PR #183, opened 2020-05-14
add back node --experimental-repl-await to mongosh bin file

lrlna:
## Description
This got accidentally removed when adding eslint to cli-repl package. Adding it
back in.

### DIFF
--- a/packages/cli-repl/bin/mongosh.js
+++ b/packages/cli-repl/bin/mongosh.js
@@ -1,4 +1,5 @@
-#!/usr/bin/env node
+#!/bin/sh
+':' //; node --experimental-repl-await "$0" "$@"
 
 /* eslint no-console: 0, no-new: 0*/
 const { CliRepl, parseCliArgs, mapCliToDriver, generateUri, USAGE } = require('../lib');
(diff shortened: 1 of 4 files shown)


packages/cli-repl/bin/mongosh.js
```diff
@@ -1,4 +1,5 @@
-#!/usr/bin/env node
+#!/bin/sh
```
mcasimir, line 1: Nice trick! Is this how it was done before?

lrlna: yes it was!

lrlna:
@mcasimir added a test as well, good point.

packages/cli-repl/test/e2e.spec.ts
```diff
@@ -72,6 +72,14 @@ describe('e2e', function() {
       });
     });
 
+    it('does not throw for a repl await function', async() => {
+      shell.stdio.stdin.write('await Promise.resolve(\'Nori-cat\');');
```
mcasimir, line 76: 👍 🐱

PR #189, opened 2020-05-18
fix: bring in node repl's top-level await

lrlna:
## Description
To package up our executables with built in top-level await support, we will now
process all input with [node's internal await
wrap](https://github.com/nodejs/node/blob/master/lib/internal/repl/await.j://github.com/nodejs/node/blob/master/lib/internal/repl/await.js).

This wrapper puts `async()` in front of an awaited function.

Because we are rewriting our own functions with `await`, this is done after our
internal rewrite.

### DIFF
--- a/packages/async-rewriter/src/async-writer-babel.ts
+++ b/packages/async-rewriter/src/async-writer-babel.ts
@@ -8,6 +8,7 @@ import {
   MongoshInvalidInputError,
   MongoshUnimplementedError
 } from '@mongosh/errors';
+import processTopLevelAwait from './await';
 
 const debug = (str, type?, indent?): void => {
   indent = indent ? '' : '  ';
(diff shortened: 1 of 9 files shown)


packages/async-rewriter/src/async-writer-babel.ts
```diff
@@ -748,4 +749,13 @@ export default class AsyncWriter {
   compile(code): string {
     return this.getTransform(code).code;
   }
+
+  /**
+   * Returns translated code.
+   * @param {string} code - string to compile.
+   */
+  process(code): string {
```
aherlihy, line 757: Could you throw a test or two in here just to make sure it's doing what it's supposed to be doing?

lrlna: Just did! Take a look again.

lrlna:
@aherlihy no, it's the same as we had before. 

Since we are using the exact same implementation as node was using when we were passing `--experimental-repl-await` flag (just moved over to our side instead), parsing performance does not change.

packages/async-rewriter/src/async-writer-babel.ts
aherlihy: Looks great!

PR #195, opened 2020-05-19
Rebased Refactor

aherlihy:
Rebased with master version, replaces #170 

**Misc**
- I went through and made sure each package's tests could be run only in that directory, because it was driving me crazy. The result is there are a few more package.json's with `mocha` or other requirements that were previously missing. There were also a few bugs picked up here and there.

**Shell API**
- Delete the YAML directory and replace with Typescript files for each shell API class.
  - The metadata is now set via decorators that are defined in `decorators.ts`. Adding metadata to a method is now as easy as a `@returnsPromise` above the method! Also since we can generate decorators based on class name, `help` and `shellApiType()` are now fully automatic.
  - The "top-level" shell API class is `Mongo`, which each contains a `ServiceProvider` instance since each `Mongo` connects to a different URI and needs a new MongoClient (/StitchClient/DataService, if we decide to implement for those repls). Every shell api class has a reference to it's containing `Mongo`. The `Mongo` class contains a reference to the internal state so that all shell api classes that need to change the state can do so (for example updating the current cursor).
- `InternalState` is a class that keeps the internal state for the shell. Includes:
  - current cursor
  - current db
  - message bus (for emitting events)
  - `asyncRewriter` (because it also tracks the state in the symbol table)
  - `initialServiceProvider` which is used to generate more service providers of the same type for new connections.
  - helper methods that are used by multiple shell-api classes. This could also go in the `Mongo` class, don't really mind.
- The methods of the shell api classes now call their `mongo`'s ServiceProvider directly (i.e. doing exactly was previously in the `mapper` class). The tests from the mapper have been moved over to the shell-api, but there might be some more we should add.
- No YAML = real enums! Used for things like `ServerVersions`, `Topologies`, etc. These are defined in `enums.ts`.
  - Add a new enum called `ReplPlatform` to each service provider to limit the API to platform-appropriate classes and functions (since, for example, `Mongo` should not work in the browser). It can also be used later for platform-aware autocomplete.
- Quite a few of the CRUD commands weren't marked as deprecated in the YAML, so I updated the `serverVersions` metadata to reflect that (namely `count`, `update`, `save`, and `remove`.)
- Add a `ExplainableCursor` class based on @mcasimir's comments, I'm not sure if that's exactly what we want or not.

**Repl packages**
- Each repl class now only has to instantiate one `ShellInternalState` class on startup that accepts as arguments: a connected service provider and a event emitter.
- The repl package also constructs a `ShellEvaluator`, but now it's only argument is the internal state constructed above. The Repl classes pass a connected instance of ServiceProvider into the internal state but *do not save it as a class property and do not do anything with it*.
- `BuildInfo` and other internal state is now stored in `ShellInternalState`. This is so that it can be automatically updated if `db` changes.

**Shell Evaluator**
- Remove everything that isn't JavaScript evaluation specific. So this package doesn't deal with any state, it only does eval.

**Service Provider packages**
- `getConnectionInfo` is removed from `cli-repl` and added to each Service Provider as a method so they can populate the connection information with whatever is useful for the context. No connection event was getting triggered for browser previously, so that is now fixed. As a result the cli service provider saves the URI now.
- `clone` was missing from the cursor implementations, so was added.
- add a `getNewConnection` method to each service provider. Since `new Mongo()` means we need to construct a new ServiceProvider of the same type that was originally passed into `InternalState`, it was easier to add a clone method than keep track of which ServiceProvider subclass needed to be instantiated. Open to other ideas if you don't like this!

**Async Rewriter**
- Previously `async-rewriter` had `shell-api` as a dev-dependency for tests. This was causing a circular dependency (even with dev, which is annoying) so I removed it and moved a copy of the shell-signatures into the repo ito test on. It shouldn't need regular updating unless we make major changes to the format of the signatures.
- Allow `db` to be reassigned. Other shell api types (like rs and sh) are still not allowed to be reassigned.
- Allow classes to return promises, since we want `m = new Mongo(...)` to be rewritten to `m = await new Mongo(...)`.

**Mapper**
- RIP <3

**Questions/TODO**
- [x] Make sure that the java-shell is updated as well
- [x] Does `shell-api/bin/report-supported-apis` still need to exist?
- [ ] More tests for the new shell-api classes?
- [x] `ExplainableCursor` the right idea?
- [ ] Add tests for new classes: `Mongo` and `ShellInternalState`
- [ ] Compass-shell tests occasionally failing due to install issue, not sure why


packages/shell-api/src/shell-internal-state.ts
```diff
@@ -0,0 +1,238 @@
...
+
+  async fetchConnectionInfo(): Promise<void> {
+    this.connectionInfo = await this.currentDb.mongo.serviceProvider.getConnectionInfo();
+    this.messageBus.emit('mongosh:connect', this.connectionInfo.extraInfo);
+  }
+
+  // TODO: do we need to close all Mongos ever made?
+  close(p): void {
+    try {
+      this.currentDb.mongo.serviceProvider.close(p);
```
mcasimir, line 47: `serviceProvider .close` returns a promise, this `try` `catch` would not defend us much from an unhandled rejection. Probably `await` or `.serviceProvider.close(p).catch(() => {})`.

packages/shell-api/src/shell-internal-state.ts
```diff
@@ -0,0 +1,238 @@
...
+        console.log(arg);
+      }
+    };
+    contextObject.printjson = contextObject.print;
+    Object.assign(contextObject, ShellBson);
+    contextObject.rs = new ReplicaSet(this.currentDb.mongo);
+    contextObject.sh = new Shard(this.currentDb.mongo);
+
+    // Add global shell objects
+    this.asyncWriter.symbols.initializeApiObjects({
```
mcasimir, line 128: This should be probably done in the constructor, or it does depend on the context.

packages/shell-api/src/shell-internal-state.ts
```diff
@@ -0,0 +1,238 @@
...
+    return this.currentDb.mongo.show(arg);
+  }
+  async it(): Promise<any> {
+    if (!this.currentCursor) {
+      // TODO: warn here
+      return new CursorIterationResult();
+    }
+    return await this.currentCursor._it();
+  }
+  public help(): Help {
```
mcasimir, line 66: I think with a bit of creativity we could generate this help method too.

packages/shell-api/src/shell-internal-state.ts
```diff
@@ -0,0 +1,238 @@
...
+    this.currentCursor = null;
+    this.currentDb = mongo.getDB('test'); // TODO: set to CLI arg
+  }
+
+  async fetchConnectionInfo(): Promise<void> {
+    this.connectionInfo = await this.currentDb.mongo.serviceProvider.getConnectionInfo();
+    this.messageBus.emit('mongosh:connect', this.connectionInfo.extraInfo);
+  }
+
+  // TODO: do we need to close all Mongos ever made?
```
mcasimir, line 44: Seems a nice thing to be doing.

packages/shell-api/src/shell-internal-state.ts
```diff
@@ -0,0 +1,238 @@
...
+   *
+   * The `contextObject` is prepared so that it can be used as global object
+   * for the repl evaluationi.
+   *
+   * @note The `contextObject` is mutated, it will retain all of its existing
+   * properties but also have the global shell api objects and functions.
+   *
+   * @param {Object} contextObject - contextObject an object used as global context.
+   */
+  setCtx(contextObject: any): void {
```
mcasimir, line 108: I'd love to have a dedicated class `TopLevelContext` with decorators to keep
`use, show, it, db, rs, Mongo, printjson, DBQuery ... all bson types` etc ..., 
in the exact same way it happens for all the other methods and apis.

For example `Decimal128` is only supported since 3.4, wereas `Symbol` is deprecated
and would be probably unavailable in future versions, same could happen about other
objects that would not make sense for compatibility going further.

Another benefit of having these methods in a similar way we get for Collection etc
is that we could add those to the help automatically and mark them as commands 
if we want (with a new annotation @command).

```
class TopLevelContext { // or ShellGlobals
  // ... all the BSON classes ()

  get ObjectId {
    return ShellBSON.ObjectId(...args) // or just define this here
  }

  get DBQuery() {
    return Cursor;
  } 

  @command
  it()

  print() {}

  printjson() {
    return this.print()
  }

  @returnPromise
  connect() {

  }

  @serverVersion(['3.4', '...'])
  get Decimal128 {

  }
}
```

In `internalState` and `setCtx` we can just do this: 

```
constructor() {
  this.topLevelContext = new TopLevelContext(serviceProvider, mongo, 'etc ...');
}

setCtx() {
  Object.entries(signatures.TopLevelContext.attributes).forEach(name => {
    contextObject[name] = typeof this.topLevelContext[name] === 'function' ?
      this.topLevelContext[name].bind(this) : this.topLevelContext[name];
  });
}
```

In shell-evaluator instead we can replace the switch with:

```
if(signatures.TopLevelContext.commands.includes(cmd)) {
  return this.internalState[cmd](...argv);
}
```

packages/shell-api/src/shell-internal-state.ts
```diff
@@ -0,0 +1,238 @@
...
+    // Add global shell objects
+    this.asyncWriter.symbols.initializeApiObjects({
+      db: signatures.Database,
+      rs: signatures.ReplicaSet,
+      sh: signatures.Shard,
+      Mongo: signatures.Mongo
+    });
+
+    try {
+      Object.defineProperty(contextObject, 'db', {
```
mcasimir, line 136: this would look so much easier if we extract it in a class

packages/shell-api/src/shell-internal-state.ts
```diff
@@ -0,0 +1,238 @@
...
+        }
+      });
+    } catch (e) { // java shell, can't use getters/setters
+      console.log(`in java shell, platform: ${this.initialServiceProvider.platform}`);
+      contextObject.db = this.currentDb;
+    }
+
+    contextObject.Mongo = async(uri?, options?): Promise<Mongo> => {
+      if (
+        this.initialServiceProvider.platform !== ReplPlatform.CLI
```
mcasimir, line 158: I love the idea to be able to negotiate with the service providers! I suggest had-hoc feature flags though, usually more helpful. Something like this:

```
if (!this.initialServiceProvider.features.multipleConnection) {
  // ...
}
```

or even a more dynamic one:

```
!this.initialServiceProvider.featureEnabled('multipleConnection')
```

packages/shell-api/src/shell-internal-state.ts
```diff
@@ -0,0 +1,238 @@
...
+      return mongo;
+    };
+
+    this.messageBus.emit(
+      'mongosh:setCtx',
+      { method: 'setCtx', arguments: { db: contextObject.db } }
+    );
+  }
+
+  public emitApiCall(event: {
```
mcasimir, line 175: This doesn't feel like belonging here somehow, I'd extract this in a separate utility class, can be an ad-hoc decorator for `messageBus` which holds all of the `emit` methods:

```
ShellApiEventsEmitter {
   constructor(bus) {}

   private emitApiCall()
   public  emitCollectionApiCall()
   public  emitDatabaseApiCall()
   public  emitCommandApiCall()
}
```
, just the `emitApiCall` function or

packages/shell-api/src/shell-internal-state.ts
```diff
@@ -0,0 +1,238 @@
...
+    this.messageBus.emit('mongosh:api-call', event);
+  }
+
+  /**
+   * Helper method to adapt aggregation pipeline options.
+   * This is here so that it's not visible to the user.
+   *
+   * @param options
+   */
+  public adaptAggregateOptions(options: any = {}): {
```
mcasimir, line 190: I'd look for a way to move this one to a util class or just as a separate function in a module, looks like completely static.

packages/shell-api/src/shell-internal-state.ts
```diff
@@ -0,0 +1,238 @@
...
+
+    if ('explain' in providerOptions) {
+      explain = providerOptions.explain;
+      delete providerOptions.explain;
+    }
+
+    return { providerOptions, dbOptions, explain };
+  }
+
+  public validateExplainableVerbosity(verbosity: string): void {
```
mcasimir, line 218: same for this one, i think it does not need to be dynamically passed down, it can be imported from a module where it is used.

packages/shell-api/src/shell-internal-state.ts
```diff
@@ -0,0 +1,238 @@
...
+    ];
+
+    if (!allowedVerbosity.includes(verbosity)) {
+      throw new MongoshInvalidInputError(
+        `verbosity can only be one of ${allowedVerbosity.join(', ')}. Received ${verbosity}.`
+      );
+    }
+  }
+
+  public checkUndefinedUpdate(...args: any): void {
```
mcasimir, line 232: same

packages/shell-api/src/shell-internal-state.ts
```diff
@@ -0,0 +1,238 @@
...
+          contextObject.rs = new ReplicaSet(this.currentDb.mongo);
+          contextObject.sh = new Shard(this.currentDb.mongo);
+          this.fetchConnectionInfo();
+          return newDb;
+        },
+        get: () => {
+          return this.currentDb;
+        }
+      });
+    } catch (e) { // java shell, can't use getters/setters
```
mcasimir, line 151: Oh boy this is unfortunate. :(

packages/shell-api/src/shard.ts
```diff
@@ -0,0 +1,50 @@
...
+  ShellApiClass,
+} from './decorators';
+
+import {
+  Document
+} from '@mongosh/service-provider-core';
+import { MongoshUnimplementedError } from '@mongosh/errors';
+
+@shellApiClassDefault
+@hasAsyncChild
```
mcasimir, line 14: this can be easily inferred from the other signatures, seems a detail we could hide in `@shellApiClassDefault`

packages/cli-repl/src/cli-repl.ts
```diff
@@ -78,10 +77,10 @@ class CliRepl {
...
    */
   async setupRepl(driverUri: string, driverOptions: NodeOptions): Promise<void> {
-    this.serviceProvider = await this.connect(driverUri, driverOptions);
-    this.ShellEvaluator = new ShellEvaluator(this.serviceProvider, this.bus, this);
-    this.buildInfo = await this.serviceProvider.buildInfo();
-    this.logBuildInfo(driverUri);
+    const initialServiceProvider = await this.connect(driverUri, driverOptions);
+    this.internalState = new ShellInternalState(initialServiceProvider, this.bus);
+    this.shellEvaluator = new ShellEvaluator(this.internalState, this);
+    await this.internalState.fetchConnectionInfo();
```
mcasimir, line 83: Wouldn't be possible for this one to return a `ConnectionInfo` object and cache  it rather than mutate this.internalState?

packages/async-rewriter/src/async-writer-babel.ts
```diff
@@ -438,6 +438,20 @@ var TypeInferenceVisitor: Visitor = { /* eslint no-var:0 */
 
       const lhsType = path.node.callee['shellType'];
       path.node['shellType'] = lhsType.returnType || { type: 'unknown', attributes: {} };
+      if (lhsType.returnsPromise) {
+        // NOTE: if need to convert top-level await into async func can do it here.
```
mcasimir, line 442: +1

packages/browser-runtime-core/src/open-context-runtime.ts
```diff
@@ -29,18 +30,17 @@ export class OpenContextRuntime implements Runtime {
...
     this.interpreter = new Interpreter(this.interpreterEnvironment);
   }
 
   async getCompletions(code: string): Promise<Completion[]> {
     if (!this.autocompleter) {
-      const buildInfo = await this.serviceProvider.buildInfo();
-      const serverVersion = buildInfo.version;
-      this.autocompleter = new ShellApiAutocompleter(serverVersion);
+      await this.internalState.fetchConnectionInfo();
+      this.autocompleter = new ShellApiAutocompleter(this.internalState.connectionInfo.buildInfo.version);
```
mcasimir, line 43: This seems done in the same way in CLI repl, maybe we should keep it in mind and see if we can factor this somehow.

packages/cli-repl/src/cli-repl.ts
```diff
@@ -152,52 +151,16 @@ class CliRepl {
     });
 
     this.repl.on('exit', () => {
-      this.serviceProvider.close(true);
+      this.internalState.close(true);
```
mcasimir, line 154: This one was the same before, but i just noticed it: this function without await can both throw or reject.

Another thing to be noted. If we don't await  `close`, `process.exit()` will likely kill the process before anything would happen about closing the service provider gracefully.

I don't see a practical downside here, just something to keep in mind.

lrlna:
@aherlihy and i were talking about this refactor on tuesday, and i made a few sketches of the API. 

Here is how this works when `Mongo` is not reassigned:
Untitled 2

And here is how this would work when `Mongo` is reassigned:
Untitled 1

packages/cli-repl/src/cli-repl.ts
lrlna: It may be that on initial connect we just keep this call entirely in internalState instead of awaiting it here.

Because if the connection information changes, technically the internal state also has to change. I think it really ought to mutate the state (and then log that change).

mcasimir: Phrased this one wrong: totally get the mutation has to happen internally. It seems safe to me to keep that 'private' inside `ShellInternalState`.

packages/shell-api/src/shell-internal-state.ts
aherlihy: Good catch! The try/catch is there in case `currentDb` isn't defined (if the user resets `db` to something strange) but you're totally right that we should also check for errors thrown by actually closing

packages/shell-api/src/shell-internal-state.ts
aherlihy: Lucy is trying to see if she can use the async-rewriter in java, but if she can't, then we will have to move the async-rewriter out of internal state entirely and put it in the shell-evaluator. I feel like if we're not setting the state in the same place (I was worried people would forget to add to both context and symbol table), might as well put it in a different package since java can't import babel and doesn't use shell-eval. That seems nicer anyway since rewriter has more to do with eval than it has to do with state imo.

packages/shell-api/src/shard.ts
aherlihy: The `hasAsyncChild` attribute is just a helper to speed up inferring if there's a child attribute with `returnsPromise` (otherwise it could get quite slow checking every child of the shell api types, every time anything is passed through the writer). We could also infer it in the decorator, but I was thinking we could do a more major refactor where we use the typescript annotations to generate the attributes.

The `returnsPromise` and `returnType` decorators are pretty much always the same thing as the type signature. We could try to infer them dynamically, but I was thinking that'd be a much bigger refactor and would slow this down significantly so it should be done separately.

packages/shell-api/src/shell-internal-state.ts
```diff
@@ -0,0 +1,238 @@
...
+} from './index';
+import { EventEmitter } from 'events';
+import { DatabaseOptions, Document, ServiceProvider, ReplPlatform } from '@mongosh/service-provider-core';
+import { MongoshInvalidInputError, MongoshUnimplementedError } from '@mongosh/errors';
+import AsyncWriter from '@mongosh/async-rewriter';
+
+/**
+ * Anything to do with the internal shell state is stored here.
+ */
+export default class ShellInternalState {
```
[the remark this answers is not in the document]
aherlihy: I've moved the utilities that are static into a `helper.ts` file, but `emitApiCall` is a helper that uses the internal state (i.e. the current bus) so I'm not sure that should be removed from the internal class.

The top-level context methods like use/show/it are all state dependent. The reason they're in internal state is because we need to bind the method to whatever the currentDb/currentCursor/etc are at the moment. To me that feels like exactly what this class is for

packages/shell-api/src/shell-internal-state.ts
aherlihy: I don't know what you mean by that, can you write an example?

packages/shell-api/src/shell-internal-state.ts
aherlihy: I moved it into a `helper.ts` file that is stateless, good idea!

packages/cli-repl/src/cli-repl.ts
aherlihy: My thinking is that the connection info is internal state, therefore it should be stored in the internal state object. It is exactly the type of information that needs to be used and updated in various places, so why hide it? 

I don't see the benefit to caching the connection info in the repl when we want that information to be stored in a place where it can be updated dynamically (so for example, when the user reassigns the `db` variable in the shell, the internal state class will automatically update it's connection info object). The repl's prompt should be based on that internal state connection information, so when we implement custom prompts then we will get that info from the internal state. Perhaps we will have the repls pass a callback to the internal state so it can update the prompt when the `db` changes, but since we haven't implemented that feature and it's not a part of this refactor, I left that unimplemented.

`fetchConnectionInfo` is a separate async function because it needs to do a DB command to get that info, and in the browser-repl the place where the ShellInternalState is instantiated isn't an async function.

packages/shell-api/src/shell-internal-state.ts
aherlihy: I've updated it so that we store all newly created mongos and close them all at the end.

packages/browser-runtime-core/src/open-context-runtime.ts
aherlihy: Do you mean the autocompleter (which irina is currently pulling out) or do you mean the step of fetchConnectionInfo?

packages/cli-repl/src/cli-repl.ts
aherlihy: I've updated it so it will await each `close` so that we get through everything

packages/shell-api/src/shell-internal-state.ts
aherlihy: it's less annoying than it seems, since there are no new `Mongo` objects in java, we don't need to do any of the things that are inside the setter except assign `db`, which we can do here. We also need to set it in `use`, but so be it

packages/shell-api/src/shell-internal-state.ts
aherlihy: I moved the other methods out, but this one I kept because it uses the bus. The alternative would be to have a method in the `helpers.ts` file that accepted a bus as the first argument, but that doesn't really change anything except make it slightly more annoying to call from the shell api classes

packages/shell-api/src/shell-internal-state.ts
aherlihy: How would you write a decorator that could generate an attribute whose name is 'show databases' or '.exit'?

packages/shell-api/src/shell-internal-state.ts
aherlihy: I don't understand what you mean by @command, can you explain or give an example?

packages/shell-api/src/shell-internal-state.ts
aherlihy: I've moved these methods into a ShellApi class in this commit: https://github.com/mongodb-js/mongosh/pull/195/commits/b6fdfd1220769050694d7a7b4f44c3bd7771526f

packages/shell-api/src/shell-internal-state.ts
aherlihy: I have been thinking about this quite a bit, and generally agree with you about using ad-hoc flags but I wonder what that would look like when we are also creating such an extensive set of feature flags based on the platform enum in shell-api. So every function/class/etc in the Shell API has a `platforms` attribute, which is one of those 3 platform enums. So would we expand everything so that we have shell API metadata that matches these ad-hoc flags? That would end up exponentially growing and I'd worry about it being fragile. Since we already have those feature flags that are sorted by platform in shell-api, I'd say we should keep the generalized platform flags.

This is a slightly special case because we want to limit the creation of `Mongo` objects based on topology, _except for the initial mongo_ which gets created for all platforms.  So usually this check would go in the constructor of a class, and be automatically generated based on the attributes set by the topologies decorator. Since we haven't implemented this yet, it's not in the decorators, but that's how I expected it would go when we do write the platform-aware feature.

packages/shell-api/src/shell-internal-state.ts
aherlihy: I moved `use`, `show`, and `it` into a top level object `ShellApi`. I got pretty far in moving the BSON types into that class as well (in this PR: https://github.com/mongodb-js/mongosh/tree/bson-pr) but realized it would affect the java-shell and I think it would be quite annoying to refactor that. So I'm going to say we should take this request and make it into a ticket to be worked on as part of the `dev prod` epic.

packages/shell-api/src/shell-internal-state.ts
mcasimir: The first thing that comes to mind is something like this: 

```
@shellApiDefaultClass({ additionalHelpKeys: ['show databases', '.exit'] })
class SomeClass {}
```

we can do that later, in a dedicated story.

packages/shell-api/src/shell-internal-state.ts
mcasimir: Ideally I'd love to have all the global variables and methods in the same place, as class accessors and methods, but I understand is probably impossible or it wont fit in the rest of the design. Something like this:

``` js
class ShellApiGlobals {
  private _currentDb: Database;
  private _replicaSetCache: Map<Mongo, ReplicaSet>;

  constructor() {
    this._currentDb = new Database(/** ... */);
    this._replicaSetCache = new Map<Mongo, ReplicaSet>();
  }
  
  get db(): Database {
    return this._currentDb;
  }

  set db(database: Database) {
    this._currentDb = database;
  }

  get rs() {
    if (!this._replicaSetCache.has(this._currentDb.mongo)) {
      this._replicaSetCache.set(new ReplicaSet(this._currentDb.mongo));
    }

    return this._replicaSetCache.get(this._currentDb);
  }
}
```

So that `setCtx` would be just like this:

``` js
class ShellInternalState {
  constructor() {
    this.shellApiGlobals = new ShellApiGlobals(/** ... */);
  }

  setCtx(contextObject) {
    const globalAttributes = Object.keys(signatures.ShellApiGlobals.attributes);

    for (const attributeName of globalAttributes) {
      contextObject[attributeName] = typeof this.shellApiGlobals[attributeName] === 'function' ?
        this.shellApiGlobals[attributeName].bind(this.shellApiGlobals) :
        this.shellApiGlobals[attributeName]
    }
  }
}
```

Totally fine to keep this for later.

packages/shell-api/src/shell-internal-state.ts
mcasimir: Glad to hear that

packages/shell-api/src/shell-internal-state.ts
mcasimir: I think is fine to keep it by platform, if that gets problematic for some reason we can figure something later.

packages/shell-api/src/shell-internal-state.ts
mcasimir: Sounds good.

packages/shell-api/src/shell-internal-state.ts
aherlihy: Ticketed here: https://jira.mongodb.org/browse/MONGOSH-215

packages/shell-api/src/shell-internal-state.ts
aherlihy: https://jira.mongodb.org/browse/MONGOSH-216

PR #196, opened 2020-05-19
apply licence to MongoDB Inc.

MaBeuLux88:
According to the previous version of this file, this is how to apply Apache 2.0 to a project:

   APPENDIX: How to apply the Apache License to your work.	

      To apply the Apache License to your work, attach the following	
      boilerplate notice, with the fields enclosed by brackets "[]"	
      replaced with your own identifying information. (Don't include	
      the brackets!)  The text should be enclosed in the appropriate	
      comment syntax for the file format. We also recommend that a	
      file or class name and description of purpose be included on the	
      same "printed page" as the copyright notice for easier	
      identification within third-party archives.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -176,18 +176,7 @@
 
    END OF TERMS AND CONDITIONS
 
-   APPENDIX: How to apply the Apache License to your work.
-
-      To apply the Apache License to your work, attach the following
-      boilerplate notice, with the fields enclosed by brackets "[]"
-      replaced with your own identifying information. (Don't include
-      the brackets!)  The text should be enclosed in the appropriate
-      comment syntax for the file format. We also recommend that a
-      file or class name and description of purpose be included on the
-      same "printed page" as the copyright notice for easier
-      identification within third-party archives.
-
-   Copyright [yyyy] [name of copyright owner]
+   Copyright 2020 MongoDB Inc.
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -1,193 +1,4 @@
...
-      boilerplate notice, with the fields enclosed by brackets "[]"
-      replaced with your own identifying information. (Don't include
-      the brackets!)  The text should be enclosed in the appropriate
-      comment syntax for the file format. We also recommend that a
-      file or class name and description of purpose be included on the
-      same "printed page" as the copyright notice for easier
-      identification within third-party archives.
-
-   Copyright [yyyy] [name of copyright owner]
+   Copyright 2020 MongoDB Inc.
```
lrlna, line 1: @MaBeuLux88 did you mean to delete the whole of this LICENSE file, or just replace line 190 with `Copyright 2020 MongoDB Inc.`?

MaBeuLux88: I meant it. The licence is already in the link.

>    You may obtain a copy of the License at
       http://www.apache.org/licenses/LICENSE-2.0

I guess it's also possible to keep it like this for example: 
https://github.com/moby/moby/blob/master/LICENSE

Let me know which one you prefer, I can amend the PR.

lrlna: We prefer to keep this file in our repos. I think we just forgot to correct the `Copyright` placeholder.

MaBeuLux88: OK done. 
I re-inserted the LICENCE text and removed the paragraph that explain how to apply the licence which isn't needed anymore.

> To apply the Apache License to your work...

PR #198, opened 2020-05-20
Allow `require` in vscode and compass shell

mcasimir:
A fix that allows to use `require` in playgrounds and compass.

**NOTE:** 

Webpack will replace `require` with its own implementation, and that would not necessarily have access to the node modules available in node (depends on the target configuration).

As a workaround we alias `require` to [`__non_webpack_require__`](https://webpack.js.org/api/module-variables/#__non_webpack_require__-webpack-specific) when available.


mmarcon:
resolves mongodb-js/vscode#112

PR #200, opened 2020-05-25
move autocompleter into its own package

lrlna:
## Description

Creates a new autocompleter package so that the browser modules are able to directly require it.

## Open Questions
@Anemy @mcasimir I am hitting a weird error in `compass-shell` when running tests. Do either of you know what it could be?
Capture d’écran, le 2020-05-25 à 12 09 33


lrlna:
closing in favour of #389

PR #201, opened 2020-05-25
format bson results

lrlna:
## Description

This is a bit (ok, a lot) of a hacky way to get to get bson types to display
better visually. It's not the best, and I'd like to redo this asap after we
release beta.


packages/cli-repl/src/format-bson.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  return fixBSONTypeOutput(string, regexp);
+}
+
+export default function bsonWriter(line) {
+  mongodb.ObjectId.prototype.toJSON = function() {
+    return `ObjectId("${this.toHexString()}")`;
+  };
+
+  if (Array.isArray(line)) return JSON.stringify(line, null, 2);
```
mcasimir, line 30: JSON.stringify does't really like circular refs, we may need to cope with that.

aherlihy:
I'm working on adding the BSON types as their own classes, (progress here: https://github.com/mongodb-js/mongosh/tree/bson-pr) and if we extend each BSON type then we can easily override the `toReplString` method. However this is work that will probably get pushed to after the initial merge of the refactor, since it would mess with what the java-shell is doing.

lrlna:
closing this as #204 is better suited to solve this issue.

PR #204, opened 2020-05-27
util inspect bson types

lrlna:
## Description
Use `util.inspect.custom` Symbol to override formatting for BSON types. 

## Open Questions
- I cannot run `this.toExtendedJSON` for `MinKey` and `MaxKey` for some reason. Anyone see what's up?
- Any other types we should be converting?

### DIFF
--- a/packages/cli-repl/src/format-output.ts
+++ b/packages/cli-repl/src/format-output.ts
@@ -3,12 +3,64 @@
 import prettyBytes from 'pretty-bytes';
 import textTable from 'text-table';
 import i18n from '@mongosh/i18n';
+import mongodb from 'mongodb';
 import util from 'util';
 import clr from './clr';
 
 type EvaluationResult = {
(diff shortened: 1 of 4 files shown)


packages/cli-repl/src/format-output.ts
```diff
@@ -11,6 +12,57 @@ type EvaluationResult = {
   type?: string;
 };
 
+const i = Symbol.for('nodejs.util.inspect.custom');
```
lrlna, line 15: The idea behind this is so us to setup a custom inspect function specifically only for mongodb types. I chose to do this here, rather than at a higher level (internal state, service provider etc), because I wanted other implementors to receive data as is, and choose what to do with it in their own way. This is perhaps flawed, as it might be better for implementors to just receive raw bson instead.

aherlihy:
Implemented by #254

PR #207, opened 2020-06-02
use node12 when running compileExec unsigned

lrlna:


packages/build/src/release.ts
```diff
@@ -24,6 +25,11 @@ const release = async(config: Config) => {
     userAgent: `mongosh ${config.version}`
   });
 
+  exec('node --version', function (err, stdout, stderr) {
```
mcasimir, line 28: `execSync`?

PR #210, opened 2020-06-04
fixes to telemetry, objectid, and local config file

lrlna:
## Description

Updates a few things in regards to telemetry:
- always call `toString()` when initially creating ObjectIds for sessions and
  userID
- Use `fast-json-parse` when parsing config file. Return and log if an error is
  present.
- run analytics.identify when updating user data if telemetry is present.
- remove warn log for analytics key
- log out node version as part of connection log event

### DIFF
--- a/packages/cli-repl/src/cli-repl.ts
+++ b/packages/cli-repl/src/cli-repl.ts
@@ -251,6 +251,8 @@ class CliRepl {
       this.writeConfigFileSync(configPath);
     } catch (err) {
       if (err.code === 'EEXIST') {
+        // make sure we catch errors for json parse and always have err and
+        // value on config result
         const config = jsonParse(fs.readFileSync(configPath, 'utf8'));
 
         if (config.err) {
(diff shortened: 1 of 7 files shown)


packages/cli-repl/src/cli-repl.ts
```diff
@@ -243,17 +244,23 @@ class CliRepl {
...
+      this.userId = new ObjectId(Date.now()).toString();
       this.enableTelemetry = true;
       this.disableGreetingMessage = false;
       this.bus.emit('mongosh:new-user', this.userId, this.enableTelemetry);
       this.writeConfigFileSync(configPath);
     } catch (err) {
       if (err.code === 'EEXIST') {
-        const config = JSON.parse(fs.readFileSync(configPath, 'utf8'));
-        this.userId = config.userId;
+        const config = jsonParse(fs.readFileSync(configPath, 'utf8'));
```
mcasimir, line 254: I think a comment explaining why we are not using `JSON.parse` here is necessary. I understand you want a tolerant version of JSON.parse that does not throw, but is not clear from this code.

PR #211, opened 2020-06-04
[Rebased] test: allow to run test on already running server

aherlihy:


packages/async-rewriter/package.json
```diff
@@ -4,8 +4,8 @@
   "description": "MongoDB Shell Async Rewriter Package",
   "main": "./lib/index.js",
   "scripts": {
-    "test": "mocha --timeout 15000 --colors -r ts-node/register \"./{src,lib}/**/*.spec.ts\"",
-    "test-ci": "mocha --timeout 15000 -r ts-node/register \"./{src,lib}/**/*.spec.ts\"",
+    "test": "mocha --timeout 60000 --colors -r ts-node/register \"./{src,lib}/**/*.spec.ts\"",
```
lrlna, line 7: do we still need this large of a timeout? or was that just for the azure pipelines windows tests?

packages/cli-repl/test/e2e.spec.ts
```diff
@@ -29,8 +29,8 @@ describe('e2e', function() {
 
     beforeEach(async() => {
       dbName = `test-${Date.now()}`;
-      const connectionString = `mongodb://localhost:27018/${dbName}`;
-      shell = TestShell.start({ args: [ connectionString ] });
+      const dbConnectionString = `${connectionString}}/${dbName}`;
```
mcasimir, line 32: I think this can just be the connectionString from `startTestServer` since concatenating is probably not going to work if we use atlas or some url that has query parameters

aherlihy: Ok, will get rid of the db in the uri

packages/async-rewriter/package.json
aherlihy: Im not sure - @mcasimir ?

mcasimir: That aligns with what was inside the `mocha` tests: `this.timeout(60000)` was overriding the 15000 from the command line. 

In general 1 minute is not a large timeout for the CI, and it won't make the tests take longer in case everything completes before, is not bad being a bit tolerant if the machines are busy.

A nice thing that we could do is see if we can override it with an environment variable, so that we can tweak it everywhere without a commit and have a more 'fail-fast' value locally.

But honestly unless it would cause an issue that should be a sensible value.

PR #225, opened 2020-06-13
Connect to DB in URI initially

aherlihy:
- Add an initialDb field to the service providers and set `db` to that DB on connect. Also add `DEFAULT_DB` to service provider core so that we have one place where we set the initial db.
- move cli-options and uri-generator into the core service provider because they both need to be used when calling `connect` or `new Mongo`, otherwise `connect('localhost:27017')` will fail since the driver doesn't accept that URI. 
- for the case `connect(<uri>/<db>)`, to avoid having to write our own uri parser (and we can't use the node driver one since the URIs need to be called from the shell-api (which is required by browser/etc), we get the initial DB via the service provider. This means that it will connect to the DB set in the URI in both the CLI and Compass.


lrlna:
I think the tests failing on windows, are actual fails.

PR #226, opened 2020-06-13
exit if node version too low

aherlihy:
Should we exit or just warn?

### DIFF
--- a/packages/cli-repl/src/cli-repl.ts
+++ b/packages/cli-repl/src/cli-repl.ts
@@ -25,6 +25,7 @@ import util from 'util';
 import read from 'read';
 import os from 'os';
 import fs from 'fs';
+import semver from 'semver';
 
 /**
  * Connecting text key.
@@ -311,10 +312,14 @@ class CliRepl {
    * The greeting for the shell.
    */
   greet(): void {
-    const { version } = require('../package.json');
+    const { version, engines } = require('../package.json');
     console.log(`Using MongoDB: ${this.internalState.connectionInfo.buildInfo.version}`);
     console.log(`${clr('Using Mongosh Beta', ['bold', 'yellow'])}: ${version}`);
     console.log(`${MONGOSH_WIKI}`);
+    if (semver.lt(process.version, engines.node.replace('^', ''))) {
+      console.log(`WARNING: mismatched node version. Minimum node version required ${engines.node}. Currently using ${process.version}. Exiting...\n\n`);
+      process.exit();
+    }
     if (!this.disableGreetingMessage) console.log(TELEMETRY);
   }
 
--- a/packages/cli-repl/src/cli-repl.ts
+++ b/packages/cli-repl/src/cli-repl.ts
@@ -51,6 +51,7 @@ class CliRepl {
    * Instantiate the new CLI Repl.
    */
   constructor(driverUri: string, driverOptions: NodeOptions, options: CliOptions) {
+    this.verifyNodeVersion();
     this.options = options;
     this.mongoshDir = path.join(os.homedir(), '.mongodb/mongosh/');
     this.lineByLineInput = new LineByLineInput(process.stdin);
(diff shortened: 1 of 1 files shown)


packages/cli-repl/src/cli-repl.ts
```diff
@@ -310,10 +311,14 @@ class CliRepl {
    * The greeting for the shell.
    */
   greet(): void {
-    const { version } = require('../package.json');
+    const { version, engines } = require('../package.json');
     console.log(`Using MongoDB: ${this.internalState.connectionInfo.buildInfo.version}`);
     console.log(`${clr('Using Mongosh Beta', ['bold', 'yellow'])}: ${version}`);
     console.log(`${MONGOSH_WIKI}`);
+    if (semver.lt(process.version, engines.node.replace('^', ''))) {
```
mcasimir, line 318: semver has a already a feature for that:

```
semver.satisfies('1.2.3', '1.x || >=2.5.0 || 5.0.0 - 7.2.3')
```

mcasimir, line 318: Very very minor. What do you think having a function `verifyNodeEngineVersion` either apart from `greet` or used inside `greet`? I feel it would be easier for a reader to know about that functionality, but I may be wrong. :)

packages/cli-repl/src/cli-repl.ts
```diff
@@ -310,10 +311,14 @@ class CliRepl {
...
+      console.log(`WARNING: mismatched node version. Minimum node version required ${engines.node}. Currently using ${process.version}. Exiting...\n\n`);
+      process.exit();
```
mcasimir, line 320: I think you can exit with `1` here

packages/cli-repl/src/cli-repl.ts
lrlna: ^ +1

packages/cli-repl/src/cli-repl.ts
```diff
@@ -310,10 +311,14 @@ class CliRepl {
    * The greeting for the shell.
    */
   greet(): void {
-    const { version } = require('../package.json');
+    const { version, engines } = require('../package.json');
     console.log(`Using MongoDB: ${this.internalState.connectionInfo.buildInfo.version}`);
     console.log(`${clr('Using Mongosh Beta', ['bold', 'yellow'])}: ${version}`);
     console.log(`${MONGOSH_WIKI}`);
+    if (semver.lt(process.version, engines.node.replace('^', ''))) {
+      console.log(`WARNING: mismatched node version. Minimum node version required ${engines.node}. Currently using ${process.version}. Exiting...\n\n`);
```
lrlna, line 319: Maybe let's throw an `MongoshUnsupportedError` here?

Message is really good though -- super clear with node versions, hopefully will lead to less bug reports.

packages/cli-repl/src/cli-repl.ts
lrlna: +1, I'd probably have this either in constructor or setupRepl before we connect to mongodb. I think if we do this after we connect to mongodb, we should make sure we close the connection as well.

lrlna:
(omg sorry for all the comments on such a tiny PR 😅 😅 😅 )

packages/cli-repl/src/cli-repl.ts
aherlihy: Oh great, ty!

aherlihy: Ok ty!

aherlihy:
Never a problem on the many comments, I opened this PR as a way of starting this conversation :)

packages/cli-repl/src/cli-repl.ts
aherlihy: done

aherlihy: done

packages/cli-repl/src/cli-repl.ts
aherlihy: We don't have unsupported, we only have unimplemented, which I think doesn't match this use case. I can change it to `RuntimeError`

packages/cli-repl/src/cli-repl.ts
aherlihy: done

packages/cli-repl/src/cli-repl.ts
aherlihy: It seems satisfies does not support `^` and we'd have to replace the carat anyway

aherlihy: Ok ok sorry, it does work, just forgot that the bool is switched. Done!

PR #227, opened 2020-06-15
Fix for method help

aherlihy:


packages/i18n/src/locales/en_US.js
```diff
@@ -1059,12 +1059,6 @@ const translations = {
               example: 'db.collection.getIndexKeys()',
               parameters: {}
             },
-            getIndices: {
-              link: '',
-              description: '',
-              example: '',
-              parameters: {}
-            },
             getIndexSpecs: {
```
mcasimir, line 1062: Should we spare some time / create tickets to fill all the missing help entries?

PR #229, opened 2020-06-15
Re-add check

aherlihy:
We were also missing linting in a few packages, including the service providers so I had the distinct displeasure of linting all those packages.

I commented out skipped tests because they were warning, up to us if we want a warning-free master or not. As is we have 0 warnings or errors. I'm okay with warnings for things that are TODOs, but we should probably discuss.

Also there's a bug with typescript linting where interfaces and types that are used get registered as undefined, so I changed most interface definitions to `export default interface xxx` which is a workaround.


packages/browser-repl/src/components/utils/inspect.spec.ts
```diff
@@ -38,18 +38,18 @@ describe('inspect', () => {
...
-      ).to.equal('Symbol(123)');
-    });
-
-    it.skip('inspects BigInt', () => {
-      expect(
-        // eslint-disable-next-line no-undef
-        inspect(BigInt(1))
-      ).to.equal('1n');
-    });
+    // it('inspects Symbol', () => {
```
mcasimir, line 41: I think these 2 we can just remove

packages/browser-runtime-core/src/interpreter/interpreter.spec.ts
```diff
@@ -71,15 +71,15 @@ describe('Interpreter', () => {
...
-    it.skip('evaluates object literal after other statements as block', async() => {
-      // TODO: for some reason this is the default behaviour in node repl and devtools,
-      // as is now it wraps everything, which breaks the code.
-      expect(
-        await testEvaluate(
-          ';{x: 1}'
-        )
-      ).to.equal(1);
-    });
+    // it('evaluates object literal after other statements as block', async() => {
```
mcasimir, line 74: this is just a difference we have with the usual node repl or chrome devtools:

in node:

```
> ;{x: 1}
1
> 
```

in browser-repl:

```
;{x: 1}
undefined
```

packages/browser-runtime-core/src/interpreter/interpreter.spec.ts
```diff
@@ -159,24 +159,24 @@ describe('Interpreter', () => {
...
-
-    it.skip('can redeclare a top level function as var', async() => {
-      expect(
-        await testEvaluate(
-          'function sum(a, b) { return a + b; }',
-          'var sum = 1'
-        )
-      ).to.equal(1);
-    });
+    // it('can redeclare a top level function as function', async() => {
```
mcasimir, line 162: this may really be an issue, again a corner case but this looks more realistic:

```
function f() { return 1; }
[Function: f]
function f() { return 2; }
[Function: f]
f()
1
```

Expected:

```
function f() { return 1; }
[Function: f]
function f() { return 2; }
[Function: f]
f()
2
```

packages/browser-runtime-core/src/interpreter/preprocessor/wrap-object-literal.spec.ts
```diff
@@ -18,10 +18,10 @@ describe('wrapObjectLiteral', () => {
...
 
-  it.skip('should not wrap multiple statements', () => {
-    // TODO: for some reason this is the default behaviour in node repl and devtools,
-    // as is now it wraps everything, which breaks the code.
-    expect(wrapObjectLiteral('{x: 2}; {x: 2}')).to.equal('{x: 2}; {x: 2}');
-  });
+  // it('should not wrap multiple statements', () => {
+  //   // TODO: for some reason this is the default behaviour in node repl and devtools,
+  //   // as is now it wraps everything, which breaks the code.
+  //   expect(wrapObjectLiteral('{x: 2}; {x: 2}')).to.equal('{x: 2}; {x: 2}');
```
mcasimir, line 24: super corner case, again we can keep it or drop it, maybe give it a look

packages/cli-repl/src/arg-mapper.spec.ts
```diff
@@ -23,15 +23,15 @@ describe('arg-mapper.mapCliToDriver', () => {
...
-    it.skip('maps to explicitlyIgnoreSession', () => {
-      expect(mapCliToDriver(cliOptions)).to.deep.equal({
-        explicitlyIgnoreSession: true
-      });
-    });
-  });
+  // context('when cli args have disableImplicitSessions', () => {
+  //   const cliOptions: CliOptions = { disableImplicitSessions: true };
+  //
+  //   it('maps to explicitlyIgnoreSession', () => {
```
mcasimir, line 29: have no clue about this one

packages/service-provider-browser/src/stitch-service-provider-browser.integration.spec.ts
```diff
@@ -44,13 +45,13 @@ describe('StitchServiceProviderBrowser [ integration ]', function() {
...
       });
 
-      context('when running against a database', () => {
-        it.skip('it rejects the action', () => {
-          return stitchTransport.aggregate('admin', null, [{ $currentOp: {}}]).catch((err) => {
-            expect(err).to.not.equal(null);
-          });
-        });
-      });
+      // context('when running against a database', () => {
```
mcasimir, line 48: no idea

packages/browser-runtime-core/src/interpreter/interpreter.spec.ts
```diff
@@ -159,24 +159,24 @@ describe('Interpreter', () => {
...
+    //   expect(
+    //     await testEvaluate(
+    //       'function f() { return 1; }',
+    //       'function f() { return 2; }',
+    //       'f()'
+    //     )
+    //   ).to.equal(2);
+    // });
+    //
+    // it('can redeclare a top level function as var', async() => {
```
mcasimir, line 172: this is also kinda common and unexpected

packages/browser-repl/src/components/utils/inspect.spec.ts
aherlihy: done

packages/browser-runtime-core/src/interpreter/interpreter.spec.ts
aherlihy: Oh interesting. It's treating `{}` as a block scope and not as an object...which would explain why if you just put in `{x: 1}` then it will give you a syntax error. We could make a ticket for it but I don't think it should be a blocker.

aherlihy: Here's the ticket: https://jira.mongodb.org/browse/MONGOSH-256

I will keep the tests as `skip` so they will warn as a reminder TODO.

packages/browser-runtime-core/src/interpreter/interpreter.spec.ts
aherlihy: I think it's worth a ticket, but will also be closed when we do the unify eval epic so I'm happy to keep this as a skip so we're reminded that there's some TODO there

packages/browser-runtime-core/src/interpreter/preprocessor/wrap-object-literal.spec.ts
aherlihy: I think this can go under the ticket https://jira.mongodb.org/browse/MONGOSH-256

packages/browser-runtime-core/src/interpreter/interpreter.spec.ts
aherlihy: Will keep it as a skip then and include it in the ticket

packages/service-provider-browser/src/stitch-service-provider-browser.integration.spec.ts
aherlihy: This seems to pass so will keep it for now, unless it breaks later on

aherlihy: Nope, fails on evergreen, so will remove it

aherlihy:
I think it was originally disabled because it was trying to lint the cli-repl/dist/mongosh.js file and it was causing a buffer overflow with no error. Once we ignore that we don't get any more random failures I think.

PR #232, opened 2020-06-16
use mongodb docs link in greet message

lrlna:


packages/cli-repl/src/cli-repl.ts
```diff
@@ -321,7 +321,7 @@ class CliRepl {
    */
   greet(): void {
     const { version } = require('../package.json');
-    console.log(`Using MongoDB: ${this.internalState.connectionInfo.buildInfo.version}`);
+    console.log(`Using MongoDB:      ${this.internalState.connectionInfo.buildInfo.version}`);
```
lrlna, line 324: I _know_ this is super pedantic, but I just _really_ like when all the values in the greeting message align ✨

PR #238, opened 2020-06-19
Changed copy of node version mismatched message.

mmarcon:
We are not checking for a minimum, we are checking if the version matches the one we expect but the error message was still talking about a minimum version required.


lrlna:
Should we actually be checking for a minimum then?

PR #239, opened 2020-06-22
if version is undefined, autocomplete everything

lrlna:
## Description

Fixes autocomplete with --nodb version


mmarcon:
Does it make sense to add a unit test?

lrlna:
Oh woops, the tests didn't get pushed, my bad

PR #240, opened 2020-06-22
MONGOSH-210: Preserve shell state when collapsing

Anemy:
https://jira.mongodb.org/browse/MONGOSH-210

This PR makes it so the shell retains its output when its collapsed and re-opened.
It also adds a hotkey `Command-L`/`Ctrl-L` (Mac/Windows) which clears the output of the shell.

We had an error in `compass-shell` where signatures from the shell api weren't loading correctly, this PR fixes that by moving the signature references into the completer so that they're referenced at runtime, since we build the signatures object from the decorator. https://github.com/mongodb-js/mongosh/pull/240/files#diff-6635c7669eeb4f2b6ed7cfb2857f389eL19

___
Jun-22-2020 13-58-38
Screen Shot 2020-06-22 at 1 56 57 PM

Before merging:
- [x] Manually tested with a large amount of output to ensure its not laggy on collapse/expand. It takes a half second to expand but it works fine.


Anemy:
@aherlihy just heads up, doing a small change on completion signatures here: https://github.com/mongodb-js/mongosh/pull/240/files#diff-6635c7669eeb4f2b6ed7cfb2857f389eL19
I think Irina had done some work on pulling the autocompleter into its own package so we can remove the cli-repl dependency from browser-repl

PR #244, opened 2020-06-26
MONGOSH-240: log out mongosh_version to log file

lrlna:
## Description
- adds `mongosh_version` (from package.json) to `getConnectInfo` function return.
- `mongosh_version` is then sent to analytics and gets logged to logger

### DIFF
--- a/packages/service-provider-server/src/cli-service-provider.ts
+++ b/packages/service-provider-server/src/cli-service-provider.ts
@@ -105,6 +105,7 @@ class CliServiceProvider implements ServiceProvider {
   async getConnectionInfo(): Promise<any> {
     const buildInfo = await this.buildInfo();
     const topology = await this.getTopology();
+    const { version } = require('../package.json');
     let cmdLineOpts = null;
     try {
       cmdLineOpts = await this.getCmdLineOpts();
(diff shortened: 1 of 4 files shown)


packages/service-provider-server/src/cli-service-provider.ts
```diff
@@ -105,6 +105,7 @@ class CliServiceProvider implements ServiceProvider {
   async getConnectionInfo(): Promise<any> {
     const buildInfo = await this.buildInfo();
     const topology = await this.getTopology();
+    const { version } = require('../package.json');
```
mcasimir, line 108: Should we use 1 source of truth for this? Say we would start to publish only packages that are changing this may cause a sneaky bug. Looks good anyway!

lrlna: I did consider that, more specifically keeping the version on `shell-internal-state`. However, that leads to a very weird API where this particular function `getConnectionInfo()` would then take version as a param. This does not make sense.

Alternatively, I have also considered storing it in the cli-service-provider and browser-service-provider, which also did not seem logical.

Finally, I have considered doing this require directly in the module that puts all the connection information together, but that would lead to have to change tests in that module _every_ time we bump the version.

This seemed like the most straightforward approach.

As for your point about publishing packages separately -- I think if we end up doing that we will likely want to split this out of the monorepo and redo our entire build process, and if we do that, this will have to be rewritten anyways.

PR #248, opened 2020-06-30
fix duplicated symbol in compiled binary

mcasimir:
`export const shellApiType = Symbol('...')` is working fine in node js with `npm run start` but in the binary we get actually 2 distinct unique symbols: one used in the class definition and one used in the shell evaluator.

To fix this we use `Symbol.for('...')` with a unique name.

### DIFF
--- a/packages/shell-api/src/enums.ts
+++ b/packages/shell-api/src/enums.ts
@@ -37,7 +37,7 @@ export const DBQuery = {
   Option: DBQueryOption
 };
 
-export const shellApiType = Symbol('shellApiType');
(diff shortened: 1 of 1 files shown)


mmarcon:
Tested on macOS, works well.

lrlna:
I am confused why the regular windows tests are running here -- i am sure we removed that task in master.

aherlihy:
Ah good catch!! Sorry I missed this - I tested with the binary before I made the final switch of adding the second symbol, but going forward will always test with the binary as a final step.

PR #249, opened 2020-07-01
Run e2e tests on compiled binaries

mcasimir:
Run e2e tests on compiled binaries

### DIFF
--- a/packages/cli-repl/test/test-shell.ts
+++ b/packages/cli-repl/test/test-shell.ts
@@ -18,7 +18,6 @@ export class TestShell {
 
   static start(options: {
     args: string[];
-    executablePath?: string;
   } = { args: [] }): TestShell {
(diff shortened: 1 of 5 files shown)


packages/cli-repl/test/test-shell.ts
```diff
@@ -8,22 +8,31 @@ import stripAnsi from 'strip-ansi';
 import assert from 'assert';
 
 const PROMPT_PATTERN = /^> /m;
-const ERROR_PATTERN = /Thrown:\n([^>]*)/m;
```
aherlihy, line 11: Should this go in a separate PR since it's a separate issue?

mcasimir: Without that tests break when running on the compiled binary since the node version is  slightly different from the one downloaded in the CI vm

PR #251, opened 2020-07-15
MONGOSH-268: Fix autocompletion in compass-shell

Anemy:
https://jira.mongodb.org/browse/MONGOSH-268

Small fix: the uri passed to the service provider is optional, however when we were fetching connection info to have the mongo version for autocomplete we were running string operations on the uri, which were erroring out. Added a small test to catch in future. Was cool to learn more of the inner workings of mongosh.

I think in future we could make this a bit more streamlined by passing the connection information in from compass into the browser repl/shell api internal state, so we can avoid the extra fetching of connection info inside of mongosh. 

Auto completion working in compass-shell:
Jul-15-2020 14-08-15

Was thinking about pulling telemetry into this ticket, but since it involves changes in compass metrics also I opened a new ticket and can work on that: https://jira.mongodb.org/browse/COMPASS-4351

Once we merge this we can have a release of mongosh and pull in the new compass-shell version in compass to pull it into the next compass beta release for 1.22.0

### DIFF
--- a/packages/service-provider-core/package.json
+++ b/packages/service-provider-core/package.json
@@ -14,6 +14,8 @@
   "scripts": {
     "compile-ts": "tsc -p tsconfig.json",
     "prepublish": "npm run compile-ts",
+    "test": "mocha -r \"../../scripts/import-expansions.js\" --timeout 60000 --colors -r ts-node/register \"./src/**/*.spec.ts\"",
+    "test-ci": "mocha -r \"../../scripts/import-expansions.js\" --timeout 60000 -r ts-node/register \"./src/**/*.spec.ts\"",
     "lint": "eslint \"**/*.{js,ts,tsx}\"",
     "check": "npm run lint"
   },
--- a/packages/service-provider-server/src/cli-service-provider.ts
+++ b/packages/service-provider-server/src/cli-service-provider.ts
@@ -69,7 +69,7 @@ class CliServiceProvider implements ServiceProvider {
   }
 
   private readonly mongoClient: MongoClient;
-  private readonly uri: string;
+  private readonly uri: string | undefined;
 
   /**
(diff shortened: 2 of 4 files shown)


packages/service-provider-server/src/cli-service-provider.ts
```diff
@@ -112,13 +112,15 @@ class CliServiceProvider implements ServiceProvider {
       // eslint-disable-next-line no-empty
     } catch (e) {
     }
+
     const connectInfo = getConnectInfo(
-      this.uri,
+      this.uri ? this.uri : '',
```
lrlna, line 117: We should perhaps add a test for an empty uri in `service-provider-core/src/connect-info.spec.ts`, just to make sure the rest of the information from this function comes out ok.

Anemy: Good call - added a test. I also noticed we didn't have a `test` script in the `service-provider-core` package. Is that intended? I added one, commented below.

packages/service-provider-core/package.json
```diff
@@ -14,6 +14,8 @@
   "scripts": {
     "compile-ts": "tsc -p tsconfig.json",
     "prepublish": "npm run compile-ts",
+    "test": "mocha -r \"../../scripts/import-expansions.js\" --timeout 60000 --colors -r ts-node/register \"./src/**/*.spec.ts\"",
+    "test-ci": "mocha -r \"../../scripts/import-expansions.js\" --timeout 60000 -r ts-node/register \"./src/**/*.spec.ts\"",
```
Anemy, line 18: Didn't see these tests being run with the lerna tests, so I added the test script.

lrlna: nice! i don't think this was intentionally left out.

PR #256, opened 2020-07-27
MONGOSH-285: build debian on evergreen

lrlna:
## Description

This PR adds Debian build variant to our build process to package debian tarballs.

There are a few key additions:
1. [`pkg-deb`](github.com/mongodb-js/pkg-deb) is used to create a debian tarball
2. Debian tarball is only created on debian build-variant
3. Debian tarball is created for `amd64` architecture. This is reflected in the [`control` file](https://www.debian.org/doc/debian-policy/ch-controlfields.html) of the packaged archive, and also in the tarball name (as per debian package spec). I have attempted to pass in `all` as architecture type to the `control` file, but that didn't not get installed properly on my debian docker image. I went with `amd64`.
4. Introduces `rootDir` to config object. This is used by `pkg-deb` to currently obtain project's `package.json`. It will also obtain the `README`, `CHANGELOG` and `LICENSE` files in the future from the same `rootDir`.
5. Introduces `buildVariant` to config object. This is used to determine which tarball to create. The difference between `platform` and `buildVariant` is as follows:

    - __platform:__ result of `os.platform()` which can return `'aix'`, `'darwin'`, `'freebsd'`, `'linux'`, `'openbsd'`, `'sunos'`, and `'win32'`. This is used to create binaries. All linux binaries will always be the same, so we do not need to distinguish between different types of linuxes.
    - __buildVariant:__ this is based on evergreen's BUILD_VARIANT, i.e. the build variants we define in our `.evergreen.yml`. Currently we have: `darwin` (renamed in this PR), `linux` (running ubuntu, renamed in this PR), win32 (running windows_ps, renamed in this PR), and `debian` (introduced in this PR). We build different tarballs based on the build variant.

6. Adds documentation to `@mongosh/build` readme.

(_I will squash commits before merging!_)


packages/build/src/tarball.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+      fs.stat(expectedTarball, (error) => {
+        expect(error).to.equal(null);
+        done();
+      });
+    });
+  });
+
+  // macos and windows build variants does not come with installed 'dpkg' so
+  // this test fails. Run this test only on linux.
+  if (os.platform() === BuildVariant.Linux) {
```
mcasimir, line 67: Why do we compare a platform with a BuildVariant?

Shouldn't this be `process.env.BUILD_VARIANT === BuildVariant.Linux`

packages/build/src/tarball.ts
```diff
@@ -0,0 +1,130 @@
...
+ * Get the path to the tarball.
+ *
+ * @param {string} outputDir - The output directory.
+ * @param {string} platform - The platform.
+ * @param {string} version - The version.
+ *
+ * @returns {string} The path.
+ */
+export const tarballPath = (outputDir: string, platform: string, version: string): string => {
+  if (platform === BuildVariant.Linux) {
```
mcasimir, line 17: should we rename platform to buildVariant here?

packages/build/src/tarball.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+  });
+
+  describe('.tarball', () => {
+    const platform = os.platform();
+    const version = '1.0.0';
+    const expectedTarball = tarballPath(__dirname, platform, version);
+    const inputFile = path.join(__dirname, '..', 'examples', 'input.js');
+    const rootDir = path.join(__dirname, '../../..');
+
+    before(() => {
```
mcasimir, line 118: super minor: there is a bit of mix different async styles here, where possible it would be great if we can stick to one across the project

packages/build/src/compile-and-zip-executable.ts
```diff
@@ -1,8 +1,8 @@
 import compileExec from './compile-exec';
-import { zip, ZipFile } from './zip';
+import { tarball, TarballFile } from './tarball';
```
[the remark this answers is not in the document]
lrlna: `createTarball` is also good! Figured I'd try to make a verb out of `tarball` first, but clearly I am not fooling anyone 😂 😂 😂

packages/build/src/tarball.spec.ts
lrlna: for *some* reason `process.env.BUILD_VARIANT` is not available here. I agree, this is a bit of hacky statement, and it's on actually dependent on the build variant, but the current platform (i.e. it can run on any linux). I can do either of these two instead:
```js
if (os.platform() === 'linux') {
```
_or_:
```
import Platform from './platform';
if (os.platform() === Platform.Linux
```

packages/build/src/tarball.ts
lrlna: yes!

packages/build/src/tarball.spec.ts
lrlna: ooooo you know I just did a `mv` on this file, and didn't actually look through previous tests that were here. Will change it up to what we are doing everywhere else.

package.json
```diff
@@ -72,7 +74,7 @@
     "browserify": "^16.5.0",
     "chai": "^4.2.0",
     "cross-env": "^6.0.3",
-    "electron-notarize": "mongodb-js/electron-notarize",
+    "electron-notarize": "git+https://github.com/mongodb-js/electron-notarize.git",
```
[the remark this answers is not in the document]
lrlna: hmmm this change I think actually came from a rebase I did in the beginning of the week. I guess master no longer has this. I'll remove.

packages/build/src/tarball.spec.ts
lrlna: did you mean for something like this instead:
```js

  describe('.tarball', () => {
    const buildVariant = process.env.BUILD_VARIANT; 
    const version = '1.0.0';
    const expectedTarball = tarballPath(__dirname, buildVariant, version);
    const inputFile = path.join(__dirname, '..', 'examples', 'input.js');
    const rootDir = path.join(__dirname, '../../..');

    after(async () =>{
      await fs.unlink(expectedTarball);
    });

    it('builds the executable', async () => {
      await createTarball(inputFile, __dirname, buildVariant, version, rootDir);

      let accessErr;
      try {
        fs.access(expectedTarball);
      } catch (err) {
        accessErr = err;
      }

      expect(accessErr).to.equal(null)
    });
  });
```

PR #262, opened 2020-07-29
COMPASS-4351: Add compass shell telemetry

Anemy:
COMPASS-4351

This PR adds telemetry events for the compass shell plugin. When the plugin is expanded it emits a `compass:compass-shell:opened`. The global appRegistry is passed as a message bus option to mongosh so that mongosh log events are emitted in compass land and can be consumed by the metrics (or possibly elsewhere later).

Here's the PR in `compass-metrics` where the events are being passed to the telemetry:  https://github.com/mongodb-js/compass-metrics/pull/147

app registry mongosh

After this is merged we'll want to do a release, we can see with the team if there's anything to be merged soon or if it can happen sooner. Then we can update the version in compass as well.


Anemy:
Looks like its the `ns` in older mongodb versions our tests are looking for. I'll see if it's a quick fix and fix or file a ticket if not.

PR #263, opened 2020-07-30
Speed local setup

mcasimir:
Changes:

1. Only download chromium in CI and use it if necessary.
2. Add few commands to the main package.json:
    - `bootstrap-with-chromium`: always download chromium
    - `bootstrap-without-compass`: do not link and install compass-shell
    - `bootstrap-ci`: used in evergreen, run lerna and npm commands in 'ci' mode: without altering package-lock.json
3. `bootstrap` will not run the `prepublish` script anymore. This means that `compile-ts` and webpack are not run automatically on every install.


mcasimir:
> SGTM, one question: is there a reason we don't want to compile-ts when we bootstrap? It pretty much needs to happen each time you bootstrap since you wouldn't bootstrap without having some code changes or wanting to build

`compile-ts` was run as part of prepublish that's why is gone. `npm run start[-*]` also does `compile-ts`, should not be an issue, but we can do a `postbootstrap: "compile everything except browser"`. Let me know.

PR #264, opened 2020-07-30
MONGOSH-180 - Support user management commands

aherlihy:
Two things I want to double check
- it won't ever be a problem that we are passing the options originally set by the user when the connect via the CLI (the auth ones are overwritten)
- command history hasn't been saving the ones with passwords, however, it's been a bit funky and has been not saving regular commands as well


mcasimir:
> * it's been a bit funky and has been not saving regular commands as well

Is it totally random or it doesn't save a specific set of commands that would be safe? We could just create a separate ticket

aherlihy:
It seems almost totally random

aherlihy:
But it never shows the commands with password so not critical

PR #266, opened 2020-07-31
MONGOSH-303 - Role management DB methods

aherlihy:


packages/cli-repl/test/e2e-auth.spec.ts
```diff
@@ -359,6 +386,309 @@ describe('Auth e2e', function() {
         });
       });
     });
+    describe('role management', () => {
+      describe('createRole', async() => {
```
mcasimir, line 390: Isn't it better to put these tests as integration tests in shell api?

aherlihy: These are the same as the integration tests for authentication, so would you want me to also move the existing auth tests out of this file? The integration tests in the shell API do virtually the same thing as these tests since it creates a CliServiceProvider. Why not do more full integration tests? We're just testing more of the path, which I think is almost always better. What exactly are the "other factors"?

mcasimir: The one about arguments belongs here. However e2e tests are affected by presentational logic too. Right now all seems quite stable, but they can get brittle / fuzzy and slow (each test does few polls). For example when we did the multiline I had to rewrite few of them. Is a tread-off.

aherlihy: But shouldn't all of those things get tested for this as well? Mostly I think it's important to test against the built executable, which these tests do but the shell API tests do not

aherlihy: What makes it brittle/slow (what do you mean by fuzzy?), just that it goes over the wire? I can move these and the auth tests into the shell-api integration, although I wish you had told me in the last PR where I put tests exactly like these ones so I wouldn't have to now update 600 lines of tests :/

aherlihy: I'm not going to move these tests because they should go with the auth tests, but I will do future tests elsewhere

PR #267, opened 2020-07-31
MONGOSH-305 - Support database and collection helper methods

aherlihy:
Not implementing Database.getSisterDB , see https://jira.mongodb.org/browse/SERVER-46231
Collection.getCollection is not mentioned in docs, plus the behavior in the old shell is weird: db.abc.getCollection('def') --> Collection("abc.def")

### DIFF
--- a/packages/shell-api/src/database.spec.ts
+++ b/packages/shell-api/src/database.spec.ts
@@ -145,6 +145,12 @@ describe('Database', () => {
       });
     });
 
+    describe('getName', () => {
+      it('returns the name of the DB', async() => {
+        expect(database.getName()).to.equal('db1');
+      });
+    });
+
     describe('runCommand', () => {
       it('calls serviceProvider.runCommand on the database', async() => {
         await database.runCommand({ someCommand: 'someCollection' });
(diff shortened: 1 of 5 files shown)


packages/shell-api/src/database.spec.ts
```diff
@@ -145,6 +145,12 @@ describe('Database', () => {
       });
     });
 
+    describe('getName', () => {
```
mcasimir, line 148: no test for getMongo?

mcasimir:
Do we need to add help?

packages/shell-api/src/database.spec.ts
aherlihy: There's no test for `db.getMongo`, and even the test for `getName` is kind of stupid since it's just returning an attribute of the class. I didn't want to write tests just for the sake of having a test, if it doesn't test anything worthwhile.

aherlihy: I can add the test if you like

mcasimir: Shouldn't hurt, is part of the spec, takes no time to add.

aherlihy: Sure, although I disagree that it's useful :)

aherlihy:
Added help

PR #274, opened 2020-08-03
fix: remove keytar, fix history when dir exists

mcasimir:
- Removes the need to have keytar as dependency of compass-shell
- Fixes history not being saved when the history folder exists


packages/compass-shell/package.json
```diff
@@ -119,8 +118,6 @@
   "less-loader": "^4.0.5",
   "mini-css-extract-plugin": "^0.8.0",
   "mocha-webpack": "^2.0.0-beta.0",
-  "mongodb-connection-model": "^14.4.3",
-  "mongodb-data-service": "^16.5.0",
```
Anemy, line 123: 🍾

packages/compass-shell/src/modules/history-storage.js
```diff
@@ -18,7 +18,14 @@ export class HistoryStorage {
    */
   async save(history) {
     const targetDir = path.dirname(this.filePath);
-    await fs.promises.mkdir(targetDir, { recursive: true });
+    try {
+      await fs.promises.mkdir(targetDir, { recursive: true });
+    } catch (error) {
+      if (error.code !== 'EEXIST') {
+        throw error;
```
lrlna, line 25: oh dang! really good catch!!

PR #276, opened 2020-08-04
MONGOSH-224: fix windows tests

mcasimir:
Changes:

- Just download and run mongo on windows instead of using `mongo-runner`
- Add `skipIfServerVersion(<semver-expr>)` helper to skip tests in case the mongodb version under test does not apply to the test suite `ie. skipIfServerVersion('< 4.4')`.

### DIFF
--- a/.evergreen/DownloadAndRunMongod.ps1
+++ b/.evergreen/DownloadAndRunMongod.ps1
@@ -1,0 +1,54 @@
+$platform = "win32"
+$version = "mongodb-win32-x86_64-2012plus-4.2.8"
+$port = "27018"
+
+$platformPath = "$PSScriptRoot\mongodb\$platform"
+$versionPath = "$platformPath\$version"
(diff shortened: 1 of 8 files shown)


.evergreen/DownloadAndRunMongod.ps1
```diff
@@ -0,0 +1,54 @@
...
+$port = "27018"
+
+$platformPath = "$PSScriptRoot\mongodb\$platform"
+$versionPath = "$platformPath\$version"
+$binaryPath = "$versionPath\bin\mongod.exe"
+$dataPath = "$versionPath\data\db"
+$logDir = "$versionPath\logs"
+
+if (!(Test-Path $binaryPath)) {
+  $url = "https://fastdl.mongodb.org/$platform/$version.zip";
```
lrlna, line 12: This is a *really* nice trick!

PR #277, opened 2020-08-06
MONGOSH-304 - Support server stats and info methods

aherlihy:


packages/cli-repl/src/format-output.ts
```diff
@@ -71,6 +75,12 @@ function formatDatabases(output): string {
   return textTable(tableEntries, { align: ['l', 'r'] });
 }
 
+function formatStats(output): string {
```
lrlna, line 78: I don't think there is a test for this result type yet.

packages/browser-repl/src/components/shell-output-line.tsx
```diff
@@ -56,6 +56,13 @@ export class ShellOutputLine extends Component<ShellOutputLineProps> {
       return <ShowDbsOutput value={value} />;
     }
 
+    if (type === 'StatsResult') {
```
lrlna, line 59: I think this also needs a test.

packages/cli-repl/src/format-output.ts
```diff
@@ -71,6 +75,12 @@ function formatDatabases(output): string {
...
+  return Object.keys(output).reduce((str, c) => {
+    return `${str}\n${c}\n${inspect(output[c])}\n---\n`;
```
lrlna, line 80: up to you if you want this suggestion:
```suggestion
    return `${str}\n${clr(c, ['bold', 'yellow'])}\n\n${inspect(output[c])}\n---\n`;
```

aherlihy: Oh woah I didn't realize I could just press 'commit suggestion', that's a sick feature! I just added it by hand.

PR #278, opened 2020-08-06
MONGOSH-300 - Free monitoring methods

aherlihy:


packages/shell-api/src/database.ts
```diff
@@ -626,4 +627,74 @@ export default class Database extends ShellApiClass {
...
+    }
+    if (result.state !== 'enabled') {
+      const urlResult = await this._mongo._serviceProvider.runCommand(
+        ADMIN_DB,
+        {
+          getParameter: 1,
+          cloudFreeMonitoringEndpointURL: 1
+        }
+      );
+      return `Unable to get immediate response from the Cloud Monitoring service. Please check your firewall settings to ensure that mongod can communicate with '${urlResult.cloudFreeMonitoringEndpointURL || '<unknown>'}'`;
```
aherlihy, line 696: Weird, but what the old shell does

PR #282, opened 2020-08-11
Add db.serverStatus()

aherlihy:


packages/shell-api/src/database.ts
```diff
@@ -688,6 +688,21 @@ export default class Database extends ShellApiClass {
...
+  async serverStatus(opts = {}): Promise<any> {
+    this._emitDatabaseApiCall('serverStatus', { options: opts });
+    const result = await this._mongo._serviceProvider.runCommand(
+      ADMIN_DB,
+      {
+        serverStatus: 1, ...opts
+      }
+    );
+    if (!result || !result.ok) {
+      throw new MongoshRuntimeError(`Error running command serverStatus ${result ? result.errmsg || '' : ''}`);
```
mcasimir, line 701: This `Error running command serverStatus ${result ? result.errmsg || '' : ''}` looks like a pattern, probably we can factor it either as a function ie. `ensureResultOk`, or with a dedicated error, something like:
`MongoshCommandExecutionError(command, result)` or both.

aherlihy: We're going to revisit this when we decide of we want ok: 0 or an error to be thrown, and for what methods. Once we decide. we can easily refactor! But since it's all gonna get rewritten with https://jira.mongodb.org/browse/MONGOSH-320 I'm going to hold off on this right now

PR #283, opened 2020-08-11
MONGOSH-296 - Bulk API

aherlihy:


packages/shell-api/src/bulk.ts
```diff
@@ -0,0 +1,217 @@
...
+import {
+  Document,
+  WriteConcern
+} from '@mongosh/service-provider-core';
+import { assertArgsDefined } from './helpers';
+import { BulkWriteResult } from './result';
+
+@shellApiClassDefault
+export class BulkFindOp {
+  _innerFind: any;
```
mcasimir, line 13: if _innerFind and _innerBulk are coming from service provider probably making that clear would help who reads to orientate a bit: `_serviceProviderBulkOp`, `_serviceProviderBulkOpFind`

packages/shell-api/src/bulk.ts
```diff
@@ -0,0 +1,217 @@
...
+  }
+}
+
+
+@shellApiClassDefault
+@hasAsyncChild
+export default class Bulk extends ShellApiClass {
+  _mongo: Mongo;
+  _collection: any; // to avoid circular ref
+  _batchCounts: any;
```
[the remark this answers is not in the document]
mcasimir: would be great to have a `type BatchCounts` with a comment on what's that for

packages/shell-api/src/bulk.ts
```diff
@@ -0,0 +1,217 @@
...
+
+  tojson(): any {
+    const batches = this._innerBulk.s.batches.length + Number(this._innerBulk.s.currentBatch !== null);
+    return {
+      ...this._batchCounts,
+      nBatches: batches
+    };
+  }
+
+  toString(): any {
```
mcasimir, line 199: isn't it safe to say this returns a `string`?

packages/shell-api/src/bulk.ts
```diff
@@ -0,0 +1,217 @@
...
+      nInsertOps: 0,
+      nUpdateOps: 0,
+      nRemoveOps: 0
+    };
+  }
+
+  /**
+   * Internal method to determine what is printed for this class.
+   */
+  _asPrintable(): string {
```
mcasimir, line 133: `tojson` looks like is returning an object, is it right that the return type is a string? 

by the way, if you have a type you know is going to be an object and just don't know its  shape you could use `Record`:

ie.
```
tojson(): Record<string, any> { ... }
```

although we could have few utility types for that to use like we've done for `Document`.

If you know which keys, you can restrict that too:

```
type BulkCounters = Record<'nInsertedOps' | 'nInsertedOps' | '...', number>
```

packages/shell-api/src/bulk.ts
```diff
@@ -0,0 +1,217 @@
...
+
+  insert(document: Document): Bulk {
+    this._batchCounts.nInsertOps++;
+    assertArgsDefined(document);
+    this._innerBulk.insert(document);
+    return this;
+  }
+
+  tojson(): any {
+    const batches = this._innerBulk.s.batches.length + Number(this._innerBulk.s.currentBatch !== null);
```
mcasimir, line 192: I find this concise but hiding a bit of the intent, wonder if this could be more clear?

```
let nBatches = this._innerBulk.s.batches.length;

if (this._innerBulk.s.currentBatch !== null) {
   this._innerBulk.s.batches.length += 1;
}
``` 

`this._innerBulk.s.currentBatch !== null` would return true also if `this._innerBulk.s.currentBatch` is `undefined`, is it intentional?

packages/shell-api/src/bulk.ts
```diff
@@ -0,0 +1,217 @@
...
+      throw new MongoshInvalidInputError('A bulk operation cannot be re-executed');
+    }
+
+    if (this._innerBulk.s !== undefined && Array.isArray(this._innerBulk.s.batches)) {
+      this._batches = [...this._innerBulk.s.batches];
+      this._batches.push(this._innerBulk.s.currentBatch);
+    }
+    const result = await this._innerBulk.execute();
+    this._executed = true;
+    this._emitBulkApiCall('execute', { operations: this._batches, writeConcern: writeConcern });
```
mcasimir, line 166: is it safe to log the operations?

packages/shell-api/src/bulk.ts
```diff
@@ -0,0 +1,217 @@
...
+  }
+
+  @returnsPromise
+  async execute(writeConcern?: WriteConcern): Promise<BulkWriteResult> {
+    if (this._executed) {
+      throw new MongoshInvalidInputError('A bulk operation cannot be re-executed');
+    }
+
+    if (this._innerBulk.s !== undefined && Array.isArray(this._innerBulk.s.batches)) {
+      this._batches = [...this._innerBulk.s.batches];
```
mcasimir, line 161: do we need to keep track of the batches in a class property or we should rather have a `_getAllBatches` method so we always use the service provider state as source of truth?

packages/shell-api/src/bulk.ts
```diff
@@ -0,0 +1,217 @@
...
+      coll: this._collection._name,
+      arguments: methodArguments
+    });
+  }
+
+  @returnsPromise
+  async execute(writeConcern?: WriteConcern): Promise<BulkWriteResult> {
+    if (this._executed) {
+      throw new MongoshInvalidInputError('A bulk operation cannot be re-executed');
+    }
```
mcasimir, line 158: to be precise in order to avoid race conditions in scripts `this._executed` should be set right here in the same "synchronous" context where is checked, or in any case before calling any async operation.

packages/cli-repl/src/cli-repl.ts
```diff
@@ -298,11 +298,12 @@ class CliRepl {
     // This checks for error instances.
     // The writer gets called immediately by the internal `this.repl.eval`
     // in case of errors.
-    if (result && result.message && typeof result.stack === 'string') {
-      this.bus.emit('mongosh:error', result);
+    if (result && (result.message || result.errmsg)) {
```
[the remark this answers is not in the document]
mcasimir: if that's the case probably would have been great to have this in a test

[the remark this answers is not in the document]
aherlihy: Yeah, that would pass. I was trying to figure out how to actually determine if something is an error, but the overlap between driver errors (specifically the bulk one) and regular errors is almost nothing. The driver error looks like:
```
{
  index: 0,
  code: 9,
  errmsg: 'multi update is not supported for replacement-style update',
  op: { q: {}, u: { x: 2 }, multi: true, upsert: false }
}
```
While the JS error has `name`, `stack`, `message`, and `code`. I suppose I could be more exclusive and check for exactly the driver error format or exactly the js error? It seems odd to me that we'd ever be in a situation where we have to determine if something is an error via its attributes, it seems like it should be a different code path. Maybe I should see if I can refactor a little to avoid the problem entirely?

packages/shell-api/src/bulk.ts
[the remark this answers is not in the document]
aherlihy: Ah yes, so for _innerBulk it would be a node driver type but the shell-api class can't import that directly. So instead you are saying we should define interfaces in the service-provider-core that indicate what the innerBulk/batches types should be? I can definitely do that

packages/shell-api/src/bulk.ts
```diff
@@ -0,0 +1,217 @@
...
+@hasAsyncChild
+export default class Bulk extends ShellApiClass {
+  _mongo: Mongo;
+  _collection: any; // to avoid circular ref
+  _batchCounts: any;
+  _executed: boolean;
+  _batches: any;
+  _innerBulk: any;
+
+  constructor(collection, innerBulk) {
```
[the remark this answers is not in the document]
aherlihy: Good idea!

packages/shell-api/src/bulk.ts
aherlihy: Yes, sorry, I thought it returned JSON for awhile which would have been confusing!

packages/shell-api/src/bulk.ts
aherlihy: Sure, I can definitely clarify the code here. If currentBatch is undefined then there's an error, so maybe we should just check that directly. Null is valid, just means there's no batch currently being created.

packages/shell-api/src/bulk.ts
aherlihy: Good point, I'll remove it

packages/shell-api/src/bulk.ts
aherlihy: We can get rid of this entirely because the driver will throw the same error. Or we could just set it to false in the constructor?

packages/shell-api/src/bulk.ts
aherlihy: You can only call getOperations after a bulk op is executed, but the bulk op class for the driver deletes the batches attribute after it's executed. So we need to save the batches locally right before we execute. Just a difference from driver to shell api.

packages/shell-api/src/bulk.ts
aherlihy: Can do!

packages/shell-api/src/bulk.ts
aherlihy: Oh cool, yes `Record` will work perfectly for this case. I should be having less `any` in my code, thanks for all the suggestions to improve my typescript!

packages/cli-repl/src/cli-repl.ts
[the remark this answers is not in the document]
aherlihy: Ah, it's a node repl thing. Why anyone thought that was a good idea is a little mystifying, oh well. I'll check for either/or instead of something that works for both.

packages/shell-api/src/bulk.ts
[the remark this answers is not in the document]
aherlihy: (I can't annotate for `collection` because the Collection class has to use the `Bulk` type for annotations, so we'd end up with a circular ref. But I can set the type of innerBulk)

packages/shell-api/src/bulk.ts
aherlihy: I removed the error check but kept it it for modifying the batch count - it wouldn't make sense if calling execute twice changed the batch count. Let me know if you think there's a problem with that

mcasimir: I think having `this._executed` there makes sense. my point was that we should not wait to set it to true though:

``` js
 async execute(writeConcern?: WriteConcern): Promise<BulkWriteResult> {
    if (this._executed) {
      throw new MongoshInvalidInputError('A bulk operation cannot be re-executed');
    }
    this._executed = true;
   ...
```

which is different from:

``` js
    if (this._executed) {
      throw new MongoshInvalidInputError('A bulk operation cannot be re-executed');
    }

    await doSomething();  // <--

    this._executed = true;
```

aherlihy: But shouldn't `this._executed` only be set true if the driver successfully executes the bulk op? That's why it's after the `await`

PR #286, opened 2020-08-13
MONGOSH-122: submit to barque

lrlna:
## Description

This PR introduces [curator](https://github.com/mongodb/curator), a tool maintained by the build team to upload bundled and packaged assets to `barque`, MongoDB's PPA service. 

Package passed to `curator` is from Evergreen's S3 bucket. `Curator` is only invoked when running on `Debian` , `Ubuntu`, and `Redhat` (rhel) build variants for the moment. `BARQUE_USERNAME` and `BARQUE_API_KEY` are currently stored in Evergreen and are extracted the same way as other `env vars`. Similar strategy is used for `NOTARY_TOKEN` , which are necessary for signing the package for a specific mongodb version. We are currently only signing for `4.4.0`.

The above variables, `BARQUE_USERNAME`, `BARQUE_API_KEY`, `NOTARY_TOKEN`, `NOTARY_KEY_NAME` are read by `curator` as env vars (using `child_process.execFile` `env` option), so we do not have pass them on as arguments to the `curator` CLI. Not passing them as arguments also prevents them from being leaked to our task logs.

## Open Questions
1. `repo-config.yml` is currently only supported for 1 ubuntu, 1 debian, 1 redhat platforms. This should perhaps be expanded to how [mongo-tools](https://github.com/mongodb/mongo-tools/blob/08dda285d3e2dbf3e2c43af3554545c4e5eb387f/etc/repo-config.yml#L63) are handling this.

3. This is currently only building for `amd64` architecture for `Debian` and `Ubuntu`, and `x86_64` for Redhat. (can also be `x86_64` very easily). I am not sure whether I should be reading current build variant's architecture, or specifically setting different architectures based on what is described in `repo-config.yml`.

4. We are only building for `4.4.0` mongodb version. I am not sure whether we should expand to other versions. Depending on which ones we have to support, we will have to adjust the `NOTARY_TOKEN` and `NOTARY_KEY_NAME`, as those only work for `4.4.0` series. (@mmarcon)

5. We are only building for `org` mongodb edition. I am not sure if we should only be building for enterprise, or both! (@mmarcon)

6. Currently each linux-like `Build Variant` will make its own `barque` request (via `curator`). Meaning, `ubuntu` will send off `ubuntu`, `debian` will send off `debian`. This is useful, as we are able to right away grab the applicable s3 path for the package, and send that with the request. I am not sure if we should be considering doing them all at once. It'll be a bit more work to get all the correct s3 paths, but if that's what others are doing, we should do the same. From my (very very basic) understanding of golang, it seems like mongo-tools is also doing it build-variant by build-variant. I could be wrong!


packages/build/src/evergreen.ts
```diff
@@ -29,9 +29,15 @@ const uploadArtifactToEvergreen = (artifact: string, awsKey: string, awsSecret:
     Key: key,
     Body: fs.createReadStream(artifact)
   };
+
   console.log(`mongosh: uploading ${artifact} to evergreen bucket:`, BUCKET, key);
   console.log(`mongosh: artifact download url: https://s3.amazonaws.com/${BUCKET}/${key}`);
   return upload(uploadParams, s3);
 };
 
+const getArtifactPath = (project: string, revision: string, artifact: string): string => {
```
mcasimir, line 38: is this more `getArtifactUrl`?

lrlna: That works better

config/repo-config.yml
```diff
@@ -0,0 +1,102 @@
...
+      - apt/debian/dists/stretch/mongodb-org
+
+  - name: debian10
+    type: deb
+    code_name: "buster"
+    bucket: repo.mongodb.org
+    edition: org
+    component: main
+    architectures:
+      - amd64
```
lrlna, line 78: With the `repo-config`'s `architectures` field, I am not sure whether I have to send all possible architectures with the `curator` CLI. Suppose the `architectures` here are:
```yml
    architectures:
      - amd64
      - ppc64el
      - s390x
      - i386
```
Do I have to send all of them? Or none at all is also ok?

Additionally, I haven't seen any of the distros to have `x86_64`, is that an implied architecture that all of them will build to?

mmarcon:
@lrlna A few questions:

1. Aren't `amd64` and `x86_64` the same thing? https://en.wikipedia.org/wiki/X86-64
2. you say "We are only building for 4.4.0 mongodb version": what does that mean exactly? Same question for "We are only building for org mongodb edition".

I feel we should probably have a quick chat about all of this, it's probably easier than discussing it in comments.

packages/build/src/barque.ts
```diff
@@ -0,0 +1,145 @@
...
+   * @param {string} repoConfig - Path to repo-config.yml used to uplaod assets
+   * to appropriate distros.
+   *
+   * @returns {Promise} The promise.
+   */
+  async execCurator(
+    curatorDirPath: string,
+    tarballURL: string,
+    repoConfig: string): Promise<any> {
+    exec(
```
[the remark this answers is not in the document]
mcasimir: A related question, shouldn't test fail?

scripts/import-expansions.js
```diff
@@ -27,10 +27,17 @@ const importExpansions = () => {
   const data = fs.readFileSync(expansionsPath, 'utf8');
   const expansions = YAML.parse(data);
   Object.keys(expansions).forEach((key) => {
+    // we need to rename signing auth token to NOTARY_TOKEN
+    // this is used for notary signing when uploading to barque
+    if (key === 'signing_auth_token_44') {
```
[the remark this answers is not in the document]
mmarcon: I'd assume so since above @lrlna said:
> we will have to adjust the NOTARY_TOKEN and NOTARY_KEY_NAME, as those only work for 4.4.0 series

I think as a first round, I'd be ok with just publishing to`4.4`-`org`. However, soon enough I think we should publish to all 4+ repos and once we go GA also to `enterprise`.

packages/build/src/build-and-release.ts
```diff
@@ -1,12 +1,15 @@
...
+import { Barque } from './barque';
+import Config from './config';
 
 export default async function buildAndRelease(
   config: Config,
   githubRepo: GithubRepo,
+  barque: Barque,
   compileAndZipExecutable: (Config) => Promise<TarballFile>,
-  uploadToEvergreen: (artifact: string, awsKey: string, awsSecret: string, project: string, revision: string) => Promise<void>,
+  uploadToEvergreen: (config: string, awsKey: string, awsSecret: string, project: string, revision: string) => Promise<void>,
```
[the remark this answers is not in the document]
lrlna: not intentional! good catch!

scripts/import-expansions.js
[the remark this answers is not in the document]
lrlna: Yep, definitely have to update this every time there is a new server version. We will also have to update keys on evergreen and those need to obtained from BUILD team as they manage those tokens. Additionally, if we are supporting mutliple versions, we will have to support and store mutliple tokens.

From what I can tell, all the teams have to go through this process.

packages/build/src/barque.ts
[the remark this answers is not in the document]
lrlna: oh whoa I have been looking at this for way too long: this should definitely be returned.

scripts/import-expansions.js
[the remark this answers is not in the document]
mcasimir: 👍. I would love to have a bit more context here, at a first glance it looks a lot like an unnecessary pain / something that could be handled differently (not talking about your implementation). Maybe I'll bug you later and try to see if there is something we could ask to coordinate a bit better, for example it would be enough to have a private repo or s3 bucket with all the token / key names that someone would update once and for everyone.

For the scope of this PR, if i've got correctly what is the use of that environment variable .. and considering what max said, I think would be better to do this translation where we use it:

``` js
await exec('${curatorDirPath}/curator ...', {
  env: {
     NOTARY_KEY_NAME: 'server-4.4',
     NOTARY_TOKEN: process.env.SIGNING_AUTH_TOKEN_44
  }
});
```

That would also prevent issues if by accident we would define a `notary_token` expansion, which would override or be overridden the other one.

rcsanchez97:
@lrlna The changes look good to me.  The only comment I have relates to your open question 5.  For libmongocrypt our approach is similar to the approach you took here.  The primary difference achieved by aggregating the publication is that the publication to barque only happens if all of the various variants successfully build their packages for a given Evergreen build.  If it is more desirable to have as many packages published as possible for any given build, then your current approach is the best.  If, however, it is more desirable to only publish when all variants have built their respective packages (and by implication this means that you wait for slow variants before publishing and any failure to build packages on any variant results in no publication), then aggregating makes more sense.  One additional consideration is that since mongosh is closely tied to the server, you might consider adopting their approach (if it is different from the current) for consistency.

scripts/import-expansions.js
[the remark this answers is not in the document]
lrlna: @mcasimir I added your suggestion to this PR. I think it works well, but lmk if you were intending for something else.

packages/build/src/barque.ts
```diff
@@ -0,0 +1,137 @@
...
+   * @param {Config} config - Config object.
+   *
+   * @returns {Promise} The promise.
+   */
+  async releaseToBarque(tarballURL: string): Promise<any> {
+    const repoConfig = path.join(this.config.rootDir, 'config', 'repo-config.yml');
+
+    if (this.config.platform === Platform.Linux) {
+      try {
+        await this.execCurator(tarballURL, repoConfig);
```
lrlna, line 58: try/catching this here so I can test this function easier. Still throw an error in the catch, so we can abort the release process if this is unsuccessful.

packages/build/src/barque.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+      evgAwsKey: 'evgAwsKey',
+      evgAwsSecret: 'evgAwsSecret',
+      downloadCenterAwsKey: 'downloadCenterAwsKey',
+      downloadCenterAwsSecret: 'downloadCenterAwsSecret',
+      githubToken: 'githubToken',
+      segmentKey: 'segmentKey',
+      rootDir: '../../../',
+      appleUser: 'appleUser',
+      applePassword: 'applePassword',
+      appleAppIdentity: 'appleAppIdentity',
```
addaleax, line 30: Since I feel like this might be relevant for me to some degree… what are these being used for? :slightly_smiling_face:

package.json
```diff
@@ -105,6 +107,8 @@
     "sinon": "^7.5.0",
     "sinon-chai": "^3.4.0",
     "tar": "^6.0.1",
+    "tar-fs": "^2.1.0",
+    "tmp-promise": "^3.0.2",
```
addaleax, line 111: This adds a few things here, but the code below doesn’t reference them … am I missing something?

addaleax (approved):

packages/build/src/barque.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+          await barque.releaseToBarque(tarballURL);
+        } catch (error) {
+          err = error;
+        }
+        expect(err).to.be.undefined;
+        expect(barque.execCurator).to.have.been.called;
+      });
+
+      it('execCurator function fails', async() => {
+        barque.execCurator = sinon.stub().returns(Promise.reject(new Error('error')));
```
mcasimir, line 61: I think we should be able to write just `sinon.stub().rejects(new Error('error'))`

packages/build/src/barque.ts
```diff
@@ -0,0 +1,137 @@
...
+   */
+  determineArch(variant: string): string {
+    // we can't use distro_id from evergreen's env variables, since those
+    // sometimes come with other attributes like -test -large -small, and are not
+    // valid distros for barque.
+    if (variant === BuildVariant.Linux) return Arch.Ubuntu;
+    if (variant === BuildVariant.Debian) return Arch.Debian;
+    if (variant === BuildVariant.Redhat) return Arch.Redhat;
+
+    return Distro.Ubuntu;
```
mcasimir, line 117: Is it intentional that other returns are `Arch` and this one is `Distro`?

scripts/import-expansions.js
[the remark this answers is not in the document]
mcasimir: The idea is for `import-expansions` to really know nothing of this curator topic. 

I'd completely remove this `if` and use directly `NOTARY_TOKEN: process.env.SIGNING_AUTH_TOKEN_44` in `exec`. Also because that is where we now have `server-4.4` which i'd assume goes in pair with this one.

package.json
lrlna: good point 😅 

this PR used to fetch a new version of curator because our build variants didn't have it. But the build folks updated the variants, so we no longer have to! I'll remove these.

packages/build/src/barque.spec.ts
lrlna: These are technically in use in `macos-sign.ts` part of the build package. Since we are not currently signing for MacOS, that file is never called.

packages/build/src/barque.ts
lrlna: nope, not intentional! thanks

scripts/import-expansions.js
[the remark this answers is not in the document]
lrlna: ah i see! I did misunderstand a little bit then. I'll fix it up to have `NOTARY_TOKEN: process.env.SIGNING_AUTH_TOKEN_44` in `barque.ts`.

packages/build/src/build-and-upload.spec.ts
```diff
@@ -99,15 +116,41 @@ describe('buildAndRelease', () => {
...
+  });
+
+  it('releases to downloads centre if a public release', async () => {
     githubRepo = createStubRepo({
       shouldDoPublicRelease: sinon.stub().returns(Promise.resolve(true)),
       releaseToGithub: sinon.stub().returns(Promise.resolve(true))
     } as any);
 
+    barque = createStubBarque({
+      releaseToBarque: sinon.stub().returns(Promise.resolve(true))
```
addaleax, line 147: I think you can do this if you like:

```suggestion
      releaseToBarque: sinon.stub().resolves(true)
```

packages/build/src/barque.ts
```diff
@@ -0,0 +1,190 @@
...
+ * Arch enum to be used when making a curator call.
+ *
+ * If we were to target a different arch for these distros, make sure
+ * config/repo-config.yml and packages/build/src/tarball.ts are changed accordingly.
+ */
+enum Arch {
+  Ubuntu = 'amd64',
+  Debian = 'amd64',
+  Redhat = 'x86_64',
+}
```
addaleax, line 42: It feels a bit odd to infer the architecture name from the distro name – maybe we should add a `buildArch` key to the config, and then take that into account here as well (in `determineArch()`)? It’s not important for this PR, but maybe the comment above could mention it?

addaleax (approved):
Nice :)

packages/build/src/build-and-upload.spec.ts
```diff
@@ -50,20 +56,26 @@ describe('buildAndRelease', () => {
...
     compileAndZipExecutable = sinon.stub().resolves(tarballFile);
+    barque = createStubBarque();
     uploadToEvergreen = sinon.spy();
     uploadToDownloadCenter = sinon.spy();
     githubRepo = createStubRepo();
   });
 
   [true, false].forEach((isPublicRelease) => {
-    it(`uploads the artifact to evergreen if is ${isPublicRelease ? 'a' : 'not a'} public release`, async() => {
+    it(`uploads the artifact to evergreen if is ${isPublicRelease ? 'a' : 'not a'} public release`, async () => {
```
mcasimir, line 66: I think we need a test to check we only release to barque if `isPublicRelease` 
and one to check that we don't if is not a public release or?

mcasimir (approved):

packages/build/src/build-and-upload.spec.ts
lrlna: are you saying you want a test that's like [this](https://github.com/mongodb-js/mongosh/pull/286/files#diff-6004d7618dcdb4b65e3f614bc3b71e98R119) but for `does not release to barque if not a public release`? Just triple checking!

packages/build/src/barque.ts
lrlna: the `config` is created based on the variables that evergreen provides for us, so if I were to put `arch` in the `config`, I would be basically doing the same thing as I am here. Which perhaps is not actually bad since we get to keep everything in one place?

mcasimir: We could also just make a ticket (or just discuss in design doc) for the release epic to just stop detecting what to do based on which runner we are on and rather pass those stuff as parameters in the scripts / as env var and make that populate the config.

What everyone thinks about that? Is probably not going to change much if we make that change only here.

lrlna: I think that it could be discussed in that next release automation epic!

packages/build/src/build-and-upload.spec.ts
mcasimir: oh yep, thanks!

PR #288, opened 2020-08-18
MONGOSH-294 - Database credentials not redacted from session logs or command history

aherlihy:


packages/shell-evaluator/src/shell-evaluator.ts
```diff
@@ -64,10 +67,13 @@ class ShellEvaluator {
         this.saveState();
         const rewrittenInput = this.internalState.asyncWriter.process(input);
 
-        this.internalState.messageBus.emit(
-          'mongosh:rewritten-async-input',
-          { original: input.trim(), rewritten: rewrittenInput.trim() }
-        );
+        const hiddenCommands = RegExp(HIDDEN_COMMANDS, 'g');
+        if (!hiddenCommands.test(input) && !hiddenCommands.test(rewrittenInput)) {
```
Anemy, line 71: Could the input here contain more commands than just the hidden commands? Are those also dropped?

packages/shell-api/src/shell-internal-state.ts
```diff
@@ -57,7 +58,10 @@ export default class ShellInternalState {
   async fetchConnectionInfo(): Promise<void> {
     if (!this.cliOptions.nodb) {
       this.connectionInfo = await this.currentDb._mongo._serviceProvider.getConnectionInfo();
-      this.messageBus.emit('mongosh:connect', this.connectionInfo.extraInfo);
+      this.messageBus.emit('mongosh:connect', {
+        ...this.connectionInfo.extraInfo,
+        uri: redactInfo(this.connectionInfo.extraInfo.uri)
```
Anemy, line 63: How is this URI used? I think the regex that would catch it would always change it to `<mongodb uri>`
https://github.com/mongodb-js/redact/blob/master/lib/regexes.js#L26

Anemy (commented):
Nice work - couple questions, I'm not too familiar with all the goings on 😅 
Is renaming the hidden functions something users might do? Could a user rename a hidden function to something and then use it which would have it in the log? Not sure that's something we need to account for.

If it's not much work - could we have a test that has a large input with a hidden command in it just to be sure its not logged when its async rewritten?

packages/shell-api/src/shell-internal-state.ts
aherlihy: The URI is used when we make a second connection for auth, since we cannot reconnect the same MongoClient with new authentication. The messageBus is only used for logging, so we don't really care that the URI is logged or not. In this case I'm just overwriting it with `redact`, but we could delete it from the object with a couple extra lines of code.  The redact should remove a little more selectively, for example `"mongodb://<ip address>:27017"` for `:"mongodb://localhost:27017"

packages/shell-evaluator/src/shell-evaluator.ts
aherlihy: If the input/output does not include any hidden commands then it will emit the log message for those commands, so they are not dropped. We want to not log for the hidden commands, which contain sensitive information like passwords.

aherlihy:
If a user renames a hidden command and it gets logged, then we don't have to worry about redacting that I think. Something like:
```
const cmd = 'changeUserPassword';
db[cmd]('anna', 'password');
```
Will log each line but I don't see an issue with that since there's not much point to renaming unless you're trying to get around the redacting. *However* I just noticed that the old shell redacts raw commands, for example `db.runCommand({ createUser: 'anna2', pwd: 'pwd', roles: []})`, while we do not do this. I will make a ticket for that since it's not a bugfix, it was never implemented. Thanks for the tip!

We currently don't test any of what gets logged, but I tested it manually and the logic for redacting is unit tested in the history package, so I think its sufficient.

packages/shell-api/src/shell-internal-state.ts
Anemy: Ah cool.
Gave it a quick test - I think the redact might do some things we don't quite expect sometimes hah
```js
const redactInfo = require('mongodb-redact');

console.log(redactInfo('mongodb://localhost:27017'));
// outputs '<mongodb uri>'

console.log(redactInfo('mongodb+srv://testUserForTesting:notMyRealPassword@some-atlas-db.mongodb.net/test'));
// outputs 'mongodb+srv://testUserForTesting:<email>/test'
```
Not much to do here I suppose but something to keep in mind for the redaction package.

Anemy (approved):
lgtm - nice!

packages/shell-api/src/shell-internal-state.ts
Anemy: Errrr thinking about this more.... Maybe our redaction package regexes could be not work properly with symbols in the password or username for connection strings and not redact properly. 🤔

packages/shell-api/src/collection.ts
```diff
@@ -1228,7 +1229,11 @@ export default class Collection extends ShellApiClass {
       throw new MongoshInvalidInputError('The "commandName" argument cannot be passed as an option to "runCommand".');
     }
 
-    this._emitCollectionApiCall('runCommand', { commandName });
+
+    const hiddenCommands = new RegExp(HIDDEN_COMMANDS);
+    if (!hiddenCommands.test(commandName)) {
+      this._emitCollectionApiCall('runCommand', { commandName });
```
Anemy, line 1235: nice

packages/shell-api/src/shell-internal-state.ts
aherlihy: I think when you have a "full" URI (meaning the generated one from the CLI options) `"mongodb://127.0.0.1:27017"` --> `"mongodb://<ip address>:27017"` it works slightly better. I do agree there's improvements to be made and I think we should address that ticket and the logging one ASAP

PR #291, opened 2020-08-31
chore(cli-repl): don't send api-call events to segment

lrlna:
## Description

As per conversations with @mmarcon and @mcasimir , we are dropping calls to
Segment when `mongosh:api-call` events happens.

This PR also reverts debounce commit since we no longer call `analytics.track`.

### DIFF
--- a/packages/cli-repl/src/logger.ts
+++ b/packages/cli-repl/src/logger.ts
@@ -158,20 +158,5 @@ export default function logger(bus: any, logDir: string): void {
 
   bus.on('mongosh:api-call', function(args: ApiEvent) {
     log.info('mongosh:api-call', redactInfo(args));
-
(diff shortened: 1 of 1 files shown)


packages/cli-repl/src/logger.ts
```diff
@@ -159,19 +159,20 @@ export default function logger(bus: any, logDir: string): void {
...
-    if (args.arguments) properties.arguments = properties.arguments;
-
-    if (telemetry) {
-      analytics.track({
-        userId,
-        event: 'Api Call',
-        properties: redactInfo(properties)
-      });
-    }
+    // const properties: ApiEvent = {};
```
mcasimir, line 164: I think we don't need to comment, we won't resume sending events as it is done here anyway.

mcasimir (approved):
👍

PR #292, opened 2020-08-31
safer publish-npm

mcasimir:
Make `publish-npm` to release master from a fresh clone:
- Make the release more stateless:
  - Avoid to accidentally publish "dirty" code and unwanted artifacts
- Correctly writes segment api key
- Sanity checks for segment api key


scripts/publish-npm.js
```diff
@@ -0,0 +1,121 @@
...
+const assert = require('assert');
+const readline = require('readline');
+
+const rootPath = path.resolve('__dirname', '..');
+
+function requireSegmentApiKey() {
+  const MONGOSH_SEGMENT_API_KEY = process.env.MONGOSH_SEGMENT_API_KEY || '';
+
+  if (!MONGOSH_SEGMENT_API_KEY) {
+    throw new Error('MONGOSH_SEGMENT_API_KEY environment variable must be set');
```
lrlna, line 13: Since you have `assert` imported, I feel like this could be an `assert` statement as well.

scripts/publish-npm.js
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  if (MONGOSH_SEGMENT_API_KEY.length !== 32) {
+    throw new Error('MONGOSH_SEGMENT_API_KEY does not seem to have the proper format');
```
lrlna, line 17: I feel like this could also be an `assert`.

package.json
```diff
@@ -34,7 +34,8 @@
     "compile-all": "npm run compile-ts && npm run compile-compass && npm run compile-exec",
     "prerelease": "npm run compile-cli",
     "release": "node scripts/release.js",
-    "publish-npm": "lerna publish --force-publish",
+    "publish-npm": "node scripts/publish-npm.js",
+    "publish-npm-lerna": "lerna publish --force-publish",
```
lrlna, line 38: I would remove this script from package.json, because it might be confusing to figure out which one to run if you're just reading the `scripts` section of package.json.

scripts/publish-npm.js
```diff
@@ -0,0 +1,121 @@
...
+
+  if (!isDone('write-segment-api-key', releaseDirPath)) {
+    writeSegmentApiKey(segmentApiKey, releaseDirPath);
+    done('write-segment-api-key', releaseDirPath);
+  } else {
+    console.info('already written segment api key .. skipping');
+  }
+
+  if (!isDone('lerna-publish', releaseDirPath)) {
+    execSync('npm run publish-npm-lerna', {cwd: releaseDirPath, stdio: 'inherit'});
```
lrlna, line 93: I would straight up use `lerna publish --force-publish` to avoid any extra confusion as to what this does.

scripts/publish-npm.js
```diff
@@ -0,0 +1,121 @@
...
+function publish() {
+  const segmentApiKey = requireSegmentApiKey();
+  const remoteUrl = getGitRemoteUrl();
+  const remoteHeadSha = getGitRemoteHeadSHA(remoteUrl);
+  const releaseDirPath = path.resolve(rootPath, 'tmp', 'releases', remoteHeadSha);
+
+  if (!isDone('clone', releaseDirPath)) {
+    console.info(`cloning '${remoteUrl}' to '${releaseDirPath}'`);
+    gitClone(remoteUrl, releaseDirPath);
+    done('clone', releaseDirPath);
```
lrlna, line 73: I keep reading this as a `callback` "done" and keep not seeing the callback 😂 😂 

The `done` function is meant to be a log of what was finished as part of this script? Maybe we can call this function `log`?

scripts/publish-npm.js
```diff
@@ -0,0 +1,121 @@
...
+
+  assert.equal(require(analyticsConfigPath).SEGMENT_API_KEY, segmentApiKey);
+}
+
+function isDone(task, releaseDirPath) {
+  return fs.existsSync(`${releaseDirPath}-${task}.done`);
+}
+
+function done(task, releaseDirPath) {
+  fs.writeFileSync(`${releaseDirPath}-${task}.done`, '');
```
lrlna, line 61: you're using files to track the state of what was completed, right? So we can restart the npm release process again if something failed?

Will this tmp dir live long enough for us to check this information after an unsuccessful publish?

lrlna (commented):
This will save us so much hassle! Thanks for putting this together so quickly 🎉

scripts/publish-npm.js
mcasimir: makes sense

package.json
mcasimir: which one `publish-npm` or `publish-npm-lerna`?

scripts/publish-npm.js
mcasimir: I like it, we'd need `node_modules/.bin/lerna`

scripts/publish-npm.js
mcasimir: `done` marks the task as done (writing a file), so running `npm publish` twice would skip what already succeeded. We could do this instead:

```
doOnce('clone', releaseDirPath, () => {
    console.info(`cloning '${remoteUrl}' to '${releaseDirPath}'`);
    gitClone(remoteUrl, releaseDirPath);
});
```

scripts/publish-npm.js
mcasimir: Is a `tmp` dir inside the repo folder (added it to `.gitignore`), will never be removed .. although probably would be nice to delete it after success, or at least do a console log "this folder can be deleted now".

package.json
lrlna: line 38, so `publish-npm-lerna`.

scripts/publish-npm.js
lrlna: Ah gotcha! I think then later on we can also do a clean up for when all tasks have successfully finished and delete the whole dir.

lrlna (approved):
Looks good!

PR #294, opened 2020-08-31
chore: generate homebrew formula

mcasimir:
Generates and print (no automated PR), the homebrew formula for the latest release of `@mongosh/cli-repl`. Gets the latest version with `npm view` and then renders a `mongosh.rb` template to the standard output with the correct `sha`.


lrlna (approved):
sweeeeet! 

For some extra brownie points, I'd add running this script at the end of a successful publish script, so you can this formula and just paste it in to the homebrew repo.

PR #296, opened 2020-09-02
MONGOSH-98 - version and load

aherlihy:


packages/i18n/src/locales/en_US.js
```diff
@@ -104,9 +104,16 @@ const translations = {
             Mongo: {
               description: 'Create a new connection and return the Mongo object. Usage: new Mongo(URI, options [optional])',
               link: 'https://docs.mongodb.com/manual/reference/method/Mongo/#Mongo'
+            },
+            version: {
+              description: 'Shell version',
+              link: 'https://docs.mongodb.com/manual/reference/method/version/'
+            },
+            load: {
```
mcasimir, line 112: will this show up as help for `.load`?

mcasimir (approved):
👍

packages/i18n/src/locales/en_US.js
aherlihy: No, the integrated repl commands like .load .exit and such can't have Help attached, but it will show up for `load`.

PR #299, opened 2020-09-07
test: fix test for 'version' command

addaleax:
This otherwise results in an unhandled rejection, because
`describe()` callbacks run immediately, without waiting for
`beforeEach()` to populate the `shell` variable.


lrlna (approved):
sweet!!

PR #300, opened 2020-09-08
MONGOSH-121: upload rpm to downloads centre

lrlna:
## Description

Uses [pkg-rpm](github.com/mongodb-js/pkg-rpm) to create an rpm package. This
package will be uploaded to download's centre when doing a release.

Additionally, this PR adds `rhel80-large` build variant. This build variant runs a
`check` and `release` tasks, and depends on tests passing on our existing linux
build variant.

### DIFF
--- a/packages/build/src/download-center.ts
+++ b/packages/build/src/download-center.ts
@@ -67,16 +67,16 @@ const CONFIG = `
           "download_link": "https://downloads.mongodb.com/compass/mongosh-{{version}}-linux.tgz"
         },
         {
-          "arch": "amd64",
+          "arch": "x64",
           "os": "debian",
           "name": "Debian 64-bit",
           "download_link": "https://downloads.mongodb.com/compass/mongosh_{{version}}_amd64.deb"
         }
         {
-          "arch": "amd64",
+          "arch": "x64",
           "os": "rhel",
           "name": "Redhat 64-bit",
-          "download_link": "https://downloads.mongodb.com/compass/mongosh-{{version}}-amd64.deb"
+          "download_link": "https://downloads.mongodb.com/compass/mongosh-{{version}}-x86_64.deb"
         }
       ]
     }
--- a/packages/build/src/download-center.ts
+++ b/packages/build/src/download-center.ts
@@ -67,10 +67,16 @@ const CONFIG = `
           "download_link": "https://downloads.mongodb.com/compass/mongosh-{{version}}-linux.tgz"
         },
         {
-          "arch": "x64",
+          "arch": "amd64",
           "os": "debian",
           "name": "Debian 64-bit",
           "download_link": "https://downloads.mongodb.com/compass/mongosh_{{version}}_amd64.deb"
+        }
+        {
+          "arch": "amd64",
+          "os": "rhel",
+          "name": "Redhat 64-bit",
+          "download_link": "https://downloads.mongodb.com/compass/mongosh-{{version}}-amd64.deb"
         }
       ]
     }
--- a/packages/build/src/download-center.ts
+++ b/packages/build/src/download-center.ts
@@ -76,7 +76,7 @@ const CONFIG = `
           "arch": "x64",
           "os": "rhel",
           "name": "Redhat 64-bit",
-          "download_link": "https://downloads.mongodb.com/compass/mongosh-{{version}}-x86_64.deb"
+          "download_link": "https://downloads.mongodb.com/compass/mongosh-{{version}}-x86_64.rpm"
         }
       ]
     }
--- a/packages/build/src/tarball.spec.ts
+++ b/packages/build/src/tarball.spec.ts
@@ -103,7 +103,7 @@ describe('tarball module', () => {
       });
 
       it('builds the executable', async() => {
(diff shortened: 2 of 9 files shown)


packages/build/src/tarball.ts
```diff
@@ -111,12 +150,19 @@ export async function createTarball(
...
+      contentType: 'application/x-rpm'
+    }
   } else if (buildVariant === BuildVariant.Debian) {
     await tarballDebian(input, outputDir, version, rootDir);
 
     return {
       path: filename,
-      // this might have to be application/gtarball MIME type
+      // this might have to be application/gzip MIME type
       contentType: 'application/vnd.debian.binary-package'
```
addaleax, line 166: I’m not sure how this is used, but fwiw, this is the official mime type for `.deb`: https://www.iana.org/assignments/media-types/application/vnd.debian.binary-package

addaleax (approved):

packages/build/src/tarball.spec.ts
```diff
@@ -92,6 +92,33 @@ describe('tarball module', () => {
...
+      after(async () => {
+        await fs.unlink(expectedTarball);
+      });
+
+      it('builds the executable', async() => {
+        await tarballDebian(inputFile, __dirname, version, path.join(__dirname, '../../..'));
+
+        let accessErr;
+        try {
+          fs.access(expectedTarball);
```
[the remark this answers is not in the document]
lrlna: nope, you're not missing anything! thanks!

packages/build/src/tarball.ts
lrlna: we use this to upload the artifact to evergreen's S3 bucket. I will be leaving it with the `content-type` I have, which is also the one you found, i.e. `application/vnd.debian.binary-package`. (removed that comment cause it's confusing at this point)

packages/build/src/download-center.ts
```diff
@@ -67,11 +67,17 @@ const CONFIG = `
           "download_link": "https://downloads.mongodb.com/compass/mongosh-{{version}}-linux.tgz"
         },
         {
-          "arch": "x64",
+          "arch": "amd64",
```
lrlna, line 70: We build debian package for `amd64`, so I think this should also be specified as `amd64`.

packages/build/src/download-center.ts
```diff
@@ -67,11 +67,17 @@ const CONFIG = `
...
           "os": "debian",
           "name": "Debian 64-bit",
           "download_link": "https://downloads.mongodb.com/compass/mongosh_{{version}}_amd64.deb"
         }
+        {
+          "arch": "amd64",
```
lrlna, line 76: I am currently building this for `amd64` as well. If we want to be building for `x86-64` instead, let me know. It's easily changeable.

packages/build/src/download-center.ts
lrlna: oh nvm this.

packages/build/src/download-center.ts
lrlna: yep, also don't mind this.

packages/build/src/download-center.ts
```diff
@@ -72,6 +72,12 @@ const CONFIG = `
           "name": "Debian 64-bit",
           "download_link": "https://downloads.mongodb.com/compass/mongosh_{{version}}_amd64.deb"
         }
+        {
+          "arch": "x64",
+          "os": "rhel",
+          "name": "Redhat 64-bit",
+          "download_link": "https://downloads.mongodb.com/compass/mongosh-{{version}}-x86_64.rpm"
```
lrlna, line 79: I want to double check that we are ok building this for this `arch`, and this naming for the downloads centre package is ok. If anyone has thoughts, please let me know.

packages/build/src/tarball.spec.ts
```diff
@@ -82,14 +83,47 @@ describe('tarball module', () => {
...
-          fs.access(expectedTarball);
+          await fs.access(expectedTarball);
         } catch (err) {
           accessErr = err;
         }
 
         expect(accessErr).to.be.undefined;
       });
     });
+
```
mcasimir, line 94: A better way than commenting this out would be this:

```
describe('.tarballRedhat', () => {
    beforeEach(async function() {
        if(!commandExists('rpmbuild')) {
            this.skip();
        }
    });
    
    it('...', () => { });
})
``` 

`this.skip()` skips all the tests in the current test block;

`commandExists` is this thing here: https://www.npmjs.com/package/command-exists

lrlna: oh! didn't know something like this existed. Will change what I currently have!

packages/build/src/tarball.ts
```diff
@@ -84,7 +123,7 @@ export const tarballWindows = (input: string, filename: string): void => {
 export type TarballFile = { path: string; contentType: string };
 
 /**
- * Create a gtarballped tarball or zip for the provided options.
```
mcasimir, line 87: "tarballped" is a great word, i think it should mean something!

lrlna: ahahhaha

PR #303, opened 2020-09-09
MONGOSH-181 - Remaining Cursor Methods

aherlihy:


lrlna (approved):
Given the title, I thought this would be a larger PR 😂

packages/shell-api/src/cursor.ts
```diff
@@ -292,4 +292,11 @@ export default class Cursor extends ShellApiClass {
     throw new MongoshUnimplementedError(
       '`maxScan()` was removed because it was deprecated in MongoDB 4.0');
   }
+
+  @returnType('Cursor')
+  @serverVersions(['3.2.0', ServerVersions.latest])
```
addaleax, line 297: If you don’t mind me asking, are these two values minimum and maximum supported version?

aherlihy:
Yeah, the other two methods `objsLeftInBatch` and `readConcern` are waiting on answers from the node driver team

packages/shell-api/src/cursor.ts
aherlihy: Definitely don't mind you asking! Those are the min/max supported *server* version, so it will still exist if the user is connected to a different server version but it won't be suggested by the autocomplete.

PR #305, opened 2020-09-09
MONGOSH-341: Fix wrong values with NumberLong for numbers > Number.MAX_SAFE_INTEGER

mcasimir:
- Allows to create `NumberLong` bigger than `Number.MAX_SAFE_INTEGER`.
- Requires the argument of `NumberLong` and `NumberDecimal` to always be a string so to prevent users to get a misrepresented value by mistake.
- Makes `NumberInt()` to be equivalent to `NumberInt(0)`, aligning the behaviour with the other constructors and with the old shell.

### DIFF
--- a/packages/service-provider-core/src/printable-bson.ts
+++ b/packages/service-provider-core/src/printable-bson.ts
@@ -65,7 +65,7 @@ export default function(bson): void {
   bson.Int32.prototype.asPrintable = bson.Int32.prototype[toString];
 
   bson.Long.prototype[toString] = function(): string {
-    return `NumberLong(${this.toString()})`;
+    return `NumberLong("${this.toString()}")`;
   };
   bson.Long.prototype.asPrintable = bson.Long.prototype[toString];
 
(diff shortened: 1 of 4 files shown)


packages/service-provider-core/src/printable-bson.ts
```diff
@@ -65,7 +65,7 @@ export default function(bson): void {
   bson.Int32.prototype.asPrintable = bson.Int32.prototype[toString];
 
   bson.Long.prototype[toString] = function(): string {
-    return `NumberLong(${this.toNumber()})`;
+    return `NumberLong(${this.toString()})`;
```
addaleax, line 68: Does it make sense to 

```suggestion
    return `NumberLong("${this.toString()}")`;
```

so that this could then be used as input again without losing preceision?

addaleax (approved):

packages/service-provider-core/src/printable-bson.ts
mcasimir: That's brilliant! I think we should do the same for NumberDecimal too then.

PR #306, opened 2020-09-10
Wait for all assets before release

mcasimir:
A temporary fix to prevent the release to take place if some of the assets where not generated properly.

We retry the validation of download centre config until succeeds for 1 hour. If after 1 hour the config would still be missing assets the release will fail.

As soon as the validation succeeds the release goes on.

### DIFF
--- a/packages/build/src/download-center.ts
+++ b/packages/build/src/download-center.ts
@@ -114,6 +114,35 @@ const uploadToDownloadCenter = (s3: S3, config: string): Promise<any> => {
   return upload(uploadParams, s3);
 };
 
+const delay = (msecs): any => new Promise((resolve) => {
+  setTimeout(resolve, msecs);
+});
+
+const waitAllDownloadAssetsAvailable = async(config): Promise<void> => {
+  const configObject = JSON.parse(config);
+  const checkInterval = 60 * 1000; // 1 minute
+
+  let attemptsRemaining = 60; // try for 1 hour
+  let succeded = false;
+
+  while (attemptsRemaining >= 0) {
+    attemptsRemaining--;
+
+    try {
+      await verifyDownloadCenterConfig(configObject);
+      succeded = true;
+      break;
+    } catch (error) {
+      console.error('verifyDownloadCenterConfig: attempt failed.', error);
+      await delay(checkInterval);
+    }
+  }
+
+  if (!succeded) {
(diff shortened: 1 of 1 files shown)


packages/build/src/download-center.ts
```diff
@@ -114,6 +117,40 @@ const uploadToDownloadCenter = (s3: S3, config: string): Promise<any> => {
   return upload(uploadParams, s3);
 };
 
+const delay = (msecs): any => new Promise((resolve) => {
```
addaleax, line 120: This could be 

```suggestion
const delay = (msecs): Promise<void> => new Promise((resolve) => {
```

, right?

addaleax (approved):

lrlna (changes requested):
I am not sure this is the right approach, since we can't restart the other build variants from this function and have to look manually as to what was wrong with the build variant that failed. So this just goes in a loop for a whole hour while we manually fix one of the build variants that fails and restart the whole build process anyways. Do you know what I mean?

packages/build/src/download-center.ts
mcasimir: yep!

mcasimir:
> I am not sure this is the right approach, since we can't restart the other build variants from this function and have to look manually as to what was wrong with the build variant that failed. So this just goes in a loop for a whole hour while we manually fix one of the build variants that fails and restart the whole build process anyways. Do you know what I mean?

Is not the right approach by all means, we discussed that with max already and concluded that we'd rather have to do a whole new release than having a broken one. 

The idea is to fail if something is not right, then to fix that we will have to do another commit / release. I really hope EVG to have a better way.

lrlna:
@mcasimir so you're using this as a way to run the `release_macos` as the very last task in the build chain?

mcasimir:
> @mcasimir so you're using this as a way to run the `release_macos` as the very last task in the build chain?

Yep, to don't go on with releasing on the download center and github unless everything was good (all platform tests and build succeded). Is something less than ideal though, but at least we have the safe guard, we can / should improve that if EVG doesn't offer a functionality. However we don't want to risk to have releases done when stuff goes off.

lrlna (approved):

mcasimir:
Going to pause on this one and see if we can move this check to evergreen deps instead

mcasimir:
Closed in favor of https://github.com/mongodb-js/mongosh/pull/311

PR #307, opened 2020-09-10
MONGOSH-313 - Remaining Mongo methods

aherlihy:


packages/shell-api/src/mongo.ts
```diff
@@ -131,4 +135,40 @@ export default class Mongo extends ShellApiClass {
...
+    // return this._serviceProvider.getReadPreference();
+    throw new MongoshUnimplementedError('getting the read pref is not currently supported, to follow the progress please see NODE-2806');
+  }
+
+  getReadConcern(): string {
+    try {
+      const rc = this._serviceProvider.getReadConcern();
+      return rc ? rc.level : undefined;
+    } catch {
+      throw new MongoshInternalError('Error retrieving ReadConcern. Please file a JIRA ticket in the MONGOSH project');
```
addaleax, line 159: Does it make sense to keep some information about the original exception here?

addaleax (approved):

packages/shell-api/src/mongo.ts
aherlihy: The only way this could error in this way is if the internal construction of the node driver changes. The next release of the node driver will have a method that handles this for us `getReadConcern`, but the version we are currently on does not make it public. So Im going into the internal state to get this information, so any error thrown would be something like "cannot get ReadConcern of undefined", so it didn't seem meaningful to me to keep. This error is highly unlikely to be ever triggered, but I wanted to wrap it in case it did happen it would be clear that something went wrong in this method. The internal state will only change when the version changes, which means it's time to change this code anyway.

addaleax: Okay, makes sense! :)

PR #308, opened 2020-09-10
feat: add ability to perform tarball dry-run

addaleax:
Provide an easy way to create distribution tarballs without also
uploading them for testing.

### DIFF
--- a/scripts/release.js
+++ b/scripts/release.js
@@ -1,16 +1,25 @@
 require('./import-expansions');
 
 const path = require('path');
-const { release } = require(path.join('..', 'packages', 'build'));
+const { release, BuildVariant } = require(path.join('..', 'packages', 'build'));
 const config = require(path.join(__dirname, '..', 'config', 'build.conf.js'));
 
 /**
  * Run the release process.
  */
 const runRelease = async() => {
+  if (process.argv.includes('--dry')) config.dryRun = true;
+  const cliBuildVariant =
+    process.argv.map((arg) => arg.match(/^--build-variant=(.+)$/)).filter(Boolean)[0];
+  if (cliBuildVariant !== undefined) config.buildVariant = cliBuildVariant[1];
+
+  // Resolve 'Windows' to 'win32' etc.
+  if (BuildVariant[config.buildVariant])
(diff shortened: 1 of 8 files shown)


scripts/release.js
```diff
@@ -1,16 +1,25 @@
...
+
   await release(config);
 };
 
-runRelease().then(() => {
-  process.exit(0);
-});
+runRelease().then(
+  () => process.exit(0),
+  (err) => process.nextTick(() => { throw err; }));
```
mcasimir, line 25: Just curious why `process.nextTick` is necessary?

scripts/release.js
```diff
@@ -1,16 +1,25 @@
...
 const path = require('path');
-const { release } = require(path.join('..', 'packages', 'build'));
+const { release, BuildVariant } = require(path.join('..', 'packages', 'build'));
 const config = require(path.join(__dirname, '..', 'config', 'build.conf.js'));
 
 /**
  * Run the release process.
  */
 const runRelease = async() => {
+  if (process.argv.includes('--dry')) config.dryRun = true;
```
mcasimir, line 11: we normally always use `{` `}` for blocks with ifs, weird that eslint didn't complain about that

mcasimir (approved):
Nice!

scripts/release.js
addaleax: Sure, done :)

scripts/release.js
addaleax: The intention here is to turn unhandled Promise rejections (which print a warning but do not result in the process exiting with a non-zero exit code) into an uncaught exception (which does result in the process exiting with code 1) – that behavior has a long, long history of debate behind it, and Node.js core will most likely change it in the future (https://medium.com/@nodejs/node-js-promise-reject-use-case-survey-98e3328340c9 has some futher reading on the current state of things, if you’re curious)

If this was re-throwing the error directly, it would just create another rejected Promise, so this is using `process.nextTick()` to make this independent of the Promise chain and avoid that

lrlna (approved):
really into this!

PR #309, opened 2020-09-10
chore: print name of missing key for required object keys

addaleax:
Some MongoDB newbies (cough cough) might not know what keys are
required or not, and just knowing that something is missing
might not be enough to figure out what’s required and what’s not. :)


lrlna (approved):
lgtm

PR #312, opened 2020-09-10
MONGOSH-346 - enable Ctrl+C in shell for sync and async operations

addaleax:
This enables Ctrl+C to interrupt both the current synchronous JS
execution and awaiting the result.

Refs: https://nodejs.org/api/repl.html#repl_repl_start_options

### DIFF
--- a/packages/cli-repl/src/cli-repl.ts
+++ b/packages/cli-repl/src/cli-repl.ts
@@ -115,7 +115,8 @@ class CliRepl {
       prompt: '> ',
       writer: this.writer,
       completer: completer.bind(null, version),
-      terminal: true
+      terminal: true,
+      breakEvalOnSigint: true,
     });
 
     const originalDisplayPrompt = this.repl.displayPrompt.bind(this.repl);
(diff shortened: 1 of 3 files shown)


lrlna (commented):
oh! I thought this doesn't work when we have a custom `eval` function? Or does this ring sort of true because we have half custom half general eval?

addaleax:
@lrlna I might need to read a bit more about how the shell works in that respect – in particular, how this interacts with the implicit `await` that’s being inserted here (probably not well). But this does work for synchronous loops, and I’m not sure why we would want to not enable this, even if it doesn’t work in every situation yet

addaleax:
Oh, just in case you were talking about the warning in the Node.js docs – it seems to work locally for a `while(true);` loop, at least

addaleax:
Let me dig a little deeper on the async part, though :)

packages/cli-repl/src/cli-repl.ts
```diff
@@ -147,7 +148,36 @@ class CliRepl {
...
+          result = await new Promise((resolve, reject) => {
+            // Handle SIGINT (Ctrl+C) that occurs while we are stuck in `await`
+            // by racing a listener for 'SIGINT' against the evalResult Promise.
+            // We remove all 'SIGINT' listeners and install our own.
+            sigintListener = (): void => {
+              // Reject with an exception similar to one thrown by Node.js
+              // itself if the `customEval` itself is interrupted.
+              reject(new Error('Asynchronous execution was interrupted by `SIGINT`'));
+            };
+            previousSigintListeners = this.repl.rawListeners('SIGINT');
```
lrlna, line 163: I _did not_ know you can just hook into the listeners o/

addaleax: Fwiw, this code is partially inspired by the way that Node.js does this internally for the REPL option :)

https://github.com/nodejs/node/blob/02e925553d079c1e70b48e5255503d8850eb1fc5/lib/vm.js#L264-L280

packages/cli-repl/test/e2e.spec.ts
```diff
@@ -238,5 +238,32 @@ describe('e2e', function() {
       shell.assertNoErrors();
     });
   });
+
+  describe('Ctrl+C aka SIGINT', () => {
+    before(function() {
+      if (process.platform === 'win32') {
```
mcasimir, line 244: how is it going to behave there?

addaleax: So, it should work there as well, there just isn’t a signal available to us for testing … but I’ll make sure to test this manually on Windows

addaleax: Confirmed that it works there just like it does on other OSes -- and I don't think there's a way to add a test for this, Node.js itself skips this too: https://github.com/nodejs/node/blob/e038d6a1cdb2ffbf666d360553c31786acaf70b2/test/parallel/test-vm-sigint.js#L3-L6

lrlna (approved):
This looks like it handles SIGINT with our 🔮 async eval magic 🔮 heaps better. I am into merging this in!

PR #313, opened 2020-09-10
MONGOSH-323 - Pretty-print BinData values

addaleax:
Print `BinData` values in a way that accounts for their subtype
(i.e. format UUID and MD5 in their common representation) and
generic `BinData` values as base64, to ensure that binary data is not
written directly to the terminal. This also enables copy-pasting of
the output for `BinData` values to be used as input for other
operations.


mcasimir (approved):
Looks great!

PR #314, opened 2020-09-11
test: unconditionally set `eventually()` timeout to 10s

addaleax:
This makes a lot more tests pass locally for me, and it shouldn’t
really increase the time it takes to run the tests.


lrlna (approved):
that's fine by me!

PR #315, opened 2020-09-11
fix: allow running mongosh with Node.js release candidates

addaleax:
When Node.js pushes a release candidate, the version will contain
a suffix like `-rc.0`, but semver doesn’t account for that in its
`satisfies()` function and rejects all versions tagged that way.


lrlna (approved):
Thanks! 🙏

PR #316, opened 2020-09-11
MONGOSH-320 - Consistently throw when command returns ok:0

aherlihy:
For all shell helpers that use runCommand, use runCommandWithCheck instead that checks the results for `ok: 0`. If so, it throws `MongoshCommandFailed` with an error message that includes the original command and a suggestion to use `runCommand` if you want the raw results.

The PR looks big because I had to rename every instance of runCommand --> runCommandWithCheck, but it's actually quite a small change.

### DIFF
--- a/packages/errors/index.ts
+++ b/packages/errors/index.ts
@@ -36,10 +36,19 @@ class MongoshWarning extends Error {
   }
 }
 
+class MongoshCommandFailed extends Error {
+  constructor(msg) {
+    super(msg);
+    this.name = 'MongoshCommandFailed';
+    this.message = `Command ${msg} returned ok: 0. To see the raw results of the command, use 'runCommand' instead.`;
+  }
+}
+
 export {
   MongoshWarning,
   MongoshRuntimeError,
   MongoshInternalError,
   MongoshInvalidInputError,
-  MongoshUnimplementedError
+  MongoshUnimplementedError,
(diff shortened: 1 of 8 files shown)


addaleax (approved):
LGTM

packages/errors/index.ts
```diff
@@ -36,10 +36,19 @@ class MongoshWarning extends Error {
   }
 }
 
+class MongoshCommandFailed extends Error {
+  constructor(msg) {
+    super(msg);
+    this.name = 'MongoshCommandFailed';
+    this.message = `Command ${msg} returned ok: 0. To see the raw results of the command, use 'runCommand' instead.`;
```
lrlna, line 43: should we very explicitly say `db.runCommand`? Or do you think that would be redundant? 
```suggestion
    this.message = `Command ${msg} returned ok: 0. To see the raw results of the command, use 'db.runCommand()' instead.`;
```

lrlna (approved):
looks good!

packages/errors/index.ts
aherlihy: There is both db.runCommand and collection.runCommand, which do the same thing under the hood. But there are collection methods that will throw this error, so I figured it's better to just name the method as opposed to both classes and the method.

lrlna: ah, fair!

PR #318, opened 2020-09-14
fix: report explicit error when invalid connection URI is passed

addaleax:
Because this gives a better error than

```
$ mongosh /x
Cannot read property '1' of null
```

:)


lrlna (approved):
nice, thanks!

PR #319, opened 2020-09-14
chore: dockerfiles to speed up linux release checks

mcasimir:


addaleax (approved):

scripts/docker/centos8-rpm.Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM centos:8
+
+ARG commit=""
+ADD https://s3.amazonaws.com/mciuploads/mongosh/${commit}/mongosh-0.2.2-x86_64.rpm /tmp
+RUN yum localinstall -y /tmp/mongosh-0.2.2-x86_64.rpm
```
lrlna, line 5: i strongly recommend also adding a command for installing `mongodb-org` (server and mongod being the most important), because otherwise we can't really test mongosh.

lrlna: (and this is for all the docker files)

mcasimir: Is run with `--network host` it will work right as it was outside docker

lrlna: oh nice!

lrlna: another thing is, should we install it with `rpm -i` instead of `yum localinstall`? So it's as closely aligned to how users will be installing it? Or are we going to be recommending users to install it with `yum localinstall`?

mcasimir: FYI: even in case you had no mongodb in the network, with docker normally you don't put more than one thing in the container but rather use networking to make more containers communicate: 

For example:

```
docker network create my-network
docker run -d --name mongo4 --network=my-network mongo:4
docker run --rm --network=my-network mongosh-centos6-rpm mongodb://mongo4:27017/db1
```

docker will make containers available to each other in the network using the container name as hostname.

with `--network host` instead (that i'm using in run.sh) the container is attached to the host network, and behave like any other process: will be able to access (and be accessed by) any other resources on localhost.

mcasimir: One final thing: you can also use `docker-compose` to run multiple container (the whole shared network is done for you)

mcasimir: > another thing is, should we install it with rpm -i instead of yum localinstall

Tha makes a lot of sense, let me try, last time `rpm` was complaining with me about something.

mcasimir: Done that too

PR #320, opened 2020-09-14
Homebrew after successful publish

mcasimir:
Create a PR in our homebrew tap automatically after the release publish succeeded.

### DIFF
--- a/scripts/utils.js
+++ b/scripts/utils.js
@@ -7,7 +7,7 @@ function getLatestVersion() {
 }
 
 function gitClone(repo, dest) {
-  return execSync(`git clone ${repo} ${dest}`);
+  return execFileSync('git', ['clone', repo, dest]`);
 }
 
 function confirm(prompt) {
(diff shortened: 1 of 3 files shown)


scripts/utils.js
```diff
@@ -0,0 +1,35 @@
+const { execSync } = require('child_process');
+const readline = require('readline');
+const { resolve } = require('path');
+
+function getLatestVersion() {
+  return execSync(`npm view @mongosh/cli-repl .dist-tags.latest`).toString().trim();
+}
+
+function gitClone(repo, dest) {
+  return execSync(`git clone ${repo} ${dest}`);
```
addaleax, line 10: I know it’s only taken from the code above, but to avoid edge cases with special characters in the expansions:

```suggestion
  return execFileSync('git', ['clone', repo, dest]`);
```

? (This also applies to other places, but I generally consider `execSync()` with template strings/string concatenation something that should be avoided unless the range of input values is fixed somehow)

mcasimir: Should be fine given we have full control over those params, however i don't see why we shouldn't improve it, we can change that together.

PR #325, opened 2020-09-15
fix: skip release if is patch

mcasimir:


addaleax (approved):
The code looks good, and I’m not seeing how this could break anything, but this feels a bit odd… what does `IS_PATCH` correspond to exactly? Does it distinguish between a tag/not a tag? Does it matter that this check isn’t done locally?

mcasimir:
> The code looks good, and I’m not seeing how this could break anything, but this feels a bit odd… what does `IS_PATCH` correspond to exactly? 

At this point I'm not sure I know. It seems evergreen gives us an `IS_PATCH` variable that should be "true" for PRs commit and false in "master".

> Does it matter that this check isn’t done locally?

What do you mean with "locally", like in our machines?

PR #327, opened 2020-09-15
chore: check user permissions to publish

mcasimir:
Check npm access to make sure the user has enough permission to publish packages before starting a release.


addaleax (approved):
Yeah, this looks just like what I would have needed yesterday :sweat_smile:

PR #329, opened 2020-09-15
MONGOSH-348 - Support collection names with dot

aherlihy:


packages/async-rewriter/src/async-writer-babel.spec.ts
```diff
@@ -186,6 +187,43 @@ class Test {
...
+            traverse(ast, {
+              Identifier(path) {
+                if (path.node.name === 'coll') {
+                  expect(path.node['shellType']).to.deep.equal(signatures.Collection);
+                  done();
+                }
+              }
+            });
+          });
+          it('decorates node.key Identifier', (done) => { // NOTE: if this ID exists in scope will be descorated with that value not undefined.
```
addaleax, line 223:
```suggestion
          it('decorates node.key Identifier', (done) => { // NOTE: if this ID exists in scope will be decorated with that value not undefined.
```

(But also, I’m not really sure what “this ID” and “that value” refer to here, tbh)

packages/async-rewriter/src/async-writer-babel.spec.ts
```diff
@@ -347,7 +419,7 @@ class Test {
           it('decorates MemberExpression', (done) => {
             traverse(ast, {
               MemberExpression(path) {
-                expect(path.node['shellType']).to.deep.equal(signatures.Collection.attributes.insertOne);
+                expect(path.node['shellType']).to.deep.equal(signatures.Cursor.attributes.hasNext);
```
addaleax, line 422: These are only being changed because the type of `c` has changed, right?

addaleax (approved):

packages/async-rewriter/src/async-writer-babel.spec.ts
aherlihy: Oops, thanks! Here's an explanation for the comment, but feel free to skip if it's too long :)

The comment is saying for a member expression node in the AST (`db.myCollection`) if the symbol for the key node (`myCollection`) exists in the scope, i.e. if there's another variable in scope that shares the same name, then the key node will be decorated with the type of the other variable. So for the entire expression `db.myCollection` in the shell, the tree will get decorated as:
1. The root MemberExpression node `db.myCollection` gets decorated with the `Collection` type because that's what the expression evaluates to
2. The key Identifier node `db` gets decorated with the `Database` type
3. The object Identifier node `myCollection` will get decorated as `unknown` UNLESS we have some other variable in scope named `myCollection`, that has a different type (say `Cursor`). If that variable does exist, the `myCollection` node will get the `Cursor` type because it will get decorated at a different point in the tree traversal.

It's inconsequential because we never read/use the type of the key node alone, but the comment is there in case someone notices that the *key* node has a decorated type that doesn't match the actual type of `db.myCollection`.

So I'm not sure if that actually clarified anything, we'll have a lunch & learn about the async rewriter at some point so that I can explain anything that doesn't make sense. The comment is just there as a note, it could also be deleted if it causes more confusion that it clears up!

packages/async-rewriter/src/async-writer-babel.spec.ts
aherlihy: Yes, I just changed the test because `Collection` was being used as a test for a type that didn't have a proxy in it. Now it does, so I changed it to `Cursor`

packages/async-rewriter/src/async-writer-babel.spec.ts
addaleax: Thank you, this is helpful indeed :)

PR #330, opened 2020-09-16
MONGOSH-329 - Add checks for BSON object constructors

aherlihy:


mcasimir (approved):
Nice!

packages/shell-api/src/helpers.ts
```diff
@@ -56,6 +56,16 @@ export function assertArgsDefined(...args: any[]): void {
   }
 }
 
+export function assertArgsType(args: any[], expectedTypes: any[]): void {
```
addaleax, line 59:
```suggestion
export function assertArgsType(args: any[], expectedTypes: string[]): void {
```

addaleax (approved):
In the issue report, this led to an unhandled rejection … that is not being change here, right? i.e. if something in the formatting methods throws, we would still receive an unhandled rejection, and this is just making the code fail earlier, correct?

aherlihy:
Yes, it changes the location where the error is thrown, but that means it will be thrown like a regular error and not an unhandled rejection. (I also updated the method that was throwing the previous error so that it won't throw). 

It seems that if an error is thrown within `inspect` the error becomes an UnhandledPromiseRejection, so I wonder if we can modify where we call inspect to handle an error. After this PR it should be almost impossible for it to throw in inspect, which I guess is okay for now, but you're right it would be better to just make it so that errors in inspect are handled and don't crash the repl. I  am not 100% sure why an error in inspect would lead to UnhandledPromiseRejection because inspect methods and other formatting methods in cli-repl shouldn't be async? I'll look into it

PR #331, opened 2020-09-17
Catch errors in repl callback; include stacktrace in log

aherlihy:


packages/cli-repl/src/cli-repl.ts
```diff
@@ -179,12 +179,24 @@ class CliRepl {
...
+          return callback(wrapError);
         }
-        return callback(err);
       }
-      callback(null, result);
+      try {
+        return callback(null, result);
+      } catch (callbackErr) {
+        const wrapError = new MongoshInternalError(callbackErr.message);
+        wrapError.stack = callbackErr.stack;
```
addaleax, line 197: Just a question, but can you explain what the benefit of wrapping the error is compared to how we were just passing it through previously?

addaleax (approved):

packages/cli-repl/src/cli-repl.ts
aherlihy: So that the user sees
```
MongoshInternalError: Cannot read property 'toString' of undefined
This is an error inside Mongosh. Please file a bug report. Please include a log file from this session.
```
instead of just 
```
TypeError: Cannot read property 'toString' of undefined
```

So they understand something went bad inside the shell, and we caught it, as opposed to it being a mystery what's going on

addaleax: Right, makes sense :)

PR #332, opened 2020-09-18
fix: Ctrl+C should work after exceptions

addaleax:
The added comment goes into a bit more detail here, but basically,
our shell evaluation function was under the wrong assumption that the
built-in eval code would always call its callback, even if evaluation
of the code threw an error. That does not turn out to be the case
for Node.js; instead, the REPL would handle the error internally
and not call its callback, leaving us with an async function that
never finishes.

This had the side effect that listeners for SIGINT would not be removed
(because from the cli-repl perspective, the inner eval would never
finish), leading to odd results when entering Ctrl+C after an earlier
statement had thrown an exception.

The solution here is hacky; we should probably take the time
at some point to find a better one.

### DIFF
--- a/packages/cli-repl/src/cli-repl.ts
+++ b/packages/cli-repl/src/cli-repl.ts
@@ -140,47 +140,8 @@ class CliRepl {
       }
     });
 
-    type REPLEvalArgs = REPLServer extends {
-      eval(...args: infer T);
-    } ? T : never;
     const replEval = this.repl.eval.bind(this.repl);
-    const originalEval = util.promisify((...args: REPLEvalArgs) => {
-      const { Domain } = require('domain');
-      const origEmit = Domain.prototype.emit;
-
-      // When the Node.js core REPL encounters an exception during synchronous
-      // evaluation, it does not pass the exception value to the callback
-      // (or in this case, reject the Promise here), as one might inspect.
-      // Instead, it skips straight ahead to abandoning evaluation and acts
-      // as if the error had been thrown asynchronously. This works for them,
-      // but for us that's not great, because we rely on the core eval function
-      // calling its callback in order to be informed about a possible error
-      // that occurred (... and in order for this async function to finish at all.)
-      // We monkey-patch `process.domain.emit()` to avoid that, and instead
-      // handle a possible error ourselves:
-      // https://github.com/nodejs/node/blob/59ca56eddefc78bab87d7e8e074b3af843ab1bc3/lib/repl.js#L488-L493
-      // It's not clear why this is done this way in Node.js, however,
-      // removing the linked code does lead to failures in the Node.js test
-      // suite, so somebody sufficiently motivated could probably find out.
-      // For now, this is a hack and probably not considered officially
-      // supported, but it works.
-      // We *may* want to consider not relying on the built-in eval function
-      // at all at some point.
-      Domain.prototype.emit = function(ev, ...eventArgs): void {
-        if (ev === 'error') {
-          throw eventArgs[0];
-        }
-        return origEmit.call(this, ev, ...eventArgs);
-      };
-
-      try {
-        return replEval(...args);
-      } finally {
-        // Reset the `emit` function after synchronous evaluation, because
-        // we need the Domain functionality for the asynchronous bits.
-        Domain.prototype.emit = origEmit;
-      }
-    });
+    const originalEval = util.promisify(this.wrapNoSyncDomainError(replEval));
 
     const customEval = async(input, context, filename, callback): Promise<any> => {
       this.lineByLineInput.enableBlockOnNewLine();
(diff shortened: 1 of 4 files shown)


packages/shell-evaluator/src/shell-evaluator.ts
```diff
@@ -74,8 +74,49 @@ class ShellEvaluator {
...
+          try {
+            syncResult = originalEval(rewrittenInput, context, filename);
+          } finally {
+            // Reset the `emit` function after synchronous evaluation, because
+            // we need the Domain functionality for the asynchronous bits.
+            Domain.prototype.emit = origEmit;
+          }
+
+          return await syncResult;
         } catch (err) {
```
addaleax, line 120: Maybe just to clarify, this catch block and all catch blocks below on the stack would never be triggered in Node.js previously if an exception occurred inside of `originalEval()`

packages/cli-repl/test/e2e.spec.ts
```diff
@@ -296,6 +297,14 @@ describe('e2e', function() {
       await result;
       shell.assertContainsError('interrupted');
     });
+    it('behaves normally after an exception', async() => {
+      await shell.executeLine('throw new Error()');
+      await new Promise((resolve) => setTimeout(resolve, 100));
+      shell.kill('SIGINT');
+      await shell.waitForPrompt();
+      await new Promise((resolve) => setTimeout(resolve, 100));
+      assert(!/interrupted/.test(shell.output));
```
mcasimir, line 306: Would it be good to have a `shell.assertNotContainsOutput` for consistency, that also outputs the diff?

packages/shell-evaluator/src/shell-evaluator.ts
```diff
@@ -74,8 +74,49 @@ class ShellEvaluator {
             { original: removeCommand(input.trim()), rewritten: removeCommand(rewrittenInput.trim()) }
           );
         }
+
+        let Domain;
```
mcasimir, line 78: Shell evaluator is not node / cli only. I don't have much context here, but would it be possible to place this in cli-repl?

lrlna: To me, this looks like it'd work as before for browser/java implementations, since we are `try`/`catch`ing domain on the line below.

packages/shell-evaluator/src/shell-evaluator.ts
```diff
@@ -74,8 +74,49 @@ class ShellEvaluator {
...
+          // We monkey-patch `process.domain.emit()` to avoid that, and instead
+          // handle a possible error ourselves:
+          // https://github.com/nodejs/node/blob/59ca56eddefc78bab87d7e8e074b3af843ab1bc3/lib/repl.js#L488-L493
+          // It's not clear why this is done this way in Node.js, however,
+          // removing the linked code does lead to failures in the Node.js test
+          // suite, so somebody sufficiently motivated could probably find out.
+          // For now, this is a hack and probably not considered officially
+          // supported, but it works.
+          // We *may* want to consider not relying on the built-in eval function
+          // at all at some point.
```
lrlna, line 102: yea, we want to have our own eval in the future -- https://jira.mongodb.org/browse/MONGOSH-197 . It's becoming increasingly difficult to try to maintain this version we have going here :/ :/ At this point it's just a lot of patching things up :/

packages/shell-evaluator/src/shell-evaluator.ts
addaleax: Yeah, it would work, but it’s also not a huge deal to move it … if cli-repl is the only user here that uses the Node.js core REPL implementation, it should be fine to only deal with this problem there

packages/cli-repl/test/e2e.spec.ts
addaleax: Sure, makes sense

packages/shell-evaluator/src/shell-evaluator.ts
addaleax: ah right, thanks for the context :+1:

packages/cli-repl/src/cli-repl.ts
```diff
@@ -140,7 +140,47 @@ class CliRepl {
       }
     });
 
-    const originalEval = util.promisify(this.repl.eval);
+    type REPLEvalArgs = REPLServer extends {
+      eval(...args: infer T);
+    } ? T : never;
+    const replEval = this.repl.eval.bind(this.repl);
+    const originalEval = util.promisify((...args: REPLEvalArgs) => {
```
mcasimir, line 147: what do you think to put this whole part with Domain as a separate function maybe with some explanatory name? Kinda feel  like the `start()` method of CLIRepl is already a lot to digest.

mcasimir (approved):

packages/cli-repl/src/cli-repl.ts
addaleax: sure, done!

PR #333, opened 2020-09-18
Help cleanup

mmarcon:
Cleaned up the `mongosh --help` output to only include things that are supported.

1. Removed options that are not yet supported (e.g. FLE, loading scripts from files, etc.)
2. Removed options that are not supported in the driver and therefore for the foreseeable future we won't support in the shell either:
   * `--ipv6`: if nothing is specified, the driver will try IPv6 first and fall back to IPv4 ([details](https://mongodb.github.io/node-mongodb-native/3.5/api/MongoClient.html#.connect))
   * `--tlsCRLFile`: this is not an option that the driver supports

Everything is removed rather than commented out to not leave big spaces in the help output. This PR is referenced as a comment in the code so we can go back and check what was removed.

### DIFF
--- a/packages/cli-repl/src/constants.ts
+++ b/packages/cli-repl/src/constants.ts
@@ -16,8 +16,7 @@ export const USAGE = `
 
   ${clr(i18n.__('cli-repl.args.options'), ['bold', 'yellow'])}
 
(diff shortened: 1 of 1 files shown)


addaleax (approved):

lrlna (approved):
Thanks Max! LGTM

PR #334, opened 2020-09-18
fix: fix and speed up nexe build

addaleax:
- Build Node.js in parallel
- Do not drop the nexe cache
- Target the latest Node.js 12 version
- Create a working executable.

Note that this is still pretty slow on the first run,
and creates rather large binaries.


lrlna (approved):
LGTM

PR #339, opened 2020-09-22
chore: add back support for generating signable binaries

addaleax:
Use `boxednode` to create binaries that are compiled in a way
that should enable us to sign them. `pkg` is retained as a
helper for local development, since it requires no native compilation.

Fixes MONGOSH-226


README.md
```diff
@@ -170,7 +176,7 @@ npm run compile-exec
 Compile a specific package, e.g. the `.deb` for Debian:
 
 ```shell
-npm run release -- --dry --build-variant=Debian
+npm run evergreen-release package -- --dry --build-variant=Debian
```
addaleax, line 179: (This is for the changes in 57c1456fbc8dc9b9f71 and just a drive-by fix, in case it’s confusing that this is in here)

packages/build/src/signable-compiler.ts
```diff
@@ -57,13 +35,8 @@ class SignableCompiler extends Compiler {
    * @returns {string} The target name.
    */
   determineTarget(): string {
-    switch(this.platform) {
-      case Platform.Windows: return Target.Windows;
-      case Platform.MacOs: return Target.MacOs;
-      default: return Target.Linux;
-    }
+    return '^12.0.0';
```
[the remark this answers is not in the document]
mcasimir: > Do we want to make this part of the build config?

I think it would be very nice.

> "determineTarget()" and "Is the lack of cross-“compilation” an issue here"

Kind of, but not only. I think that can be changed anyway. 

We had some initial struggle to tame evergreen. That brought us to rely a lot on what we knew we could rely on, and to have this "detection of platform" approach in all sort of places. In all fairness that has the great merit of having brought us this far and have a release available in all the channels we have to.

On a less bright note, that creates a fair amount of non-obvious coupling, It would be really nice in a future iteration to make use of what we now know evergreen can offer, and have a more explicit way to instruct the CI what to build and when.


@lrlna i think you have most context here.

packages/build/src/compile-exec.ts
```diff
@@ -51,23 +52,24 @@ const compileExec = async(
...
   await generateInput(input, execInput, analyticsConfig, segmentKey);
 
   const executable = executablePath(outputDir, platform);
   console.log('mongosh: creating binary:', executable);
 
-  if (platform === Platform.MacOs) {
-    const { exec } = require('pkg');
-    await new UnsignableCompiler(input, executable, platform)
-      .compile(exec);
+  if (signable) {
```
mcasimir, line 66: will we ever use the non-signable one in the ci?

mcasimir (approved):
I'd just remove the non signable one if we don't use it

packages/build/src/signable-compiler.ts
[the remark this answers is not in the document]
addaleax: > > Do we want to pin a specific version
> 
> I'd really recommend to have the same version both here and in `.evergreen/.install_node` and `InstallNode.ps1`

Okay, that’s a good point :+1: But still, do we want a specific version or just “latest v12.x/latest v14.x”?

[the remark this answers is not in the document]
lrlna: > @lrlna i think you may have more context here.

`determineTarget()` in this particular case was only used for `pkg` and `nexe`. So, if a binary that `boxednode` builds can be used to package for `.tgz`, `.deb` and `.rpm`, we do not need this information.
> "detection of platform" approach in all sort of places
We use `os.platform()` and build variant information (`rhel`, `macos`, `debian` etc) when creating a tarball/archive. I haven't found a better way for us to use this from the CI directly, which is why we are detecting current platform or build variant in `tarball.ts`

packages/build/src/compile-exec.ts
addaleax: I don’t think so, no … happy to remove `pkg` support altogether if that makes sense to you

.evergreen.yml
```diff
@@ -273,8 +286,8 @@ buildvariants:
       - name: test
       - name: release_linux
   - name: rhel
-    display_name: "RHEL 8.0"
-    run_on: rhel80-large
+    display_name: "RHEL 8.0 (in Ubuntu docker)"
+    run_on: ubuntu1804-test
```
mcasimir, line 290: `ubuntu1804-test` has docker installed?

addaleax: Apparently :woman_shrugging:

mcasimir (approved):

lrlna (approved):
Excited for this to land!

PR #340, opened 2020-09-23
Create cursors with expected Mongo instance

kornilova203:
Cursors were created with `Collection` and `Database` instances instead of expected `Mongo`


addaleax (approved):
LGTM

Curious that typescript doesn’t catch this, though

PR #342, opened 2020-09-25
MONGOSH-278 - Notarize macOS package

addaleax:


.evergreen.yml
```diff
@@ -263,21 +265,25 @@ tasks:
...
     run_on: macos-1014-codesign
+    tasks:
+      - name: release_macos
+  - name: darwin
+    display_name: "MacOS Mojave"
+    run_on: macos-1014
     tasks:
       - name: check
       - name: test
-      - name: release_macos
```
addaleax, line 280: The changes here are mostly to move work off the codesign machine – there’s only one of them versus 70 other `macos-1014` hosts, and we’re already hogging it a lot.

This also helps with parallelism, because we don’t want to have to wait for that single machine to become available and process `check` and `test` before we can proceed with the release, even for other platforms

package.json
```diff
@@ -95,7 +95,7 @@
     "mocha": "^7.1.2",
     "mongodb-js-precommit": "^2.0.0",
     "mongodb-runner": "^4.7.5",
-    "node-codesign": "durran/node-codesign",
+    "node-codesign": "github:addaleax/node-codesign",
```
addaleax, line 98: This is just so that the `entitlements` option becomes available … this should probably become some official npm package at some point, I might ping Durran about it after this is merged :)

lrlna: yea i was looking at that before, and we should just publish it.

addaleax: Yeah, I’ll ping Durran and if nothing speaks against it I’ll move this into the org (i.e. move it back, but as `node-codesign`) and publish it :)

.evergreen.yml
addaleax: Actually, I had to undo this because it messed with the `BUILD_VARIANT` variable to have two of these in there … so it’s not this straightforward and probably best kept out of here

mcasimir (approved):
Great!

PR #345, opened 2020-09-29
chore: try to use sccache in CI

addaleax:


.evergreen.yml
```diff
@@ -26,9 +29,15 @@ variables:
             -e EVERGREEN_EXPANSIONS_PATH=/tmp/build/tmp/expansions.yaml \
             -e NODE_JS_VERSION \
             -e BUILD_VARIANT \
+            -e AWS_ACCESS_KEY_ID -e AWS_SECRET_ACCESS_KEY -e SCCACHE_BUCKET \
```
mcasimir, line 32: this is going to be done by `scripts/import-expansions.js`

addaleax: That wouldn’t load it under the right names, though (unless we also add these here to the project config)

mcasimir: That's right, let's see how it goes then!

mcasimir (approved):

PR #346, opened 2020-09-29
chore(build): upload mongosh PPA to ubuntu2004 dist

lrlna:
We were building previously only for 18.04. This adds an additional
20.04 upload.


mcasimir (approved):
👍

PR #347, opened 2020-09-29
MONGOSH-638 - Enable sharding methods

aherlihy:
- I didn't include integration tests in this PR because it will make more sense to do so when there are more methods to test as one (i.e. enableSharding, then disableSharding, etc)
- I didn't include tests for sh.status since it's mostly legacy and just very complicated

### DIFF
--- a/packages/shell-api/src/helpers.ts
+++ b/packages/shell-api/src/helpers.ts
@@ -1,3 +1,4 @@
+/* eslint-disable complexity */
 /**
  * Helper method to adapt aggregation pipeline options.
  * This is here so that it's not visible to the user.
@@ -7,6 +8,7 @@
 import { DatabaseOptions, Document } from '@mongosh/service-provider-core';
 import { MongoshInvalidInputError } from '@mongosh/errors';
 import crypto from 'crypto';
+import Mongo from './mongo';
 
 export function adaptAggregateOptions(options: any = {}): {
   providerOptions: Document;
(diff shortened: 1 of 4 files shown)


packages/shell-api/src/helpers.ts
```diff
@@ -120,3 +122,322 @@ export function processDigestPassword(username, passwordDigestor, command): any
...
+  const settingsColl = configDB.getCollection('settings');
+  const changelogColl = configDB.getCollection('changelog');
+
+  const version = await versionColl.findOne();
+  if (version === null) {
+    throw new MongoshInvalidInputError('This db does not have sharding enabled.' +
+      ' be sure you are connecting to a mongos from the shell and not to a mongod.');
+  }
+
+  result.shardingVersion = await versionColl.findOne();
```
addaleax, line 152: This could be

```suggestion
  result.shardingVersion = version;
```

right?

packages/shell-api/src/helpers.ts
```diff
@@ -120,3 +122,322 @@ export function processDigestPassword(username, passwordDigestor, command): any
...
+
+  if (activeMigrations.length > 0) {
+    balancerRes['Collections with active migrations'] = activeMigrations.map((migration) => {
+      return `${migration._id} started at ${migration.when}`;
+    });
+  }
+
+  // Actionlog and version checking only works on 2.7 and greater
+  let versionHasActionlog = false;
+  const metaDataVersion = (await versionColl.findOne()).currentVersion;
```
addaleax, line 246:
```suggestion
  const metaDataVersion = version.currentVersion;
```

packages/shell-api/src/helpers.ts
```diff
@@ -120,3 +122,322 @@ export function processDigestPassword(username, passwordDigestor, command): any
...
+    activeLocks.forEach((lock) => {
+      activeMigrations.push({ _id: lock._id, when: lock.when });
+    });
+  }
+
+  if (activeMigrations.length > 0) {
+    balancerRes['Collections with active migrations'] = activeMigrations.map((migration) => {
+      return `${migration._id} started at ${migration.when}`;
+    });
+  }
```
addaleax, line 242: A lot of the db queries here are run sequentially, e.g. those that fill `balancerRes`, could we run them in parallel to reduce runtime?

packages/shell-api/src/helpers.ts
aherlihy: Yes, thanks!

packages/shell-api/src/helpers.ts
aherlihy: Sure, what would that look like? I haven't actually done that before :D

addaleax: So … I haven’t tested this, but I the general idea would be to turn e.g.

```js
// Is the balancer currently enabled
const balancerRes = {};
const balancerEnabled = await settingsColl.findOne({ _id: 'balancer' }) as any;
balancerRes['Currently enabled'] = balancerEnabled === null || !balancerEnabled.stopped ? 'yes' : 'no';

// Is the balancer currently active
let balancerRunning = 'unknown';
try {
  const balancerStatus = await configDB.adminCommand({ balancerStatus: 1 });
  balancerRunning = balancerStatus.inBalancerRound ? 'yes' : 'no';
} catch (err) {
  // pass, ignore all error messages
}
balancerRes['Currently running'] = balancerRunning;

// Output the balancer window
const settings = await settingsColl.findOne({ _id: 'balancer' });
if (settings !== null && settings.hasOwnProperty('activeWindow')) {
  const balSettings = settings.activeWindow;
  balancerRes['Balancer active window is set between'] = `${balSettings.start} and ${balSettings.stop} server local time`;
}

// Output the list of active migrations
const activeLocks = await configDB.getCollection('locks').find({ state: { $eq: 2 } }).toArray();
const activeMigrations = [];
if (activeLocks !== null) {
  activeLocks.forEach((lock) => {
    activeMigrations.push({ _id: lock._id, when: lock.when });
  });
}

if (activeMigrations.length > 0) {
  balancerRes['Collections with active migrations'] = activeMigrations.map((migration) => {
    return `${migration._id} started at ${migration.when}`;
  });
}
```

into something like

```js
// Is the balancer currently enabled
const balancerRes = {};

await Promise.all([
  (async () => {
    const balancerEnabled = await settingsColl.findOne({ _id: 'balancer' }) as any;
    balancerRes['Currently enabled'] = balancerEnabled === null || !balancerEnabled.stopped ? 'yes' : 'no';
  })(),
  
  (async () => {
    // Is the balancer currently active
    let balancerRunning = 'unknown';
    try {
      const balancerStatus = await configDB.adminCommand({ balancerStatus: 1 });
      balancerRunning = balancerStatus.inBalancerRound ? 'yes' : 'no';
    } catch (err) {
      // pass, ignore all error messages
    }
    balancerRes['Currently running'] = balancerRunning;
  })(),
  
  (async () => {
    // Output the balancer window
    const settings = await settingsColl.findOne({ _id: 'balancer' });
    if (settings !== null && settings.hasOwnProperty('activeWindow')) {
      const balSettings = settings.activeWindow;
      balancerRes['Balancer active window is set between'] = `${balSettings.start} and ${balSettings.stop} server local time`;
    }
  })(),

  (async () => {
    // Output the list of active migrations
    const activeLocks = await configDB.getCollection('locks').find({ state: { $eq: 2 } }).toArray();
    const activeMigrations = [];
    if (activeLocks !== null) {
      activeLocks.forEach((lock) => {
        activeMigrations.push({ _id: lock._id, when: lock.when });
      });
    }

    if (activeMigrations.length > 0) {
      balancerRes['Collections with active migrations'] = activeMigrations.map((migration) => {
        return `${migration._id} started at ${migration.when}`;
      });
    }
  })()
]);
```

This is probably not the only part here where we could do this, and maybe not ideal, but I think the idea is visible?

aherlihy: yes, that makes sense, thanks!

aherlihy: Ah I didn't push the change before merging, I will make a new PR

PR #349, opened 2020-09-29
feat(shell-api): MONGOSH-358 add rs.config, rs.intiate, rs.reconfig commands

lrlna:
## Description

This PR removes MongoshUnimplementedError for Replica Set API and adds
rs.config, rs.initiate, rs.reconfig APIs. 

This PR exludes end-to-end tests, as it would be good to have a few more
implemented APIs so we can test replication properly.

### DIFF
--- a/packages/shell-api/src/replica-set.spec.ts
+++ b/packages/shell-api/src/replica-set.spec.ts
@@ -1,25 +1,203 @@
-import { expect } from 'chai';
-import sinon from 'ts-sinon';
+import { bson, ServiceProvider } from '@mongosh/service-provider-core';
+import { StubbedInstance, stubInterface } from 'ts-sinon';
+import ShellInternalState from './shell-internal-state';
+import { signatures } from './decorators';
 import ReplicaSet from './replica-set';
-// import { asShellResult } from './enums';
+import { EventEmitter } from 'events';
+import { expect } from 'chai';
+import Mongo from './mongo';
+import {
+  ADMIN_DB,
+  ALL_PLATFORMS,
+  ALL_SERVER_VERSIONS,
+  ALL_TOPOLOGIES,
+  asShellResult
+} from './enums';
 
 describe('ReplicaSet', () => {
-  describe('unimplemented', () => {
-    it('throws', () => {
-      const mSpy = sinon.spy();
-      const rs = new ReplicaSet(mSpy);
-      try {
-        (rs as any).anything();
-      } catch (e) {
-        expect(e.name).to.be.equal('MongoshUnimplementedError');
-      }
+  describe('help', () => {
+    const apiClass: any = new ReplicaSet({});
+
+    it('calls help function', async() => {
+      expect((await apiClass.help()[asShellResult]()).type).to.equal('Help');
+      expect((await apiClass.help[asShellResult]()).type).to.equal('Help');
+    });
+
+    it('calls help function for methods', async() => {
+      expect((await apiClass.initiate.help()[asShellResult]()).type).to.equal('Help');
+      expect((await apiClass.initiate.help[asShellResult]()).type).to.equal('Help');
     });
   });
-  // describe('help', () => {
-  //   const apiClass: any = new ReplicaSet({});
-  //   it('calls help function', async() => {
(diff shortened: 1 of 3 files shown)


lrlna:
@aherlihy @mcasimir @addaleax if this looks good, feel free to merge it while I am away. If this needs changes, also feel free to push them to this branch as well

packages/shell-api/src/replica-set.ts
```diff
@@ -18,15 +19,49 @@ export default class ReplicaSet extends ShellApiClass {
...
+  }
+
+  /**
+   *  rs.reconfig calls replSetReconfig admin command.
+   *
+   *  @param config
+   *  @param options
+   */
+  @returnsPromise
+  async reconfig(config: any, options?: any): Promise<any> {
```
addaleax, line 53:
```suggestion
  async reconfig(config: object, options?: object): Promise<any> {
```

packages/shell-api/src/replica-set.ts
```diff
@@ -18,15 +19,49 @@ export default class ReplicaSet extends ShellApiClass {
...
-    return proxy;
+  }
+
+  /**
+   *  rs.initiate calls replSetInitiate admin command.
+   *
+   * @param config
+   */
+  @returnsPromise
+  async initiate(config?: any): Promise<any> {
```
addaleax, line 30:
```suggestion
  async initiate(config?: object): Promise<any> {
```

addaleax (approved):
LGTM, I only have two minor comments about the typescript signatures … if somebody else agrees that those suggestions make sense, I’d apply them, but it’s definitely optional either way :)

packages/shell-api/src/replica-set.ts
mcasimir: I totally agree with replacing `any`, although `object` is not what we want (it only matches `{}`). `Record<string, any>` could be an option.

addaleax: > Agree with replacing `any`, although `object` is not what we want (it only matches `{}`)

Right, but `{}` matches a superset of all objects. :) I think the difference between it any `Record<string, any>` is basically that the latter excludes symbol keys, which also seems fine to me.

packages/shell-api/src/replica-set.spec.ts
```diff
@@ -1,25 +1,203 @@
...
+        ]
+      };
+
+      it('calls serviceProvider.runCommandWithCheck without optional arg', async() => {
+        await rs.initiate();
+
+        expect(serviceProvider.runCommandWithCheck).to.have.been.calledWith(
+          ADMIN_DB,
+          {
+            replSetInitiate: undefined
```
aherlihy, line 88: Nit: this should probably default to `{}` instead of undefined, right?

packages/shell-api/src/replica-set.ts
aherlihy: Could also use `={}` for default argument to avoid the TS issue and prevent sending undefined to the server?

aherlihy (approved):

packages/shell-api/src/replica-set.ts
mcasimir: > {} matches a superset of all objects

Oh no .. i mean `object` in typescript only matches an empty object. That it would match any object is what anyone would expect .. for some reason it doesn't seem to be the case.

```
function foo1(x: object) {
  return x.bar; // complains
}

function foo2(x: Object) {
  return x.bar; // complains
}

function foo3(x: {[key: string]: any}) {
  return x.bar; // ok
}

function foo4(x: Record<string, any>) {
  return x.bar; // ok
}
```

https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABMOcCMAKAHgLkXAIwCsBTaASkQG8AoRRAJxKhAaSwDoCBDBgbhoBfGjVCRYCZKgBM2PAHliZKJVr0mLNok49+QkWOjwkKOAGY51ANoBrEgE88AZygMYYAOYBdPNzD3BVTpGZlZ2Ll4BYVFwI0lTABZLACUyOAYAEwAeFzdPABpEP3sAPiD1UK0dSKEgA

PR #353, opened 2020-10-02
MONGOSH-369 - DB/Coll sharding methods

aherlihy:


packages/shell-api/src/collection.ts
```diff
@@ -1418,4 +1425,92 @@ export default class Collection extends ShellApiClass {
...
+
+    for (const extShardStats of collStats) {
+      // Extract and store only the relevant subset of the stats for this shard
+      const shardStats = {
+        shardId: extShardStats.shard,
+        host: (await config.getCollection('shards').findOne({ _id: extShardStats.shard })).host,
+        size: extShardStats.storageStats.size,
+        count: extShardStats.storageStats.count,
+        numChunks:
+          await config.getCollection('chunks').countDocuments({ ns: extShardStats.ns, shard: extShardStats.shard }),
```
addaleax, line 1468: These `await`s could also be pararellized here, considering that it’s used in a loop?

addaleax (approved):

packages/shell-api/src/collection.ts
aherlihy: Just pushed the change, let me know if you think it looks ok!

addaleax (approved):

PR #355, opened 2020-10-02
MONGOSH-364 - Add shard tag/zone methods

aherlihy:


packages/shell-api/src/shard.ts
```diff
@@ -72,12 +79,91 @@ export default class Shard extends ShellApiClass {
...
+  @returnsPromise
+  @serverVersions(['3.4.0', ServerVersions.latest])
+  async addShardTag(shard: string, tag: string): Promise<any> {
+    try {
+      return await this.addShardToZone(shard, tag);
+    } catch (error) {
+      if (error.codeName === 'CommandNotFound') {
+        error.message = `${error.message}. Are you connected to version > 3.4?`;
+      }
+      throw error;
```
addaleax, line 130: Not necessarily something for this PR, but is this something that would make sense to implement inside of `@serverVersions`?

packages/shell-api/src/shard.ts
```diff
@@ -72,12 +79,91 @@ export default class Shard extends ShellApiClass {
...
+    });
+  }
+
+  @returnsPromise
+  @serverVersions(['3.4.0', ServerVersions.latest])
+  async addTagRange(namespace: string, min: Document, max: Document, zone: string): Promise<any> {
+    assertArgsDefined(namespace, min, max, zone);
+    this._emitShardApiCall('updateZoneKeyRange', { namespace, min, max, zone });
+
+    await getConfigDB(this._mongo);
```
addaleax, line 154: If these calls are there for the side-effect of checking whether we are connected to a `mongos`, it might be worth leaving a comment about that :)

addaleax (approved):

packages/shell-api/src/shard.ts
aherlihy: This one is a bit of an odd case because the old shell changed the behavior of this helper in 3.4, but we only support 4.0+ in the new shell so we don't want to handle the case that the user is calling this helper on an old server version. 

If we added it to the decorator, it would apply for all methods and I don't think we want to add this hint to every time we try a command that doesn't exist when we decorate the method with a specific version (since the decorations, for now, are just for autocomplete). I think the error message could be a bit clearer, since what I'm really trying to say is 'are you expecting this to work the way the old shell did, on your old version of the server, because it won't work because we don't support that'.

packages/shell-api/src/shard.ts
aherlihy: Will do!

packages/shell-api/src/shard.ts
addaleax: Okay, makes sense, I didn’t have that context :)

aherlihy: I've updated the message to make it clearer what's going on, I think its better now

PR #356, opened 2020-10-02
First go at parallelizing sh.status

aherlihy:
I forgot this was in progress before merging the last pr, so making a new one.

I see there are a few other places in this method where we could parallelize, but I wanted to keep the changes minimal here since the code is mostly from the old shell and since we weren't also able to port the tests, I'm hesitant to change too much of the code without introducing new tests.


addaleax (approved):
LGTM :)

PR #358, opened 2020-10-05
fix: MONGOSH-355 - error if config folder can't be created

mcasimir:
Properly exits with an error if the config / log folder / can't be created.

Before:

Screenshot 2020-10-05 at 12 01 28

After:

Screenshot 2020-10-05 at 12 01 41

Other changes:

- Changed a test for barque which timeouts on a slow connection. Replaced it with a mocked response, and checking that the file can actually be downloaded doing an HEAD request.


packages/build/src/barque.spec.ts
```diff
@@ -163,8 +165,32 @@ describe('Barque', () => {
...
+      const response = await fetch(LATEST_CURATOR, {
+        method: 'HEAD'
+      });
+
+      expect(response.ok).to.be.true;
+    });
+  });
+
   describe('.extractLatestCurator', () => {
-    it('extractss latest curator to tmp directory', async() => {
```
mcasimir, line 167: this used to try and download the file for real, which timeouts with a slow connection,  and .. unfortunately my connection is slow :)

packages/cli-repl/src/cli-repl.ts
```diff
@@ -50,25 +50,25 @@ class CliRepl {
...
+    this.lineByLineInput = new LineByLineInput(process.stdin);
+
     if (this.isPasswordMissing(driverOptions)) {
       this.requirePassword(driverUri, driverOptions);
     } else {
       this.setupRepl(driverUri, driverOptions).catch((error) => {
-        this.bus.emit('mongosh:error', error);
-        console.log(formatError(error));
-        return process.exit();
+        return this._fatalError(error);
```
aherlihy, line 71: Nice!

aherlihy (approved):

PR #361, opened 2020-10-05
MONGOSH-376 - Include metadata about data source in result

addaleax:
For results from calls on collections, add metadata about the source of
the returned data (database name, collection name, method, arguments)
so that consumers of the shell can figure out from what source
the result was generated, e.g. in order to understand how to modify
that original data.

This also adds a fake `Document` type to the `ShellResult` reported
for e.g. `findOne()`, so that it is possible to tell those apart
from other plain JavaScript objects, e.g. in order to be able to
know whether the result is a modifiable object in the database or not.

### DIFF
--- a/packages/shell-api/src/aggregation-cursor.ts
+++ b/packages/shell-api/src/aggregation-cursor.ts
@@ -5,7 +5,8 @@ import {
   returnType,
   hasAsyncChild,
   ShellApiClass,
-  ShellResult
+  ShellResult,
+  resultSource
 } from './decorators';
 import {
   Cursor as ServiceProviderCursor,
@@ -47,7 +48,8 @@ export default class AggregationCursor extends ShellApiClass {
   async [asShellResult](): Promise<ShellResult> {
     return {
       type: 'AggregationCursor',
-      value: this._mongo._serviceProvider.platform === ReplPlatform.JavaShell ? this : await this._asPrintable()
+      value: this._mongo._serviceProvider.platform === ReplPlatform.JavaShell ? this : await this._asPrintable(),
+      source: this[resultSource] ?? undefined
     };
   }
 
--- a/packages/shell-api/src/decorators.ts
+++ b/packages/shell-api/src/decorators.ts
@@ -9,6 +9,13 @@ import {
   asShellResult
 } from './enums';
 import { MongoshInternalError } from '@mongosh/errors';
+import { addHiddenDataProperty } from './helpers';
+
+const addSourceToResultsSymbol = Symbol('@@mongosh.addSourceToResults');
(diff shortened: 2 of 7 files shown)


addaleax:
I’ve checked that this does make the necessary information available on the VSCode side

Also, the naming of some of the things in here could probably be improved, feel free to make suggestions :)

packages/shell-api/src/decorators.ts
```diff
@@ -34,6 +50,55 @@ export class ShellApiClass implements ShellApiInterface {
...
+// We also attach the `shellApiType` and `asShellResult` properties to the
+// return type (if that is possible and they are not already present), so that
+// we can also provide sensible information for methods that do not return
+// shell classes, like db.coll.findOne() which returns a Document (i.e. a plain
+// JavaScript object).
+function wrapWithAddSourceToResult(fn: Function, functionName: string): Function {
+  function addSource<T extends {}>(result: T, obj: any, args: any[]): T {
+    if (typeof result === 'object' && result !== null) {
+      const resultSourceInformation: ShellResultSourceInformation = {
+        dataSource: obj._dataSourceIdentifier(),
```
mcasimir, line 67: we don't have a concept of `dataSource` i think the same is called namespace in our domain area, is probably better to keep the language consistent and use that name.

addaleax: `namespace` sounds good :+1:

packages/shell-api/src/decorators.ts
```diff
@@ -34,6 +50,55 @@ export class ShellApiClass implements ShellApiInterface {
...
+        call: functionName,
```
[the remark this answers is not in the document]
addaleax: Yeah, I’ve gone back and forth on this a bit myself … it feels like something that could be really useful, e.g. for figuring out how to get the data that this call resulted in a second time or displaying what call resulted in this output. I’d lean towards keeping it since it’s essentially free and data that is always going to be available anyway, but I’m also happy to remove it

packages/shell-api/src/aggregation-cursor.ts
```diff
@@ -47,7 +48,8 @@ export default class AggregationCursor extends ShellApiClass {
   async [asShellResult](): Promise<ShellResult> {
     return {
       type: 'AggregationCursor',
-      value: this._mongo._serviceProvider.platform === ReplPlatform.JavaShell ? this : await this._asPrintable()
+      value: this._mongo._serviceProvider.platform === ReplPlatform.JavaShell ? this : await this._asPrintable(),
+      source: this[resultSource] ?? undefined
```
aherlihy, line 52: Oh cool, I've never seen '??' before.

aherlihy (approved):

PR #362, opened 2020-10-05
MONGOSH-367 - Chunk management methods

aherlihy:


aherlihy:
Closed in favor of #363

PR #363, opened 2020-10-05
MONGOSH-367 - Chunk management methods [rebased]

aherlihy:
Sometimes it's easier to make a new branch than rebasing

### DIFF
--- a/packages/shell-api/src/shard.ts
+++ b/packages/shell-api/src/shard.ts
@@ -215,4 +215,61 @@ export default class Shard extends ShellApiClass {
       throw error;
     }
   }
+
+  @returnsPromise
+  @serverVersions(['3.4.0', ServerVersions.latest])
+  async enableAutoSplit(): Promise<any> {
+    assertArgsDefined();
+    this._emitShardApiCall('enableAutoSplit', {});
+    const config = await getConfigDB(this._mongo);
+    return await config.getCollection('settings').updateOne(
+      { _id: 'autosplit' },
+      { $set: { enabled: true } },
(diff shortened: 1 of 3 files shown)


packages/shell-api/src/shard.ts
```diff
@@ -215,4 +215,61 @@ export default class Shard extends ShellApiClass {
...
+  @returnsPromise
+  @serverVersions(['3.4.0', ServerVersions.latest])
+  async disableAutoSplit(): Promise<any> {
+    assertArgsDefined();
+    this._emitShardApiCall('disableAutoSplit', {});
+    const config = await getConfigDB(this._mongo);
+    return await config.getCollection('settings').updateOne(
+      { _id: 'autosplit' },
+      { $set: { enabled: false } },
+      { upsert: true, writeConcern: { w: 'majority', wtimeout: 30000 } }
```
addaleax, line 241: Is it worth making the `writeConcern` configurable or will users generally not care about that?

addaleax (approved):

packages/shell-api/src/shard.ts
aherlihy: writeConcern is generally configurable for write commands, but in this case it's not _really_ a write command, it's a configuration command so we want to enforce that it reaches the majority. (Writing to a config collection is sort of a legacy way of making a configuration command, it used to be that there were system collections that users could write to in order to make configuration changes, but the server has been moving away from that and towards specific config commands).

There could be unexpected consequences if a user doesn't fully understand write concern, sets it to something weak, then starts writing data expecting autosplit to be fully turned on. So we do it for them, and ensure that can't happen.

PR #364, opened 2020-10-05
MONGOSH-366 - Balancer methods

aherlihy:


packages/i18n/src/locales/en_US.js
```diff
@@ -1251,6 +1251,47 @@ const translations = {
...
+            },
+            disableBalancing: {
+              link: 'https://docs.mongodb.com/manual/reference/method/sh.disableBalancing',
+              description: 'disable balancing',
+              example: 'sh.disableBalancing(ns)',
+            },
+            getBalancerState: {
+              link: 'https://docs.mongodb.com/manual/reference/method/sh.getBalancerState',
+              description: 'Enables or disables the balancer. Starting in MongoDB 4.2, the method also enables auto-splitting if enabling the balancer and disables auto-splitting if disabling the balancer. ' +
+                'Use sh.getBalancerState() to determine if the balancer is currently enabled or disabled and sh.isBalancerRunning() to check its current state.',
```
addaleax, line 1272: Is this accurate? I would not expect a method named `getBalancerState()` to enable or disable the balancer

packages/i18n/src/locales/en_US.js
```diff
@@ -1251,6 +1251,47 @@ const translations = {
...
+            },
+            startBalancer: {
+              link: 'https://docs.mongodb.com/manual/reference/method/sh.startBalancer',
+              description: 'Calls startBalancer for a sharded DB. uses the balancerStart command',
+              example: 'sh.startBalancer(timeout)',
+            },
+            stopBalancer: {
+              link: 'https://docs.mongodb.com/manual/reference/method/sh.stopBalancer',
+              description: 'Calls stopBalancer for a sharded DB. uses the balancerStop command',
+              example: 'sh.stopBalancer(timeout)',
```
addaleax, line 1288: Maybe we could explicitly mention that these don’t wait for the balancer state to actually change, because at least the NHTT makes claims to the contrary (for the old shell, which don’t even seem to be true there anymore, but still…)? :)

addaleax (approved):

packages/i18n/src/locales/en_US.js
aherlihy: Oh good catch! it got mixed up with setBalancerState

packages/i18n/src/locales/en_US.js
aherlihy: Sure, I'll get the copy from the docs directly

aherlihy: Went and improved all the descriptions to be the ones here: https://docs.mongodb.com/manual/reference/method/#sharding. Not sure how I thought those descriptions were intelligible before!!

aherlihy: When I close the ticket I will mark it for the docs team so they know that the behavior is changed and that it does not wait

PR #368, opened 2020-10-06
MONGOSH-360 - Replica set management

aherlihy:


packages/shell-api/src/replica-set.ts
```diff
@@ -125,6 +125,48 @@ export default class ReplicaSet extends ShellApiClass {
...
+      }
+    );
+  }
+
+  @returnsPromise
+  async stepDown(stepdownSecs?: number, catchUpSecs?: number): Promise<any> {
+    assertArgsType([stepdownSecs, catchUpSecs], ['number', 'number']);
+    this._emitReplicaSetApiCall('stepDown', { stepdownSecs, catchUpSecs });
+    const cmd = {
+      replSetStepDown: stepdownSecs || 60,
```
addaleax, line 146: I assume `0` would not be a meaningful value for `stepdownSecs` here?

addaleax (approved):

packages/shell-api/src/replica-set.ts
aherlihy: That's a good point - if it was 0 I'm not sure what the server would do. The old shell would replace with 60 but I think that was just a side effect of how it was written so I'm going to update it only to default to 60 when undefined. Thanks!

PR #341, opened 2020-09-24
Use ShellEvaluator in java shell

kornilova203:
There are couple of commits that were cherry-picked from my [java-shell-0.0.3](https://github.com/kornilova203/mongosh/tree/java-shell-0.0.3) branch and adapted to current code in master.
Java shell from this branch is used in DataGrip consoles.
Now I want to pull changes from this branch to master.

Most of the changes are in Java codebase, there is no need to review them, all java tests pass. But there is a change in mongosh core:
Shell evaluator fetches cursors and returns arrays with result. In DataGrip we want to have control over cursor and fetch it ourselves. This way we are able to start rendering results without waiting for whole fetch to be done.

See:
1. [`asShellResult` implementation in `AggregationCursor`](https://github.com/mongodb-js/mongosh/compare/master...kornilova203:use-shellevaluator-in-java-shell?expand=1#diff-685046018c1a23ff5d025a0eb4f1cb5aR47)
2. [`asShellResult` implementation in `Cursor`](https://github.com/mongodb-js/mongosh/compare/master...kornilova203:use-shellevaluator-in-java-shell?expand=1#diff-bfce9978e5907f6b28eb5446165997ebR33)

PR #371, opened 2020-10-07
chore: clean up ShellResult usage

addaleax:
Provide more encapsulation by using a fixed interface for other
packages to work with `shell-api`, without the need to use internal
methods of `shell-api` objects.

- Inside `shell-api`:
  - Consistently use `async [asPrintable]()`, `[resultSource]`,
    `[shellApiType]` (all as symbol properties)
    - These are all non-exported, local symbol properties, i.e.
      not for outside access
  - Deprecate `_asPrintable`
  - Deprecate `ShellResult.value` in favor of more expressively named
    `.rawValue` and `.printable` (`.value` is very generic, and in
    particular I would have expected it to be the evaluation result
    and not a formatting-ready version of it)
  - Remove `[asShellResult]()`
- On the API surface of `shell-api`:
  - Remove `asShellResult` and `shellApiType`
  - Provide `getShellApiType()` and `toShellResult()` as the sole APIs
    to turn an evaluation result into a `ShellResult`
  - In particular, this applies for *all* values – no special handling
    is necessary in `shell-evaluator` anymore

This is partly in preparation for MONGOSH-237, so that we do not start
introducing calls to half-internal methods like `_asPrintable()`
in more places than we already do.


packages/shell-api/src/decorators.ts
```diff
@@ -36,27 +36,81 @@ export interface ShellResultSourceInformation {
...
   source?: ShellResultSourceInformation;
+
+  /// @deprecated Use either `printable` or `rawValue`.
+  value: any;
 }
 
 export class ShellApiClass implements ShellApiInterface {
   help: any;
-  [asShellResult](): any {
+  get [shellApiType](): string {
```
mcasimir, line 61: @aherlihy did we have issues with getters and setters in the past with he java shell?

packages/shell-api/src/decorators.ts
```diff
@@ -36,27 +36,81 @@ export interface ShellResultSourceInformation {
...
   }
-  _asPrintable(): any {
+  set [shellApiType](value: string) {
+    addHiddenDataProperty(this, shellApiType, value);
+  }
+  [asPrintable](): any {
     return Object.assign({}, this);
   }
+
+  /// @deprecated: Use `toShellResult(value).printable` instead.
```
mcasimir, line 71: I'd just drop it and fix it in the consumers, not even sure anyone is using it right now, didn't Jetbrains switch to use the `ShellEvaluator`?

packages/shell-api/src/enums.ts
```diff
@@ -26,8 +26,11 @@ export enum DBQueryOption {
...
 
 export const shellApiType = Symbol.for('@@mongosh.shellApiType');
-// TODO: Would require changes to java-shell
-// export const asPrintable = Symbol('asPrintable');
-export const asShellResult = Symbol.for('@@mongosh.asShellResult');
+// TODO: This requires changes to java-shell.
+export const asPrintable = Symbol.for('@@mongosh.asPrintable');
+export const namespaceInfo = Symbol.for('@@mongosh.namespaceInfo');
+
+/// @deprecated toShellResult() returns both the raw value and the printable version now.
```
mcasimir, line 33: same here, would not bother deprecating and just try to make the change in the java shell, then drop a message in mongosh-jetbrains.

mcasimir (commented):
Looks like a good improvement to me! @aherlihy though is probably a better reviewer here :). Please before merging make sure getter and setters won't cause issues with Jetbrains, i have a vague memory of that being a thing. Also probably we could spare the deprecation if the change in Kotlin is simple enough.

Otherwise let's make a ticket for that and make sure we'd follow up and don't leave it around.

packages/shell-api/src/decorators.ts
aherlihy: Exactly, you would have to update the java shell since they use this method directly (which is why it was the only non-symbol of these internal methods)

aherlihy:
Hi! Have you tried running the java shell tests? In the repo there are instructions for how to run them (it's pretty straight forward)

packages/shell-api/src/decorators.ts
addaleax: Are you thinking about `_asPrintable()` here? That is still left as an alias for the java shell (`asShellResult` is not mentioned in their source, and always was a symbol)

packages/shell-api/src/decorators.ts
addaleax: They are still calling `_asPrintable()` in a few places right now … I can try to figure out how to move them away from it, but for me, the `java-shell` test suite is not passing locally on `master` to begin with

addaleax:
> Hi! Have you tried running the java shell tests? In the repo there are instructions for how to run them (it's pretty straight forward)

Yeah, I did, but they failed for me locally even on `master` … maybe that changed after #372 this morning, the description at least says that they do pass now :)

packages/shell-api/src/decorators.ts
aherlihy: Yes I was! Sorry the comment was super unclear. I just meant that the reason _asPrintable was different from the others was because of the java shell

aherlihy:
They do pass for me on current master, you have to create a user before the tests run on the DB you're running locally. So if you start a standalone at 27017 and run `use admin; db.createRole({user: 'admin', pwd: 'admin', roles: ['root']})` then the tests should pass. (It won't be forever - hopefully soon we'll add the java tests to the CI so we can automatically tell if something is off. Which means we'll have to update the java tests so they set up / tear down the servers they test against, but that's future work)

I rebased your branch with master and it looks like the tests are passing, so it seems there's nothing to worry about there :)

addaleax:
> I rebased your branch with master (locally) and it looks like the tests are passing, so it seems there's nothing to worry about there :)

Yeah, I’m currently trying to figure out how to get rid of `_asPrintable` there as well, as @mcasimir suggested – seems like that should be doable :)

addaleax:
Okay, so:

- Removed `_asPrintable` as a non-symbol/semi-public method altogether, including from the Java shell
- Removed the special handling for the Cursor classes that was introduced in https://github.com/mongodb-js/mongosh/pull/341, the Java shell doesn’t need it anymore

The only remaining deprecation introduced here is the one for `ShellResult.value` here, the VSCode extension uses that, but we can also remove it altogether here and just take care of that at the next mongosh update in VSCode.

packages/java-shell/src/main/js/all.js
```diff
@@ -4,5 +4,7 @@ const ShellEvaluator = require('../../../../shell-evaluator/').default;
 /** temporal object that is used to access symbols from closures generated by browserify */
 _global = {
     ShellInternalState: ShellApi.ShellInternalState,
-    ShellEvaluator: ShellEvaluator
+    ShellEvaluator: ShellEvaluator,
+    toShellResult: ShellApi.toShellResult,
```
aherlihy, line 8: Nice!

packages/shell-api/src/enums.ts
```diff
@@ -26,8 +26,8 @@ export enum DBQueryOption {
 }
 
 export const shellApiType = Symbol.for('@@mongosh.shellApiType');
-// TODO: Would require changes to java-shell
-// export const asPrintable = Symbol('asPrintable');
-export const asShellResult = Symbol.for('@@mongosh.asShellResult');
+// TODO: This requires changes to java-shell.
+export const asPrintable = Symbol.for('@@mongosh.asPrintable');
```
aherlihy, line 30: I think you can get rid of the TODO here since it's been updated

packages/shell-api/src/result.ts
```diff
@@ -102,32 +97,10 @@ export class DeleteResult extends ShellApiClass {
...
-    Object.defineProperty(this, asShellResult, {
-      value: () => {
-        return {
-          type: 'CursorIterationResult',
-          value: this._asPrintable()
-        };
-      },
-      enumerable: false
-    });
+    addHiddenDataProperty(this, shellApiType, 'CursorIterationResult');
```
aherlihy, line 104: Much nicer!

aherlihy (approved):
Looks great! I always thought this was a bit clunky and it's much better now.

packages/shell-api/src/enums.ts
addaleax: Ah right, thanks for noticing – done! :)

packages/java-shell/src/main/kotlin/com/mongodb/mongosh/MongoShellContext.kt
```diff
@@ -161,13 +165,31 @@ internal class MongoShellContext(client: MongoClient) {
...
+    fun extract(printable: Value?, rawValue: Value? = null): MongoShellResult<*> {
+        var v: Value
+        var type: String? = null
+        if (printable == null) {
+            type = getShellApiType(rawValue as Value)
+            val shellResult = toShellResult(rawValue as Value)
+            v = shellResult["printable"]!!
+        } else {
+            v = printable
+        }
```
kornilova203, line 186: We can use `val` here
```suggestion
        val v: Value
        val type: String?
        if (printable == null) {
            type = getShellApiType(rawValue as Value)
            v = toShellResult(rawValue)["printable"]!!
        } else {
            type = null
            v = printable
        }
```

kornilova203 (changes requested):
Overall looks good
I'm happy that shell implementations now have more control over results handling

addaleax:
@kornilova203 Thanks for the review, made the changes you suggested :)

PR #377, opened 2020-10-09
MONGOSH-383 - e2e repl tests

aherlihy:


packages/shell-api/src/replica-set.spec.ts
```diff
@@ -13,6 +13,7 @@ import {
   ALL_TOPOLOGIES,
   asShellResult
 } from './enums';
+import { CliServiceProvider } from '../../service-provider-server';
```
addaleax, line 16: Is there a reason not to use 

```suggestion
import { CliServiceProvider } from '@monogsh/service-provider-server';
```

like we usually do? If it’s about the fact that it’s not in the `package.json` here, we could list it under `devDependencies`, right?

packages/shell-api/src/replica-set.spec.ts
```diff
@@ -483,4 +484,157 @@ describe('ReplicaSet', () => {
...
+            if ((await rs.isMaster()).ismaster) {
+              expect((await rs.conf()).members.length).to.equal(3);
+              done();
+            } else {
+              ensureMaster(done, timeout * 2);
+            }
+          }, timeout);
+        }
+      });
+    };
```
addaleax, line 522: This might honestly be a bit easier to understand as an `async` function… something like this?

```js
    const ensureMaster = async(timeout): void => {
      while (!(await rs.isMaster()).ismaster) {
        if (timeout > 8000) 
          return expect.fail(`Waited for ${host}:${port0} to become master, never happened`);
        await delay(timeout);  // delay = util.promisify(setTimeout)
        timeout *= 2; // try again but wait double
      }
      expect((await rs.conf()).members.length).to.equal(3);
    };
```

addaleax (approved):

packages/shell-api/src/replica-set.spec.ts
aherlihy: We had issues with having circular dependency problems (even with devDependency) since the `service-provider-server package` requires the `shell-api` package. We do the same thing in `integration.spec.ts`, since the  only place we use `CliServiceProvider` in this package is in tests.

packages/shell-api/src/replica-set.spec.ts
aherlihy: Will do!

packages/shell-api/src/replica-set.spec.ts
```diff
@@ -585,4 +587,151 @@ describe('ReplicaSet', () => {
       });
     });
   });
+
+  xdescribe('integration', () => {
```
mcasimir, line 591: Doesn't `eslint` complain about this?

Should we rather use environment variables to control this? `process.env.MONGOSH_TEST_RS_HOST1` ...

aherlihy: eslint does complain, but its a warning not a error. We can do the env variable and everything but it seems overkill if we're just going to fix the problem with CI soon enough.

PR #378, opened 2020-10-09
MONGOSH-237 - Allow custom print/printjson for individual platforms

addaleax:
Fwiw, alternatives that I’ve considered are a) passing this over the `messageBus` instead or b) changing the return value of `ShellEvaluator.customEval()` (to e.g. an async iterable instead of a plain promise), but in the end this seemed like the best choice to me because this might be helpful if we ever need another way to customize shell evaluation.

### DIFF
--- a/packages/browser-repl/package.json
+++ b/packages/browser-repl/package.json
@@ -42,7 +42,7 @@
     "@mongosh/history": "^0.4.2",
     "@mongosh/i18n": "^0.4.2",
     "@mongosh/service-provider-core": "^0.4.2",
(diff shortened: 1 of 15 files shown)


mcasimir:
If i understood how it works, i like it a lot. Is the idea for the platform to pass down an object implementing the interface of an `EvaluationListener` right? Just making sure :).

mcasimir (approved):
👍

addaleax:
> If i understood how it works, i like it a lot. Is the idea for the platform to pass down an object implementing the interface of an `EvaluationListener` right? Just making sure :).

Yes, exactly :)

PR #381, opened 2020-10-14
WIP - Add passwordPrompt() support for the CLI

addaleax:
This is missing support in the async rewriter
for auto-awaiting `passwordPrompt()`.

Partial implementation of:

- MONGOSH-314
- MONGOSH-315

### DIFF
--- a/packages/shell-api/src/shell-internal-state.ts
+++ b/packages/shell-api/src/shell-internal-state.ts
@@ -26,6 +26,11 @@ export interface EvaluationListener {
    * Called when print() or printjson() is run from the shell.
    */
   onPrint?: (value: ShellResult[]) => Promise<void> | void;
+
+  /**
+   * Called when e.g. passwordPrompt() is called from the shell.
+   */
+  onPrompt?: (question: string, type: 'password') => Promise<string> | string;
 }
 
 /**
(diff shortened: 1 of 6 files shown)


packages/shell-api/src/shell-internal-state.ts
```diff
@@ -169,6 +174,13 @@ export default class ShellInternalState {
       });
     }
 
+    contextObject.passwordPrompt = async(): Promise<string> => {
```
aherlihy, line 177: Hi! I wasn't sure if you wanted input on this yet since its a WIP, apologies if it's too early!

I would consider adding this method to the top level `ShellApi` class (in shell-api.ts) if possible. It should then get automatically added to ContextObject on line 114/119. That way it would be as simple as using the `returnsPromise` decorator for the async rewriter to know it needs an await inserted before each use. The ShellApi object has access to the internal state object via `this._internalState` so the `evaluationListener` should still be accessible.

aherlihy: If I'm understanding it all right, it should work pretty much the same as the `connect` method

addaleax: Oh nice, thank you, I think that’s exactly what I was looking for!

I think all this still needs are tests then. :slightly_smiling_face:

aherlihy: Awesome, it looks good!

addaleax:
Going to revisit this at some point, but this has a few too many merge conflicts to do that now :)

PR #385, opened 2020-10-19
MONGOSH-269 - Add nyc for test coverage

addaleax:


packages/browser-repl/package.json
```diff
@@ -15,6 +15,7 @@
     "start": "npm run storybook",
     "storybook": "start-storybook -p 6006",
     "test": "karma start ./config/karma.conf.js --colors",
+    "test-ci": "echo Not tested in CI",
```
mcasimir, line 18: wait really? this wasn't tested? I think we should re-add it. Do we need a ticket or would just work?

mcasimir (approved):

package.json
```diff
@@ -17,8 +17,8 @@
     "clean": "lerna clean -y && rm -Rf node_modules",
     "check": "lerna run check --since HEAD --exclude-dependents",
     "check-ci": "lerna run check",
-    "test": "lerna run test",
-    "test-ci": "lerna run test-ci",
+    "test": "rimraf .nyc_output && lerna exec -- nyc --no-clean --cwd ../.. --reporter=none npm run test && npm run report-coverage",
```
mcasimir, line 20: you may want to use some `pretest` or magic package to allow `&&` in windows

packages/cli-repl/test/e2e.spec.ts
```diff
@@ -280,11 +280,20 @@ describe('e2e', function() {
 
     let shell;
     beforeEach(async() => {
-      shell = TestShell.start({ args: [ '--nodb' ] });
+      // We set CLEAR_SIGINT_LISTENERS to remove all `process.on('SIGINT')`
```
mcasimir, line 283: I'd do this in TestShell itself only if the binary was not passed: we run e2e tests on the built binary too which will make the coverage meaningless and removing listeners will possibly taint the tests.  Would be nice to leave everything intact in that case.

mcasimir (changes requested):

package.json
mcasimir, line 20: do we want to have coverage always or as a separate job like `test-cov`? won't it take more to run tests with coverage? or is intentional?

packages/browser-repl/package.json
mcasimir: also great you spotted it :)

addaleax: Err, I can change it and see what happens? (Fwiw, I'm not entirely clear on why test and test-ci are seperate scripts anyway... it doesn't usually seem like a good idea to me to have local tests do something different than CI)

package.json
addaleax: `&&` should work just fine in npm scripts in Windows ... there are some issues with coverage generation that only seem to affect Windows, though, I'm currently looking into that

packages/cli-repl/test/e2e.spec.ts
addaleax: > and removing listeners will possibly taint the tests

Yeah, I've thought about this tainting the test but it just seems highly unlikely that this might ever become an issue. I still like the suggestion of doing this only if we're not running the executable, that seems just fine to me :+1:

package.json
addaleax: It'll take a bit longer, yes... but `npm test` for the full repo already takes at least 10 minutes, so it's not like it's something that one would run on a regular basis during development (I only do it before submitting a PR, and admittedly only sometimes ... CI is there to do that work for us)

packages/cli-repl/test/e2e.spec.ts
addaleax: done :+1:

packages/browser-repl/package.json
addaleax: Opened https://github.com/mongodb-js/mongosh/pull/386 to see if this works

mcasimir: > it doesn't usually seem like a good idea to me to have local tests do something different than CI

I'm not sure it is doing something different. If it is, at least till now is not on purpose. Then again if it will be useful for some reason I don't know, feel free to remove it.

package.json
mcasimir: Oh i see, the coverage runs only on the root package .. didn't notice.

addaleax: Yeah, I was referring to `test` and `test-ci` of the root package :)

mcasimir (approved):

PR #387, opened 2020-10-19
chore: clean up color/tty handling core in cli-repl

addaleax:
Don’t assume that `process.stdout.isTTY` is a good indicator
for coloring strings:

- Not all terminals have color support
- We should honour environment variables like `NO_COLOR`
- Not all formatted strings end up printed to stdout

Also, don’t tell the Node.js repl that we are working with a terminal
if we don’t know so for sure.

### DIFF
--- a/packages/cli-repl/src/arg-parser.ts
+++ b/packages/cli-repl/src/arg-parser.ts
@@ -2,7 +2,7 @@ import { CliOptions } from '@mongosh/service-provider-server';
 import { USAGE } from './constants';
 import i18n from '@mongosh/i18n';
 import minimist from 'minimist';
-import clr from './clr';
+import { colorizeForStderr as clr } from './clr';
 
 /**
  * Unknown translation key.
@@ -73,7 +73,7 @@ const OPTIONS = {
       return true;
     }
     throw new Error(
-      `  ${clr(i18n.__(UNKNOWN), ['red', 'bold'], { colors: true })} ${clr(parameter, 'bold', { colors: true })}
+      `  ${clr(i18n.__(UNKNOWN), ['red', 'bold'])} ${clr(parameter, 'bold')}
       ${USAGE}`
     );
   }
--- a/packages/cli-repl/src/arg-parser.ts
+++ b/packages/cli-repl/src/arg-parser.ts
@@ -73,7 +73,7 @@ const OPTIONS = {
       return true;
     }
     throw new Error(
-      `  ${clr(i18n.__(UNKNOWN), ['red', 'bold'])} ${clr(parameter, 'bold')}
+      `  ${clr(i18n.__(UNKNOWN), ['red', 'bold'], { colors: true })} ${clr(parameter, 'bold', { colors: true })}
       ${USAGE}`
     );
   }
--- a/packages/cli-repl/src/cli-repl.ts
+++ b/packages/cli-repl/src/cli-repl.ts
@@ -97,7 +97,7 @@ class CliRepl {
    */
   async connect(driverUri: string, driverOptions: NodeOptions): Promise<any> {
     if (!this.options.nodb) {
-      console.log(i18n.__(CONNECTING), '    ', clr(retractPassword(driverUri), ['bold', 'green']));
+      console.log(i18n.__(CONNECTING), '    ', this.clr(retractPassword(driverUri), ['bold', 'green']));
     }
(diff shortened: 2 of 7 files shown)


packages/cli-repl/src/cli-repl.ts
```diff
@@ -475,6 +477,25 @@ class CliRepl {
...
+  }
+
+  clr(text: string, style: string|string[]): string {
+    return clr(text, style, this.getFormatOptions());
+  }
+
+  getFormatOptions(): { colors: boolean } {
+    return {
+      colors: this.repl ? this.repl.useColors :
+        process.stdout.isTTY && process.stdout.getColorDepth() > 1
```
mcasimir, line 496: nice stuff!

packages/cli-repl/src/arg-parser.ts
```diff
@@ -73,7 +73,7 @@ const OPTIONS = {
       return true;
     }
     throw new Error(
-      `  ${clr(i18n.__(UNKNOWN), ['red', 'bold'])} ${clr(parameter, 'bold')}
+      `  ${clr(i18n.__(UNKNOWN), ['red', 'bold'], { colors: true })} ${clr(parameter, 'bold', { colors: true })}
```
mcasimir, line 76: why here is ok to set `colors= true`?

packages/cli-repl/src/cli-repl.ts
```diff
@@ -116,9 +116,9 @@ class CliRepl {
       prompt: '> ',
       writer: this.writer,
       completer: completer.bind(null, version),
-      terminal: true,
       breakEvalOnSigint: true,
       preview: false,
+      terminal: process.env.MONGOSH_FORCE_TERMINAL ? true : undefined
```
mcasimir, line 121: nice and simple!

packages/cli-repl/src/cli-repl.ts
```diff
@@ -128,12 +128,14 @@ class CliRepl {
       this.lineByLineInput.nextLine();
     };
 
-    const originalEditorAction = this.repl.commands.editor.action.bind(this.repl);
+    if (this.repl.commands.editor) {
```
mcasimir, line 131: 👍

packages/cli-repl/src/clr.ts
```diff
@@ -1,5 +1,8 @@
 import ansi from 'ansi-escape-sequences';
 
-export default function clr(text: string, style: any): string {
-  return process.stdout.isTTY ? ansi.format(text, style) : text;
+export default function clr(text: string, style: string|string[], options: { colors: boolean }): string {
```
mcasimir, line 3: Wouldn't be a bit confusing without context for a `clr` function (that i believe stands for `color`) to take a `color` `true | false` argument? 

Should we rename something, or do something to make this clear?

mcasimir (approved):
Nice!

packages/cli-repl/src/arg-parser.ts
addaleax: It’s not really, but we’re also not explicitly printing anything here … if I understand correctly, this gets passed down to the Node.js error handler itself

So a “clean” solution would be e.g. creating an error object here that has a colored and a non-colored message, and then handling that at the `parseCliArgs` call side in `src/run.ts` and deciding what to print from there … but I’m not sure if that’s worth the extra complexity? Especially that we’re targeting scripted usage with the color-on/color-off handling, and wouldn’t typically contain argument name misspellings (or not for a long time :))

packages/cli-repl/src/clr.ts
addaleax: I would absolutely be in favor of renaming `clr` to something more expressive :) I’ll try to think of something

packages/cli-repl/src/arg-parser.ts
addaleax: Fwiw, I’ve made this use the same thing as above, i.e. basically assume that this ends up on stderr (which I think is accurate)

packages/cli-repl/src/clr.ts
addaleax: I’ve renamed the functions themselves but not the places where they were imported, still not ideal but at least not that strongly abbreviated

lrlna (approved):
This looks good to me! Thanks for this clean up ^__^

PR #388, opened 2020-10-20
MONGOSH-248 - Run java-shell tests as part of the test suite

addaleax:


lrlna (approved):
This looks good to me @addaleax! Thanks for tackling this \o/

packages/java-shell/package.json
```diff
@@ -12,7 +12,8 @@
   },
   "scripts": {
     "browserify-shell-api": "browserify src/main/js/all.js -o src/main/resources/js/all-standalone.js -dv",
-    "test-ci": "npm run test"
+    "test": "mocha -r \"../../scripts/import-expansions.js\" --timeout 60000 --colors -r ts-node/register \"./src/test/js/run-tests.ts\"",
```
mcasimir, line 15: What do i need locally to run this? Should we add some docs?

addaleax: It’s the same as the other packages – it’s part of `npm test` on the root module. Whether that’s good or not is another question, it requires Java, but then again `npm test` already requires stuff like headless chrome, so … :woman_shrugging:

PR #393, opened 2020-10-26
PoC: link kerberos addon into binary

addaleax:
Mostly opening this to see whether this a) works in CI and
b) to see whether we can upstream the changes to the build
config of the kerberos module early.

Refs: https://github.com/mongodb-js/kerberos/commit/0040c1b25af29307cce64612221b9fa5085153eb

### DIFF
--- a/packages/build/src/signable-compiler.ts
+++ b/packages/build/src/signable-compiler.ts
@@ -1,5 +1,8 @@
 import Platform from './platform';
 import os from 'os';
+import Module from 'module';
+import pkgUp from 'pkg-up';
+import path from 'path';
 
 /**
  * A compiler that can produce an executable that is actually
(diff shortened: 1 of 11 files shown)


packages/build/src/signable-compiler.ts
```diff
@@ -38,7 +47,10 @@ class SignableCompiler {
         // Custom env vars for sccache:
         AWS_ACCESS_KEY_ID: process.env.DEVTOOLS_CI_AWS_KEY,
         AWS_SECRET_ACCESS_KEY: process.env.DEVTOOLS_CI_AWS_SECRET
-      }
+      },
+      addons: [
```
mcasimir, line 51: wow! this is all we need to add native modules? 😮

addaleax: Yes – the trickier bit here is modifying the kerberos build config so that it supports being used as a static library in the first place :)

packages/build/src/signable-compiler.ts
```diff
@@ -22,6 +25,12 @@ class SignableCompiler {
    * @param {Function} exec - The boxednode compile function.
    */
   async compile(exec: Function) {
+    const cliReplRequire = Module.createRequire(path.resolve(__dirname, '..', '..', 'cli-repl', 'src'));
```
mcasimir, line 28: Would it be possible to put `Module.createRequire` + `path.dirname` `pkgUp` in some function like `findModulePath`? Just so it is immediate what it does

PR #395, opened 2020-11-06
fix: correctly format several values in browser-repl

addaleax:
Update the formatting for several values in the browser REPL.
Since fixing this for Date objects consistently enabled doing
straightforward fixes for a few other issues, this addresses:

- COMPASS-4440 (Consistently formatting `Date` objects)
- COMPASS-4441 (Ignoring `undefined` results)
- MONGOSH-410 (`sh.status()` formatting)
- COMPASS-4465 (`ObjectId` formatting in nested objects)

### DIFF
--- a/packages/browser-repl/src/components/shell-output-line.tsx
+++ b/packages/browser-repl/src/components/shell-output-line.tsx
@@ -13,9 +13,9 @@ import { ShowCollectionsOutput } from './types/show-collections-output';
 import { CursorOutput } from './types/cursor-output';
 import { CursorIterationResultOutput } from './types/cursor-iteration-result-output';
 import { ObjectOutput } from './types/object-output';
+import { StatsResultOutput } from './types/stats-result-output';
 import { SimpleTypeOutput } from './types/simple-type-output';
 import { ErrorOutput } from './types/error-output';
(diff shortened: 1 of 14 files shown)


packages/browser-repl/src/components/shell-output-line.tsx
```diff
@@ -13,9 +13,9 @@ import { ShowCollectionsOutput } from './types/show-collections-output';
 import { CursorOutput } from './types/cursor-output';
 import { CursorIterationResultOutput } from './types/cursor-iteration-result-output';
 import { ObjectOutput } from './types/object-output';
+import { StatsResultOutput } from './types/stats-result-output';
```
mcasimir, line 16: what should we do in future to avoid forgetting to implement these here?

mcasimir (approved):

packages/browser-repl/src/components/shell-output-line.tsx
addaleax: Well … we didn’t forget to implement this, we just didn’t check properly whether it works as intended in the browser repl, I think

I’m not really worried about this being a common occurrence, tbh

PR #396, opened 2020-11-11
[WIP][java-shell] Don't fetch Java cursor in toShellResult method

kornilova203:
Java shell users want to have more control over cursor. E.g. execute `db.coll.find()`, then in Java code set batch size to the cursor and fetch it in Java code.

[toShellResult](https://github.com/mongodb-js/mongosh/blob/master/packages/shell-api/src/decorators.ts#L75) method fetch cursor itself. When `rawValue` cursor is returned by Java shell, it's already exhausted.

I added test case `testSetBatchSize` that throws [`query already executed`](https://github.com/mongodb-js/mongosh/blob/b709e3e4b4a21871bd3356c1f476d8711b5bc439/packages/java-shell/src/main/kotlin/com/mongodb/mongosh/service/Cursor.kt#L242) exception when `toShellResult` fetches cursor.

I think there should be a way to override `asPrintable` method for cursor in Java shell.


kornilova203:
New PR #428

PR #401, opened 2020-11-11
MONGOSH-400 - Use mlaunch for spawning test mongod processes

addaleax:
This is meant to finish up MONGOSH-400, i.e. investigating whether we can use mlaunch, rather than providing an actual working implementation, although this already works quite well imo. Evergreen Windows hosts apparently (?) do not have Python 3.x installed, that’s something we might need to solve before this could theoretically be merged. All in all, I’m pretty happy with how mlaunch itself works and the capabilities we get from it, in particular, it does enable running the replica set/sharding tests :slightly_smiling_face: 

---

Switch to using `mlaunch` for spwaning mongod processes
and enable the replica set and sharding tests.

If merged in its current form, this would resolve

- MONGOSH-400 (Decide how to spawn mongod processes)
- MONGOSH-411 (Switch mongosh tests to use mlaunch)
- MONGOSH-420 (Enable sharding tests)
- MONGOSH-421 (Enable replica set tests)


packages/shell-api/src/replica-set.spec.ts
```diff
@@ -588,19 +589,17 @@ describe('ReplicaSet', () => {
...
-      ]
-    };
+
+    const testServers = startTestCluster(
+      ['--single', '--replSet', replId],
+      ['--single', '--replSet', replId],
+      ['--single', '--replSet', replId],
+      ['--single', '--replSet', replId]
+    );
+    let cfg: {_id: string, members: {_id: number, host: string, priority: number}[]};
```
aherlihy, line 601: This is so cool I haven't seen this syntax before!

packages/shell-api/src/replica-set.spec.ts
```diff
@@ -609,34 +608,43 @@ describe('ReplicaSet', () => {
...
+
+      serviceProvider = await CliServiceProvider.connect(srv0.connectionString());
       internalState = new ShellInternalState(serviceProvider);
       mongo = internalState.currentDb.getMongo();
       rs = new ReplicaSet(mongo);
 
       // check replset uninitialized
       try {
         await rs.status();
+        expect.fail();
```
aherlihy, line 641: So mlaunch supports starting up uninitialized replsets? If it's not possible then we can potentially just skip the init test for now

aherlihy (commented):
This looks great! Nice job

packages/shell-api/src/replica-set.spec.ts
addaleax: @aherlihy Yes, the tests here run & pass :) It doesn’t have a designated feature for starting uninitialized replsets, but you can pass pretty much any mongod server option through over it

testing/integration-testing-hooks.ts
```diff
@@ -1,92 +1,370 @@
...
+      }
+    }
+  }
+  if (!python) {
+    throw new Error('Could not find Python 3.x installation, install mlaunch manually');
+  }
+
+  // Install mlaunch, preferably locally and otherwise attempt to do so globally.
+  try {
+    await execFile('pip3', ['install', '--target', tmpdir, 'mtools[mlaunch]']);
```
rose-m, line 78: I'm afraid just installing that is not enough. I just tried it on my macOS with basically a really plain setup (no PIP packages or anything so far). When trying to run the tests I was presented with `ImportErrors` for the following packages:
- `psutil`
- `pymongo`

After manually installing these packages using PIP it worked like a charm 😄 So we should add them to the list of installed packages here.

rose-m (commented):
Looks amazing and seems to make those tests super easy! I tried running it locally, however, and found we probably have to add 2 other PIP packages (see my comment) to make it work out of the box 👍

rose-m:
Ah and there's one failing test:

```
@mongosh/shell-api:   1) Shard
@mongosh/shell-api:        integration
@mongosh/shell-api:          sharding info
@mongosh/shell-api:            returns the status:
@mongosh/shell-api:       AssertionError: expected [ Array(6) ] to be a superset of [ Array(6) ]
@mongosh/shell-api:       + expected - actual
@mongosh/shell-api:        [
@mongosh/shell-api:          "shardingVersion"
@mongosh/shell-api:          "shards"
@mongosh/shell-api:       -  "most recently active mongoses"
@mongosh/shell-api:       +  "active mongoses"
@mongosh/shell-api:          "autosplit"
@mongosh/shell-api:          "balancer"
@mongosh/shell-api:          "databases"
@mongosh/shell-api:        ]
@mongosh/shell-api:
@mongosh/shell-api:       at Context.<anonymous> (src/shard.spec.ts:1140:54)
```

testing/integration-testing-hooks.ts
addaleax: @rose-m This is very odd, pip should install dependencies automatically … I’ll try to look into that

addaleax: @rose-m I could not really reproduce this, cleared out the installed python modules on my mac and ran the tests succesfully … so I’m not sure how to progress here :confused: We can add them, but it’s odd that that would be necessary. Can you try to reproduce the problem, and maybe add a `console.log()` here that prints whatever `pip3` logged (just wrapping this line in it should be enough)?

rose-m: Yup, I'll give it another try!

rose-m: Okay I cannot reproduce anymore... it's clear that PIP automatically downloads and installs the dependencies as you said... Tried cleaning everything up and removing caches - works fine... So I'd say we can mark this comment as resolved then. Sorry 🤷

addaleax:
Ready for review, assuming that I didn’t manage to regress into breaking CI :)

rose-m (approved):
Tried it again - and worked like a charm on my machine 👍 Code (as far as I know the context and understood it) looks good to me 😄

aherlihy (approved):
Works for me! Nice job, this is a huge improvement :) I wonder if it' would be possible to silence the mlaunch 'Running command' printouts for every test? Unless you think it's useful to see. Feel free to merge right away.

testing/integration-testing-hooks.ts
```diff
@@ -1,92 +1,398 @@
...
+
+  connectionString(): string {
+    return this._connectionString;
+  }
+
+  host(): string {
+    return new URL(this.connectionString()).hostname;
+  }
+
+  port(): string {
```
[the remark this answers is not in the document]
addaleax: I'm not sure I understand -- the default port for `mongodb://` URLs is `27017`, so we should return that if there's no port in the URL, right?

addaleax:
> I wonder if it' would be possible to silence the mlaunch 'Running command' printouts for every test? Unless you think it's useful to see. Feel free to merge right away.

Hm yeah ... it's been helpful for debugging this PR itself, but I can remove them/comment them out until we run into problems with these :+1:

testing/integration-testing-hooks.ts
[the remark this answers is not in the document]
aherlihy: Yes - that's right. I was thinking maybe we'd want to prevent tests from starting a server on 27017 (since most people have something else running on 27017 so if you start your test processes it may cause collisions), but it looks like none of the tests are doing that anyway and it would be incorrect/unexpected to always avoid 27017. So ignore my comment!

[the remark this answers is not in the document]
addaleax: The general idea here is that you can still specify `MONGOSH_TEST_SERVER_URL` for debugging if that helps to point to a specific server, and that in that case, if that URL does not have a port, we should use `27017` :)

PR #405, opened 2020-11-13
fix(shell-api): do not include CursorIterationResult in output MONGOSH-430

addaleax:


mcasimir (approved):

packages/shell-api/src/decorators.ts
```diff
@@ -87,8 +87,14 @@ export async function toShellResult(rawValue: any): Promise<ShellResult> {
     return toShellResult(await rawValue);
   }
 
-  const printable =
-    typeof rawValue[asPrintable] === 'function' ? await rawValue[asPrintable]() : rawValue;
+  // Convert objects to their printed representation, if necessary repeatedly
+  // until we've reached a plain object.
+  let printable = rawValue;
```
aherlihy, line 92: I'm confused the case where calling asPrintable would return an API object, can you give me an example?

addaleax: We’re currently returning a `CursorIterationResult` from `Cursor.prototype[asPrintable]` (and the same for `AggregationCursor`) here

aherlihy: If a cursor asPrintable returns a CursorIterationObject, shouldn't it just call asPrintable on that object within it's own asPrintable method? Or maybe I'm just confused

addaleax: Yeah, we could do that here as well… no strong preferences, but it sounds like you have one? ;)

aherlihy: haha I don't feel strongly! Sorry for any terse wording :) I think it's maybe a little nicer if we call it twice in cursor's asPrintable since it appears that's the only time it would need to be called, but I certainly don't mind the way it is now.

addaleax: @aherlihy Updates with your suggestion :slightly_smiling_face:

PR #409, opened 2020-11-16
chore(cli-repl): split out async eval bits of CliRepl MONGOSH-436

addaleax:
Split out the parts of the `CliRepl` class that are specific to async
evaluation and test them entirely.

### DIFF
--- a/packages/cli-repl/src/async-repl.ts
+++ b/packages/cli-repl/src/async-repl.ts
@@ -1,0 +1,131 @@
+import type { REPLServer, ReplOptions } from 'repl';
+import { Recoverable, start as originalStart } from 'repl';
+import isRecoverableError from 'is-recoverable-error';
+import { promisify } from 'util';
+import { Domain } from 'domain';
+
+// Utility, inverse of Readonly<T>
+type Mutable<T> = {
+  -readonly[P in keyof T]: T[P]
+};
+
+export type OriginalEvalFunction = (input: string, context: any, filename: string) => Promise<any>;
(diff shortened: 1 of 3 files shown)


packages/cli-repl/src/async-repl.ts
```diff
@@ -0,0 +1,131 @@
...
+    // It's not clear why this is done this way in Node.js, however,
+    // removing the linked code does lead to failures in the Node.js test
+    // suite, so somebody sufficiently motivated could probably find out.
+    // For now, this is a hack and probably not considered officially
+    // supported, but it works.
+    // We *may* want to consider not relying on the built-in eval function
+    // at all at some point.
+    Domain.prototype.emit = function(ev: string, ...eventArgs: any[]): boolean {
+      if (ev === 'error') {
+        this.exit();
```
addaleax, line 116: This line was missing and causing Node.js core deprecation warnings to be emitted that some of us noticed, yay for testing to catch this :)

aherlihy: Nice!!

aherlihy (approved):
Awesome!

aherlihy:
Do you think the windows tests are failing due to a race condition or is there something widows-specific about recoverable syntax errors? I restarted the windows test just in case, since I can't really see from looking at it why it would be failing now

addaleax:
> Do you think the windows tests are failing due to a race condition or is there something widows-specific about recoverable syntax errors? I restarted the windows test just in case, since I can't really see from looking at it why it would be failing now

… I’m not sure, but I definitely messed up how to skip the SIGINT tests (as I always do), this might have just been some cross-test interation. Maybe doing that properly is enough to fix this :)

PR #414, opened 2020-11-17
chore(cli-repl): split out config/log management from CliRepl MONGOSH-439

addaleax:
Not super happy with the class/file naming and happy to take suggestions :)


packages/cli-repl/src/config-directory.ts
```diff
@@ -0,0 +1,76 @@
+import { promises as fs } from 'fs';
+import path from 'path';
+import { EventEmitter } from 'events';
+
+export class ConfigAndLogDirectoryManager<Config> extends EventEmitter {
```
mcasimir, line 5: What about decoupling it even more, moving the accent to the fact that there is a Config and some Logs: 

``` js
class ShellHomeDir {
    ensure(...) {} ...
    path() {}
}
```

``` js
class ConfigManager<Config> {
     constructor(shellHomeDir) {}
 }
```

``` js
class LogStuff {
     constructor(shellHomeDir) {}
 }
```

mcasimir, line 5: Nice idea to have generics here

packages/cli-repl/src/cli-repl.ts
```diff
@@ -353,7 +281,7 @@ class CliRepl {
     console.log(`Using MongoDB:      ${this.internalState.connectionInfo.buildInfo.version}`);
     console.log(`${this.clr('Using Mongosh Beta', ['bold', 'yellow'])}: ${version}`);
     console.log(`${MONGOSH_WIKI}`);
-    if (!this.disableGreetingMessage) console.log(TELEMETRY);
+    if (!this.config.disableGreetingMessage) console.log(TELEMETRY);
```
mcasimir, line 284: Is `TELEMETRY` a some form of `TELEMETRY_GREETING_MESSAGE`?

addaleax: It’s 

> To help improve our products, anonymous usage data is collected and sent to MongoDB periodically (https://www.mongodb.com/legal/privacy-policy). You can opt-out by running the `disableTelemetry()` command.

so, yes. :+1:

packages/cli-repl/src/config-directory.ts
addaleax: I mean, the log stuff is basically just a single call to `path` :woman_shrugging: In that case it probably makes sense to drop the `Log` part from the name, even though most files in that path will be log files

packages/cli-repl/src/cli-repl.ts
```diff
@@ -207,81 +218,11 @@ class CliRepl {
...
-   * @returns {string} Status of telemetry logging: disabled/enabled
-   */
-  toggleTelemetry(enabled: boolean): string {
-    this.enableTelemetry = enabled;
-    this.disableGreetingMessage = true;
-
-    this.bus.emit('mongosh:update-user', this.userId, this.enableTelemetry);
-    const configPath = path.join(this.mongoshDir, 'config');
-    this.writeConfigFileSync(configPath);
+  async toggleTelemetry(enabled: boolean): Promise<string> {
```
mcasimir, line 221: I see why this may be convenient if it stays here, but it kinda seems this may even go somewhere else.

Would it be a bad idea to have a config that is just this interface:

```
interface Config {
    async isTelemetryEnabled(): boolean;
    async setTelemetryEnabled(boolean);
    async getHistoryPath();
    ...
}
```

Essentially abstracting away anything file system related or even the internal structure of the config to some more dedicated place?

addaleax: So … this particular function is being called from the shell evaluator, so I think we’re not really going to get around having a function that updates the config and triggers the store-to-filesystem operation in some way

But yeah, this is a good point, and I’ll think about it and either address it here or as part of MONGOSH-441 if I come up with something good

packages/cli-repl/src/config-directory.ts
mcasimir: Well for the purpose of making it more testable i think we nailed it by doing what you've done. 

From the perspective of isolating concerns it still feels a bit mixed. I mean there are still some fragments in cli repl, something here is both about the config and logs as well. 

Don't know, I have this feeling we could go one step further and completely isolate CliRepl from any knowledge about directories, how things are stored etc. 

Leaving the CliRepl code with only knowledge that's needed for it to operate: like the CliRepl only needs to know if the telemetry is enabled to show the greeting message, is not necessary to also know how that's stored.

packages/cli-repl/src/cli-repl.ts
mcasimir: I see so this is a different ticket. Got it.

`toggleTelemetry` probably could be called in the same way I guess in the shell evaluator. Anyway just leaving some random comment, all looks great!

packages/shell-evaluator/src/shell-evaluator.ts
```diff
@@ -7,7 +7,7 @@ import {
 } from '@mongosh/shell-api';
 
 interface Container {
```
mcasimir, line 9: Just noticed `Container` here sounds super generic, this one maybe deserves a more clarifying name or at least a comment. That one could actually even be `Config`. Would be appropriate as a contract to redefine it here as the subset of the config that needs to be accessed from ShellEvaluator.

But yeah, up to you if you want to do anything about it.

mcasimir (approved):

packages/cli-repl/src/config-directory.ts
addaleax: @mcasimir Yeah, I just changed this to basically adopt your suggestion here :slightly_smiling_face: It makes sense to separate the directory management from config management, I just left out a specific class for logging facilities because that would basically be empty for now

packages/shell-evaluator/src/shell-evaluator.ts
addaleax: I thought about it, and I think it makes sense to just turn `enableTelemetry()` and `disableTelemetry()` into actual functions that then call a method on the `EvaluationListener`, just like we do for `console.log()` etc.

PR #418, opened 2020-11-19
chore(cli-repl): split CliRepl and test it MONGOSH-441

addaleax:
Split the CliRepl class into one part that is not concerned with
external I/O and one part that is. Add unit tests to cover them both.

### DIFF
--- a/packages/cli-repl/src/cli-repl.spec.ts
+++ b/packages/cli-repl/src/cli-repl.spec.ts
@@ -96,7 +96,11 @@ describe('CliRepl', () => {
         await onerror;
       });
 
-      it('emits error for inaccessible home directory', async() => {
+      it('emits error for inaccessible home directory', async function() {
+        if (process.platform === 'win32') {
+          this.skip(); // TODO: Figure out why this doesn't work on Windows.
+          return;
+        }
         cliReplOptions.shellHomePath = '/nonexistent/inaccesible';
         cliRepl = new CliRepl(cliReplOptions);
         const onerror = once(cliRepl.bus, 'mongosh:error');
--- a/packages/cli-repl/src/cli-repl.spec.ts
+++ b/packages/cli-repl/src/cli-repl.spec.ts
@@ -1,9 +1,189 @@
-import { expect } from 'chai';
+import { PassThrough, Duplex } from 'stream';
+import { promisify } from 'util';
+import path from 'path';
+import { promises as fs } from 'fs';
+import { once } from 'events';
+import rimraf from 'rimraf';
+import CliRepl, { CliReplOptions } from './cli-repl';
+import { startTestServer } from '../../../testing/integration-testing-hooks';
+import { expect, useTmpdir, waitEval } from '../test/repl-helpers';
 
-describe('cli-repl', () => {
-  context('placeholder', () => {
-    it('passes', () => {
-      expect(true).to.be.true;
+describe('CliRepl', () => {
+  let cliReplOptions: CliReplOptions;
+  let cliRepl: CliRepl;
+  let input: Duplex;
+  let outputStream: Duplex;
+  let output = '';
+  let exitCode: null|number = null;
+  const tmpdir = useTmpdir();
+
+  beforeEach(async() => {
+    input = new PassThrough();
+    outputStream = new PassThrough();
+    output = '';
+    outputStream.setEncoding('utf8').on('data', (chunk) => { output += chunk; });
+
+    cliReplOptions = {
+      shellCliOptions: {},
+      input: input,
+      output: outputStream,
+      shellHomePath: tmpdir.path,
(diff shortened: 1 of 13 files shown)


packages/cli-repl/src/cli-repl.spec.ts
```diff
@@ -1,9 +1,189 @@
...
+      await waitEval(cliRepl.bus);
+      expect(output).to.include('acknowledged: true');
+
+      output = '';
+      input.write('db.cats.find()\n');
+      await waitEval(cliRepl.bus);
+      expect(output).to.include('pia');
+
+      input.write('.exit\n');
+    });
```
addaleax, line 149: This can do basically anything our e2e tests can also do, but without the overhead of spawning a new process or spinning a timer to wait for output from the child process. For some e2e tests, we could consider migrating them over to a pattern like this, if we want to.

packages/cli-repl/src/cli-repl.ts
```diff
@@ -1,74 +1,57 @@
...
-import { ShellInternalState } from '@mongosh/shell-api';
-import { ShellEvaluator, ShellResult } from '@mongosh/shell-evaluator';
-import formatOutput, { formatError } from './format-output';
-import { LineByLineInput } from './line-by-line-input';
-import { TELEMETRY_GREETING_MESSAGE, MONGOSH_WIKI } from './constants';
 import { ConfigManager, ShellHomeDirectory } from './config-directory';
 import { MongoshInternalError, MongoshWarning } from '@mongosh/errors';
-import { changeHistory, retractPassword } from '@mongosh/history';
-import completer from '@mongosh/autocomplete';
+import { retractPassword } from '@mongosh/history';
```
mcasimir, line 4: isn't this supposed to be `redactPassword`? I think this may be some kind of typo.

addaleax: Yeah, uh … this seems like a typo :smile:

https://github.com/mongodb-js/mongosh/blob/4de0838bcd27207a8a149ace6811d5cb83c9a6e4/packages/history/src/index.ts#L2-L3

packages/cli-repl/src/cli-repl.ts
```diff
@@ -295,56 +171,33 @@ class CliRepl {
...
-  formatOutput(value: any): string {
-    return formatOutput(value, this.getFormatOptions());
-  }
+  private _fatalError(error: any): never {
+    this.bus.emit('mongosh:error', error);
 
-  formatError(value: any): string {
-    return formatError(value, this.getFormatOptions());
+    this.output.write(this.mongoshRepl.formatError(error) + '\n');
+    this.bus.emit('mongosh:exit', 1);
```
mcasimir, line 194: Is there a reason we are doing this with the bus and not in some more explicit way, like:

``` js
this.exit(1);
```

Maybe even injecting `process` if that was the issue. Even if not I'd wrap `this.bus.emit('mongosh:exit')` in `exit(statusCode)` regardless.

addaleax: The reason that this uses the bus is that it keeps us from having to add yet another property to the `MongoshRepl` and that it’s a bit easier to test this way, but if we want to, we can do that, sure.



> Even if not I'd wrap `this.bus.emit('mongosh:exit')` in `exit(statusCode)` regardless.

Sure, done! :)

packages/cli-repl/src/run.ts
```diff
@@ -1,36 +1,46 @@
...
+      const repl = new CliRepl({
+        shellCliOptions: options,
+        input: process.stdin,
+        output: process.stdout,
+        onExit: process.exit,
+        shellHomePath: path.join(os.homedir(), '.mongodb/mongosh/')
+      });
+      await repl.start(driverUri, { appname, ...driverOptions });
+    }
+  } catch (e) {
```
mcasimir, line 42: Don't we have a `process.exit(1)` here?

packages/cli-repl/src/run.ts
```diff
@@ -1,36 +1,46 @@
...
+    // eslint-disable-next-line no-console
+    console.log(e.message);
```
mcasimir, line 44: Should we direct this to `stderr` or it would make no sense? 🤔

packages/cli-repl/src/run.ts
addaleax: Yup :+1:

packages/cli-repl/src/run.ts
addaleax: Ah yeah, that definitely is the correct thing to do here :+1:

mcasimir (approved):

mcasimir (commented):
This is great, helps testing lots of stuff that were though to deal with otherwise.

PR #419, opened 2020-11-19
chore(test): simplify mlaunch spawning a bit

addaleax:
Instead of having an odd double-async structure where you’d first
await the test server, then await the results of async methods on it,
just have a single class that provides async methods.


gribnoysup (approved):

aherlihy (commented):
Great!

PR #420, opened 2020-11-20
chore(test): add vscode extension integration tests MONGOSH-408

addaleax:


scripts/docker/ubuntu18.04-xvfb.Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+RUN apt-get -y -qq install git curl apt-transport-https ca-certificates apt-utils software-properties-common
+
+# Add Node.js
+RUN curl -sSL https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add -
+RUN echo "deb https://deb.nodesource.com/node_12.x bionic main" | tee /etc/apt/sources.list.d/nodesource.list
+RUN echo "deb-src https://deb.nodesource.com/node_12.x bionic main" | tee -a /etc/apt/sources.list.d/nodesource.list
+
+# Install Node.js and vscode dependencies
+RUN apt-get update
+RUN apt-get -y -qq install nodejs libnss3 gnupg libxkbfile1 libsecret-1-0 libgtk-3-0 libxss1 libgbm1 libasound2 xvfb
```
addaleax, line 14: We’re using Docker because otherwise these would not be available in evergreen :woman_shrugging:

Anemy:
Cool! I dig this - going to have to make sure we don't have any flaky tests in vscode hah. Should we also offer this as an npm script? Looks like it isn't platform dependent because we're using docker. Might be useful for debugging if this test ever does fail.
In the case where we're updating the APIs that vscode talks to, we do expect these tests to fail. Do we just ignore them in the meantime? I think we've chatted on this before but I don't remember the outcome 😅 .

addaleax:
> Should we also offer this as an npm script? Looks like it isn't platform dependent because we're using docker. Might be useful for debugging if this test ever does fail.

You can use the raw shell script that’s included here, that works locally and doesn’t need Docker (just some graphical interface in order to be able to run vscode) – we could also provide one for Windows, if we want to, though.

> In the case where we're updating the APIs that vscode talks to, we do expect these tests to fail. Do we just ignore them in the meantime? I think we've chatted on this before but I don't remember the outcome sweat_smile .

Yeah, I’m not sure. We could either temporarily disable the tests, or just try to always have an upgrade path available instead of clean breakages.

Anemy (approved):
Screen Shot 2020-11-20 at 12 02 39 PM
lgtm 👍

PR #422, opened 2020-11-20
feat(cli-repl): add passwordPrompt() support MONGOSH-315

addaleax:
This is a partial implementation of

* MONGOSH-314
* MONGOSH-315

and picks up https://github.com/mongodb-js/mongosh/pull/381 again
now that it’s easier to test this. :)

### DIFF
--- a/packages/cli-repl/src/cli-repl.ts
+++ b/packages/cli-repl/src/cli-repl.ts
@@ -8,12 +8,11 @@ import MongoshNodeRepl from './mongosh-repl';
 import Nanobus from 'nanobus';
 import setupLoggerAndTelemetry from './setup-logger-and-telemetry';
 import type { StyleDefinition } from './clr';
-import read from 'read';
+import askpassword from 'askpassword';
 import semver from 'semver';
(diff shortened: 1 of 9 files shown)


packages/cli-repl/src/cli-repl.ts
```diff
@@ -172,15 +171,14 @@ class CliRepl {
...
       input: this.input,
-      output: this.output
-    };
+      output: this.output,
+      replacementCharacter: '*'
+    });
+    this.output.write('Enter password: ');
     try {
-      (driverOptions.auth as any).password = await promisify(read)(readOptions);
+      (driverOptions.auth as any).password = (await passwordPromise).toString();
```
addaleax, line 181: We could keep `read` if we want to, but it doesn’t feel like it makes sense to have two different libraries for achieving the same task. We can’t use `read` for `passwordPrompt()`, though, because that doesn’t work with the REPL (because it doesn’t remove the REPL’s `.on('data')` listeners for the input).

mcasimir (approved):
👍

PR #423, opened 2020-11-23
Support for Sessions & Transactions

aherlihy:
This PR includes tickets:
- MONGOSH-444 - Create Session Object
- MONGOSH-445 - Session Object Methods
- MONGOSH-448 - Throw for unsupported sessions methods
- MONGOSH-446 - Transaction Methods
- MONGOSH-342 - Mongo.readPreference methods
- MONGOSH-449 - Investigate if reconnecting MongoClient will interfere with sessions
  - If a user has an active session, then calls a command like `db.auth` or `db.setReadPreference` then nothing will error. However if they try to use the session again, any command will fail with `inactive session`. I think this behavior is acceptable and no further work is required from us because the user should expect that resetting connection-wide options or reauthenticating should close a current session.

For more information about driver sessions, see [the spec](https://github.com/mongodb/specifications/blob/master/source/sessions/driver-sessions.rst#high-level-summary-of-the-api-changes-for-sessions).

The changes include:

- A new `Session` object that wraps the driver session.
- The `Database` class now has a set of `baseOptions` that are passed into every driver call, including anything from the `Collection` class. Right now it only contains `session` but in the future could contain any database-wide option that needs to be tracked outside the driver. 
- Some ServiceProvider methods did not accept an options object, so they have been updated to forward options to the driver. NOTE: this required changes to the java-shell project so I would like @kornilova203 to approve this PR to make sure I didn't do anything wrong there, and to see if there are any other changes required.
- The CliServiceProvider now forwards a `baseCmdOptions` object to all driver methods.
- [NODE-2807](https://jira.mongodb.org/browse/NODE-2807) was completed so I updated the shell code to reflect the change [MONGOSH-342](https://jira.mongodb.org/browse/MONGOSH-342)
- Previously some shell-api methods were using other shell-api methods as helpers, which would end up with double the telemetry events. Added internal helper methods like `_runCommand` to avoid that for most cases (but not all).
- Since the `rs` and `sh` objects are session-aware, forward all their service-provider calls via the `Database` class instead of the `Mongo` class.
- The driver does not accept an empty writeConcern, so methods now send nothing in the options object instead of `{}` if the writeConcern is not set by the user.
- 2 bugfixes involving errors states and reconnecting the MongoClient

### DIFF
--- a/packages/java-shell/src/main/kotlin/com/mongodb/mongosh/service/JavaServiceProvider.kt
+++ b/packages/java-shell/src/main/kotlin/com/mongodb/mongosh/service/JavaServiceProvider.kt
@@ -32,7 +32,7 @@ internal class JavaServiceProvider(private val client: MongoClient, private val 
     }
 
     @HostAccess.Export
-    override fun runCommandWithCheck(database: String, spec: Value): Value = promise {
+    override fun runCommandWithCheck(database: String, spec: Value, options: Value?): Value = promise {
         getDatabase(database, null).map { db ->
             val res = if (spec.isString) {
                 db.runCommand(Document(spec.asString(), 1))
@@ -154,7 +154,7 @@ internal class JavaServiceProvider(private val client: MongoClient, private val 
     }
 
     @HostAccess.Export
-    override fun dropDatabase(database: String, writeConcern: Value?, dbOptions: Value?): Value = promise<Any?> {
+    override fun dropDatabase(database: String, option: Value?, dbOptions: Value?): Value = promise<Any?> {
         Left(NotImplementedError())
     }
 
@@ -449,7 +449,7 @@ internal class JavaServiceProvider(private val client: MongoClient, private val 
     }
 
     @HostAccess.Export
-    override fun getIndexes(database: String, collection: String): Value = promise {
+    override fun getIndexes(database: String, collection: String, options: Value?): Value = promise {
         getDatabase(database, null).map { db ->
             db.getCollection(collection).listIndexes()
         }
@@ -534,7 +534,7 @@ internal class JavaServiceProvider(private val client: MongoClient, private val 
     }
 
     @HostAccess.Export
-    override fun dropIndexes(database: String, collection: String, indexes: Value?): Value = promise<Any?> {
+    override fun dropIndexes(database: String, collection: String, indexes: Value?, options: Value?): Value = promise<Any?> {
         val indexes = if (indexes != null && !indexes.isNull) indexes else throw IllegalArgumentException("Indexes parameter must not be null")
         val indexesList = if (indexes.hasArrayElements()) toList(indexes, "indexes")!!
         else listOf(context.extract(indexes).value)
@@ -557,7 +557,7 @@ internal class JavaServiceProvider(private val client: MongoClient, private val 
     }
 
     @HostAccess.Export
-    override fun dropCollection(database: String, collection: String): Value = promise {
+    override fun dropCollection(database: String, collection: String, option: Value?): Value = promise {
         getDatabase(database, null).map { db ->
             db.getCollection(collection).drop()
         }
--- a/packages/java-shell/src/main/kotlin/com/mongodb/mongosh/service/JavaServiceProvider.kt
+++ b/packages/java-shell/src/main/kotlin/com/mongodb/mongosh/service/JavaServiceProvider.kt
@@ -154,7 +154,7 @@ internal class JavaServiceProvider(private val client: MongoClient, private val 
     }
 
     @HostAccess.Export
-    override fun dropDatabase(database: String, option: Value?, dbOptions: Value?): Value = promise<Any?> {
+    override fun dropDatabase(database: String, options: Value?, dbOptions: Value?): Value = promise<Any?> {
         Left(NotImplementedError())
     }
 
@@ -557,7 +557,7 @@ internal class JavaServiceProvider(private val client: MongoClient, private val 
     }
 
     @HostAccess.Export
-    override fun dropCollection(database: String, collection: String, option: Value?): Value = promise {
+    override fun dropCollection(database: String, collection: String, options: Value?): Value = promise {
         getDatabase(database, null).map { db ->
             db.getCollection(collection).drop()
         }
--- a/packages/service-provider-server/src/cli-service-provider.ts
+++ b/packages/service-provider-server/src/cli-service-provider.ts
@@ -50,7 +50,9 @@ import {
   ServiceProviderCore,
   AuthOptions,
   ReadConcern,
-  ReadPreference
+  ReadPreference,
+  ServiceProviderSession,
+  SessionOptions
 } from '@mongosh/service-provider-core';
 
 import NodeOptions from './node/node-options';
@@ -83,8 +85,6 @@ const DEFAULT_BASE_OPTIONS = Object.freeze({
  * Encapsulates logic for the service provider for the mongosh CLI.
  */
 class CliServiceProvider extends ServiceProviderCore implements ServiceProvider {
-  public readonly platform: ReplPlatform;
-  public readonly initialDb: string;
   /**
    * Create a new CLI service provider from the provided URI.
    *
@@ -114,9 +114,12 @@ class CliServiceProvider extends ServiceProviderCore implements ServiceProvider 
     return new CliServiceProvider(mongoClient, clientOptions, uri);
   }
 
+  public readonly platform: ReplPlatform;
+  public readonly initialDb: string;
   public mongoClient: MongoClient; // public for testing
   private readonly uri?: string;
   private initialOptions: any;
+  public baseCmdOptions: any; // public for testing
 
   /**
    * Instantiate a new CliServiceProvider with the Node driver's connected
@@ -136,6 +139,7 @@ class CliServiceProvider extends ServiceProviderCore implements ServiceProvider 
       this.initialDb = DEFAULT_DB;
     }
     this.initialOptions = clientOptions;
+    this.baseCmdOptions = { ... DEFAULT_BASE_OPTIONS }; // currently do not have any user-specified connection-wide command options, but I imagine we will eventually
   }
 
   async getNewConnection(uri: string, options: NodeOptions = {}): Promise<CliServiceProvider> {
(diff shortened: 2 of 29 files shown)


packages/service-provider-server/src/cli-service-provider.ts
```diff
@@ -1092,10 +1098,15 @@ class CliServiceProvider extends ServiceProviderCore implements ServiceProvider
       authMechanism: authDoc.mechanism,
       authSource: authDoc.authDb
     };
-    this.mongoClient = await MongoClient.connect(
```
aherlihy, line 1095: This was a bug where we were not closing old MongoClients upon reauthentication.

packages/service-provider-server/src/cli-service-provider.ts
```diff
@@ -1165,16 +1168,37 @@ class CliServiceProvider extends ServiceProviderCore implements ServiceProvider
...
     };
-    try {
-      await this.mongoClient.close();
-      // eslint-disable-next-line no-empty
-    } catch {}
     const mc = await MongoClient.connect(
       this.uri as string,
       clientOptions
     );
+    try {
```
aherlihy, line 1175: This was a bug where if the reconnection failed for whatever reason, we had already closed the old client so could not restore the connection

packages/shell-api/src/collection.spec.ts
```diff
@@ -1170,4 +1175,115 @@ describe('Collection', () => {
...
+          }
+          expect(serviceProvider[method].calledOnce).to.equal(true, `expected ${method} to be called but it was not`);
+          expect((serviceProvider[method].getCall(-1).args[3] as any).session).to.equal(internalSession);
+        }
+      }
+    });
+    it('all commands that use other methods', async() => {
+      for (const method of Object.getOwnPropertyNames(Collection.prototype).filter(
+        k => Object.keys(exceptions).includes(k)
+      )) {
```
addaleax, line 1268: Do we just want to
```suggestion
      for (const method of Object.keys(exceptions)) {
```

(and ditto for `database`)? It would be unexpected for any key of the exceptions object to be missing as a Collection method, right?

packages/shell-api/src/mongo.ts
```diff
@@ -174,4 +171,24 @@ export default class Mongo extends ShellApiClass {
...
+  setCausalConsistency(): void {
+    throw new MongoshInvalidInputError('It is not possible to set causal consistency for an entire connection due to the driver, use startSession({causalConsistency: <>}) instead.');
+  }
+
+  isCausalConsistency(): void {
+    throw new MongoshInvalidInputError('Causal consistency for drivers is set via Mongo.startSession and can be checked via session.getOptions. The default value is true');
+  }
+
+  setSlaveOk(): void {
+    throw new MongoshInvalidInputError('setSlaveOk is deprecated.');
```
addaleax, line 188: Is this also being replaced by `setReadPreference`?

```suggestion
    throw new MongoshInvalidInputError('setSlaveOk is deprecated, use setReadPreference instead.');
```

packages/shell-api/src/session.spec.ts
```diff
@@ -0,0 +1,325 @@
...
+    const delay = util.promisify(setTimeout);
+    const ensureMaster = async(timeout): Promise<void> => {
+      while (!(await mongo.getDB(ADMIN_DB).isMaster()).ismaster) {
+        if (timeout > 32000) {
+          return expect.fail(`Waited for ${cfg.members[0].host} to become master, never happened`);
+        }
+        await delay(timeout);
+        timeout *= 2; // try again but wait double
+      }
+    };
```
addaleax, line 141: Do we want to put this into a helper file, to avoid having duplicate code between here and the other replset tests?

packages/java-shell/src/main/kotlin/com/mongodb/mongosh/service/JavaServiceProvider.kt
```diff
@@ -154,7 +154,7 @@ internal class JavaServiceProvider(private val client: MongoClient, private val
     }
 
     @HostAccess.Export
-    override fun dropDatabase(database: String, writeConcern: Value?, dbOptions: Value?): Value = promise<Any?> {
+    override fun dropDatabase(database: String, option: Value?, dbOptions: Value?): Value = promise<Any?> {
```
addaleax, line 157:
```suggestion
    override fun dropDatabase(database: String, options: Value?, dbOptions: Value?): Value = promise<Any?> {
```

for consistency?

packages/java-shell/src/main/kotlin/com/mongodb/mongosh/service/JavaServiceProvider.kt
```diff
@@ -557,7 +557,7 @@ internal class JavaServiceProvider(private val client: MongoClient, private val
     }
 
     @HostAccess.Export
-    override fun dropCollection(database: String, collection: String): Value = promise {
+    override fun dropCollection(database: String, collection: String, option: Value?): Value = promise {
```
addaleax, line 560:
```suggestion
    override fun dropCollection(database: String, collection: String, options: Value?): Value = promise {
```

addaleax (approved):
:tada:

> NODE-2807 was completed so I updated the shell code to reflect the change MONGOSH-342

Do we need to bump the dependency version in `packages/service-provider-server/package.json`?

packages/shell-api/src/mongo.ts
aherlihy: setting slaveOk and setting read preference are slightly different operations, and from what I could tell / what Matt indicated, there's no straightforward way to set slaveOk using readPreference that might not end up with some unexpected behavior (for example just setting readPreference to secondaryOk isn't the same as setting primaryPreferred)

aherlihy:
The NODE ticket https://jira.mongodb.org/browse/NODE-2807 indicates it was fixed in 3.6.3, which we are already on

packages/shell-api/src/collection.spec.ts
aherlihy: Most of the database methods are actually _not_ present in `exceptions`. The methods present in `exceptions` are the methods that behave differently from the standard methods, for example if they have their options argument in a different location or they don't use runCommand and instead use some other driver method.

I also want to apply this test to _all_ database and collection methods, including future ones.

packages/shell-api/src/session.spec.ts
aherlihy: will do!

addaleax:
> The NODE ticket https://jira.mongodb.org/browse/NODE-2807 indicates it was fixed in 3.6.3, which we are already on

```sh
$ grep mongodb-3.6 packages/service-provider-server/package-lock.json 
			"resolved": "https://registry.npmjs.org/mongodb/-/mongodb-3.6.1.tgz",
```

Am I missing something?

aherlihy:
Oh! I was going off https://github.com/mongodb-js/mongosh/blob/master/package.json#L121

addaleax:
> Oh! I was going off https://github.com/mongodb-js/mongosh/blob/master/package.json#L121, so we should remove the requirement form the service-provider-server package entirely

I think we should remove it from the main package.json, it’s a devDependency there and `service-provider-server` needs it as an actual dependency when we publish it to npm

packages/shell-api/src/collection.spec.ts
aherlihy: Oh oops nevermind, I totally misread the suggestion!

packages/java-shell/src/main/kotlin/com/mongodb/mongosh/service/JavaServiceProvider.kt
aherlihy: Yes, ty!

packages/java-shell/src/main/kotlin/com/mongodb/mongosh/service/JavaServiceProvider.kt
aherlihy: Yes, ty!

aherlihy:
Unfortunately it doesn't look like I can remove it from the top-level dev dependency in package.json, since running tests gives me 'cannot find module mongodb'. Do you know if there's a better way to link it to the version set in service-provider-server?

addaleax:
@aherlihy I’m not sure there’s any way to ensure that the two of them match (besides adding an explicit test that just parses the JSON files) … for now I would probably just update it in `service-provider-server` only

aherlihy:
Ok, will do

kornilova203 (approved):

packages/shell-api/src/database.ts
```diff
@@ -181,15 +234,15 @@ export default class Database extends ShellApiClass {
...
+      aggOptions,
       dbOptions,
       explain
     } = adaptAggregateOptions(options);
 
     const providerCursor = this._mongo._serviceProvider.aggregateDb(
       this._name,
       pipeline,
-      providerOptions,
+      { ...aggOptions, ...this._baseOptions },
```
rose-m, line 245: Maybe this is super nitpicky - but I think the order of arguments is swapped here and it should be `{ ...this._baseOptions, ...aggOptions}` 🤔 Other wise the `_baseOptions` would override anything in the `aggOptions`. Inside `collections.ts` the order is as basically everywhere else with `_baseOptions` first: https://github.com/mongodb-js/mongosh/pull/423/files#diff-c975be9c4e061c8d6cf6aac2dd4da42c6096bdf5e8e62601eca89f85e22302c3R130.

packages/shell-api/src/database.ts
```diff
@@ -235,12 +287,12 @@ export default class Database extends ShellApiClass {
   async dropDatabase(writeConcern?: WriteConcern): Promise<any> {
     return await this._mongo._serviceProvider.dropDatabase(
       this._name,
-      writeConcern
+      { writeConcern, ...this._baseOptions }
```
rose-m, line 290: Same here - I'd naively expect `{...this._baseOptions, writeConcern }`.

rose-m (commented):
Huge work! 🎉 I just looked over the code and didn't run it - spotted 2 tiny things that maybe you could take a look at :)

packages/shell-api/src/database.ts
aherlihy: Great catch! Not a nitpick at all :)

PR #424, opened 2020-11-23
fix(test): allow running tests with mongod 4.0, 4.2, 4.4 MONGOSH-312

addaleax:
Make tests pass for all of the target server versions.

### DIFF
--- a/packages/shell-api/src/integration.spec.ts
+++ b/packages/shell-api/src/integration.spec.ts
@@ -7,7 +7,6 @@ import AggregationCursor from './aggregation-cursor';
 import ShellApi from './shell-api';
 import { startTestServer, skipIfServerVersion } from '../../../testing/integration-testing-hooks';
 import { toShellResult } from './index';
-import semver from 'semver';
 
 // Compile JS code as an expression. We use this to generate some JS functions
 // whose code is stringified and compiled elsewhere, to make sure that the code
@@ -18,7 +17,6 @@ const compileExpr = (templ, ...subs): any => {
 
 describe('Shell API (integration)', function() {
   const testServer = startTestServer('shared');
-  let serverVersion;
   this.timeout(60000);
   let serviceProvider: CliServiceProvider;
 
@@ -92,7 +90,6 @@ describe('Shell API (integration)', function() {
   };
 
   before(async() => {
-    serverVersion = await testServer.serverVersion();
     serviceProvider = await CliServiceProvider.connect(await testServer.connectionString());
   });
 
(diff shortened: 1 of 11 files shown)


packages/shell-api/src/integration.spec.ts
```diff
@@ -492,9 +495,13 @@ describe('Shell API (integration)', function() {
     });
 
     describe('stats', () => {
+      const extraStatsKeys: string[] = [];
```
mcasimir, line 498: I think this is not really necessary. I believe i've done this test initially, and i was probably a bit lazy and only copy/pasted the result that matched.

Honestly we can just use a different strategy and match only few keys, just to make sure the result is a 'stats' result and not some other command, after all we don't fiddle with the command result in any way.

Long story short, i think is easier and more stable to just remove  `extraStatsKeys` and avoid having to support the differences between versions here.

addaleax: Sure, I mean, that’s also easier for me :slightly_smiling_face:

testing/download-mongodb.ts
```diff
@@ -64,12 +67,21 @@ async function lookupDownloadUrl(versionInfo: VersionInfo): Promise<string> {
...
 }
 
+const downloadPromises: Record<string, Promise<string>> = {};
 // Download mongod + mongos and return the path to a directory containing them.
-export async function downloadMongoDb() {
-  const resp = await fetch('https://downloads.mongodb.org/full.json');
-  const json: FullJSON = await resp.json();
-  const productionVersions = json.versions
-    .filter((version: VersionInfo) => version.production_release)
+export async function downloadMongoDb(targetVersion = '*'): Promise<string> {
```
mcasimir, line 72: `'*'` why not `latest`?

packages/shell-api/src/shard.spec.ts
```diff
@@ -1210,8 +1210,11 @@ describe('Shard', () => {
         expect((await sh.startBalancer()).ok).to.equal(1);
         expect((await sh.isBalancerRunning()).mode).to.equal('full');
       });
-      it('reports state for collection', async() => {
-        expect(Object.keys(await sh.balancerCollectionStatus(ns))).to.include('balancerCompliant');
+      describe('balancerCollectionStatus', () => {
+        skipIfServerVersion(mongos, '< 4.4');
```
mcasimir, line 1214: is `mongos` a string here?

addaleax: It’s the reference to the test server (i.e. a `MongodSetup` from `integration-testing-hooks.ts`)

testing/download-mongodb.ts
addaleax: because `latest` is not a meaningful semver string … I’ll update the variable name to reflect that this uses semver :)

mcasimir (approved):

PR #425, opened 2020-11-23
chore(async-rewriter): remove unused typescript deps

addaleax:
We list typescript and ts-node as dependencies in the top-level
package.json, so they should not be specified in the async-rewriter
package.json (and, more importantly, not as full dependencies
and rather as devDependencies).

### DIFF
--- a/packages/async-rewriter/src/async-writer-babel.ts
+++ b/packages/async-rewriter/src/async-writer-babel.ts
@@ -526,7 +526,8 @@ var TypeInferenceVisitor: Visitor = { /* eslint no-var:0 */
         if (classPath.node['shellType'].returnType.type === 'unknown') {
           classPath.node['shellType'].returnType.type = 'object';
         }
-        classPath.node['shellType'].returnType.attributes[path.node.key.name] = path.node['shellType'];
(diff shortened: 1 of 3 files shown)


packages/async-rewriter/src/async-writer-babel.ts
```diff
@@ -526,7 +526,8 @@ var TypeInferenceVisitor: Visitor = { /* eslint no-var:0 */
         if (classPath.node['shellType'].returnType.type === 'unknown') {
           classPath.node['shellType'].returnType.type = 'object';
         }
-        classPath.node['shellType'].returnType.attributes[path.node.key.name] = path.node['shellType'];
+        // TODO: The 'as any' here is probably not really a valid thing to do.
+        classPath.node['shellType'].returnType.attributes[(path.node.key as any).name] = path.node['shellType'];
```
aherlihy, line 530: I'm surprised that this would change due to dependencies being moved from local to top-level Do you know why it's an issue now and not before?

addaleax: @aherlihy I assume it’s caused by a mismatch in the typescript version that we’re using here

aherlihy (approved):

PR #428, opened 2020-11-24
[java-shell] Don't fetch cursor twice

kornilova203:
This PR includes:
1. commit 170142901052e0e167dae068c8945c2fcf450f81 from @addaleax that allows to set custom result handler
2. my commit that adds custom result handler to java shell so cursor is not fetched before it's returned to user

BTW: after these commits are merges java shell will not need `rawValue` field in shell result, so maybe it should be removed.


addaleax (approved):

addaleax:
> BTW: after these commits are merges java shell will not need rawValue field in shell result, so maybe it should be removed.

I’d keep it. This seems like a reasonable thing to provide, and it might be useful in other contexts for the same reasons that it was useful in the java-shell. :slightly_smiling_face:

PR #429, opened 2020-11-24
[java-shell] remove URI.txt file

kornilova203:


kornilova203:
@addaleax, please activate tests so we check that they don't fall without URI.txt

addaleax (approved):

PR #432, opened 2020-11-24
[java-shell] in tests parse host and port from uri

kornilova203:


kornilova203:
@addaleax, please start tests

addaleax (approved):

PR #433, opened 2020-11-24
chore(ci): add jobs for testing across mongod versions MONGOSH-312

addaleax:
This is basically a revamp the evergreen config:

- Rename CI tasks to be more descriptive
- Add CI tasks to verify that the Linux artifacts actually run on
  their target platforms (we previously had jobs in CI that were
  *named* to imply that they were using RHEL 8.0 and Debian 10
  but did not actually run those OSes in any way, only built
  packages targeting those)
- Add test_m??x tasks where ?? represents the target server version
  for testing
- Update the mongod download script to account for those different
  server versions across the supported OSes
- Move darwin testing off the codesign machine (because there is
  only one host for that) – compiling still takes place there
- Update the build config script to account for the fact that
  the `revision` expansion only represents the *base* revision for
  patches, and use a different format for patches so that S3
  artifacts do not accidentally get overridden with new artifacts
  from patches targeting the same base revision
- Actually wait for all other tasks to succeed before continuing
  with a release

### DIFF
--- a/.evergreen.yml
+++ b/.evergreen.yml
@@ -2,9 +2,9 @@ exec_timeout_secs: 7200
 
 # Variables for common functions.
 #
-# compile_and_release: Runs the exec compile and release process.
+# compile_and_upload_artifact: Runs the exec compile and release process.
 variables:
-  - &compile_and_release
+  - &compile_and_upload_artifact
     command: shell.exec
     params:
       working_dir: src
@@ -48,7 +48,22 @@ variables:
         echo "$MONGOSH_TEST_EXECUTABLE_PATH"
         npm run test-e2e-ci
 
-  - &compile_and_release_ps
+  - &test_linux_artifact
+    command: shell.exec
+    params:
+      working_dir: src
+      shell: bash
+      script: |
+        set -e
+        source .evergreen/.setup_env
+        export EVERGREEN_EXPANSIONS_PATH="$(pwd)/../tmp/expansions.yaml"
+        node scripts/download-linux-artifact
+        dist/mongosh --version
+        export MONGOSH_TEST_EXECUTABLE_PATH="$(pwd)/dist/mongosh"
+        echo "$MONGOSH_TEST_EXECUTABLE_PATH"
+        npm run test-e2e-ci
+
+  - &compile_and_upload_artifact_ps
     command: shell.exec
     params:
       working_dir: src
@@ -66,15 +81,20 @@ variables:
         npm run test-e2e-ci
 
 # Functions are any command that can be run.
+# Variables ending with _ps do the same thing as their non-suffixed counterparts
+# but use PowerShell instead.
 #
 # Current functions:
 #   checkout - Checks out the project from git.
 #   install - Installs Node and all dependencies.
 #   check - Performs linter and dependency checks.
 #   test - Runs all tests.
-#   release_macos - Publishes the npm packages and uploads the tarballs.
-#   release_linux - Publishes the npm packages and uploads the tarballs.
-#   release_win_ps - Publishes the npm packages and uploads the tarballs.
+#   test_vscode - Clones the vscode extension repository and runs its tests.
+#   compile_and_upload_artifact - Compile the release binary and upload it to S3.
+#   test_linux_artifact - Test that the built artifact works where we expect it to.
+#                         We use this to verify that e.g. the Ubuntu-built release
+#                         binary also works on RHEL and Debian.
+#   release_publish - Publishes the npm packages and uploads the tarballs.
 functions:
   checkout:
     - command: git.get_project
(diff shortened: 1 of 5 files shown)


.evergreen.yml
```diff
@@ -218,115 +226,173 @@ tasks:
...
-      - name: test
+      - name: test_m40x
+      - name: test_m42x
+      - name: test_m44x
       - name: test_vscode
-      - name: release_linux
+      - name: compile_and_upload_artifact_linux
   - name: rhel
-    display_name: "RHEL 8.0 (in Ubuntu docker)"
+    display_name: "Ubuntu 18.04 (rpm target)"
```
rose-m, line 366: Stupid question: what are `rhel` and `debian`? They are both running on `ubuntu1804-test` and thus the same as `linux`? 🤔 Can we throw them out?

rose-m (approved):
Testing all versions 👏 😄 Just got a minor question regarding `rhel` vs `rhel_real` that's got me like 🤔

.evergreen.yml
```diff
@@ -151,36 +171,24 @@ functions:
             -e BUILD_VARIANT \
             --rm -v $PWD:/tmp/build ubuntu18.04-xvfb \
             -c 'cd /tmp/build && ./testing/test-vscode.sh'
-  release_macos:
+  compile_and_upload_artifact:
```
mcasimir, line 174: Nice! Super clear

.evergreen.yml
```diff
@@ -218,115 +226,173 @@ tasks:
...
+      - name: compile_and_upload_artifact_linux
+        variant: linux
+    commands:
+      - func: checkout
+      - func: install
+      - func: test_linux_artifact
   - name: release_publish
     depends_on:
-      - name: release_win_ps
+      - name: compile_and_upload_artifact_ps
```
mcasimir, line 323: Probably at this point we can rename `ps` to `win` and `powershell` to `windows`?
I think the only reason we had `ps` was cause at some point we were transitioning from `cygwin` which was the default windows shell in evergreen

config/build.conf.js
```diff
@@ -1,3 +1,5 @@
+'use strict';
```
mcasimir, line 1: just curious, why is this needed here?

config/build.conf.js
```diff
@@ -41,6 +43,13 @@ const ANALYTICS_CONFIG = path.join(MONGOSH, 'lib', 'analytics-config.js');
  */
 const APPLE_NOTARIZATION_BUNDLE_ID = 'com.mongodb.mongosh';
 
+/**
+ * The SHA for the current git HEAD.
+ */
+const REVISION = process.env.IS_PATCH ?
+  `pr-${process.env.GITHUB_PR_NUMBER}-${process.env.REVISION_ORDER_ID}` :
```
mcasimir, line 50: How is this used? This will be probably super-useful for releases too.

mcasimir (approved):
👍

.evergreen.yml
addaleax: Ah, right :slightly_smiling_face: So … our `@mongosh/build` package makes a distinction on how to behave based on the build variant name, and, in particular, it packages as a tarball/deb/rpm based on it. The variable gets picked up in `config/build.conf.js` from an environment variable, and passed around throughout the build package.

That’s why the names here, `rhel` and `debian`, are not freely changeable:

https://github.com/mongodb-js/mongosh/blob/dde7b239f0cef50d217a4bf1766a89c8890a2e9d/packages/build/src/build-variant.ts#L7-L13

This is obviously not an ideal situation, no. It means that we’re currently building the Linux binary three separate times, only differing in the final packaging step, which wastes lots of CPU time in CI :slightly_smiling_face: But I also wanted to get the specific task in question done, and this doesn’t make anything worse in this respect as far as I can tell.

Ultimately, it would be nice imo if:

- We didn’t make choices on how to behave purely based on the build variant name, and
- We only built the Linux binary exactly once.

config/build.conf.js
addaleax: @mcasimir It’s not needed for anything in particular, it’s just good practice to always write JS code in strict mode (TS should already always be strict mode for us)

config/build.conf.js
addaleax: Err, it might be easier to show this once CI has run :) It’s used for uploading artifacts to S3; e.g. the one from this PR would now become available at 

https://s3.amazonaws.com/mciuploads/mongosh/pr-433-anna.henningsen_158/mongosh-0.5.2-linux.tgz

instead of 

https://s3.amazonaws.com/mciuploads/mongosh/dde7b239f0cef50d217a4bf1766a89c8890a2e9d/mongosh-0.5.2-linux.tgz

(where dde7b239f0cef50d217a4bf1766a89c8890a2e9d is the *base* revision, not a commit from this PR). That matters because we would otherwise overwrite the already-existing tarball for dde7b239f0cef50d217a4bf1766a89c8890a2e9d – we’re lucky we didn’t have problems with this, but in the worst case, I think this could previously have potentially made builds from a PR override a release build :grimacing:

.evergreen.yml
rose-m: Thank you very much! Didn't expect these names to be relevant outside the config, but that's good to know 👍

.evergreen.yml
addaleax: Sure, done :+1:

PR #435, opened 2020-11-24
fix(cli-repl): turn LineByLineInput into Readable stream MONGOSH-401

addaleax:
Instead of controlling whether `data` events are emitted or not
in the LineByLineInput implementation, make LineByLineInput
an actual Readable stream and emit data that way.

This ensures that the relative timing of `'data'` events is the
same as it would otherwise be for streams, which the Node.js
readline implementation expects.


mcasimir (approved):
Much nicer! Thanks!

PR #436, opened 2020-11-24
[java-shell] support db.createView, db.dropDatabase Date.getTime, ISODate constructor with offset, insertMany options

kornilova203:
* support Date.getTime method 
* support db.createView() 
* Date() should return readable value e.g. 'Tue Nov 24 14:54:31 UTC 2020' 
* support offset in ISODate 
* support insertMany options 
* add java test for MONGOSH-341 (wrong NumberLong with big values) 
* support db.dropDatabase()


kornilova203:
@addaleax, please start tests :)

PR #441, opened 2020-11-25
feat(browser-repl): support passwordPrompt() MONGOSH-315

addaleax:
Adds a password prompt feature to the browser-repl for parity with
the CLI.

Example run

### DIFF
--- a/packages/browser-repl/src/components/password-prompt.less
+++ b/packages/browser-repl/src/components/password-prompt.less
@@ -1,0 +1,12 @@
+@import '~@leafygreen-ui/palette/dist/ui-colors.less';
+@import './common.less';
+
+.password-prompt input {
+  .font();
+
+  background-color: @leafygreen__gray--dark-3;
+  color: @leafygreen__gray--light-3;
+  padding: 0 3px;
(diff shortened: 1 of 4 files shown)


packages/browser-repl/src/components/password-prompt.less
```diff
@@ -0,0 +1,12 @@
+@import '~@leafygreen-ui/palette/dist/ui-colors.less';
+@import './common.less';
+
+.password-prompt input {
```
gribnoysup, line 4: Feels like in the current implementation the prompt is too close to the left side of the shell. Maybe we should add some padding to align it with text above, what do you think?

```suggestion
.password-prompt {
  padding-left: 23px;
}

.password-prompt input {
```

This is how it looks:

image

gribnoysup (approved):
Overall looks and works great! I have a tiny suggestion to add some padding to the prompt

packages/browser-repl/src/components/password-prompt.less
addaleax: Sure, that works for me :)

PR #444, opened 2020-11-26
chore: add meaningful error codes MONGOSH-431

rose-m:
## Description
This PR adds the concept of explicit error codes to all `Mongosh...Error` classes. Explicit error codes follow the pattern of `<SCOPE>-<number>` where `SCOPE` allows to group error codes, e.g. to a specific package.

Both the complete error code as well as the extracted scope are used in telemetry to allow faster identification of errors.

This PR also adds a utility to auto-generate a documentation overview of all error codes and their JSDoc in all packages.

### Adding error codes to `Mongosh...Error`s
Most constructors have a new optional argument `code: string` that can be filled to signal the explicit error code. Some of the errors use a pre-defined error code (e.g. `MongoshInternalError`).

All constructors also take an additional `metadata: Object` argument that can be provided to send additional info to telemetry.

Example:
```
throw new MongoshInvalidInputError('Unable to handle nested assignment to this keyword', 'ASYNC-01006');
```

The error code will be printed on the console, too:
```
MongoshUnimplementedError: [ASYNC-10006] Unable to handle nested assignment to 'this' keyword
     at Object.exit (.../async-rewriter/src/async-writer-babel.ts:312:21)
     at NodePath._call (.../async-rewriter/node_modules/@babel/traverse/lib/path/context.js:55:20)
     at NodePath.call (.../async-rewriter/node_modules/@babel/traverse/lib/path/context.js:42:17)
     at NodePath.visit (.../async-rewriter/node_modules/@babel/traverse/lib/path/context.js:99:8)
     at TraversalContext.visitQueue (.../async-rewriter/node_modules/@babel/traverse/lib/context.js:112:16)
```

### Error code and scope
In general, an error code is an arbitrary string. However, the code automatically parses the error code to find a potential _scope_. Error codes are parsed using `/^([a-zA-Z0-9]+)-/` where the group denotes the _scope_.

Example:
```
Code: ASYNC-01005 ---> Scope: ASYNC
```

### Declaring errors and overview generation
To automatically generate an overview of all error codes used in the `mongosh` packages, the errors have to be declared in a specific form:

1. The error codes MUST be declared as members of an `enum` with string literal initializers.
2. The `enum` declaration MUST have a JSDoc tag `@mongoshErrors`.
3. The enum members SHOULD have a JSDoc comment explaining their meaning.

Example error enum:
```
/**
 * @mongoshErrors
 */
enum AsyncRewriterErrors {
  /**
   * Identifies an error where object destructuring was used with Mongosh API objects which we currently do not support.
   *
   * To prevent this error, refrain from using object destructuring.
   */
  DestructuringNotImplemented = 'ASYNC-10001'
}
```

To run the generation of the overview page, run in the top-level directory:
```
npm run generate-error-overview
```


packages/errors/scripts/extract-errors.ts
```diff
@@ -0,0 +1,189 @@
+import * as fs from 'fs';
```
addaleax, line 1: Feel free to ignore, but I think you can also 

```suggestion
import { promises as fs } from 'fs';
```

since you’re always using them as async functions anyway :)

packages/errors/scripts/extract-errors.ts
```diff
@@ -0,0 +1,189 @@
...
+      ux.quiet('No mongosh error enums found.\n');
+    }
+  }
+
+  await renderErrorOverview(packageErrors);
+
+  ux.success('👏👏👏👏');
+  ux.success('Wrote generated overview page to: error-overview.md');
+  ux.success('👏👏👏👏');
+})();
```
addaleax, line 40:
```suggestion
})().catch(err => process.nextTick(() => { throw err; }));
```

Otherwise errors will not lead to an exit with a non-zero exit code (yet, that will happen in a future version of Node.js)

addaleax (approved):
This approach works for me if everyone’s good with it :)

packages/async-rewriter/src/async-writer-babel.spec.ts
```diff
@@ -1214,6 +1215,7 @@ function f() {
           compileCheckScopes(writer, 'let {a} = {a: db}');
         } catch (e) {
           expect(e.name).to.be.equal('MongoshUnimplementedError');
+          expect(e.code).to.be.equal(AsyncRewriterErrors.DestructuringNotImplemented);
```
aherlihy, line 1218: This is great! It will be a decent amount of lines of code that need to be updated for the tests, but it makes it so much more precise.

aherlihy (commented):
This is awesome!

packages/errors/scripts/extract-errors.ts
rose-m: This exists? 😲

packages/shell-api/src/error-codes.ts
```diff
@@ -0,0 +1,572 @@
...
+   * **Solution: Make sure you are connected to a replica set.**
+   */
+  DatabasePrintSecondaryReplicationInfoNoReplicaSet = 'SHAPI-10314',
+
+  /**
+   * Signals calling `getReplicationInfo` of the Database API without being connected to a replica set.
+   *
+   * **Solution: Make sure you are connected to a replica set.**
+   */
+  DatabaseGetReplicationInfoNoReplicaSet = 'SHAPI-10315',
```
addaleax, line 223: Maybe we could group errors where the cause and the solution is identical, e.g. all those which require being connected to a replica set under one error code?

packages/shell-api/src/error-codes.ts
```diff
@@ -0,0 +1,572 @@
...
+
+  /**
+   * Signals calling `advanceClusterTime` of the Session API which is currently not yet supported.
+   *
+   * **Solution: Do not call `advanceClusterTime`.**
+   */
+  SessionAdvanceClusterTimeUnsupported = 'SHAPI-10802',
+
+  /**
+   * Signals an error while converting a BSON string to a valid ISODate.
```
addaleax, line 324:
```suggestion
   * Signals an error while parsing a string as a valid ISO-8610 date.
```

packages/cli-repl/src/error-codes.ts
```diff
@@ -0,0 +1,23 @@
...
+enum CliReplErrors {
+  /**
+   * Signals that the currently installed Node version does not match the one expected by mongosh.
+   *
+   * See the output for further details on the required Node version.
+   */
+  NodeVersionMismatch = 'CLIREPL-10001',
+
+  /**
+   * Signals that mongosh continued to run despite an exit being trigger.
```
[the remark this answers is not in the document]
aherlihy: Do you think it would be useful to have a 'this is very bad' error code so that we can see from our telemetry if something impossible is happening?

packages/shell-api/src/database.ts
```diff
@@ -1156,6 +1226,9 @@ export default class Database extends ShellApiClass {
 
   @returnsPromise
   async printSlaveReplicationInfo(): Promise<CommandResult> {
-    throw new MongoshInvalidInputError('Method deprecated, use db.printSecondaryReplicationInfo instead');
+    throw new MongoshInvalidInputError(
+      'Method deprecated, use db.printSecondaryReplicationInfo instead',
+      ShellApiErrors.DatabasePrintSlaveReplicationInfoRemoved
```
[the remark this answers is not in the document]
aherlihy: I also really am not sure we actually need individual errors for every type of deprecation warning, would one be enough? We don't _really_ care to follow how often people are trying to use deprecated methods right?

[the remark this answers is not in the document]
rose-m: True, I think that's definitely a good point to unify the deprecations and use a single code for them.

aherlihy (changes requested):
Nice job!! this is a huge change and it looks great. I have a few pieces of feedback, mostly to cut down on the number of unique error codes we need to keep track of.

The first is any time a user passes an invalid argument to a function, we don't need any more information other than 'invalid argument'. So all the places in shell-api that we check for a certain type of an argument, or that an object has a required subfield, etc. can all be condensed into one error code. Especially since we so often use the helper methods to check args and those already have generic error codes, and we have the error code depend on an implementation detail.

Secondly (which we should double check with Max) I think we can have one deprecated error code, since we don't care if someone is using something that is deprecated. However we do care if something is removed/unimplemented/deprecated *because the node driver doesn't support it*. So the last bit of feedback is it would be really useful if we could find a way to track all the times a user runs into a limitation that is due to the driver so we can use that data to support our feature requests.

packages/cli-repl/src/error-codes.ts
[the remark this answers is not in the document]
rose-m: I would leave an "OH CRAP" error code in there to make sure we see when one of our assumptions is violated...

packages/cli-repl/src/cli-repl.ts
```diff
@@ -198,7 +199,7 @@ class CliRepl {
   exit(code: number): never {
     this.bus.emit('mongosh:exit', code);
     // Emitting mongosh:exit never returns. If it does, that's a bug.
-    throw new MongoshInternalError('mongosh:exit unexpectedly returned');
+    throw new MongoshInternalError('mongosh:exit unexpectedly returned', CliReplErrors.UnexpectedExitReturn);
```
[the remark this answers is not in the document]
rose-m: Good point - it wasn't :D I explicitly emit the error event on the bus now before throwing and that seems to solve it.

packages/shell-api/src/mongo.ts
```diff
@@ -156,7 +158,10 @@ export default class Mongo extends ShellApiClass {
       const rc = this._serviceProvider.getReadConcern();
       return rc ? rc.level : undefined;
     } catch {
-      throw new MongoshInternalError('Error retrieving ReadConcern. Please file a JIRA ticket in the MONGOSH project');
+      throw new MongoshInternalError(
+        'Error retrieving ReadConcern. Please file a JIRA ticket in the MONGOSH project',
```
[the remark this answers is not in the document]
rose-m: Yes indeed, I only saw that when pointed to by your comment. I'll streamline that.

packages/errors/package.json
```diff
@@ -23,5 +23,11 @@
     "check": "npm run lint",
     "compile-ts": "tsc -p tsconfig.json",
     "prepublish": "npm run compile-ts"
+  },
+  "devDependencies": {
+    "chalk": "^4.1.0",
+    "handlebars": "^4.7.6",
+    "ts-node": "^9.0.0",
+    "typescript": "^4.1.2"
```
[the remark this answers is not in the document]
rose-m: Thanks - I was confused as I could not just run the task inside the `errors` package as I wanted to and added them. What I learned: I have to use `lerna run` on the top level and then it has those dependencies from the top level `package.json` available, too. Solved it 👍

packages/errors/scripts/error-overview.tmpl.md
```diff
@@ -0,0 +1,18 @@
+# mongosh Error Codes Overview
+
+To quickly find an error by it's code just search for the code in this overview.
```
addaleax, line 3:
```suggestion
To quickly find an error by its code, search for the code in this overview.
```

addaleax (approved):

packages/errors/package.json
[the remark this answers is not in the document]
aherlihy: Ah yes! I had that issue earlier too, and now I just run `lerna run test --scope=@mongosh/xxx` instead. And use `describe.only` very liberally when I'm testing locally :)

packages/shell-api/src/collection.ts
```diff
@@ -1493,7 +1525,10 @@ export default class Collection extends ShellApiClass {
       dropped: { $ne: true }
     }));
     if (!isSharded) {
-      throw new MongoshInvalidInputError(`Collection ${this._name} is not sharded`);
+      throw new MongoshInvalidInputError(
+        `Collection ${this._name} is not sharded`,
+        ShellApiErrors.NoSharding
```
aherlihy, line 1530:
```suggestion
        ShellApiErrors.NotConnectedToShardedCluster
```

packages/shell-api/src/collection.spec.ts
```diff
@@ -1123,6 +1184,15 @@ describe('Collection', () => {
         expect(catchedError).to.equal(expectedError);
       });
     });
+    describe('getShardDistribution', () => {
+      it('throws when collection is not sharded', async() => {
+        const error = await collection.getShardDistribution().catch(e => e);
+
+        expect(error).to.be.instanceOf(MongoshInvalidInputError);
+        expect(error.message).to.contain('is not sharded');
+        expect(error.code).to.equal(ShellApiErrors.NoSharding);
```
aherlihy, line 1193:
```suggestion
        expect(error.code).to.equal(ShellApiErrors.NotConnectedToShardedCluster);
```

packages/shell-api/src/database.ts
```diff
@@ -1036,7 +1081,7 @@ export default class Database extends ShellApiClass {
       for (const node of status.members) {
         const nodeResult = {} as any;
         if (node === null || node === undefined) {
-          throw new MongoshInternalError('Member returned from command replSetGetStatus is null');
```
aherlihy, line 1039: Nice catch!

packages/shell-api/src/error-codes.ts
```diff
@@ -0,0 +1,51 @@
...
+ * @mongoshErrors
+ */
+enum ShellApiErrors {
+  /**
+   * Signals calling a method that requires sharding for a collection that is not sharded
+   * or a database that does not have sharding enabled.
+   *
+   * **Solution: Be sure to enable sharding on the database or that the collection is sharded.**
+   */
+  NoSharding = 'SHAPI-10001',
```
aherlihy, line 24: SInce this and the error below are the same, maybe we can rename this to be `NotConnectedToShardedCluster` for consistency?
```suggestion
  NotConnectedToShardedCluster = 'SHAPI-10001',
```

packages/shell-api/src/mongo.ts
```diff
@@ -83,7 +84,7 @@ export default class Mongo extends ShellApiClass {
     assertArgsDefined(name);
     assertArgsType([name], ['string']);
     if (!name.trim()) {
-      throw new MongoshInvalidInputError('Database name cannot be empty.');
+      throw new MongoshInvalidInputError('Database name cannot be empty.', CommonErrors.NotImplemented);
```
aherlihy, line 87: This should be `InvalidArgument` I think?
```suggestion
      throw new MongoshInvalidInputError('Database name cannot be empty.', CommonErrors.InvalidArgument);
```

packages/shell-api/src/shell-internal-state.ts
```diff
@@ -167,7 +167,7 @@ export default class ShellInternalState {
 
     const setFunc = (newDb: any): Database => {
       if (getShellApiType(newDb) !== 'Database') {
-        throw new MongoshInvalidInputError('Cannot reassign \'db\' to non-Database type');
+        throw new MongoshInvalidInputError('Cannot reassign \'db\' to non-Database type', CommonErrors.InvalidArgument);
```
aherlihy, line 170: Possible nit - it's a bit odd to have an error `InvalidArgument` for something that's not an argument. However I'm not sure what else to call, so we can keep this as Argument if you think its' the closest fit. It's inconsequential since we don't really need telemetry on this specific error

packages/shell-api/src/helpers.ts
```diff
@@ -156,8 +161,10 @@ export async function getPrintableShardStatus(db: Database, verbose: boolean): P
 
   const version = await versionColl.findOne();
   if (version === null) {
-    throw new MongoshInvalidInputError('This db does not have sharding enabled.' +
-      ' be sure you are connecting to a mongos from the shell and not to a mongod.');
+    throw new MongoshInvalidInputError(
+      'This db does not have sharding enabled. Be sure you are connecting to a mongos from the shell and not to a mongod.',
+      ShellApiErrors.NoSharding
```
aherlihy, line 166: It has to be connected to mongos and sharded, so I think better this error:
```suggestion
      ShellApiErrors.NotConnectedToMongos
```

packages/shell-api/src/mongo.ts
```diff
@@ -174,7 +176,7 @@ export default class Mongo extends ShellApiClass {
       const rc = this._serviceProvider.getReadConcern();
       return rc ? rc.level : undefined;
     } catch {
-      throw new MongoshInternalError('Error retrieving ReadConcern. Please file a JIRA ticket in the MONGOSH project');
+      throw new MongoshRuntimeError('Error retrieving ReadConcern.', CommonErrors.CommandFailed);
```
aherlihy, line 179: This one isn't a runtime error, this is an internal error since this should never happen. If it does I think we have a drivers error.
```suggestion
      throw new MongoshInternalError('Error retrieving ReadConcern.', CommonErrors.CommandFailed);
```

aherlihy (approved):
Awesome job!! This was a heroic effort.

packages/shell-api/src/shell-internal-state.ts
rose-m: Ah I see - since it's an assignment. I switched it to `InvalidOperation`.

PR #445, opened 2020-11-26
[java-shell] fix findOne command. Wrap resulting doc in JS object so properties can be defined on it

kornilova203:
`findOne` returns single document as a result.
Mongosh core will try to defineProperty on it and fail if value is not wrapped in JS object.


addaleax (approved):
This seems good, but I’m surprised that this was an issue – what kind of object did we previously attempt to use `Object.defineProperty()` on? That might be good to know so we can avoid this problem in the future on our side

kornilova203:
@addaleax 
`JavaServiceProvider` always returns JS objects so you can access their properties and define new ones. `Object.defineProperty` works well on them.
These objects are converted to Java only when returning them to user.

`Cursor.next` is an exception, it returns Java object because it would be too slow to convert collections back and forth.
[`findOne`](https://github.com/mongodb-js/mongosh/blob/master/packages/shell-api/src/collection.ts#L432) method returns this Java object from `Cursor.next` and fails to define property on it.

I see it as GraalVM interop flaw, ideally JS and Java object should be indistinguishable.
Core should not care about these flaws when there is a simple workaround that doesn't affect performance.

In this PR I wrap object in `Cursor.next` only if limit size is 1, so there are no performance issues.

addaleax:
Okay, that makes sense – thanks for explaining! And let us know if there’s anything we can do to catch this earlier in the future.

PR #451, opened 2020-11-26
[Do not merge] Upgrade driver to 4.0

aherlihy:


packages/java-shell/src/main/kotlin/com/mongodb/mongosh/service/Cursor.kt
```diff
@@ -10,7 +10,9 @@ import org.graalvm.polyglot.Value
 
 internal class Cursor(private var helper: BaseMongoIterableHelper<*>, private val converter: MongoShellConverter, private val wrapper: ValueWrapper) : ServiceProviderCursor {
     private var iterator: MongoCursor<out Any?>? = null
-    private var closed = false
+
+    @HostAccess.Export
+    override var closed = false
```
kornilova203, line 15: By default Kotlin will generate a private field with getter and setter, so you need `@JvmField` annotation to force it to be a public field.
But this annotation is not applicable to overridden properties, so you'll need to remove the property from `ServiceProviderCursor` inteface.
```suggestion
    @JvmField
    @HostAccess.Export
    override var closed = false
```

kornilova203 (changes requested):

packages/shell-api/src/collection.ts
```diff
@@ -1573,4 +1568,15 @@ export default class Collection extends ShellApiClass {
     result.Totals = totalValue;
     return new CommandResult('StatsResult', result);
   }
+
+  @serverVersions(['3.1.0', ServerVersions.latest])
+  watch(pipeline: Document[] = [], options: ChangeStreamOptions = {}): ChangeStreamCursor {
+    this._emitCollectionApiCall('watch', { pipeline, options });
+    const cursor = new ChangeStreamCursor(
+      this._mongo._serviceProvider.watch(pipeline, options, {}, this._database._name, this._name),
+      this[asPrintable]()
```
addaleax, line 1577:
```suggestion
      this._name
```

to avoid mixing presentation with functionality?

packages/shell-api/src/database.ts
```diff
@@ -1158,4 +1149,15 @@ export default class Database extends ShellApiClass {
   async printSlaveReplicationInfo(): Promise<CommandResult> {
     throw new MongoshInvalidInputError('Method deprecated, use db.printSecondaryReplicationInfo instead');
   }
+
+  @serverVersions(['3.1.0', ServerVersions.latest])
+  watch(pipeline: Document[] = [], options: ChangeStreamOptions = {}): ChangeStreamCursor {
+    this._emitDatabaseApiCall('watch', { pipeline, options });
+    const cursor = new ChangeStreamCursor(
+      this._mongo._serviceProvider.watch(pipeline, options, {}, this._name),
+      this[asPrintable]()
```
addaleax, line 1158: ditto here, 
```suggestion
      this._name
```

?

(Fwiw, it stumped me a bit that we print Databases and Collections as if they were plain strings when I first started using mongosh … I don’t necessarily have a better suggestion, though :))

addaleax (commented):
This looks good to me :) I went through everything and left a few comments, but I know that this is a ton of work and you can feel free to ignore them if that makes your life easier

packages/i18n/src/locales/en_US.ts
```diff
@@ -591,6 +599,41 @@ const translations: Catalog = {
...
+              description: 'Returns the ResumeToken of the change stream'
+            },
+            hasNext: {
+              description: 'WARNING: on change streams this method will block unless the cursor is closed. Use tryNext to check if there are any documents in the batch. This is a breaking change'
+            },
+            next: {
+              description: 'WARNING: on change streams this method will block unless the cursor is closed. Use tryNext to get the next document in the batch. This is a breaking change'
+            },
+            tryNext: {
+              description: 'If there is a document in the change stream, it will be returned. Otherwise returns null.'
```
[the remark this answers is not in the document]
aherlihy: When you call `tryNext` it iterates the cursor, so you could implement `hasNext` in a way that would act the same but the side effect is that you can no longer get the document out at all, since if you then call `next` the cursor won't have the doc anymore that was returned by the first `tryNext`

packages/shell-api/src/collection.ts
aherlihy: Sure!

packages/shell-api/src/database.ts
aherlihy: Yeah it's a bit odd, we can revisit in the shell 2.0 :D

packages/i18n/src/locales/en_US.ts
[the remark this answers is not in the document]
addaleax: We could cache the document and return it from `next`/hide that as an implementation detail if we wanted to though, right? Not saynig we should, just wondering if that’s a possibility

packages/browser-repl/src/components/utils/inspect.spec.ts
```diff
@@ -45,7 +45,7 @@ describe('inspect', () => {
     });
   });
 
-  context('with BSON types', () => {
+  context.skip('with BSON types', () => { // TODO: fix BSON types
```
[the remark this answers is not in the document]
aherlihy: Good idea

packages/i18n/src/locales/en_US.ts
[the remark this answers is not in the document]
aherlihy: That's a good point. I think because there are so many other backwards breaking changes to the tailable cursors (for example, isExhausted no longer makes sense since the driver does not allow you call tryNext after the cursor has closed) that I think maybe a more dramatic break would be better so people don't end up making subtle mistakes. I asked Max this morning and he linked me to this comment by dan: https://docs.google.com/document/d/1tGB65zRhH3CKSlzg8YIpzCtShOpY8sWDDsgZJZL4wkI/edit?disco=AAAAHXcXDJ0. So I don't think backwards-breaking changes to this API are a big problem as long as we document well

packages/service-provider-core/src/printable-bson.ts
```diff
@@ -1,8 +1,8 @@
-import * as BSON from 'bson';
+import { bson as BSON } from './index';
 const inspectCustom = Symbol.for('nodejs.util.inspect.custom');
 
 export const bsonStringifiers: Record<string, (this: any) => string> = {
-  ObjectID: function(): string {
+  ObjectId: function(): string {
```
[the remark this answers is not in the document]
aherlihy: Nice, will remove that line too

packages/shell-api/src/collection.ts
```diff
@@ -1561,6 +1561,7 @@ export default class Collection extends ShellApiClass {
         totalValue[`Shard ${shardStats.shardId}`] = [
           `${estDataPercent} % data`,
           `${estDocPercent} % docs in cluster`,
+          `${estDocPercent} % docs in cluster`,
```
addaleax, line 1564: Is this intentional?

aherlihy: Nope!! Time for me to stop working today :)

packages/service-provider-server/package.json
```diff
@@ -39,12 +39,10 @@
   "dependencies": {
     "@mongosh/errors": "^0.6.1",
     "@mongosh/service-provider-core": "^0.6.1",
+    "@types/bl": "^2.1.0",
```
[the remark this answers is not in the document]
aherlihy: Will do!

PR #452, opened 2020-11-26
feat(cli-repl): upgrade Node.js to v14.x MONGOSH-438

addaleax:


packages/cli-repl/src/cli-repl.ts
```diff
@@ -142,6 +142,9 @@ class CliRepl {
   }
 
   verifyNodeVersion(): void {
+    if (process.env.MONGOSH_SKIP_VERSION_CHECK === 'i-will-not-complain-if-mongosh-does-not-work') {
```
mcasimir, line 145: I mean i guess `"true"` or `"1"`  here is ok, after all it would be even half-ok as a feature.

addaleax: I personally would be fine with that as well, but I also didn’t want to ignore the concerns raised in the planning call about this :) Happy to switch this to something else if you like.

.evergreen.yml
```diff
@@ -220,104 +229,229 @@ tasks:
...
+          mongosh_skip_version_check: "i-will-not-complain-if-mongosh-does-not-work"
+  - name: test_m40x_n12_win
+    commands:
+      - func: checkout
+      - func: install_win
+        vars:
+          node_js_version: "12.20.0"
+      - func: test_win
+        vars:
+          mongosh_server_test_version: "4.0.x"
```
mcasimir, line 287: Isn't it clearer to use all caps for environment vars?

packages/cli-repl/src/cli-repl.ts
mcasimir: We could be more specific and add node to the name, say: `MONGOSH_SKIP_NODE_VERSION_CHECK`

.evergreen.yml
addaleax: So… these are not environment variables per se, but evergreen expansions, which are usually all lowercase, and it’s only the import-expansions script that turns them to uppercase

So I’d probably like to be consistent between `node_js_version` and `mongosh_server_test_version`, but I guess if we wanted to we could turn them all uppercase

mcasimir (approved):
👍  Cool! added only few comments

.evergreen.yml
mcasimir: Oh no then is great, so are just interpolated strings, makes sense

packages/cli-repl/src/cli-repl.ts
addaleax: Makes sense, renamed the variables :slightly_smiling_face:

mcasimir: > concerns raised in the planning call about this
I don't remind it, .. and maybe it was even me :)

Buh i really don't see a reason to be concerned here, it seems totally fine. If a user gets here and then sets  `MONGOSH_SKIP_NODE_VERSION_CHECK` to true they may probably know what are they going after.

We do specify a supported version range. If we want to really be on the safe side I guess we can add a comment specifying that this is an internal variable used for testing, but honestly i'd just ignore it.

For what i recall we introduced the version check more for our own safety, to make sure we don't  accidentally build with the wrong version of node, since it happened in the past.

addaleax: I think @mmarcon had some stronger concerns here… but yeah, I also don’t care much personally

mcasimir: buh let's wait and see what he says :)

mmarcon: I think it was actually @aherlihy who brought it up as a concern.

I think that as long as that is undocumented outside of our team (and I mean really undocumented so not even support knows about it) if someone is advanced enough to read the code and figure out that they can use that variable they are also probably smart enough to understand that we do this runtime check for a reason.

Worst case, if we see that we start seeing bug reports related to this we can always change the way we do that check.

mcasimir: Yeah let's use a normal value. I want to make it clear again though, this line:

```
MONGOSH_SKIP_NODE_VERSION_CHECK=1 mongosh
```

by a user will not have any effect, is not like they can change the version of node that is in the binary or the one linked in the brew tap. Is just doing nothing, so is essentially harmless.

addaleax: Okay, switched this to be a plain `if (process.env.MONGOSH_SKIP_NODE_VERSION_CHECK)` check :+1:

PR #456, opened 2020-12-02
fix(cli-repl): fix piping into mongosh (really) MONGOSH-401

addaleax:
Handle the `'end'` event correctly by adding it to the queue of pending
characters if appropriate. Otherwise, the stream would have ended too
early, with potentially pending characters still in the `charQueue`.


aherlihy (approved):

rose-m (approved):
I hope I remember those tiny things when I'll have to implement something like that at some point :D 👍

PR #457, opened 2020-12-02
chore(test): test against latest server version MONGOSH-470

addaleax:


packages/shell-api/src/replica-set.spec.ts
```diff
@@ -710,6 +710,9 @@ describe('ReplicaSet', () => {
     });
 
     describe('add member', () => {
+      // TODO: Fix these tests? They are currently failing with
```
aherlihy, line 713: Oh interesting, just with 4.4? We could make the tests wait and retry.

addaleax: I think with 4.4 they still work … I’m not sure what the change in behavior is coming from, maybe we could just open a ticket about this after this is merged?

aherlihy (approved):

packages/shell-api/src/replica-set.spec.ts
aherlihy: SGTM

PR #459, opened 2020-12-03
chore: remove printable bson package

aherlihy:
Now that the bson lib has it's own util inspect methods, we can get rid of our monkey patching. Unfortunately, we still need the bson package to be imported at the service-provider implementation package (i.e. service-provider-server) and stored on the ServiceProvider instance because we need the package to be modified by shell-api so we can have `Help` and other metadata on all BSON types, including ones returned from the server.


addaleax (approved):

packages/browser-repl/src/components/utils/inspect.spec.ts
```diff
@@ -69,15 +69,6 @@ describe('inspect', () => {
         inspect({ p: new bson.Binary('abcdefghiklmnopq', 4) })
       ).to.equal('{ p: UUID("61626364-6566-6768-696b-6c6d6e6f7071") }');
     });
-
-    it('does not require BSON types to be instances of the current bson library', () => {
```
aherlihy, line 73: I'm not sure I understand the purpose of this test @mcasimir?

addaleax: I’m not sure if I added this test because I needed to, but the idea was that we can also inspect bson objects that come from e.g. Compass itself, which would use `bson@1.x`

If this does not affect us, then great, but we may want to check that this does not regress COMPASS-4465 (in Compass itself)?

addaleax (approved):

packages/browser-repl/src/components/utils/inspect.spec.ts
aherlihy: Ah. So it seems we can't get rid of printable-bson it seems at all, at least until Compass upgrades the driver. Will revert that too

PR #461, opened 2020-12-03
chore(service-provider-core): remove extra methods MONGOSH-451

addaleax:
There are currently a few service provider methods (buildInfo,
getCmdLineOpts, dropIndexes, reIndex, convertToCapped)
that just do runCommand so those should not be in the SP API.


aherlihy (approved):
Nice!

kornilova203 (approved):

addaleax:
So … the java-shell tests were failing on mongod 4.0.x, because `dropIndexes` does not support array arguments there. I’ve done what the java-shell previously did here, namely to simply iterate over the arguments, is that okay for us?

PR #462, opened 2020-12-03
chore(shell-api): refactor and test toIterator

addaleax:
- Make all cursor types async iterables (this is just for us,
  the async rewriter does not support async iteration at this point)
- Use the same iteration in `toIterator` for all input types
- Add tests for `toIterator`
- Add an e2e test to cover this feature


aherlihy (approved):

aherlihy:
Awesome!

PR #471, opened 2020-12-07
chore(shell-api): clean up options passing and types MONGOSH-422

addaleax:


packages/shell-api/src/collection.spec.ts
```diff
@@ -265,6 +265,24 @@ describe('Collection', () => {
...
+
+        await collection.bulkWrite(requests, {
+          writeConcern: { w: 'majority' }
+        });
+
+        expect(serviceProvider.bulkWrite).to.have.been.calledWith(
+          'db1',
+          'coll1',
+          requests,
+          { writeConcern: { w: 'majority' } },
```
aherlihy, line 281: Do we want to update the code so that we don't send writeConcern twice? We only really need to send it as a command option, not a database option I think

aherlihy (approved):

packages/shell-api/src/collection.spec.ts
addaleax: But then why do we have code that explicitly adds it to the db options? Like, I’m happy to remove code instead of adding tests if that also works here :)

aherlihy: Not sure why that was added but I don't think we still need it, since it seems like all the options in the new driver take in subclasses of `CommandOperationOptions`, which extends `WriteConcernOptions`: https://github.com/mongodb/node-mongodb-native/blob/master/src/operations/command.ts#L18-L20 so it should be enough to just pass in the command options instead of the db options. My guess why it was that way before is that the old driver either didn't document or did not have a way to pass writeConcern via command options

addaleax: Ok, that makes sense to me – maybe we can also use somewhat stricter typescript here to avoid any future confusion, I’ll check and update this accordingly

aherlihy: The test is definitely worth keeping btw, just the extra writeConcern in the arguments caught my eye and made me realize maybe we're doing something unnecessary here :)

packages/shell-api/src/database.ts
```diff
@@ -264,8 +272,8 @@ export default class Database extends ShellApiClass {
   getSiblingDB(db: string): Database {
     assertArgsDefined(db);
     this._emitDatabaseApiCall('getSiblingDB', { db });
-    if (this._baseOptions.session) {
-      return this._baseOptions.session.getDatabase(db);
+    if (this._session) {
```
aherlihy, line 275: I don't know if there will be any other db-wide options we need to send to every command in the future, but you're right it's better to be more specific now and then just update if we ever need it

packages/shell-api/src/session.spec.ts
```diff
@@ -72,6 +78,11 @@ describe('Session', () => {
         expect(db).to.deep.equal(new Database(mongo, 'test', session));
         expect(session.getDatabase('test')).to.equal(db); // reuses db
       });
+      it('also affects Database.getSiblingDB', () => {
```
aherlihy, line 81: Nice catch!

aherlihy (approved):
This is much nicer than what we had in master!

packages/shell-api/src/database.ts
addaleax: Ultimately, I was just going for “something that documents that `_baseOptions` actually might contain” here, because I was a bit confused by that in the past :)

PR #474, opened 2020-12-08
Update node driver to latest master

aherlihy:
Includes fixes for:
- some out of date TODOs were removed
- NODE-2923 - Bulk operation.execute returns undefined for insertedIds
- NODE-2937 - package.json missing when bundled
- NODE-2913 - Cursor max/min
- NODE-2921 - Auth needs password not pass
- NODE-2918 - remove owner in DriverSessionOptions
- NODE-2936 - CRUD spec conformance
- Implements withReadPref and withReadConcern on cursors
- NODE-2932 means we can remove the bulk typings in service-provider-core

- NODE-2912, NODE-2920, NODE-2926, NODE-2927, NODE-2757 essentially went away

The withReadPref, withReadConcern, and CRUD spec compliance Node driver changes required some minimal changes to be made to the java service provider.


packages/service-provider-server/src/cli-service-provider.ts
```diff
@@ -888,7 +888,7 @@ class CliServiceProvider extends ServiceProviderCore implements ServiceProvider
    * @returns {Promise} topology.
    */
   getTopology(): Topology | undefined {
-    return this.mongoClient.topology;
+    return this.mongoClient.topology; // TODO: maybe return topology description instead, see NODE-2910
```
addaleax, line 891: I think I’m missing something re: this comment/the corresponding one in `service-provider-core` – Right now, we can do `serviceProvider.getTopology()?.description` to access the description, and that’s publicly accessible without hitting any private properties. What do we need from the driver here?

addaleax (approved):

packages/service-provider-server/src/cli-service-provider.ts
aherlihy: From our conversation last week they said that topology description was something they'd consider making public directly in the future, https://jira.mongodb.org/browse/NODE-2910, but if that's gone away bceause it's already public we can should the ticket.

aherlihy: I think the difference between `getTopologyDescription` and just `topology.description` is minor enough we should just close the ticket, save everyone some time

addaleax: Yeah, I’ve closed the ticket. Maybe they were referring to making only the topology description available, rather than the public `Toplogy` class + topology description? But I think in that case we should also be fine.

aherlihy: Yeah, either way the comment should be gone :)

PR #476, opened 2020-12-08
fix(service-provider-core): Default to direct connection to specified node MONGOSH-465

rose-m:
## Description
This fixes the breaking behavioral change when connecting to a specific node. The old shell directly connected to a node as specified whereas we so far tried to establish a connection to e.g. the primary node of a replica set.

The default behavior now is to add the `directConnection=true` query parameter to the connection string automatically _except_:
* the `replicaSet` query parameter is present in the connection string
* the connection string uses the `mongodb+srv://` scheme
* the connection string contains a seed list with multiple hosts

### DIFF
--- a/packages/cli-repl/test/e2e-direct.spec.ts
+++ b/packages/cli-repl/test/e2e-direct.spec.ts
@@ -1,0 +1,99 @@
+import { MongodSetup, startTestCluster } from '../../../testing/integration-testing-hooks';
+import { eventually } from './helpers';
+import { TestShell } from './test-shell';
+
+describe('e2e direct connection', () => {
+  afterEach(async() => await TestShell.killall());
+
+  context('to a replica set', async() => {
+    const replSetId = 'replset';
+    const [rs0, rs1, rs2] = startTestCluster(
+      ['--single', '--replSet', replSetId],
+      ['--single', '--replSet', replSetId],
+      ['--single', '--replSet', replSetId]
+    );
+
+    [ [rs0, 'rs0'], [rs1, 'rs1'], [rs2, 'rs2'] ].forEach((elem) => {
+      const server = elem[0] as MongodSetup;
+      const node = elem[1] as string;
+      it(`works when connecting to node ${node} of uninitialized set`, async() => {
+        const shell = TestShell.start({ args: [await server.connectionString()] });
+        await shell.waitForPrompt();
+        await shell.executeLine('db.isMaster()');
+        shell.assertContainsOutput('ismaster: false');
+        shell.assertNotContainsOutput(`setName: '${replSetId}'`);
+      });
+    });
+
+    it('allows to initialize the replica set', async() => {
+      const replSetConfig = {
+        _id: replSetId,
+        version: 1,
+        members: [
+          { _id: 0, host: await rs0.hostport(), priority: 1 },
+          { _id: 1, host: await rs1.hostport(), priority: 0 },
+          { _id: 2, host: await rs2.hostport(), priority: 0 },
+        ]
+      };
+
+      const shell = TestShell.start({ args: [await rs0.connectionString()] });
+      await shell.waitForPrompt();
+      await shell.executeLine(`rs.initiate(${JSON.stringify(replSetConfig)})`);
+      shell.assertContainsOutput('ok: 1');
+      await eventually(async() => {
+        await shell.executeLine('db.isMaster()');
+        shell.assertContainsOutput('ismaster: true');
+        shell.assertContainsOutput(`me: '${await rs0.hostport()}'`);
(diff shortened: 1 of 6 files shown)


packages/cli-repl/test/e2e-direct.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+    const replSetId = 'replset';
+    const [rs0, rs1, rs2] = startTestCluster(
+      ['--single', '--replSet', replSetId],
+      ['--single', '--replSet', replSetId],
+      ['--single', '--replSet', replSetId]
+    );
+
+    [ [rs0, 'rs0'], [rs1, 'rs1'], [rs2, 'rs2'] ].forEach((elem) => {
+      const server = elem[0] as MongodSetup;
+      const node = elem[1] as string;
```
addaleax, line 18: With something like

```suggestion
    for (const { server, node } of [
      { server: rs0, node: 'rs0' },
      { server: rs1, node: 'rs1' },
      { server: rs2, node: 'rs2' }
    ]) {
```

you should be able to avoid the type casting here :)

addaleax (commented):
Looks good so far :)

packages/cli-repl/test/e2e-direct.spec.ts
rose-m: (unfortunately I got a TypeScript error here so explicitly added it - will check again :/ )

rose-m: Yup works when making it an object 👍

addaleax (approved):
Nice :)

packages/service-provider-core/src/uri-generator.ts
```diff
@@ -142,7 +147,61 @@ function generateUri(options: CliOptions): string {
...
+
+/**
+ * Parses a given mongodb:// connection string and adds the `directConnection=true` query parameter if required.
+ * See: https://github.com/mongodb/specifications/blob/master/source/connection-string/connection-string-spec.rst#reference-implementation
+ * @param uri mongodb:// connection string
+ */
+function addDirectConnectionQueryParameterToMongoUriIfRequired(uri: string): string {
+  const uriNoScheme = uri.substr(Scheme.Mongo.length);
+
+  // Split URI at first "/"
```
addaleax, line 161: Btw, I think it should be fine to use `new URL(uri)` here for parsing and modification. I’m not sure if that makes things easier or not, but it might be nice to have an more URL-parser-y thing here?

packages/service-provider-core/src/uri-generator.ts
```diff
@@ -142,7 +147,61 @@ function generateUri(options: CliOptions): string {
...
+  if (hostInfo.indexOf(',') > -1) {
+    // multiple hosts, i.e. a seed list is present -> return original uri
+    return uri;
+  }
+
+  if (authDbAndOptions) {
+    // Check if a replicaSet or directConnection parameter is already present
+    return `mongodb://${userAndHostInfo}` + getDbAndQueryStringWithDirectConnectionIfRequired(authDbAndOptions);
+  }
+  return `${uri.endsWith('/') ? uri : uri + '/'}?directConnection=true`;
```
addaleax, line 183: What happens if `directConnection=false` is explicitly specified?

packages/service-provider-core/src/uri-generator.ts
rose-m: I'll give it a try! I didn't bother using it since the connection string spec allows for multiple hosts which would be an invalid URI... but worth trying if I would just get all in the host field.

packages/service-provider-core/src/uri-generator.ts
rose-m: Then `authDbAndOptions` will be set and you'll return in line 181 which calls the `getDbAndQueryStringWithDirectConnectionIfRequired(...)` method which should take care of that. See this test case: https://github.com/mongodb-js/mongosh/pull/476/files/ee6dcba4f152178c047372ea42af0757fb07264f#diff-779e8029eb04be4763e9ba6da3c47a259c239920374b91ad80bb514db98987a2R100

packages/service-provider-core/src/uri-generator.ts
rose-m: Okay I've tried it with `new URL(uri)` - the problem indeed are the multiple hosts which will throw an exception:
```
TypeError [ERR_INVALID_URL]: Invalid URL: mongodb://192.42.42.42:27017,192.0.0.1:27018/db?readPreference=primary
```

aherlihy (approved):
Great job! I would make sure to mark the associated JIRA ticket with 'doc changes needed' and communicate to the docs team how the default behavior for the shell has changed. We should make sure users can find that we've added the default direct connection param if they go looking.

PR #485, opened 2020-12-10
chore(service-provider-server) remove returnNonCachedInstance option

aherlihy:
See https://jira.mongodb.org/browse/NODE-2931

### DIFF
--- a/packages/service-provider-server/src/cli-service-provider.ts
+++ b/packages/service-provider-server/src/cli-service-provider.ts
@@ -262,30 +262,13 @@ class CliServiceProvider extends ServiceProviderCore implements ServiceProvider 
    * @returns {Db} The database.
    */
   private db(name: string, dbOptions: DbOptions = {}): Db {
-    // Usually, we'd let the driver do the caching of the Database objects, but
-    // due to our particular needs, we need to cache DBs both by name *and* by
-    // options (so that subsequent calls to this method have their db options
-    // respected, instead of receiving the instance that was cached only by name
-    // and might have been configured with different options, e.g. read/write
-    // concern, read preference, etc.).
-    // We still need caching, though, since otherwise every call to
-    // this method creates a new driver Database object that registers itself
-    // and effectively cannot be garbage collected on its own, which would
-    // lead to memory leaks. We thus maintain our own cache.
     const key = `${name}-${JSON.stringify(dbOptions)}`;
     const dbcache = this.getDBCache();
     const cached = dbcache.get(key);
     if (cached) {
       return cached;
     }
-    const optionsWithForceNewInstace: DbOptions = {
-      ...dbOptions,
-
-      // Ensure that we actually get a fresh database instance from the driver.
-      returnNonCachedInstance: true
-    } as any; // TODO: returnNonCachedInstance supported? see NODE-2931
-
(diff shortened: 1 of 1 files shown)


packages/service-provider-server/src/cli-service-provider.integration.spec.ts
```diff
@@ -661,26 +661,6 @@ describe('CliServiceProvider [integration]', function() {
...
-      const db2 = serviceProvider._dbTestWrapper('foo', { readPreference: { mode: 'secondary' } } as DbOptions);
-      expect(db1).not.to.equal(db2);
-    });
-
-    it('returns the different db instances when used with different options', () => {
-      const db1 = serviceProvider._dbTestWrapper('foo', { readPreference: { mode: 'primary' } } as DbOptions);
-      const db2 = serviceProvider._dbTestWrapper('foo', { readPreference: { mode: 'secondary' } } as DbOptions);
-      expect(db1).not.to.equal(db2);
-    });
-  });
```
addaleax, line 682: I can see that they were failing in CI, but that surprises me, because these are not supposed to be affected by the driver’s caching at all and instead are supposed to test our *own* caching, and that the tests failed because `db()` returned the same `Db` instance for different options does seem a bit concerning to me … do you know more about this?

addaleax: It looks like for now the caching feature is still alive and well in the driver’s master branch: https://github.com/mongodb/node-mongodb-native/blob/cb9ee9e6175a6654c3c300801884e4a3c3a653ac/src/mongo_client.ts#L404

packages/service-provider-server/src/cli-service-provider.ts
```diff
@@ -256,30 +256,13 @@ class CliServiceProvider extends ServiceProviderCore implements ServiceProvider
...
     const cached = dbcache.get(key);
     if (cached) {
       return cached;
     }
-    const optionsWithForceNewInstace: DbOptions = {
-      ...dbOptions,
-
-      // Ensure that we actually get a fresh database instance from the driver.
-      returnNonCachedInstance: true
-    } as any; // TODO: returnNonCachedInstance supported? see NODE-2931
```
addaleax, line 280: I think for now what we could probably do is to remove the `as any` along with `: DbOptions`? That should let typescript figure out that this type is “`DbOptions` + this one other thing that is not mentioned there but that the `.db()` method still accepts”

packages/service-provider-server/src/cli-service-provider.integration.spec.ts
aherlihy: Ah, it seems this is in code review https://jira.mongodb.org/browse/NODE-2931 so not quite in master yet, so that was a little presumtious of me to make this PR

packages/service-provider-server/src/cli-service-provider.ts
aherlihy: I think we can just put this PR on hold until it's merged into node master, I didn't realize it hadn't been already done. Then we can revert some of the tests so that it makes sure the behavior's still happening

aherlihy: Until this is merged: https://github.com/mongodb/node-mongodb-native/pull/2672

addaleax (approved):

PR #486, opened 2020-12-10
[java-shell] support cursor.readConcern

kornilova203:


addaleax (approved):

addaleax:
Looks like CI failed in a relevant way:

```
 [2020/12/10 13:54:08.811] com.mongodb.mongosh.CursorTest > testReadConcern FAILED
 [2020/12/10 13:54:08.811]     org.junit.ComparisonFailure: expected:<...ClassCastException: [class org.bson.Document cannot be cast to class java.lang.String (org.bson.Document is in unnamed module of loader 'app'; java.lang.String is in module java.base of loader 'bootstrap')]> but was:<...ClassCastException: [org.bson.Document cannot be cast to java.lang.String]>
```

Maybe that’s because of a different Java version or something?

kornilova203:
@addaleax, seems so
I've just removed this particular test case

PR #489, opened 2020-12-10
PoC for embedding FLE addon MONGOSH-488

addaleax:


scripts/prep-fle-addon.sh
```diff
@@ -0,0 +1,81 @@
...
+  case `uname` in
+      Darwin*)                          PREBUILT_OSNAME=macos;;
+      CYGWIN*|MINGW32*|MSYS*|MINGW*)    PREBUILT_OSNAME=windows-test;;
+  esac
+fi
+
+if [ x"$PREBUILT_OSNAME" != x"" ]; then
+  # Download and extract prebuilt binaries.
+  curl -L https://s3.amazonaws.com/mciuploads/libmongocrypt/all/master/latest/libmongocrypt-all.tar.gz | \
+    tar --strip-components=1 -xzvf - $PREBUILT_OSNAME
```
addaleax, line 33: This downloads prebuilt binaries from the latest version on the `master` branch of libmongocrypt, as I understand it, and the git commands below also pick up the latest `master`. Are we okay with that? I feel like if we pin specific versions of the dependencies, we’ll forget to update them, and adding some logic along the lines of “pick the most recent release version”, that’s going to come with some extra complexity that I’m not sure is worth it.

aherlihy: I think that's fine, but I would double check with the team that manages it, so I think drivers?

packages/cli-repl/package.json
```diff
@@ -69,5 +69,8 @@
     "@types/pino": "^6.3.3",
     "@types/read": "^0.0.28",
     "@types/text-table": "^0.2.1"
+  },
+  "optionalDependencies": {
+    "mongodb-client-encryption": "^1.1.0"
```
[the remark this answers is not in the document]
aherlihy: I think it probably should be in service-provider-server so that we can test it fully there and within the shell-api package, but if it's any additional complexity we can keep it here for now

scripts/prep-fle-addon.sh
addaleax: I asked this:

> Is the lack of versioning going to be a risk for us in any way? Should we pin to specific versions/tags?

and the answer was:

> You can just pin to a stable version of the c-driver to get the bson library. The flexibility of versions is for drivers that like the c-driver and those that derive from it.

Which I don’t feel tells us a lot, I feel? :smile:

packages/build/src/signable-compiler.ts
```diff
@@ -38,8 +67,15 @@ class SignableCompiler {
         ...process.env,
         // Custom env vars for sccache:
         AWS_ACCESS_KEY_ID: process.env.DEVTOOLS_CI_AWS_KEY,
-        AWS_SECRET_ACCESS_KEY: process.env.DEVTOOLS_CI_AWS_SECRET
-      }
+        AWS_SECRET_ACCESS_KEY: process.env.DEVTOOLS_CI_AWS_SECRET,
+        ...(os.platform() === Platform.MacOs ? {
+          LDFLAGS: '-framework CoreFoundation -framework Security'
```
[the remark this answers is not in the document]
addaleax: Okay, this has been cleared up. The Node.js addon does some things to make itself work that probably go under “technical debt”, but for now I’ve made it build in the same way that the addon does, and added an explanatory comment.

packages/cli-repl/package.json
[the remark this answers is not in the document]
addaleax: @aherlihy Done, moved it and added a dummy test in the shell-api package :slightly_smiling_face:

scripts/prep-fle-addon.sh
aherlihy: hmmm that answer was confusing, I think they possibly misunderstood your question

aherlihy: Ok just read through the thread in #fle, I'm not sure I understand why specifically the bson library is relevant, since we'd care about pinning the entirety of libmongocrypt

rose-m (approved):

PR #496, opened 2020-12-15
chore(test): Add TLS connectivity tests MONGOSH-418

rose-m:


packages/cli-repl/src/arg-mapper.ts
```diff
@@ -11,18 +13,33 @@ const MAPPINGS = {
...
   tlsCertificateKeyFilePassword: 'tlsCertificateKeyFilePassword',
   username: 'auth.username',
-  verbose: [ 'loggerLevel', 'debug' ]
+  verbose: { opt: 'loggerLevel', val: 'debug' }
 };
 
+async function readCrlFileContent(crlFilePath: string): Promise<string | Buffer> {
+  let error: Error | undefined = undefined;
+  try {
+    if ((await fs.stat(crlFilePath)).isFile()) {
```
addaleax, line 32: I think you can skip this, if it’s not a readable file then `fs.readFile()` will throw anyway (there’s also some general context around this type of check being an anti-pattern – it enables [TOCTOU race conditions](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) that can usually be avoided by not checking first, and it’s also the reason why `fs.exists()` is deprecated. But none of that really matters all that much here, I think).

packages/cli-repl/test/fixtures/certificates/README.md
```diff
@@ -0,0 +1,135 @@
...
+   openssl req -new -x509 -key invalid-client.key -out invalid-client.crt -days 99999
+   ```
+   * Organization Name: `MongoDB`
+   * Organizational Unit Name: `DevTools Testers`
+   * Common Name: `Wonderwoman`
+   * E-Mail: `tester@example.com`
+3. Create a bundle with client key and certificate to use for connecting:
+   ```
+   cat invalid-client.crt invalid-client.key > invalid-client.bundle.pem
+   ```
```
addaleax, line 135: Thanks for this documentation file! This is definitely going to be helpful in the future :+1:

addaleax (approved):
This looks really nice! ~~And it doesn’t need to be `WIP` anymore, right?~~ You were faster. :smile:

packages/cli-repl/src/arg-mapper.ts
rose-m: Now that you say it it totally makes sense :D And I think I kind of knew... but never heard of `TOCTOU`, thanks for the link!

testing/integration-testing-hooks.ts
```diff
@@ -1,17 +1,18 @@
...
-import {MongoClient, MongoClientOptions} from 'mongodb';
-import { promisify } from 'util';
 import child_process from 'child_process';
 import crypto from 'crypto';
+import { once } from 'events';
 import { promises as fs } from 'fs';
+import { Test } from 'mocha';
+// Installing @types/mongodb at the lerna root breaks some of the tests...
+// Yikes. Stick with ts-ignore for now.
+// @ts-ignore
```
addaleax, line 8: You can remove this, I think. With the 4.0 driver, we’re not using the `@types` package anymore anyway.

addaleax (approved):

addaleax (approved):
Still looks good :)

addaleax:
Looks like this failed some of the executable compilation tasks:

https://evergreen.mongodb.com/task/mongosh_linux_compile_and_upload_artifact_linux_0ab86916c6a8cfdf9ced2a1ae295a1545fec22ee_20_12_18_14_39_47

PR #499, opened 2020-12-16
chore(deps): bump mongodb to 2c03b782 

addaleax:
See https://github.com/mongodb/node-mongodb-native/compare/f6d9b81f...2c03b782.

:crossed_fingers:


aherlihy:
I think there are two tickets that we requested, https://jira.mongodb.org/browse/NODE-2768 and https://jira.mongodb.org/browse/NODE-2931 in the current master that would have changes for us. The driver now exposes some batch information that we should take advantage of, so we can stop looking inside the internal `s` object in our bulk API implementation. Also they should be no longer caching db objects so we can either just merge this PR after this one is done: https://github.com/mongodb-js/mongosh/pull/485 or we can include those changes here.

addaleax:
@aherlihy Is it good with you if I merge this after triage (CI should be green by then) and then merge #485 and open a separate PR for the bulk stuff?

aherlihy:
Yup!

aherlihy (approved):

PR #500, opened 2020-12-16
chore(shell-api): remove dependency on bulk op driver internals

addaleax:
See https://jira.mongodb.org/browse/NODE-2768.

### DIFF
--- a/packages/shell-api/src/bulk.spec.ts
+++ b/packages/shell-api/src/bulk.spec.ts
@@ -1,7 +1,8 @@
 import { CommonErrors } from '@mongosh/errors';
 import { bson, ServiceProvider } from '@mongosh/service-provider-core';
 import { fail } from 'assert';
-import { expect } from 'chai';
+import chai, { expect } from 'chai';
+import sinonChai from 'sinon-chai';
 import { EventEmitter } from 'events';
 import sinon, { StubbedInstance, stubInterface } from 'ts-sinon';
 import Bulk, { BulkFindOp } from './bulk';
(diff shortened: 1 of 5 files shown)


packages/shell-api/src/bulk.spec.ts
```diff
@@ -114,29 +120,6 @@ describe('Bulk API', () => {
                 nInsertOps: 0, nUpdateOps: 0, nRemoveOps: 0, nBatches: 4
               });
             });
-            it('returns unknown if batches cannot be counted', () => {
-              const bulk2 = new Bulk({} as any, { insert: () => {} } as any, t === 'ordered').insert({}).insert({});
-              expect(bulk2.tojson()).to.deep.equal({
-                nInsertOps: 2, nUpdateOps: 0, nRemoveOps: 0, nBatches: 'unknown'
-              });
-            });
-            it('counts current batches', () => {
```
aherlihy, line 123: This test should be alright to keep, right?

aherlihy (approved):

packages/shell-api/src/bulk.spec.ts
addaleax: Well … it becomes fairly artificial at that point, something like this:

```js
it('counts current batches', () => {
  const batches: any = [{}];
  const bulk2 = new Bulk({} as any, {
    insert: () => {
      if (t === 'unordered') {
        batches.push({});
      }
    },
    batches
  } as any,
  t === 'ordered'
  ).insert({}).insert({});
  expect(bulk2.tojson()).to.deep.equal({
    nInsertOps: 2, nUpdateOps: 0, nRemoveOps: 0, nBatches: t === 'ordered' ? 1 : 3
  });
});
```

and at that point, it doesn’t seem like we’re actually testing the `Bulk` class anymore?

aherlihy: Yeah good point :)

PR #501, opened 2020-12-17
feat(shell-api): support non-cursor explainable methods MONGOSH-185

addaleax:
Add `.explain()` support for non-cursor methods. In the process,
also fix up the options handling for `collection.remove()`, which
previously passed the `justOne` option to the driver instead of
handling it in mongosh (which was ignored by the driver).


aherlihy (approved):
Nice!

PR #505, opened 2020-12-17
fix(cli-repl): wait for output drain before exiting

addaleax:
We’re explicitly calling `process.exit()` in the CLI repl once the
REPL has been closed, stops all asynchronous operations, including
writing to asynchronous stdio streams.

We should at least wait until all currently pending output has been
written to stdout before doing so. This seems fix a problem that
I cannot reproduce locally, but @gribnoysup can. :)


gribnoysup (approved):
Works great! As a proof, here's before the fix applied:

image

And here's after the fix:

image


Notice how in the "after" there is a command output visible and how it's missing in the "before"

PR #506, opened 2020-12-18
chore(test): Add Analytics node connectivity tests MONGOSH-418

rose-m:
Adds connectivity tests for analytics nodes.

The test spawns 4 nodes in a replica set, 3 nodes with voting, 1 passive node (priority: 0 and votes: 0). The passive node also has the `nodeType: 'ANALYTICS'` tag.

### DIFF
--- a/packages/cli-repl/test/e2e-analytics.spec.ts
+++ b/packages/cli-repl/test/e2e-analytics.spec.ts
@@ -1,0 +1,69 @@
+import { expect } from 'chai';
+import { startTestCluster } from '../../../testing/integration-testing-hooks';
+import { eventually } from './helpers';
+import { TestShell } from './test-shell';
+
+describe('e2e Analytics', () => {
+  const replSetName = 'replicaSet';
+  const [rs0, rs1, rs2, rs3] = startTestCluster(
+    [ '--single', '--replSet', replSetName ],
+    [ '--single', '--replSet', replSetName ],
+    [ '--single', '--replSet', replSetName ],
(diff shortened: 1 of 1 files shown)


packages/cli-repl/test/e2e-analytics.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+import { startTestCluster } from '../../../testing/integration-testing-hooks';
+import { eventually } from './helpers';
+import { TestShell } from './test-shell';
+
+describe('e2e Analytics', () => {
+  const replSetName = 'replicaSet';
+  const [rs0, rs1, rs2, rs3] = startTestCluster(
+    [ '--single', '--replSet', replSetName ],
+    [ '--single', '--replSet', replSetName ],
+    [ '--single', '--replSet', replSetName ],
```
addaleax, line 11: Do we need the other secondaries? It’s not a big deal, but it does save some time when running the tests to only spawn the servers that we need

addaleax (approved):

packages/cli-repl/test/e2e-analytics.spec.ts
rose-m: Good point, I'll test that... The first one will just become the primary... it might be nice to leave them in to have "real secondaries" since we also have to specify `readPreference=secondary` for the tags to properly work.

addaleax: Sure, works for me :+1:

PR #507, opened 2020-12-18
chore(deps): bump mongodb to 634ae4f9

addaleax:
See https://github.com/mongodb/node-mongodb-native/compare/2c03b782...634ae4f9.

### DIFF
--- a/packages/shell-api/src/aggregation-cursor.ts
+++ b/packages/shell-api/src/aggregation-cursor.ts
@@ -57,7 +57,7 @@ export default class AggregationCursor extends ShellApiClass {
 
   @returnsPromise
   tryNext(): Promise<Document | null> {
-    return (this._cursor as any).tryNext() as any; // TODO: tryNext private, see NODE-2952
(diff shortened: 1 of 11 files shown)


packages/shell-api/src/aggregation-cursor.ts
```diff
@@ -57,7 +57,7 @@ export default class AggregationCursor extends ShellApiClass {
 
   @returnsPromise
   tryNext(): Promise<Document | null> {
-    return (this._cursor as any).tryNext() as any; // TODO: tryNext private, see NODE-2952
+    return this._cursor.tryNext() as any;
```
aherlihy, line 60: Does thsi still have to be `as any` or will it be `Promise<Document | null>` without the cast?

aherlihy (approved):

packages/shell-api/src/aggregation-cursor.ts
addaleax: Right, good catch :) Dropped the `as any`

PR #511, opened 2020-12-21
chore(test): fix determining server version when no env var + local server present MONGOSH-416

rose-m:
I broke the `compile_and_upload` part in #496 inside docker. This fixes it.


testing/integration-testing-hooks.ts
```diff
@@ -458,7 +461,17 @@ export function skipIfEnvServerVersion(semverCondition: string): void {
...
-      testServerVersion = await getInstalledMongodVersion();
+      try {
+        testServerVersion = await getInstalledMongodVersion();
+      } catch(e) {
+        // no explicitly specified version but also no local mongod installation
+        // -> spawn a server to get the "live" version
+        console.log('Starting server to detect runtime version...');
+        const server = startTestServer('shared');
+        await server.start();
+        skipIfServerVersion(server, semverCondition);
```
mcasimir, line 472: this runs a `before` inside a `before`, how is it going to work?

testing/integration-testing-hooks.ts
```diff
@@ -458,7 +461,17 @@ export function skipIfEnvServerVersion(semverCondition: string): void {
   before(async function() {
     let testServerVersion = process.env.MONGOSH_SERVER_TEST_VERSION;
     if (!testServerVersion) {
-      testServerVersion = await getInstalledMongodVersion();
+      try {
+        testServerVersion = await getInstalledMongodVersion();
+      } catch(e) {
+        // no explicitly specified version but also no local mongod installation
+        // -> spawn a server to get the "live" version
```
mcasimir, line 468: What is the 'live' version here :)?

Starting a mongod server just for the version probably is unnecessary.

My understanding of how the tests are setup here is that mlaunch without an env var would:

1. try to see if a mongod binary is in the path an use it
2. if not download and start the latest

In that case maybe we could do something like this (super-inaccuarte-pseudo-code):

``` js
const ver = await commandExists('mongod')? parseVersion(await exec('mongod --version')) : 9999.9999.9999
```

rose-m: The "live" version is the one we really end up using :D But good point I could default to a super high version number since it is the "latest".

Shutdown is not necessary as it is a "shared" instance and will be re-used and later shut down automatically.

testing/integration-testing-hooks.ts
rose-m: Good question 😅 It did work locally tho - I'd assume since it's still in "before" the other before hook gets registered and executed. With your suggestion below it would be a lot simpler and not required to launch it here.

mcasimir (changes requested):
Had no idea such a corner case existed! Cool that you spotted it. Left some comments maybe we can avoid starting a new server for detecting the version (if i understood the use case correctly .. which i may not :))

testing/integration-testing-hooks.ts
mcasimir: Oh nice!

testing/integration-testing-hooks.ts
mcasimir: Obscure mocha voodoo, well if it works .. :)

testing/integration-testing-hooks.ts
mcasimir: > since it is the "latest"
Oh is always latest? When i run it locally it uses the mongod in the path, which i have as 4.0. 

Anyway if this approach works better then we can also merge that.

rose-m: Yeah _if_ it finds one on the path, mlaunch will use that. Otherwise it downloads `latest`.

But that's unfortunately exactly the corner case: there is nothing on the path and nothing in the env vars. So we "dynamically" end up with what gets downloaded for `latest`. So in the case here it will work even with the hard-coded "9999.9999.9999" version since I'm comparing against a "past version". I'll hard code that version and put it in the comment since needing this function is already an edge case by itself 😅 
We only need this since the server parameters changed from one version to the other inside our testing matrix...

mcasimir: I see ok!

mcasimir: Ping me if you need another review, going to pre-approve it so you can merge if i'm not around

mcasimir (approved):

PR #512, opened 2020-12-21
chore(cli-repl): replace minimist with yargs-parser

rose-m:
As stumbled upon during MONGOSH-418 (see #496), the `minimist` arguments parser will explicitly set default values for boolean flags even when the flags are not provided (see also https://github.com/substack/minimist/issues/94). This makes it impossible to distinguish between a default value and a value specified by the user.

Passing on these flags automatically interferes with parsing the query parameters of the connection string. Explicit options passed to `MongoClient` will always override connection string query parameters.

This PR replaces `minimist` with `yargs-parser`. `yargs-parser` does not set default values for flags that are not specified. All our flags defaulted to `false` with `minimist`. Their value now still qualifies as _falsly_ and the migration thus should have no user visible impact apart from allowing query parameters to take effect. To prove the expected result, the so-far failing (and therefore skipped) TLS connectivity tests are now activated.

### DIFF
--- a/packages/build/src/generate-input.ts
+++ b/packages/build/src/generate-input.ts
@@ -37,11 +37,23 @@ async function generateInput(input: string, execInput: string, analyticsConfig: 
   // of 'package.json' instead of the 'main' key. This works fine for most of
   // our direct code, because TypeScript handles the discrepancy for us,
   // but it's causing trouble for nested dependencies like is-recoverable-error.
+  // On the other hand, it also cannot deal with '.cjs' files and would place them
+  // in a separate "bundle" which won't work, e.g. for yargs-parser. So we need to
+  // make sure that we take the '.js' exported main or module.
   const originalGPE = (bundler as any).resolver.getPackageEntries;
   (bundler as any).resolver.getPackageEntries = function(packageJsonContent: any): any {
+    const mainField = packageJsonContent.main;
     const moduleField = packageJsonContent.module;
+    if (mainField?.endsWith('.cjs')) {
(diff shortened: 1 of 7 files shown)


mcasimir (approved):
👍

packages/build/src/generate-input.ts
```diff
@@ -37,11 +37,23 @@ async function generateInput(input: string, execInput: string, analyticsConfig:
...
+  // On the other hand, it also cannot deal with '.cjs' files and would place them
+  // in a separate "bundle" which won't work, e.g. for yargs-parser. So we need to
+  // make sure that we take the '.js' exported main or module.
   const originalGPE = (bundler as any).resolver.getPackageEntries;
   (bundler as any).resolver.getPackageEntries = function(packageJsonContent: any): any {
+    const mainField = packageJsonContent.main;
     const moduleField = packageJsonContent.module;
+    if (mainField?.endsWith('.cjs')) {
+      if (!moduleField || !moduleField.endsWith('.js')) {
+        console.warn('mongosh: bundling detected potential issue - ' + packageJsonContent.name + ' only has a .cjs main file...');
```
addaleax, line 49: I think we could also just temporarily copy the `.cjs` file to a `.js` file with the same name, parcel should just accept that I think (but as long as this works I think it’s fine, this bit here is already a hack anyway :smile:)

addaleax (approved):
Nice, I think we can just merge this and fix the remaining tests once the driver bug is resolved

PR #513, opened 2020-12-23
chore(ci): move powershell scripts into dedicated files

rose-m:
I noticed due to some other build failures that the scripts we were executing inside powershell do not fail as expected.

Code declared in a `shell.exec` step is piped into the powershell executable via stdin (see https://github.com/evergreen-ci/evergreen/blob/master/agent/command/shell.go#L137). This treats every line completely separate in powershell and powershell will always continue execution. A failing line is disregarded (regardless of `$ErrorActionPreference`).

The scripts are now moved to dedicated `.ps1` files and are started using `subprocess.exec`.

### DIFF
--- a/.evergreen.yml
+++ b/.evergreen.yml
@@ -67,22 +67,15 @@ variables:
         npm run test-e2e-ci
 
   - &compile_and_upload_artifact_win
-    command: shell.exec
+    command: subprocess.exec
     params:
       working_dir: src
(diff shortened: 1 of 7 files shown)


addaleax (approved):

addaleax:
Well … looks like this is uncovering yet another actual bug – I’ll push a fix :)

PR #517, opened 2020-12-28
fix(service-provider-server): fix bsonsize MONGOSH-517

addaleax:
`any` strikes again :)
Make sure that the necessary method used by `bsonsize` is
always available.


aherlihy (approved):
Ooof! Nice one

PR #521, opened 2020-12-29
feat(shell-api): add cls support MONGOSH-239

addaleax:
Add support for the `cls` (clear screen command) through using
`console.clear` in Node.js and using the existing method for the
browser REPL.

As drive-by cleanups:

- Move other global methods to the `ShellApi` class and document
  them accordingly
- Use decorators on the `ShellApi` methods that are exposed as
  bash-like commands instead of special-casing them in the
  shell-evaluator package.

### DIFF
--- a/packages/shell-evaluator/src/shell-evaluator.ts
+++ b/packages/shell-evaluator/src/shell-evaluator.ts
@@ -37,37 +37,29 @@ class ShellEvaluator<EvaluationResultType = ShellResult> {
    * @param {String} filename
    */
   private async innerEval(originalEval: EvaluationFunction, input: string, context: object, filename: string): Promise<any> {
-    const argv = input.trim().replace(/;$/, '').split(' ');
-    const cmd = argv[0];
-    argv.shift();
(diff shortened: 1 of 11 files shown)


packages/shell-evaluator/src/shell-evaluator.ts
```diff
@@ -37,37 +37,29 @@ class ShellEvaluator<EvaluationResultType = ShellResult> {
...
-      case 'exit':
-      case 'quit':
-        return await this.internalState.shellApi.exit();
-      default:
-        this.saveState();
-        const rewrittenInput = this.internalState.asyncWriter.process(input);
+    const { shellApi } = this.internalState;
+    const argv = input.trim().replace(/;$/, '').split(/\s+/g);
+    const cmd = argv.shift() as keyof typeof shellApi;
+    if (shellApi[cmd]?.isDirectShellCommand) {
```
aherlihy, line 43: Much nicer than before!!

aherlihy (approved):

PR #522, opened 2020-12-30
fix(cli-repl): store files in APPDATA on Windows MONGOSH-230

addaleax:
Store config and history in `%APPDATA%` on Windows, and the log
files in `%LOCALAPPDATA%` since roaming them is not meaningful.
The location for other OSes remains `~/.mongodb/mongosh`.


rose-m (approved):
Didn't know that difference between roaming and local... only knew about APPDATA 👍

PR #524, opened 2020-12-30
feat(shell-api): add collation to Bulk API MONGOSH-322

addaleax:


rose-m (approved):
Nice, looks good to me... now need ☕ 😄

PR #525, opened 2020-12-30
feat: Topology aware prompt MONGOSH-196

rose-m:
This adds the topology aware prompt but in such a way that we can easily extend that to support a user-customizable prompt in the same way the old shell did.

**Connecting to an Atlas Cluster:**

image

**Connecting to a local node with replica set, not initialized, then initializing:**

image

**Compass Browser Repl:**

image


packages/cli-repl/src/mongosh-repl.ts
```diff
@@ -228,8 +229,22 @@ class MongoshNodeRepl {
...
-      return await shellEvaluator.customEval(originalEval, input, context, filename);
+      let error: Error | undefined = undefined;
+      let result: any = undefined;
+      try {
+        result = await shellEvaluator.customEval(originalEval, input, context, filename);
+      } catch (e) {
+        error = e;
+      }
+
+      await this.updatePrompt();
```
addaleax, line 242: We could also move this into the `finally`, right?

rose-m: Yep, true... will move it.

packages/shell-api/src/shell-internal-state.ts
```diff
@@ -231,4 +239,94 @@ export default class ShellInternalState {
...
+
+      if (isMaster.setName) {
+        if (isMaster.ismaster) {
+          role = 'PRIMARY';
+        } else if (isMaster.secondary) {
+          role = 'SECONDARY';
+        } else if (isMaster.arbiterOnly) {
+          role = 'ARBITER';
+        } else {
+          role = 'OTHER';
```
addaleax, line 321: Do we want to keep the uppercase names? That has a very 80s feel to it for me … :)

rose-m: :D I'm fine with either casing... would you suggest just having it lowercase?

addaleax: I think either lowercase or capitalized would be a bit nicer, but I don’t have a strong preference anyway :)

aherlihy:
LGTM! Can you remind me if you looked into using the SDAM monitoring to determine the prompt? I remember chatting briefly about it during standup but I don't remember if you did a spike or it was determined to be too much work for the beta. The benefit of using SDAM would be the driver would notify us if there was a topology change, so we would be able to automatically update the prompt to match the topology, and we would not have to potentially run an extra command for every single line of user input.

rose-m:
@aherlihy as we discussed yesterday, we could move the default prompt also in the `ServiceProvider` and then rely on the topology of the driver instead of running additional commands. However, right now their types are not properly exported - or rather only the types and not certain enums - which would make it pretty fragile to implement currently. I'd rather suggest merging what we already have and once the driver fixes their exports go on and move it to the `ServiceProvider`.

Of course we can also just put this on hold until it is resolved.

See: https://mongodb.slack.com/archives/C0V8RU15L/p1609926543248900

packages/browser-repl/src/components/shell-input.tsx
```diff
@@ -106,14 +106,30 @@ export class ShellInput extends Component<ShellInputProps, ShellInputState> {
...
+      />);
+    } else if (this.props.prompt) {
+      const trimmed = this.props.prompt.trim();
+      if (trimmed.endsWith('>')) {
+        prompt = (<>
+          <span>{trimmed.replace(/>$/g, '')}</span>
+          <Icon
+            size={12}
+            glyph={'ChevronRight'}
+          />
```
addaleax, line 122: Nice trick :)

packages/cli-repl/src/mongosh-repl.ts
```diff
@@ -282,10 +282,25 @@ class MongoshNodeRepl implements EvaluationListener {
...
+    } catch (e) {
+      if (e.name === 'MongoNetworkError' || e.name === 'MongoServerSelectionError') {
+        // we track these errors to prevent getting the default prompt which would
+        // otherwise cause the driver to try and reconnect which will timeout after 30s
+        // - right now we don't have any other way of detecting "disconnection"
+        connectionIssues = true;
+      }
+      throw e;
     } finally {
+      if (connectionIssues) {
```
addaleax, line 299: I don’t think we need to care about the prompt for programmatic usage, so something like this might be nice for saving some DB calls when we’re running a script:

```suggestion
      if (connectionIssues || !this.input.isTTY) {
```

packages/cli-repl/test/test-shell.ts
```diff
@@ -113,7 +113,23 @@ export class TestShell {
 
   async waitForPrompt(start = 0): Promise<void> {
     await eventually(() => {
-      if (!this._output.slice(start).match(PROMPT_PATTERN)) {
+      const output = this._output.slice(start);
+      const lines = output.split('\n');
+      const found = !!lines.filter(l => l.match(PROMPT_PATTERN)) // a line that is the prompt must at least match the pattern
+        .find(l => {
+          // in some situations the prompt occurs multiple times in the line (but only in tests!)
+          const prompts = l.trim().replace(/>$/g, '').split('>').map(m => m.trim().replace(/>$/g, ''));
```
addaleax, line 121: There should be no `>` characters left after the `split('>')`, so I think that last `.replace()` can be dropped?

addaleax (approved):

packages/browser-runtime-core/src/open-context-runtime.ts
```diff
@@ -62,4 +64,8 @@ export class OpenContextRuntime implements Runtime {
     this.internalState.setEvaluationListener(listener);
     return prev;
   }
+
+  async getShellPrompt(): Promise<string> {
+    return await this.serviceProvider.getDefaultPrompt();
```
addaleax, line 69: Hm… maybe this could still forward to something like `this.internalState.getShellPrompt()`, which then forwards to the service provider? That way we can avoid storing it here just for this

packages/service-provider-server/src/cli-service-provider.ts
```diff
@@ -213,19 +216,102 @@ class CliServiceProvider extends ServiceProviderCore implements ServiceProvider
...
+    return '';
+  }
+
+  private getTopologySpecificPrompt(): string {
+    const description = this.mongoClient.topology?.description;
+    if (!description) {
+      return '';
+    }
+
+    // TODO: replace with proper TopologyType constants - NODE-2973
```
addaleax, line 260: Tbh, I think using the strings as-is is just fine – I had understood the problem as us having to do some sort of casting, but since TypeScript would complain about invalid values in the `switch` I think we’re good here

addaleax (approved):

addaleax:
Btw, `ShellInternalState` seemed like the right place to implement this… we’re already allowing `getTopology()` to return `undefined`, so there’s no additional requirements to service providers in general, and I don’t think relying on driver internals is something we need to worry about since this is all public. If you want to make this a bit easier for the shell, one thing that could be nice (also for #532), would be processing the topology and server descriptions in the service provider, and then returning another enum value based on that to the shell-api package?

packages/browser-runtime-core/src/open-context-runtime.ts
rose-m: Yeah I cleaned up the remnants of the `internalState` now. All places now go to the `serviceProvider` directly. The forwarding didn't really make sense to me anymore since also the user-specific prompt will be implemented and evaluated in `MongoshNodeRepl` which basically has the `serviceProvider` directly already, too. So it made sense to me to get rid of the unnecessary level of indirection.

packages/service-provider-server/src/cli-service-provider.ts
rose-m: Yeah indeed - in the switch it works but it complained in the tests when I tried to use it typed. I'd leave it in there just to be sure to make it consistent.

packages/browser-runtime-core/src/open-context-runtime.ts
addaleax: I’m just a bit concerned because this now means that this package has to have knowledge about what the service providers look like, rather than it being just something that it passes through to the `ShellInternalState` constructor

rose-m: Ah very valid point... mh... I've no hard feelings on that one. True, the runtimes just forward that to the internal state...

addaleax (approved):

PR #526, opened 2020-12-30
feat: control iteration batch size + add 'Type "it"' prompt MONGOSH-361

addaleax:
Make the iteration batch size for cursors controllable through
`.batchSize()` (the ticket suggested using `DBQuery.shellBatchSize`
for this, like the old shell did, but we’ve deprecated that and
are already explicitly telling users to use `.batchSize()` instead)
and print `Type "it" for more` in the shell when the current cursor
is iterable.

### DIFF
--- a/packages/browser-repl/src/components/types/cursor-iteration-result-output.tsx
+++ b/packages/browser-repl/src/components/types/cursor-iteration-result-output.tsx
@@ -8,7 +8,7 @@ export interface Document {
 }
 
 interface CursorIterationResultOutputProps {
-  value: Document[];
+  value: Document[] & { cursorHasMore: boolean };
 }
 
 export class CursorIterationResultOutput extends Component<CursorIterationResultOutputProps> {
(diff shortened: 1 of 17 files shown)


aherlihy (approved):
LGTM, do we want to also set the driver cursor's batchSize? https://github.com/mongodb/node-mongodb-native/blob/master/src/cursor/abstract_cursor.ts#L55

addaleax:
@aherlihy You mean, do the same for `AggregationCursor` as for `Cursor`? Sure, that sounds good :+1:

addaleax:
Hm, doesn’t look like that’s possible – can you maybe clarify a bit more for me?

aherlihy:
Sorry for the ambiguous comment - I was looking at https://github.com/mongodb/node-mongodb-native/blob/master/src/cursor/abstract_cursor.ts#L507-L519 and thought since batchSize was in AbstractCursor, it would be in all the cursors, but you're right that doesn't seem to be the case. Can ignore my comment!

PR #527, opened 2020-12-30
chore(cli-repl): ensure mongosh works w/o network/fs access MONGOSH-380

addaleax:
- Ensure that the REPL works without being able to write
  its config/log files to the filesystem
- Ensure that it works when the telemetry host is not reachable
- Make sure to flush the telemetry data on exit
  - Refactor `exit`/`quit` so that they work like `.exit` in the
    cli repl, i.e. close down all resources and flush telemetry data


aherlihy:
Sorry for the ambiguous comment - I was looking at https://github.com/mongodb/node-mongodb-native/blob/master/src/cursor/abstract_cursor.ts#L507-L519 and thought since batchSize was in AbstractCursor, it would be in all the cursors, but you're right that doesn't seem to be the case. Can ignore my comment!

mcasimir (approved):

PR #532, opened 2021-01-07
feat(autocomplete): autocomplete collection names MONGOSH-335

addaleax:
Also implicitly address MONGOSH-291 by adding a 200ms timeout after
which cached results are used (which may seem either high or low,
so I’m happy to adjust it – see also the comment about that value).

### DIFF
--- a/packages/shell-api/src/database.ts
+++ b/packages/shell-api/src/database.ts
@@ -53,6 +53,7 @@ export default class Database extends ShellApiClass {
   _collections: Record<string, Collection>;
   _baseOptions: CommandOperationOptions;
   _session: Session | undefined;
+  _cachedCollectionNames: string[] = [];
 
   constructor(mongo: Mongo, name: string, session?: Session) {
     super();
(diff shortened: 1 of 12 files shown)


packages/shell-api/src/database.ts
```diff
@@ -138,6 +139,30 @@ export default class Database extends ShellApiClass {
...
+        return await this._getCollectionNames();
+      })(),
+      (async() => {
+        // 200ms should be a good compromise between giving the server a chance
+        // to reply and responsiveness for human perception. It's not the end
+        // of the world if we end up using the cached results; usually, they
+        // are not going to differ from fresh ones, and even if they do, a
+        // subsequent autocompletion request will almost certainly have at least
+        // the new cached results.
+        await new Promise(resolve => setTimeout(resolve, 200).unref());
```
rose-m, line 160: Nice, the `unref` bit checks the "I learned something new today" 😄

rose-m (approved):
LGTM 😊

packages/shell-api/src/database.ts
geezerP: @rose-m it sure does check the box for me too

PR #535, opened 2021-01-08
chore: publish NPM packages from CI MONGOSH-246

mcasimir:
- Bump and Publish NPM packages from CI
- Use a placeholder version in development
- Remove `publish-npm.js`


packages/types/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mongosh/types",
-  "version": "0.6.1",
+  "version": "0.0.0-dev.0",
   "description": "Types for mongosh internals",
   "author": "Anna Henningsen <anna.henningsen@mongodb.com>",
```
addaleax, line 5: Err, sorry about this, this might have been lerna defaults :)

packages/build/src/release.ts
```diff
@@ -25,6 +27,16 @@ export default async function release(
...
   const githubRepo = new GithubRepo(config.repo, octokit);
+  const commitTag = await githubRepo.getTagByCommitSha(config.revision);
+
+  config = {
+    ...config,
+    version: await getReleaseVersionFromTag(commitTag?.name) || config.version
+  };
+
+  // updates the version of internal packages to reflect the tagged one
+  bumpNpmPackages(config.version);
```
addaleax, line 38: I think we only want to do these things in the `comand === 'publish'` conditional

packages/build/src/npm-packages.ts
```diff
@@ -0,0 +1,65 @@
...
+    '--yes'
+  ], {
+    cwd: PROJECT_ROOT,
+    stdio: 'inherit'
+  });
+}
+
+function listNpmPackages(): {version: string}[] {
+  const lernaListOutput = spawnSync(
+    LERNA_BIN, [
```
rose-m, line 58: Aren't you missing the `'list'` here? 🤔

packages/build/src/npm-packages.ts
```diff
@@ -0,0 +1,65 @@
...
+const LERNA_BIN = path.resolve(PROJECT_ROOT, 'node_modules', '.bin', 'lerna');
+
+export function bumpNpmPackages(version: string): void {
+  if (!version || version === PLACEHOLDER_VERSION) {
+    return;
+  }
+
+  spawnSync(LERNA_BIN, [
+    'version',
+    '--bump', version,
```
rose-m, line 15: I think it's just `lerna version <version>` without the `--bump`.

```
lerna version [bump]

Bump version of packages changed since the last release.

Positionals:
  bump  Increment version(s) by explicit version _or_ semver keyword,
        'major', 'minor', 'patch', 'premajor', 'preminor', 'prepatch', or 'prerelease'.     
```

packages/build/src/npm-packages.ts
```diff
@@ -0,0 +1,65 @@
...
+    throw new Error(`Refusing to publish packages with multiple versions: ${versions}`);
+  }
+
+  if (versions[0] === PLACEHOLDER_VERSION) {
+    throw new Error('Refusing to publish packages with placeholder version');
+  }
+
+  spawnSync(LERNA_BIN, [
+    'publish',
+    '--bump', 'from-package',
```
rose-m, line 43: Same here `from-package` is a positional argument - `--bump` is not a flag 😊

packages/build/src/release.ts
```diff
@@ -25,6 +27,16 @@ export default async function release(
   });
 
   const githubRepo = new GithubRepo(config.repo, octokit);
+  const commitTag = await githubRepo.getTagByCommitSha(config.revision);
```
rose-m, line 30: What if there are multiple tags on that revision? 🤔 
The proposed process is to use the `rc` tags and then re-tag the latest rc for publishing. However, this will lead to the same commit hash having multiple tags. The `getTagByCommitSha` would then just take the "first one" it `.find()`s. Is this deterministic?

packages/build/src/publish.ts
```diff
@@ -24,6 +25,7 @@ export default async function publish(
 
   await githubRepo.promoteRelease(config);
 
+  publishNpmPackages();
```
rose-m, line 28: Maybe stupid question, but: right now, the `npm run evergreen-release` happens in its own task on evergreen. After that, a new task for the `release_publish` is trigger which checks out the code again, runs install and then triggers the publish.
At this point, the versions in the npm packages are fresh from checkout, i.e. they are the "dev" versions. We would have to trigger setting the version inside `publishNpmPackages()`, too, right? 🤔

rose-m (commented):
Nice, getting to full CI publication soon :P Got some questions and remarks :)

packages/build/src/release.ts
mcasimir: This is intentional, i think packages consider their package.json version as the current mongosh version, so in the package 'phase' we need the version to be reflected.

packages/build/src/release.ts
mcasimir: No this is just going away. We should just use the evergreen env var, but we probably need another ticket to wire that in.

packages/build/src/publish.ts
mcasimir: That is the reason why we have `bumpNpmPackages` outside the if in `release.ts`, essentially the bump is done again.

packages/build/src/npm-packages.ts
mcasimir: I really gotta find a way to test it out, at least manually once, i'm sure that the first time we use it, it won't work.

scripts/publish-npm.js
```diff
@@ -1,192 +0,0 @@
-const { execSync, execFileSync } = require('child_process');
```
addaleax, line 1: This script is referenced from the root `package.json`, I think you’ll want to remove it from there as well

addaleax (approved):
:crossed_fingers:

rose-m (approved):

PR #536, opened 2021-01-08
feat(ci): split compilation and packaging up MONGOSH-469

addaleax:
- Split the `compile_and_upload_artifact` evergreen functions/tasks
  into two steps, named `compile_artifact` and
  `package_and_upload_artifact`. The former only compiles the binary,
  the latter creates the tarball/pkg/deb/zip from it and uploads
  it to S3.
  - On the build package level, the latter is also subdivided into
    creating the tarball and uploading it. This makes it easier to
    perform local testing, and replaces the “dry run” mode that we
    previously had.
- Run the different tasks on appropriate machines:
  - On Windows and Linux, the build steps are moved to separate,
    more powerful machines that are intended for compilation steps.
    All other tasks remain on the same hosts. For Linux, this means
    compiling the executable once and packaging it three times,
    rather than compiling it three different times as well, and
    faster compilation times overall.
  - On macOS, the packaging step is kept on the codesign machine,
    while building takes place on regular macOS hosts. This helps
    us be nice to other evergreen consumers, because it reduces the
    amount of time that we’re hogging the single codesign host
    that evergreen provides.
- In order to move executables from the compilation to the packaging
  steps, store them temporarily in S3 using evergreen’s built-in
  storage mechanism.


mcasimir (commented):
Looks great so far, there should be no problem with the npm stuff as well

rose-m (commented):
Nice! Looks good to me, too - should not interfere with homebrew.

addaleax:
I think this should pass now, removing the WIP label :)

.evergreen.yml
```diff
@@ -27,31 +26,60 @@ variables:
...
           if [ `uname` == Darwin ]; then
             # Verify signing
             spctl -a -vvv -t install dist/mongosh
           fi
+          npm run evergreen-release upload
         fi
         RELEASE_MONGOSH
-
         if [ `uname` == Darwin ]; then
           ssh -v -p 2222 localhost "bash ~/release_mongosh.sh"
```
rose-m, line 78: Independent from the PR: but why is there this `ssh` command on Mac? 🤔

rose-m (approved):

PR #539, opened 2021-01-08
(feat) Core FLE functionality

aherlihy:
Note the tests for the FLE functionality in the `Mongo` class are in `shell-api.spec.ts` instead of `mongo.spec.ts` because FLE options are only available to the user via the `ShellApi` class methods for `Mongo`, and not any internal construction of the `Mongo` object.

Tickets: 

- [x] https://jira.mongodb.org/browse/MONGOSH-492
- [x] https://jira.mongodb.org/browse/MONGOSH-493
- [x] https://jira.mongodb.org/browse/MONGOSH-515
- [x] https://jira.mongodb.org/browse/MONGOSH-494
- [x] https://jira.mongodb.org/browse/MONGOSH-495


packages/shell-api/src/csfle-options.ts
```diff
@@ -0,0 +1,27 @@
+import { Document, BinaryType } from '@mongosh/service-provider-core';
```
addaleax, line 1:
```suggestion
import type { Document, BinaryType } from '@mongosh/service-provider-core';
```

addaleax (approved):

packages/shell-api/src/field-level-encryption.ts
```diff
@@ -0,0 +1,165 @@
...
+  kmsProvider: awsKms | localKms,
+  schemaMap?: Document,
+  bypassAutoEncryption?: boolean;
+}
+
+@shellApiClassDefault
+@hasAsyncChild
+@classPlatforms([ ReplPlatform.CLI ] )
+export class ClientEncryption extends ShellApiClass {
+  public _mongo: any; // Mongo but any to avoid circular ref
```
addaleax, line 39: I think this is also something that `import type` would solve

packages/shell-api/src/field-level-encryption.ts
```diff
@@ -0,0 +1,165 @@
...
+  public _libmongocrypt: any;
```
addaleax, line 40: Is it worth asking the driver team for TS definitions for the libmongocrypt bindings, or something that we would potentially want to do ourselves? (I wouldn’t be opposed to writing them, and the node driver already has some available in its own source tree, but also don’t want to feel like that would be wasting my time)

packages/shell-api/src/field-level-encryption.ts
```diff
@@ -0,0 +1,165 @@
...
+
+  constructor(mongo: any) {
+    super();
+    this._mongo = mongo;
+    this._libmongocrypt = new mongo._serviceProvider.fle.ClientEncryption(
```
addaleax, line 45: Btw, I just picked `serviceProvider.fle` as a place to put the libmongocrypt bindings because that was easy – feel free to move it/put it behind a method/… if you like

packages/shell-api/src/field-level-encryption.ts
```diff
@@ -0,0 +1,165 @@
...
+  public _mongo: any; // Mongo but any to avoid circular ref
+  public _clientEncryption: ClientEncryption;
+  private _keyColl: Collection;
+  constructor(mongo: any, clientEncryption: ClientEncryption) {
+    super();
+    this._mongo = mongo;
+    this._clientEncryption = clientEncryption;
+    const [ db, coll ] = mongo._fleOptions.keyVaultNamespace.split('.');
+    if (coll.includes('.')) {
+      throw new MongoshInvalidInputError('keyVaultNamespace must be <db>.<coll>');
```
addaleax, line 91: This error is never going to be thrown, I think

addaleax (approved):

packages/shell-api/src/field-level-encryption.ts
aherlihy: Oh nice! Thanks!

packages/shell-api/src/field-level-encryption.ts
aherlihy: I think I remember asking and they said since it's not part of the driver, they don't write it but I don't think they'd mind us adding them. See this thread for the convo: https://mongodb.slack.com/archives/C0V8RU15L/p1608128060225500. I made this ticket https://jira.mongodb.org/browse/NODE-2989 to see if they'd be alright with us adding the types

packages/shell-api/src/field-level-encryption.ts
aherlihy: This works fine for me, since for now this is the only way we'll be doing FLE. In the future if we end up with different FLE backends for different platforms we can abstract some of the API behind methods.

packages/shell-api/src/field-level-encryption.ts
aherlihy: Ok can be removed

addaleax: I mean, I think, you probably want a condition like `if (!coll)` here instead – but a string split at `'.'` won’t contain `'.'` in the resulting parts

aherlihy: Oops, the ts compiler complains "Object is possibly 'undefined'" so I think the error check should stay there? The namespace would get validated upon `MongoClient` connection but there's no way to indicate that to the compiler, from what I can tell?

packages/shell-api/src/field-level-encryption.ts
aherlihy: It seems they're cool with us doing it, so I made a MONGOSH ticket that we can put in the upcoming sprint https://jira.mongodb.org/browse/MONGOSH-527

addaleax (approved):
Looks good!

packages/shell-api/src/field-level-encryption.ts
```diff
@@ -0,0 +1,183 @@
...
+  public _mongo: Mongo;
+  public _libmongocrypt: any;
+
+  constructor(mongo: any) {
+    super();
+    this._mongo = mongo;
+    this._libmongocrypt = new mongo._serviceProvider.fle.ClientEncryption(
+      mongo._serviceProvider.getRawClient(),
+      {
+        options: this._mongo._fleOptions
```
addaleax, line 58: Should this be
```suggestion
        ...this._mongo._fleOptions
```
?

aherlihy: Oh nice, yes it should absolutely be that

aherlihy: I updated it and added a test

PR #540, opened 2021-01-11
chore(deps): bump mongodb to b1bdb06c

addaleax:
Fixes MONGOSH-418.

See https://github.com/mongodb/node-mongodb-native/compare/16c15749...b1bdb06c.

---

feat(service-provider-core): add ConnectionString class MONGOSH-528

Add a ConnectionString class for easier URI manipulation.

### DIFF
--- a/packages/cli-repl/src/arg-mapper.ts
+++ b/packages/cli-repl/src/arg-mapper.ts
@@ -1,6 +1,4 @@
-import { MongoshInvalidInputError } from '@mongosh/errors';
 import { CliOptions, MongoClientOptions } from '@mongosh/service-provider-server';
-import { promises as fs } from 'fs';
 import setValue from 'lodash.set';
 
 /**
@@ -19,22 +17,12 @@ const MAPPINGS = {
   tlsAllowInvalidCertificates: 'tlsAllowInvalidCertificates',
   tlsAllowInvalidHostnames: 'tlsAllowInvalidHostnames',
   tlsCAFile: 'tlsCAFile',
-  tlsCRLFile: { opt: 'sslCRL', val: readCrlFileContent },
+  tlsCRLFile: 'sslCRL',
   tlsCertificateKeyFile: 'tlsCertificateKeyFile',
   tlsCertificateKeyFilePassword: 'tlsCertificateKeyFilePassword',
   username: 'auth.username',
   verbose: { opt: 'loggerLevel', val: 'debug' }
 };
-
-async function readCrlFileContent(crlFilePath: string): Promise<string | Buffer> {
-  try {
-    return await fs.readFile(crlFilePath, { encoding: 'utf-8' });
-  } catch (error) {
-    throw new MongoshInvalidInputError(
-      `The file specified by --tlsCRLFile does not exist or cannot be read${error ? ': ' + error.message : ''}`
-    );
-  }
-}
 
 function isExistingMappingKey(key: string, options: CliOptions): key is keyof typeof MAPPINGS {
   return MAPPINGS.hasOwnProperty(key) && options.hasOwnProperty(key);
@@ -59,11 +47,7 @@ async function mapCliToDriver(options: CliOptions): Promise<MongoClientOptions> 
         const cliValue = (options as any)[cliOption];
         if (cliValue) {
           const { opt, val } = mapping;
-          if (typeof val === 'function') {
-            setValue(nodeOptions, opt, await val(cliValue));
-          } else {
-            setValue(nodeOptions, opt, val);
-          }
+          setValue(nodeOptions, opt, val);
         }
       } else {
         setValue(nodeOptions, mapping, (options as any)[cliOption]);
--- a/packages/cli-repl/test/e2e-analytics.spec.ts
+++ b/packages/cli-repl/test/e2e-analytics.spec.ts
@@ -16,31 +16,31 @@ describe('e2e Analytics', () => {
     await TestShell.killall();
   });
 
-  context('setting up the replica set', () => {
-    it('works', async() => {
-      const rsConfig = {
-        _id: replSetName,
-        members: [
-          { _id: 0, host: `${await rs0.hostport()}`, priority: 2 },
-          { _id: 1, host: `${await rs1.hostport()}`, priority: 1 },
-          { _id: 2, host: `${await rs2.hostport()}`, priority: 1 },
-          { _id: 3, host: `${await rs3.hostport()}`, priority: 0, votes: 0, tags: { nodeType: 'ANALYTICS' } }
-        ]
-      };
+  before(async() => {
+    const rsConfig = {
+      _id: replSetName,
+      members: [
+        { _id: 0, host: `${await rs0.hostport()}`, priority: 2 },
+        { _id: 1, host: `${await rs1.hostport()}`, priority: 1 },
+        { _id: 2, host: `${await rs2.hostport()}`, priority: 1 },
+        { _id: 3, host: `${await rs3.hostport()}`, priority: 0, votes: 0, tags: { nodeType: 'ANALYTICS' } }
+      ]
+    };
 
-      const shell = TestShell.start({
-        args: [await rs0.connectionString()]
-      });
-      await shell.waitForPrompt();
-      await shell.executeLine(`rs.initiate(${JSON.stringify(rsConfig)})`);
-      shell.assertContainsOutput('ok: 1');
-      await eventually(async() => {
-        const isMaster = await shell.executeLine('db.isMaster().ismaster');
-        expect(isMaster).to.contain('true');
-      });
+    const shell = TestShell.start({
+      args: [await rs0.connectionString()]
     });
+    await shell.waitForPrompt();
+    await shell.executeLine(`rs.initiate(${JSON.stringify(rsConfig)})`);
+    shell.assertContainsOutput('ok: 1');
+    await eventually(async() => {
+      const isMaster = await shell.executeLine('db.isMaster().ismaster');
+      expect(isMaster).to.contain('true');
+    });
+  });
 
-    it('a direct connection without readPreference ends up at primary', async() => {
+  context('without readPreference', () => {
+    it('a direct connection ends up at primary', async() => {
       const shell = TestShell.start({
         args: [ await rs0.connectionString() ]
       });
--- a/packages/cli-repl/test/e2e-direct.spec.ts
+++ b/packages/cli-repl/test/e2e-direct.spec.ts
@@ -27,74 +27,76 @@ describe('e2e direct connection', () => {
       });
     });
 
-    it('allows to initialize the replica set', async() => {
-      const replSetConfig = {
-        _id: replSetId,
-        version: 1,
-        members: [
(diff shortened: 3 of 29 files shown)


packages/cli-repl/src/arg-mapper.ts
```diff
@@ -19,23 +17,13 @@ const MAPPINGS = {
   tlsAllowInvalidCertificates: 'tlsAllowInvalidCertificates',
   tlsAllowInvalidHostnames: 'tlsAllowInvalidHostnames',
   tlsCAFile: 'tlsCAFile',
-  tlsCRLFile: { opt: 'sslCRL', val: readCrlFileContent },
+  tlsCRLFile: 'sslCRL',
```
addaleax, line 20: They fixed it :) If there’s an issue associated with this, I can’t find it (/cc @rose-m ?)

packages/cli-repl/test/e2e-analytics.spec.ts
```diff
@@ -16,31 +16,31 @@ describe('e2e Analytics', () => {
...
+      args: [await rs0.connectionString()]
+    });
+    await shell.waitForPrompt();
+    await shell.executeLine(`rs.initiate(${JSON.stringify(rsConfig)})`);
+    shell.assertContainsOutput('ok: 1');
+    await eventually(async() => {
+      const isMaster = await shell.executeLine('db.isMaster().ismaster');
+      expect(isMaster).to.contain('true');
     });
+  });
```
addaleax, line 40: Unrelated, but without this one can’t run one of the other individual tests in this file

packages/cli-repl/test/e2e-tls.spec.ts
```diff
@@ -114,9 +112,7 @@ describe('e2e TLS', () => {
         shell.assertContainsOutput('MongoServerSelectionError');
       });
 
-      // TODO: enable when NODE-2977 is solved
-      // eslint-disable-next-line mocha/no-skipped-tests
-      it.skip('fails when not using --tls (connection string)', async() => {
+      it('fails when not using --tls (connection string)', async() => {
```
addaleax, line 115: Not sure if actually fixed, it worked when I added a `:` into the filename on Linux but we’ll see what CI says.

packages/cli-repl/test/e2e-tls.spec.ts
```diff
@@ -221,14 +215,12 @@ describe('e2e TLS', () => {
...
-      // TODO: enable when NODE-2977 is solved
-      // eslint-disable-next-line mocha/no-skipped-tests
-      it.skip('works with valid cert (connection string)', async() => {
+      it('works with valid cert (connection string)', async() => {
         const shell = TestShell.start({
           args: [
             `${await server.connectionString()}?serverSelectionTimeoutMS=1500`
             + '&authMechanism=MONGODB-X509'
-            + `&tls=true&tlsCAFile=${encodeURIComponent(CA_CERT)}&tlsCertificateKeyFile=${encodeURIComponent(CLIENT_CERT)}`
+            + `&tls=true&tlsCAFile=${encodeURIComponent(CA_CERT)}&tlsCertificateKeyFile=${encodeURIComponent(CLIENT_CERT)}&tlsCertificateFile=${encodeURIComponent(CLIENT_CERT)}`
```
addaleax, line 223: Waiting for the driver team to get back to us about this on slack

packages/service-provider-server/src/cli-service-provider.ts
```diff
@@ -1016,15 +1018,14 @@ class CliServiceProvider extends ServiceProviderCore implements ServiceProvider
...
       ...this.initialOptions,
-      auth: auth
-    };
+      auth
+    });
     if (authDoc.mechanism) clientOptions.authMechanism = authDoc.mechanism as AuthMechanismId;
     if (authDoc.authDb) clientOptions.authSource = authDoc.authDb;
     const mc = await MongoClient.connect(
-      this.uri as string,
+      Object.assign(new URL(this.uri as string), { username: '', password: '' }).href,
```
addaleax, line 1028: Also waiting for the driver team on slack here

packages/cli-repl/test/e2e-tls.spec.ts
addaleax: CI says it’s fixed :)

packages/service-provider-server/src/cli-service-provider.ts
addaleax: So, it looks like this was intentional on their end. Is `this.uri` something sufficiently URL-like at this point? Do you know @aherlihy @rose-m, or do you have any other ideas for how to do this well?

packages/cli-repl/test/e2e-tls.spec.ts
addaleax: Also looks like this is something we need to deal with for now. Do we want to do some URL rewriting for cases like this? Do we want to just ignore this and accept the driver change as it is?

packages/service-provider-server/src/cli-service-provider.ts
rose-m: As mentioned in Slack, the `uri` might contain multiple hosts in the seed list which is not URL compliant.
Furthermore, @mcasimir mentioned the escaping difference.
So I'm afraid we are stuck with re-structuring the URL again 😞 Maybe we can provide some better utility function and move it to a common dependency? Right now the `cli-repl` package knows quite well how to de-structure the URL and also uses `URL` to deal with parameter manipulation already.

packages/cli-repl/test/e2e-direct.spec.ts
```diff
@@ -27,74 +27,76 @@ describe('e2e direct connection', () => {
...
-      const replSetConfig = {
-        _id: replSetId,
-        version: 1,
-        members: [
-          { _id: 0, host: await rs0.hostport(), priority: 1 },
-          { _id: 1, host: await rs1.hostport(), priority: 0 },
-          { _id: 2, host: await rs2.hostport(), priority: 0 },
-        ]
-      };
+    context('after rs.initiate()', () => {
```
addaleax, line 30: Same as above, these changes are necessary so that the individual test cases below can be run on their own but are otherwise unrelated

packages/cli-repl/test/e2e-tls.spec.ts
addaleax: Okay, we’re doing URL rewriting now. The test change here is undone now :)

packages/service-provider-server/src/cli-service-provider.ts
addaleax: Also replaced this with a cleaner variant

packages/service-provider-core/src/connection-string.ts
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  toString(): string {
+    return super.toString().replace(DUMMY_HOSTNAME, this.hosts.join(','));
+  }
+
+  clone(): ConnectionString {
+    return new ConnectionString(this.toString());
+  }
+}
```
addaleax, line 112: Talked with @mcasimir and, if we like this, we can also split this out into a separate package that compass and/or the driver itself could then use

packages/cli-repl/src/arg-mapper.ts
rose-m: No there wasn't an issue with that - `sslCRL` was officially documented by them in their API to be the _content_ of the CRL. That's why I added the transformation on our side. But even better now that they do that :D

packages/service-provider-core/src/connection-string.ts
```diff
@@ -0,0 +1,130 @@
...
+  }
+  class TextDecoder {
+    decode(bytes: Uint8Array): string {
+      const str = Buffer.from(bytes).toString('utf8');
+      return str.slice(+str.startsWith('\ufeff'));
+    }
+  }
+  Object.assign(require('util'), { TextDecoder, TextEncoder });
+}
+import { URL as WhatWGURL } from 'whatwg-url';
```
mcasimir, line 18: Is the url from browserify not compliant?

packages/service-provider-core/src/connection-string.ts
mcasimir: @Anemy you may want to take a look if this is something viable for compass

packages/service-provider-core/src/connection-string.ts
addaleax: The problem is that there isn’t any :grimacing:

testing/integration-testing-hooks.ts
```diff
@@ -239,9 +239,7 @@ export class MongodSetup {
 
     let client;
     try {
-      client = await MongoClient.connect(await this.connectionString(), {
-        useNewUrlParser: true
-      } as MongoClientOptions);
+      client = await MongoClient.connect(await this.connectionString(), {});
```
mcasimir, line 242: What is the issue with `useNewUrlParser: true`?

addaleax: It has been removed, there is only one URL parser now as I understand it

mcasimir: Makes sense

packages/service-provider-core/src/connection-string.spec.ts
```diff
@@ -0,0 +1,138 @@
...
+          protocol: 'mongodb:',
+          username: 'cat',
+          password: 'meow',
+          pathname: '/db',
+          search: '',
+          hash: '',
+          isSRV: false,
+          hosts: ['localhost:12345']
+        }
+      }
```
rose-m, line 62: Could we add one more test case with multiple hosts and a query string?
```
{
  uri: 'mongodb://localhost:12345,anotherHost/?directConnection=true',
  match: {
    href: 'mongodb://localhost:12345,anotherHost/?directConnection=true',
    protocol: 'mongodb:',
    username: '',
    password: '',
    pathname: '/',
    search: '?directConnection=true',
    hash: '',
    isSRV: false,
    hosts: ['localhost:12345', 'anotherHost']
  }
}
```

rose-m (approved):
Nice! :) Just maybe add another test case for the URI parser to cover exactly the case why we can't use the regular URL 👍

packages/service-provider-core/src/connection-string.spec.ts
addaleax: Sure, done!

aherlihy:
Apologies if this is a dumb question, but would it be possible for us to use the driver's URI parser directly?

addaleax:
> Apologies if this is a dumb question, but would it be possible for us to use the driver's URI parser directly?

We talked about it... apparently Compass uses some not-really-public parsing code of the 3.x driver: https://github.com/mongodb-js/connection-model/blob/802fae6731ad7d16c403f601efe154323da67192/lib/model.js#L18 But I wouldn't expect them to consider that public API. So, probably, we'd need to ask them to expose something first, or find some other way to use shared code for this.

aherlihy:
Cool, this seems fine to me but in the future we should consider merging the implementations and asking the driver to expose it to us. Since it's spec'd behavior, it's always better to rely on the drivers implementation bc they use the spec tests and we don't wanna reproduce work if we can avoid it

PR #545, opened 2021-01-12
MONGOSH-523 - AutoEncryption only run on enterprise

aherlihy:
The best solution I could find was connecting twice, once with the autoencryption options entirely removed. This connection runs buildInfo, and if it's not enterprise, then it will close the connection and error. If it is enterprise, then it will close the connection and reconnect with the autoencryption options included.

TODO: the message that is thrown should be written & reviewed by the legal team per @mmarcon

### DIFF
--- a/packages/service-provider-server/src/cli-service-provider.ts
+++ b/packages/service-provider-server/src/cli-service-provider.ts
@@ -74,7 +74,7 @@ import {
   ConnectionString
 } from '@mongosh/service-provider-core';
 
-import { MongoshCommandFailed, MongoshInternalError } from '@mongosh/errors';
+import { MongoshCommandFailed, MongoshInternalError, MongoshRuntimeError } from '@mongosh/errors';
 
 const bsonlib = {
   Binary,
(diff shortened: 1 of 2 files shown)


packages/service-provider-server/src/cli-service-provider.ts
```diff
@@ -119,7 +119,35 @@ const DEFAULT_BASE_OPTIONS = Object.freeze({
...
+    const optionsWithoutFLE = { ...clientOptions };
+    delete optionsWithoutFLE.autoEncryption;
+    const client = await mClient.connect(uri, optionsWithoutFLE);
+    const buildInfo = await client.db('admin').admin().command({ buildInfo: 1 });
+    if (
+      !(buildInfo.gitVersion && buildInfo.gitVersion.match(/enterprise/)) &&
+      !(buildInfo.modules && buildInfo.modules.indexOf('enterprise') !== -1)
+    ) {
+      await client.close();
+      throw new MongoshRuntimeError('Cannot turn on automatic encryption for connections to non-enterprise hosts');
```
aherlihy, line 142: This is TBD

packages/service-provider-server/src/cli-service-provider.ts
```diff
@@ -119,7 +119,35 @@ const DEFAULT_BASE_OPTIONS = Object.freeze({
...
 /**
- * Encapsulates logic for the service provider for the mongosh CLI.
+ * Connect a MongoClient. If AutoEncryption is requested, first connect without the encryption options and verify that
+ * the connection is to an enterprise cluster. If not, then error, otherwise close the connection and reconnect with the
+ * options the user initially specified. Provide the client class as an additional argument in order to test.
+ * @param uri {String}
+ * @param clientOptions {MongoClientOptions}
+ * @param mClient {MongoClient}
+ */
+export async function connectMongoClient(uri: string, clientOptions: MongoClientOptions, mClient = MongoClient): Promise<MongoClient | void> {
```
addaleax, line 129: When would this not return a `Promise<MongoClient>`?

packages/service-provider-server/src/cli-service-provider.ts
mmarcon: I am not sure about the regex on `gitVersion`. At least for our Atlas cluster, this is the first part of the `buildInfo` output:
```javascript
{
  version: '4.2.11',
  gitVersion: 'ea38428f0c6742c7c2c7f677e73d79e17a2aab96',
  modules: [ 'enterprise' ],
 ...
}
```
Why were you thinking of checking `gitVersion` in addition to `modules`?

Another minor thing: `buildInfo.modules.indexOf('enterprise') !== -1` can be rewritten as  `buildInfo.modules.includes('enterprise')`. I _think_ Typescript also supports optional chaining so you can probably write the second condition as
```typescript
buildInfo.modules?.includes('enterprise')
```

As for the copy of the error message, I'd go for something like:
`Automatic encryption is only available with Atlas and MongoDB Enterprise`. I'll get that reviewed.

mmarcon: Other minor thing: not sure if we do it for the other errors, and not sure if this is special, but would it make sense to put this in the i18n package and read it from there?

packages/service-provider-server/src/cli-service-provider.ts
aherlihy: Yes, it would! Good catch - I previously had it looking like
```
if (x) {
    return y;
} 
throw z
```
which I just learned you can't have the function signature return just the type of `y`. Then I refactored to avoid that issue, but then forgot to update the signature  🤦

packages/service-provider-server/src/cli-service-provider.ts
aherlihy: I took it directly from the [mongodb-build-info package](https://github.com/mongodb-js/mongodb-build-info/blob/master/index.js#L18-L28), which is what we're using in the rest of the shell to determine if it's an enterprise cluster. I also found only the git hash without 'enterprise' when I tested locally, but I thought it could possibly be from older versions before they had `modules` in `buildInfo`. I have a very vague memory of Irina mentioning the git hash with enterprise, but I can't find that conversation. We may want to improve the check there as well ([placeholder jira ticket](https://jira.mongodb.org/browse/MONGOSH-536)). 

Per the i18n error message, we don't yet use i18n for error messages everywhere but that's certainly a future feature we should track in the future.

aherlihy: From this stack overflow it seems like it may happen sometimes that the gitversion has enterprise: https://stackoverflow.com/questions/33160352/how-to-check-if-mongodb-enterprise-is-in-use but it may be an older version than we support

mmarcon: Ah, ok, it's cool then. Thank you for clarifying.

PR #553, opened 2021-01-14
chore: update FLE API MONGOSH-540

rose-m:
This PR includes the use of the types provided by `mongodb-client-encryption` as well as support of the new Azure and GCP KMS providers.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "posttest": "npm run clear-mlaunch",
     "posttest-ci": "npm run clear-mlaunch -- --killAllMongod",
     "posttest-e2e-ci": "npm run clear-mlaunch -- --killAllMongod",
-    "test": "rimraf .nyc_output && lerna exec --scope @mongosh/shell-api -- nyc --no-clean --cwd ../.. --reporter=none npm run test && npm run report-coverage",
+    "test": "rimraf .nyc_output && lerna exec -- nyc --no-clean --cwd ../.. --reporter=none npm run test && npm run report-coverage",
     "test-ci": "rimraf .nyc_output && lerna exec -- nyc --no-clean --cwd ../.. --reporter=none npm run test-ci && npm run report-coverage",
     "test-ci-nocoverage": "lerna exec -- npm run test-ci",
     "test-e2e": "lerna run --stream test-e2e",
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "posttest": "npm run clear-mlaunch",
     "posttest-ci": "npm run clear-mlaunch -- --killAllMongod",
     "posttest-e2e-ci": "npm run clear-mlaunch -- --killAllMongod",
-    "test": "rimraf .nyc_output && lerna exec -- nyc --no-clean --cwd ../.. --reporter=none npm run test && npm run report-coverage",
+    "test": "rimraf .nyc_output && lerna exec --scope @mongosh/shell-api -- nyc --no-clean --cwd ../.. --reporter=none npm run test && npm run report-coverage",
     "test-ci": "rimraf .nyc_output && lerna exec -- nyc --no-clean --cwd ../.. --reporter=none npm run test-ci && npm run report-coverage",
     "test-ci-nocoverage": "lerna exec -- npm run test-ci",
     "test-e2e": "lerna run --stream test-e2e",
--- a/packages/service-provider-core/src/index.ts
+++ b/packages/service-provider-core/src/index.ts
@@ -24,6 +24,7 @@ import { bsonStringifiers } from './printable-bson';
 import ShellAuthOptions from './shell-auth-options';
 import { ConnectionString } from './connection-string';
 export * from './all-transport-types';
+export * from './all-fle-types';
 
 const bson = {
   ObjectId,
--- a/packages/shell-api/src/field-level-encryption.spec.ts
+++ b/packages/shell-api/src/field-level-encryption.spec.ts
@@ -1,10 +1,10 @@
 import { signatures, toShellResult } from './decorators';
 import { ALL_PLATFORMS, ALL_SERVER_VERSIONS, ALL_TOPOLOGIES } from './enums';
-import { KeyVault, ClientEncryption } from './field-level-encryption';
+import { KeyVault, ClientEncryption, ClientSideFieldLevelEncryptionOptions } from './field-level-encryption';
 import Mongo from './mongo';
 import { expect } from 'chai';
 import sinon, { StubbedInstance, stubInterface } from 'ts-sinon';
-import { ServiceProvider, bson, BinaryType } from '@mongosh/service-provider-core';
+import { ServiceProvider, bson, ClientEncryption as FLEClientEncryption } from '@mongosh/service-provider-core';
 import { EventEmitter } from 'events';
 import ShellInternalState from './shell-internal-state';
 import Database from './database';
@@ -41,29 +41,24 @@ const AWS_KMS = {
 
 const ALGO = 'AEAD_AES_256_CBC_HMAC_SHA_512-Deterministic';
 
-// TODO: see NODE-2989
-interface lmc {
-  encrypt(value: any, options: { keyId: BinaryType, keyAltName: string, algorithm: string }): Promise<void>;
-  decrypt(): Promise<any>;
-  createDataKey(): Promise<any>;
-}
-
 const RAW_CLIENT = { client: 1 } as any;
 
 describe('Field Level Encryption', () => {
   let sp: StubbedInstance<ServiceProvider>;
   let mongo: Mongo;
   let internalState: ShellInternalState;
-  let libmongoc: StubbedInstance<lmc>;
+  let libmongoc: StubbedInstance<FLEClientEncryption>;
   let clientEncryption: ClientEncryption;
   let keyVault: KeyVault;
   let clientEncryptionSpy;
   describe('Metadata', () => {
     before(() => {
-      libmongoc = stubInterface<lmc>();
+      libmongoc = stubInterface<FLEClientEncryption>();
       sp = stubInterface<ServiceProvider>();
       sp.bsonLibrary = bson;
-      sp.fle = { ClientEncryption: function() { return libmongoc; } };
+      sp.fle = {
+        ClientEncryption: function() { return libmongoc; }
+      } as any;
       sp.initialDb = 'test';
       internalState = new ShellInternalState(sp, stubInterface<EventEmitter>());
       internalState.currentDb = stubInterface<Database>();
@@ -119,7 +114,7 @@ describe('Field Level Encryption', () => {
   describe('commands', () => {
     beforeEach(() => {
       clientEncryptionSpy = sinon.spy();
-      libmongoc = stubInterface<lmc>();
+      libmongoc = stubInterface<FLEClientEncryption>();
       sp = stubInterface<ServiceProvider>();
       sp.getRawClient.returns(RAW_CLIENT);
       sp.bsonLibrary = bson;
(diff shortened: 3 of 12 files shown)


packages/shell-api/src/field-level-encryption.ts
```diff
@@ -47,12 +50,16 @@ export interface ClientSideFieldLevelEncryptionOptions {
...
+  public _libmongocrypt: MongoCryptClientEncryption;
 
-  constructor(mongo: any) {
+  constructor(mongo: Mongo) {
     super();
     this._mongo = mongo;
-    this._libmongocrypt = new mongo._serviceProvider.fle.ClientEncryption(
+    const fle = mongo._serviceProvider.fle;
+    if (!fle) {
+      throw new MongoshRuntimeError('FLE API is not available');
```
addaleax, line 60: I think this is where https://jira.mongodb.org/browse/MONGOSH-514 comes into play :)

rose-m: Yep :) Just wanted to get rid of the TypeScript error with undefined things... :D

aherlihy: Yeah this is literally what I was going to put in, so once this is merged I'll just close the ticket. Nice!

packages/service-provider-core/src/index.ts
```diff
@@ -24,6 +24,7 @@ import { bsonStringifiers } from './printable-bson';
 import ShellAuthOptions from './shell-auth-options';
 import { ConnectionString } from './connection-string';
 export * from './all-transport-types';
+export * from './all-fle-types';
```
aherlihy, line 27: Will this cause an error for the contexts where we can't import the fle package? Just checking bc previously we had everything FLE in service-provider-server and not core

packages/shell-api/src/field-level-encryption.ts
```diff
@@ -75,17 +82,17 @@ export class ClientEncryption extends ShellApiClass {
       value,
       {
         keyId: encryptionId,
-        algorithm: encryptionAlgorithm
```
aherlihy, line 78: If you wanted you could make the type of encryptionAlgorithm more specific to match the ClientEncryption types, to avoid any, but it's not a big deal

packages/shell-api/src/field-level-encryption.ts
```diff
@@ -75,17 +82,17 @@ export class ClientEncryption extends ShellApiClass {
...
       }
     );
   }
 
   @returnsPromise
-  decrypt(
+  async decrypt(
     encryptedValue: any
-  ): Promise<BinaryType> {
+  ): Promise<any> {
```
aherlihy, line 93: Shouldn't this stay Promise<Binary>?

packages/shell-api/src/field-level-encryption.ts
```diff
@@ -113,6 +120,7 @@ export class KeyVault extends ShellApiClass {
 
   @returnsPromise
   createKey(
+    // TODO kms cannot be a document
```
aherlihy, line 123: See this conversation for the context: https://docs.google.com/document/d/1v0UUDvK4Ch8WM03XAFw1Ci-hem84vrgIw07RWCwmyJ4/edit?disco=AAAALAO7L-U

packages/service-provider-core/src/index.ts
addaleax: Uh, good point – I think these should both be `export type` (and other lines as well)? I think so long as we depend on it only at build time we should be fine.

rose-m: Yeah, I'm not sure... Inside `all-fle-types` it's already just "types" and I've checked the compiled JavaScript - it's basically an empty file for `all-fle-types.ts`.

packages/shell-api/src/field-level-encryption.ts
rose-m: Inside `mongodb-client-encryption` it's not guaranteed to be `Binary`, at least reading through the code and looking at the tests in there it can just come out as a string. One of the tests stuffs a `"hello"` into `encrypt(...)`, then `decrypt(...)`s that result, to verify it matches `"hello"` again which obviously isn't a binary 😅

aherlihy: Huh, that's odd! We can just use `Document` then instead of `any`, for consistency.

addaleax:
I think we’d want to re-open https://jira.mongodb.org/browse/MONGOSH-527, or create another ticket for tracking this?

rose-m:
Should update the title - yeah I've commented in MONGOSH-540 and would use this for tracking. As the lacking parts are essentially down to updating that API part now.

packages/shell-api/src/field-level-encryption.ts
```diff
@@ -119,18 +108,11 @@ export class KeyVault extends ShellApiClass {
 
   @returnsPromise
   createKey(
-    kms: string | Document,
-    customMasterKey?: Document, // doc defined here: https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.rst#masterkey
-    keyAltName?: string[]
+    kms: ClientEncryptionDataKeyProvider,
+    options?: ClientEncryptionCreateDataKeyProviderOptions
```
addaleax, line 112: https://docs.mongodb.com/manual/reference/method/KeyVault.createKey/ seems to want `keyAltName` as the third argument -- do we want to break from that?

packages/shell-api/src/helpers.ts
```diff
@@ -618,29 +618,13 @@ export function processFLEOptions(fleOptions: ClientSideFieldLevelEncryptionOpti
...
-      autoEncryption.kmsProviders.local = {
-        ...autoEncryption.kmsProviders.local,
-        key: rawBuff
-      };
-    } else {
-      // Future TODO: allow binary types that are not from a string
-      throw new MongoshInvalidInputError('key field of local kmsProvider must be a BinData type created from a base64 encoded string');
-    }
-  }
+  };
```
addaleax, line 627: This also seems to be breaking with the old shell, right?

packages/shell-api/src/field-level-encryption.ts
rose-m: Yes, the documentation is invalid with the new specification of supporting GCP/Azure, too. I've updated the Technical Design document for that, too, after a small discussion we had: https://docs.google.com/document/d/1v0UUDvK4Ch8WM03XAFw1Ci-hem84vrgIw07RWCwmyJ4/edit?disco=AAAALAO7L-U. The API now matches the official spec and thus just forwards it to libmongocrypt.

packages/shell-api/src/helpers.ts
rose-m: Yes, I missed that... 🙈 Does it make sense to break here to then re-use exactly the driver types exposed AutoEncryption types?

rose-m: I've put it back 😄

packages/shell-api/src/field-level-encryption.spec.ts
```diff
@@ -311,20 +307,20 @@ describe('Field Level Encryption', () => {
...
       internalState.currentDb = stubInterface<Database>();
     });
     it('accepts the same local key twice', () => {
-      const localKmsOptions = {
+      const localKmsOptions: ClientSideFieldLevelEncryptionOptions = {
         keyVaultNamespace: `${DB}.${COLL}`,
         kmsProvider: {
           local: {
-            key: new bson.Binary(Buffer.alloc(96).toString('base64'))
+            key: Buffer.alloc(96).toString('base64')
```
addaleax, line 323: Ah, maybe there should have been a comment here leaving some context – this is a regression test case that makes sure that we don’t modify `kmsProvider.local.key` by converting it from `Binary` to `Buffer` in-place, so moving this to a plain `Buffer` more or less invalidates the test :)

addaleax (dismissed):
:shipit:

packages/shell-api/src/field-level-encryption.ts
addaleax: I’d still have a mild preference for backwards compatibility here as well, but as long as this is documented, I think it should be fine :)

packages/shell-api/src/field-level-encryption.spec.ts
rose-m: Ah okay :D I've put these tests into `shell-api`: https://github.com/mongodb-js/mongosh/pull/553/files/fce05fba6c22c207cd9aa6acf22dcb633b276ad1#diff-6567f9efa0de147b837a1b2c94bc14ca6b1f8d88dd7b5ccde016fa71735281bbR215 to make sure we support all 3 variants - binary, buffer, or plain string :) I think we can leave this as is then? 🤔

addaleax: Yeah, I saw that – so this should remain:
```suggestion
            key: new bson.Binary(Buffer.alloc(96).toString('base64'))
```

packages/shell-api/src/field-level-encryption.ts
addaleax: Actually, could we throw an exception when a third argument is passed then? I think that should make the upgrade path a lot easier than running the risk of silently ignoring options

packages/shell-api/src/field-level-encryption.spec.ts
rose-m: It's back ✅

packages/shell-api/src/field-level-encryption.ts
rose-m: Yup sounds good, will add the check for additional args!

packages/shell-api/src/field-level-encryption.ts
rose-m: @addaleax I'd be fine to change it to `Document` if you want - but the API in itself also just works on strings for example... See also the official "encrypt" example: https://docs.mongodb.com/manual/reference/method/ClientEncryption.encrypt/#example. It just passes in a string so when I'd decrypt that I'd get back my string right?

addaleax: yeah, I think that’s fine then :) just wanted to make sure

packages/shell-api/src/field-level-encryption.ts
addaleax: Sorry, I just realized that this is not just affecting `keyAltNames`, but also `masterKey` … I would really suggest not doing this breakage here, if we don’t have a good reason to

packages/shell-api/src/field-level-encryption.ts
```diff
@@ -113,22 +116,63 @@ export class KeyVault extends ShellApiClass {
...
-      );
+
+    if (typeof masterKeyOrAltNames === 'string') {
+      if (kms === 'local' && masterKeyOrAltNames === '') {
+        // allowed in the old shell - even enforced prior to 4.2.3
+        // https://docs.mongodb.com/manual/reference/method/KeyVault.createKey/
+        masterKeyOrAltNames = undefined;
+      } else {
+        throw new MongoshInvalidInputError(
+          'KeyVault.createKey does not support providing masterKey as string anymore. For AWS please use createKey("aws", { region: ..., masterKey: ... })',
```
addaleax, line 138: Yeah, I think this is going to be super helpful :+1:

addaleax (approved):
Thank you :)

package.json
```diff
@@ -27,7 +27,7 @@
     "posttest": "npm run clear-mlaunch",
     "posttest-ci": "npm run clear-mlaunch -- --killAllMongod",
     "posttest-e2e-ci": "npm run clear-mlaunch -- --killAllMongod",
-    "test": "rimraf .nyc_output && lerna exec -- nyc --no-clean --cwd ../.. --reporter=none npm run test && npm run report-coverage",
+    "test": "rimraf .nyc_output && lerna exec --scope @mongosh/shell-api -- nyc --no-clean --cwd ../.. --reporter=none npm run test && npm run report-coverage",
```
rose-m, line 30: Stupid mistake here...
```suggestion
    "test": "rimraf .nyc_output && lerna exec -- nyc --no-clean --cwd ../.. --reporter=none npm run test && npm run report-coverage",
```

addaleax:
The win32 smoke test failures are from https://jira.mongodb.org/browse/EVG-13983, if everything else is green I’ll merge this :)

PR #554, opened 2021-01-14
chore(ci): wrap bash scripts in curly brackets

addaleax:
Evergreen implements bash script execution by piping into bash.
This is not good, because it means that commands might actually
end up in a subprocess’s stdin instead of bash itself, leading to
commands being skipped. Use curly brackets to make sure that bash
has read the entire script before it starts executing it.
(See https://jira.mongodb.org/browse/EVG-13667)

### DIFF
--- a/.evergreen.yml
+++ b/.evergreen.yml
@@ -12,6 +12,7 @@ variables:
       shell: bash
       script: |
         set -e
+        { # curly brackets to work around EVG-13667
         cd $(pwd)
         export NODE_JS_VERSION=${node_js_version}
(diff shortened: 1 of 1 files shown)


addaleax:
(To be clear, I’m not sure that this is currently causing any trouble, but in the past it did; e.g. before https://github.com/mongodb-js/mongosh/commit/837eb064f4773c5d0c099b4fe3b855cded2e0f10 we ran the e2e tests on the generated executable on Linux but not on macOS because of this.)

aherlihy (approved):

PR #556, opened 2021-01-15
chore(deps): bump mongodb to 62717a1d

addaleax:
See https://github.com/mongodb/node-mongodb-native/compare/85d8d097...62717a1d


aherlihy (approved):
This has got to be the most minimalist driver update as of yet :D

PR #569, opened 2021-01-25
chore(build): add way to write uploaded URL to file

addaleax:
Add a way to write the S3 URL for an uploaded package to a specified
location on disk. This is useful for MONGOSH-506, because it makes it
possible to store that file in a location that evergreen can understand
on its own (i.e. it doesn't need to use JS code to figure out where
the artifact is stored) and use it as a pointer to the actual artifact
URL.

### DIFF
--- a/config/build.conf.js
+++ b/config/build.conf.js
@@ -98,6 +98,7 @@ module.exports = {
     owner: 'mongodb-js',
     repo: 'mongosh'
   },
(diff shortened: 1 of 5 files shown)


addaleax:
(Split out from the actual MONGOSH-506 work in order to make @rose-m's life easier :))

PR #575, opened 2021-01-26
chore(build): ensure git 2.24 is installed in centos MONGOSH-505

rose-m:
Adapted the CentOS docker images to include git 2.24. By default, CentOS comes with git v1.8.3.1 which is way too old for `lerna` to be able to bump the package versions.

It went undetected, as we only bump the version when we tag a commit.

### DIFF
--- a/packages/build/src/npm-packages.ts
+++ b/packages/build/src/npm-packages.ts
@@ -10,14 +10,6 @@ export function bumpNpmPackages(version: string): void {
     return;
   }
 
-  execFileSync('git', ['--version'], { stdio: 'inherit' });
-  try {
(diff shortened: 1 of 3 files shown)


packages/build/src/npm-packages.ts
```diff
@@ -10,14 +10,6 @@ export function bumpNpmPackages(version: string): void {
     return;
   }
 
-  execFileSync('git', ['--version'], { stdio: 'inherit' });
```
rose-m, line 13: Note: I had this in directly on master for testing as it took me a while to understand the problem...

mcasimir (approved):

PR #579, opened 2021-01-26
chore(build): always bump package version number MONGOSH-505

rose-m:
Added a new "evergreen-release" command `bump` which just bumps the package.json versions to the potentially present tag version.

I chose to always bump the version now, as for example the E2E tests verify the version that is output by `mongosh --version`. To avoid any additional issues with the version, it's now always bumped. I left the explicit bump also in for the other commands to ensure any other potential changes to the version will always be forced correctly when releasing.

### DIFF
--- a/packages/build/src/release.ts
+++ b/packages/build/src/release.ts
@@ -22,15 +22,30 @@ import uploadToDownloadCenter from './upload-to-download-center';
 import { downloadMongocrypt } from './download-mongocryptd';
 
 /**
- * Run the release process.
- * zip, release internally on evergreen, and, if applicable do a public release
- * (download centre and github.
- * @param {Config} config - the configuration, usually config/build.config.js.
+ * Run release specific commands.
+ * @param command The command to run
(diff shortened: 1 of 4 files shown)


packages/build/src/release.ts
```diff
@@ -22,37 +22,39 @@ import uploadToDownloadCenter from './upload-to-download-center';
...
-
-  let tarballFile: TarballFile;
-
   console.info(
     `mongosh: running command '${command}' with config:`,
     redactConfig(config)
   );
 
+  // updates the version of internal packages to reflect the tagged one
+  bumpNpmPackages(config.version);
```
mcasimir, line 44: why is not just:

```
  if (command === 'bump') {
    return bumpNpmPackages(config.version);
  }
```

shouldn't that work already once `evergreen-release bump` is called on the install task?

mcasimir (approved):

packages/build/src/release.ts
rose-m: I'll give it a shot :D

PR #580, opened 2021-01-27
chore(cli-repl): add --smokeTests option MONGOSH-506

addaleax:
We can use this to expand the test suite more easily in the future.

### DIFF
--- a/packages/cli-repl/src/smoke-tests.spec.ts
+++ b/packages/cli-repl/src/smoke-tests.spec.ts
@@ -1,0 +1,12 @@
+import { runSmokeTests } from './';
+import path from 'path';
+
+describe('smoke tests', () => {
+  it('self-test passes', async() => {
+    // Use ts-node to run the .ts files directly so nyc can pick them up for
+    // coverage.
+    await runSmokeTests(
+      process.execPath, '-r', 'ts-node/register', path.resolve(__dirname, 'run.ts')
+    );
+  });
+});
--- a/packages/cli-repl/src/smoke-tests.ts
+++ b/packages/cli-repl/src/smoke-tests.ts
@@ -1,0 +1,33 @@
+import { spawn } from 'child_process';
+import assert from 'assert';
+import { once } from 'events';
+
+// Run smoke tests on a executable, e.g. runSmokeTests("/path/to/mongosh.exe")
+// or runSmokeTests("/path/to/node", "packages/cli-repl/bin/mongosh.js").
+export async function runSmokeTests(executable: string, ...args: string[]): Promise<void> {
+  for (const { input, output, testArgs } of [{
+    input: 'print("He" + "llo" + " Wor" + "ld!")',
+    output: /Hello World!/,
+    testArgs: ['--nodb']
+  }]) {
+    await runSmokeTest(executable, [...args, ...testArgs], input, output);
+  }
+  // eslint-disable-next-line no-console
+  console.log('all tests passed');
+}
+
+async function runSmokeTest(executable: string, args: string[], input: string, output: RegExp): Promise<void> {
+  const proc = spawn(executable, [...args], {
+    stdio: ['pipe', 'pipe', 'inherit']
+  });
+  let stdout = '';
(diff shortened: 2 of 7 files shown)


packages/cli-repl/src/smoke-tests.spec.ts
```diff
@@ -0,0 +1,12 @@
+import { runSmokeTests } from './';
+import path from 'path';
+
+describe('smoke tests', () => {
+  it('self-test passes', async() => {
+    // Use ts-node to run the .ts files directly so nyc can pick them up for
+    // coverage.
+    await runSmokeTests(
+      process.execPath, '-r', 'ts-node/register', path.resolve(__dirname, 'run.ts')
```
mcasimir, line 9: How does this work? I immagine smoke tests to be something that gets compiled in, does not require dev deps, or even node to be present on the machine, and uses assert to check things up.

Is this allowing us to do something like:

```
mongosh --smokeTests smoketest.ts
```

packages/cli-repl/src/smoke-tests.ts
```diff
@@ -0,0 +1,33 @@
...
+    testArgs: ['--nodb']
+  }]) {
+    await runSmokeTest(executable, [...args, ...testArgs], input, output);
+  }
+  // eslint-disable-next-line no-console
+  console.log('all tests passed');
+}
+
+async function runSmokeTest(executable: string, args: string[], input: string, output: RegExp): Promise<void> {
+  const proc = spawn(executable, [...args], {
```
rose-m, line 20: Won't this suffer the same issue on Windows that we had e.g. with lerna where we would have to rely on `cross-spawn` to get it working? I'd assume on a Windows machine the executable (when self-tested) is `mongosh.exe` and then `spawn` will fail?

packages/cli-repl/src/smoke-tests.spec.ts
mcasimir: I understood this by reading the next file :)

packages/cli-repl/src/smoke-tests.ts
```diff
@@ -0,0 +1,33 @@
+import { spawn } from 'child_process';
+import assert from 'assert';
+import { once } from 'events';
+
+// Run smoke tests on a executable, e.g. runSmokeTests("/path/to/mongosh.exe")
+// or runSmokeTests("/path/to/node", "packages/cli-repl/bin/mongosh.js").
+export async function runSmokeTests(executable: string, ...args: string[]): Promise<void> {
```
mcasimir, line 7: Now I see .. makes sense!

packages/cli-repl/src/smoke-tests.spec.ts
addaleax: Ok, glad that this makes sense then :smile:

mcasimir (approved):

rose-m (commented):
Interesting question on "how" we want to write the smoke tests and in the end run them... With this solution the smoke tests would be baked into all mongosh distributables we build (based on the `cli-repl`). I'm not sure if I really like that since we would distribute test code into production 😅 
On the other hand if the requirement is that we want to keep the machines we run the smoke tests on clean from other stuff...

addaleax:
>  On the other hand if the requirement is that we want to keep the machines we run the smoke tests on clean from other stuff

Yeah, as @mcasimir mentioned in the comments above, I think that’s the best solution – I’m also not super worried about including test code in the executable, tbh.

packages/cli-repl/src/smoke-tests.ts
addaleax: Err, we can see what CI says, but I’m not sure why this would fail? I think `cross-spawn` mostly is for situations in which we don’t have the full path to what we want to run and want to emulate `cmd`’s lookup algorithm for things like `PATH_EXT`, but that isn’t really a concern here

rose-m: Ah okay, thank you :)

PR #581, opened 2021-01-27
fix(autocomplete): return fewer false positives MONGOSH-507

addaleax:
- Use word-boundary matching instead of substring tests for
  top-level identifiers.
- Make sure that completions for collections aren’t accidentally
  provided for function calls.
- Provide completions for `rs.` and `sh.` only if immediate methods
  on these objects are being completed.


rose-m (approved):
LGTM 👍

PR #587, opened 2021-01-28
feat(shell-api): support hedge/tagSet in cursor.readPref MONGOSH-511

addaleax:


aherlihy (approved):
Nice! Did we want to update `setReadPref` in the Mongo object to also use `readPrefFromOptions`?

PR #589, opened 2021-01-28
feat(shell-api,autocomplete): skip deprecated methods MONGOSH-509

addaleax:
Mark methods as deprecated and skip them for autocompletion.


gribnoysup (approved):
Looks good!

PR #607, opened 2021-02-03
fix: print AggregationCursor output correctly MONGOSH-562

addaleax:
Treat AggregationCursor instances like Cursor instances in the output
handling functions.

Simplify `CursorIterationResult` so that we don’t need a separate
class for its printable version, since it is already supposed
to be printable.

### DIFF
--- a/packages/shell-api/src/result.ts
+++ b/packages/shell-api/src/result.ts
@@ -1,6 +1,5 @@
 import { ShellApiClass, shellApiClassDefault } from './decorators';
 import { shellApiType, asPrintable } from './enums';
-import { addHiddenDataProperty } from './helpers';
 import { Document, ObjectIdType } from '@mongosh/service-provider-core';
 
 @shellApiClassDefault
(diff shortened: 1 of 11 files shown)


mcasimir (approved):
cool!

packages/shell-api/src/result.ts
```diff
@@ -108,28 +107,14 @@ export class DeleteResult extends ShellApiClass {
...
-    this.cursorHasMore = cursor.hasMore;
-  }
-}
-
-// NOTE: because this is inherited, the decorator does not add attributes. So no help() function.
 @shellApiClassDefault
-export class CursorIterationResult extends Array {
-  [shellApiType]: string;
-  hasMore: boolean;
+export class CursorIterationResult extends ShellApiClass {
```
mcasimir, line 111: Do we still need this to be a class? perhaps we can just ditch this thing / turn it into an interface now that there is no special behaviour, would be one concept less to understand.

addaleax: I don’t think turning this into an interface would make this easier to deal with necessarily – for now, we’re at least still using the feature that this automatically assigns the right shell API type this way.

(Or, another way around … I think if we got rid of this, I think we’d still want to have a factory function that generates these objects in order to avoid code duplication, and classes are basically factory functions)

PR #610, opened 2021-02-03
fix(shell-api): fix KeyVault.createKey signature MONGOSH-546

addaleax:
This was incorrect before (as mentioned in the ticket), but now
that we’ve updated `mongodb-client-encryption` this actually
becomes necessary, since no valid value for `masterKey` can
be specified when the kms is `local`.


rose-m (approved):
LGTM 👍

PR #612, opened 2021-02-03
chore(build): separate draft from publish MONGOSH-554

rose-m:
This PR separates the following stages / commands:
* `package`: As previously this produces the packaged artifacts (but does not upload them anywhere)
* `upload`: Uploads the artifacts _only to evergreen_ so they are available for smoke and verification tests and later draft and publish steps
* `draft`: Only runs when the commit was tagged with a draft release tag. Grabs the artifacts from evergreen and uploads them to Download Center as well as a draft GitHub release. The `release_draft` task now uses task tags to declare its dependencies.
* `publish`: Only runs when the commit was tagged with a release tag. Uses the known pre-uploaded artifacts to publish to Download Center, releases via Barque, etc.

For the `publish` to work as intended, the evergreen configuration has to be adapted so that we _only_ run the `release_publish` task by using the task tag pattern `publish`.


packages/build/src/github-repo.ts
```diff
@@ -60,7 +57,15 @@ export class GithubRepo {
         this.repo,
       );
 
-    return tags.find(({ commit }) => commit.sha === sha);
+    const sortedDraftTags = tags
+      .filter(t => t.name && t.name.startsWith(`v${releaseVersion}`) && t.name.match(/^v\d+\.\d+\.\d+-draft\.\d+/))
```
addaleax, line 61:
```suggestion
      .filter(t => t.name && t.name.startsWith(`v${releaseVersion}-draft`) && t.name.match(/^v\d+\.\d+\.\d+-draft\.\d+/))
```

Otherwise this would match `v0.7.10` for a `releaseVersion` of `v0.7.1`?

packages/build/src/github-repo.ts
```diff
@@ -60,7 +57,15 @@ export class GithubRepo {
...
+      .map(t => ({
+        name: t.name,
+        draftVersion: semver.prerelease(t.name)?.[1] as string
+      }))
+      .filter(t => t.draftVersion !== undefined)
+      .sort((t1, t2) => parseInt(t1.draftVersion, 10) < parseInt(t2.draftVersion, 10) ? 1 : -1);
```
addaleax, line 67: I think this can just be

```suggestion
      .sort((t1, t2) => t1.draftVersion - t2.draftVersion);
```

packages/build/src/github-repo.ts
```diff
@@ -60,7 +57,15 @@ export class GithubRepo {
...
+    return sortedDraftTags.length ? { name: sortedDraftTags[0].name } : undefined;
```
addaleax, line 68:
```suggestion
    return sortedDraftTags[0]?.name;
```

addaleax (commented):
Nice job!

Since this publishes the npm packages from the current working tree, should this also verify that the git sha for the latest draft release and the current HEAD when running `publish` are identical?

packages/build/src/run-package.ts
```diff
@@ -0,0 +1,31 @@
...
+export async function runPackage(
+  config: Config,
+): Promise<TarballFile> {
+  await fs.copyFile(await downloadMongocrypt(), config.mongocryptdPath, fsConstants.COPYFILE_FICLONE);
+
+  const runCreateTarball = async(): Promise<TarballFile> => {
+    return await createTarball(
+      config.outputDir,
+      config.buildVariant,
+            config.packageInformation as (Required<Config>['packageInformation'])
```
addaleax, line 18:
```suggestion
      config.packageInformation as (Required<Config>['packageInformation'])
```

packages/build/src/tarball.ts
```diff
@@ -322,52 +330,40 @@ export type TarballFile = { path: string; contentType: string };
...
+    default:
+      throw new Error(`Unhandled build variant: ${buildVariant}`);
   }
-  await tarballWindows(packageInformation, tarballFilename);
 
   return {
-    path: tarballFilename,
-    contentType: 'application/zip'
+    ...tarballFile,
+    path: fullTarballFilePath
```
addaleax, line 367: I opened a separate PR with the changes to this file in https://github.com/mongodb-js/mongosh/pull/615, because I’ll have to modify it and this patch is not something I’d want to rebase against :D

packages/build/src/github-repo.ts
rose-m: Looks different now since I'm returning both tag name as well as SHA.

packages/build/src/github-repo.ts
rose-m: Thanks for spotting 👍

rose-m:
I've added the check for the commit hash. This should work currently using `config.revision` as we are not releasing from patch builds (yet?).

addaleax (approved):
:+1:

PR #619, opened 2021-02-04
feat(build): add MSI artifact MONGOSH-125

addaleax:


packages/build/src/tarball.spec.ts
```diff
@@ -155,6 +155,24 @@ describe('tarball module', () => {
       });
     });
 
+    describe('.tarballWindowsMSI', () => {
+      // rpmbuild is not available everywhere. When it's not there, we skip
```
rose-m, line 159:
```suggestion
      // WIX is not available everywhere. When it's not there, we skip
```

packages/build/src/tarball.spec.ts
```diff
@@ -155,6 +155,24 @@ describe('tarball module', () => {
...
+      // the part of the test that would actually generate the RPM.
```
rose-m, line 160:
```suggestion
      // the part of the test that would actually generate the MSI.
```

packages/build/src/build-variant.ts
```diff
@@ -6,6 +6,7 @@
  */
 enum BuildVariant {
   Windows = 'win32',
+  WindowsMSI = 'win32msi',
```
rose-m, line 9: Mh should add this to `ALL_BUILD_VARIANTS` just immediately below? 🤔 Or are there reasons we would not want it in there - then I'd have to think of a better variable name for that array...

There are tests checking that array so everything is accounted for - but I should probably extend them so that they should already fail when you modify the enum without adapting the array :D

Once that is in there `run-draft` will automatically pick it up and upload to GitHub / DownloadCenter.

rose-m (commented):
That reminds me of the times a decade ago where I was once tasked with creating an MSI - it was already WiX back then... I'm getting old... 👴

packages/build/src/build-variant.ts
addaleax: Well... I'll add it and we see what happens 🤷‍♀️

packages/build/src/build-variant.spec.ts
```diff
@@ -4,42 +4,13 @@ import BuildVariant, { ALL_BUILD_VARIANTS } from './build-variant';
...
-    it('returns debian', () => {
-      expect(BuildVariant.Debian).to.equal('debian');
-    });
-  });
-
-  describe('BuildVariant.Redhat', () => {
-    it('returns rhel', () => {
-      expect(BuildVariant.Redhat).to.equal('rhel');
-    });
-  });
```
addaleax, line 44: For context, I've deleted this because this just checks that TypeScript enums work, not any mongosh functionality that we rely on

rose-m (approved):

PR #620, opened 2021-02-04
chore(ci): add FLE smoke test MONGOSH-543

addaleax:
Run a FLE script that verifies that automatic FLE using mongocryptd
works. This uses the same endpoint as the connectivity tests.

This also involves changing the mongocryptd binary we use for .deb
packages: The one from Ubuntu 16.04 only works there, so we use
the one from Ubuntu 18.04 instead. Ubuntu 16.04 users will have
to install mongocryptd manually (unless we decide to offer multiple
.deb packages at some point).


rose-m (approved):
👍👍

PR #629, opened 2021-02-08
chore(shell-api): add FLE integration tests MONGOSH-496

addaleax:
(Blocked on #553, and thus still hitting a slightly moving target here)

Provide fake KMS services in order to test local, AWS, GCP and Azure
FLE integration.

Also introduce an undocumented (unless we want to document it)
`bypassAutoEncryptionFully` that lets us create `Mongo` objects
which do not try to perform automatic encryption themselves,
but which still provide working `.getClientEncryption()` and
`.getKeyVault()` methods (which is otherwise impossible).

### DIFF
--- a/packages/shell-api/src/mongo.ts
+++ b/packages/shell-api/src/mongo.ts
@@ -62,6 +62,7 @@ export default class Mongo extends ShellApiClass {
   private _keyVault: KeyVault | undefined; // need to keep it around so that the ShellApi ClientEncryption class can access it
   private _clientEncryption: ClientEncryption | undefined;
   private _readPreferenceWasExplicitlyRequested = false;
+  private _bypassAutoEncryptionFully = false;
 
   constructor(
     internalState: ShellInternalState,
@@ -79,6 +80,11 @@ export default class Mongo extends ShellApiClass {
     }
     this._readPreferenceWasExplicitlyRequested = /\breadPreference=/.test(this._uri);
     if (fleOptions) {
+      if (fleOptions.bypassAutoEncryptionFully !== undefined) {
+        fleOptions = { ...fleOptions };
+        this._bypassAutoEncryptionFully = !!fleOptions.bypassAutoEncryptionFully;
+        delete fleOptions.bypassAutoEncryptionFully;
+      }
       this._fleOptions = processFLEOptions(fleOptions, this._serviceProvider);
       const { mongocryptdSpawnPath } = this._internalState;
       if (mongocryptdSpawnPath) {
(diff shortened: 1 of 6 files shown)


packages/shell-api/src/mongo.ts
```diff
@@ -79,6 +80,11 @@ export default class Mongo extends ShellApiClass {
     }
     this._readPreferenceWasExplicitlyRequested = /\breadPreference=/.test(this._uri);
     if (fleOptions) {
+      if (fleOptions.bypassAutoEncryptionFully !== undefined) {
```
addaleax, line 83: I think we’ll want to turn this concept into an officially supported option, otherwise there would currently be no way to get a `KeyVault` or `ClientEncryption` object without (enterprise-only) automatic encryption enabled

testing/fake-kms.ts
```diff
@@ -0,0 +1,75 @@
...
+    }
+  });
+}
+
+export function fakeGCPKMS(): NodeJS.ReadableStream & NodeJS.WritableStream {
+  return makeFakeHTTP(({ body, url }) => {
+    if (url.endsWith('/token')) {
+      return oauthToken;
+    } else if (url.endsWith(':encrypt')) {
+      // Here we also just perform noop encryption.
```
rose-m, line 69: That's my preferred encryption - noop encryption. Such performance, much wow 🐶

packages/shell-api/src/mongo.ts
rose-m: I'd also be fine with doing that - the key thing we don't want to expose is auto encryption but IMHO there's nothing against using the encryption primitives (KeyVault / ClientEncryption) directly... Probably clarify with @mmarcon ?

rose-m (approved):

packages/shell-api/src/mongo.ts
mmarcon: Can I get a 5 minutes overview of what is happening here and what y'all are suggesting we do?

rose-m: So the suggestion is to enable access to instances of `ClientEncryption` and `KeyVault` event if the mongo instance was not created with auto-encryption enabled. This will allow us to better test things but of course does not make auto-encryption available to everyone (it's still the enterprise only thing).

mmarcon: Sounds good to me, I don't see any harm in doing so.

addaleax: Ok, I’ll try to come up with a good name + open a separate ticket/PR so that this can be documented as public API

PR #635, opened 2021-02-09
fix(compass-shell): make max height 95% view height instead of 800 pixels COMPASS-4629

Anemy:
Screen Shot 2021-02-09 at 9 50 20 PM


rose-m (approved):
Space 🛰️

PR #637, opened 2021-02-10
feat(shell-api): add/rename explicitEncryptionOnly option MONGOSH-590

addaleax:
Rename the previously introduced option with a better name, in order
to provide a good public API, and add a check that verifies that this
is not used in combination with `schemaMap` (because the options would
contradict each other).


rose-m (approved):
Good point - that sounds a lot better and matches the `automatic` 👍

PR #651, opened 2021-02-11
chore(build): streamline release command execution MONGOSH-534

rose-m:
Use `ts-node` to directly run the entry script in `build/src/index.ts`.


addaleax (commented):
Fwiw, there’s nothing in here that would stick out to me…

packages/build/src/run-publish.ts
```diff
@@ -26,11 +26,6 @@ export async function runPublish(
   shouldDoPublicRelease: typeof shouldDoPublicReleaseFn = shouldDoPublicReleaseFn,
   getEvergreenArtifactUrl: typeof getArtifactUrlFn = getArtifactUrlFn
 ): Promise<void> {
-  console.info(
-    'mongosh: beginning publish release with config:',
-    redactConfig(config)
-  );
```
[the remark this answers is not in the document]
rose-m: Yeah it is already there at the top of `release` so it was duplicate in some cases

addaleax (approved):

.evergreen/package-and-upload-artifact.sh
```diff
@@ -4,11 +4,14 @@ set -e
 cat <<RELEASE_MONGOSH > ~/release_mongosh.sh
 set -e
 cd $(pwd)
+
 export NODE_JS_VERSION=${NODE_JS_VERSION}
-export ARTIFACT_URL_FILE=artifact-url.txt
+export ARTIFACT_URL_FILE="$PWD/artifact-url.txt"
```
addaleax, line 9: If you want to use absolute paths on Windows, you’ll need to use `cygpath -w` on the filepath before passing it to Node.js (because this would be `/cygdrive/c/.../artifact-url.txt`, which on Windows would be interpreted as a path relative to the FS root, e.g. `C:\cygdrive\c\...\artifact-url.txt`)

rose-m: I hope the way I've included it now makes that part work...

.evergreen/package-and-upload-artifact.sh
```diff
@@ -37,6 +40,10 @@ else
     # Verify signing
     spctl -a -vvv -t install dist/mongosh
   fi
+  if [ "$(uname)" == CYGWIN ]; then
+    # Fix absolute path before handing over to node
+    export ARTIFACT_URL_FILE=$(cygwin -w "$ARTIFACT_URL_FILE")
+  fi
```
addaleax, line 46:
```suggestion
  if [ "$OS" == "Windows_NT" ]; then
    # Fix absolute path before handing over to node
    export ARTIFACT_URL_FILE="$(cygpath -w "$ARTIFACT_URL_FILE")"
  fi
```

(We use the same check for cygwin elsewhere too, and there might be spaces in the paths as well, even though that’s currently not a problem on evergreen itself I think)

addaleax (approved):

PR #658, opened 2021-02-15
chore(ci): skip latest server alpha on macOS for now

addaleax:
See MONGOSH-592, this should be skipped until SERVER-54508 is resolved.

### DIFF
--- a/.evergreen.yml
+++ b/.evergreen.yml
@@ -441,16 +441,18 @@ tasks:
         variant: darwin
       - name: test_m44x_n12
         variant: darwin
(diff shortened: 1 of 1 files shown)


rose-m (approved):
LGTM! Alternative is we could also just change the `depends_on` statuses to `*` so it will ignore failures. But that would be a larger change presumably.

addaleax:
No strong preferences -- but I think it's fine to omit these entirely, they have been failing at a pretty high rate and we know that they won't work until the server ticket is fixed anyway

PR #663, opened 2021-02-16
chore(build): decouple tasks from build variant names MONGOSH-534

rose-m:
Uses an explicit expansion to define the "variant" the distribution should be built for. Also uses task tags for easier dependency definition.

### DIFF
--- a/.evergreen.yml
+++ b/.evergreen.yml
@@ -140,13 +140,13 @@ functions:
         env:
           NODE_JS_VERSION: ${node_js_version}
           DISTRO_ID_OVERRIDE: ${distro_id}
-          BUILD_VARIANT: ${build_variant}
+          DISTRIBUTION_BUILD_VARIANT: ${distribution_build_variant}
     - command: s3.put
       params:
         aws_key: ${aws_key}
         aws_secret: ${aws_secret}
         local_file: src/artifact-url.txt
-        remote_file: mongosh/binaries/${revision}/${revision_order_id}/artifact-url-${build_variant}.txt
+        remote_file: mongosh/binaries/${revision}/${revision_order_id}/artifact-url-${distribution_build_variant}.txt
         bucket: mciuploads
         permissions: public-read
         content_type: application/x-gzip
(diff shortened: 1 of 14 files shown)


packages/build/src/barque.spec.ts
```diff
@@ -112,44 +112,44 @@ describe('Barque', () => {
 
   describe('.determineDistro', () => {
     it('determines distro for ubuntu', async() => {
-      const distro = barque.determineDistro('linux');
+      const distro = barque.determineDistro(BuildVariant.Linux);
```
addaleax, line 115: Can we make this simpler rather than more complex, and just make `type BuildVariant = 'linux' | 'debian" ...`, or, ideally, `type BuildVariant = 'tgz' | 'deb' | 'rpm' | 'msi' | ...`?

.evergreen.yml
```diff
@@ -431,28 +443,16 @@ tasks:
...
+      # Remove the !.mlatest when this issue is resolved
       # https://jira.mongodb.org/browse/MONGOSH-592
-      #- name: test_mlatest_n12
-      #  variant: darwin
-      - name: test_m40x_n14
-        variant: darwin
-      - name: test_m42x_n14
-        variant: darwin
-      - name: test_m44x_n14
+      - name: ".unit-test !.mlatest"
```
addaleax, line 454: This means we're going to return to having failing runs in the waterfall, just without other tasks depending on them?

rose-m: Yes, this would have the red boxes back but they would not depend on it. So we'd just see when it works again. I could still leave them commented out of course!

addaleax: Ok -- I don't care much, but I think we'll see when this works again anyway when the SERVER ticket is resolved :)

packages/build/src/barque.spec.ts
addaleax: Yeah, I think that's the point of the ticket -- try to actually be intentional about what we're differentiating here ... we can also replace the name "build variant" if you have a good idea for an alternative :)

rose-m: Mh okay - unfortunately it's not really replaceable directly. We're using the Build variant as also the system identifier to construct our tarball names. So we have `mongosh-v1.0.0-win32.zip` for example. In the end that change you commented is not making it more complex but more explicit. I changed the signature to just use the type of the enum.

addaleax: Hm yeah -- it's just confusing that neither `linux` nor `rhel` nor `debian` actually do what they stand for, they're just differentiating the package formats and not the OSes... I don't think it's bad to leave this as-is, but if we can move away from it in some way, we probably should

PR #667, opened 2021-02-18
chore(build): add local release command MONGOSH-530

rose-m:


packages/build/src/helpers/user-input.ts
```diff
@@ -0,0 +1,37 @@
...
+      });
+
+      rl.question(`${prompt} `, (answer) => {
+        rl.close();
+        resolve(answer);
+      });
+    } catch (e) {
+      reject(e);
+    }
+  });
```
addaleax, line 19:
```suggestion
  return new Promise(resolve => {
    const rl = readline.createInterface({
      input: process.stdin,
      output: process.stdout
    });

    rl.question(`${prompt} `, (answer) => {
      rl.close();
      resolve(answer);
    });
  });
```

(Exceptions from inside the `Promise` constructor’s callback are turned into rejections automatically :))

packages/build/src/local/get-latest-tag.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    return {
+      semverName: semverTag,
+      releaseVersion: `${semver.major(semverTag)}.${semver.minor(semverTag)}.${semver.patch(semverTag)}`,
+      draftVersion: prerelease ? parseInt(prerelease[1], 10) : undefined
+    };
+  }).filter(t => !!t) as TagDetails[];
+
+  return { commit, tags: tagDetails };
+}
```
addaleax, line 67: Would we maybe better start out with `git tag`, looking for the tag that we actually care about, and then use `git log -n1 --pretty='%H' <tag>` to get the commit hash?

packages/build/src/local/trigger-release-draft.ts
```diff
@@ -0,0 +1,98 @@
...
+  if (!repositoryStatus.branch?.local) {
+    throw new Error('Could not determine local repository information - please verify your repository is intact.');
+  }
+
+  if (repositoryStatus.branch?.local !== 'master') {
+    throw new Error('Triggering a draft release is only allowed from master branch');
+  }
+  if (!repositoryStatus.branch.tracking) {
+    throw new Error('The branch you are on is not tracking any remote branch. Ensure you are on the master branch of the repository.');
+  }
```
addaleax, line 63: These things are actually moving us further away from being able to do support releases, right? Could we maybe prompt the user instead to make sure they’re on the right branch?

packages/build/src/local/trigger-release-draft.ts
```diff
@@ -0,0 +1,98 @@
...
+  if (!confirmed) {
+    throw new Error('User aborted.');
+  }
+
+  console.info('... creating and pushing tag ...');
+  spawnSync('git', ['tag', nextTagName], {
+    cwd: repositoryRoot,
+    encoding: 'utf-8'
+  });
+  spawnSync('git', ['push', '--tags'], {
```
addaleax, line 44: I would maybe recommend only pushing the tag that we created here, instead of any tags the user might have lying around as well

rose-m: I currently fail if the user has unpushed tags - so here it would not push anything additional. But good point, might change that to only push that specific tag.

packages/build/src/local/trigger-release-draft.ts
rose-m: Yeah it depends probably on how we want to do them. At least a tracking branch should be there. Right now I've put in the `master` guard just for good measure. I'd assume we create a separate branch for a support release and tag on that branch. But for that it should definitely be pushed to the remote (thus we should have a `tracking` branch).

packages/build/src/local/get-latest-tag.ts
rose-m: Yeah sounds good to me, too. Probably the main difference is that `git tag` gives us all tags in the repo we already know whereas the `git log` check returns only those from the current branch. 

How is that in the case of future support releases? 🤔 <- That question I'd rather answer in a different ticket.

So changing to `git tag` and grabbing hash afterwards for now!

packages/build/src/helpers/user-input.ts
rose-m: Huh, that makes sense :D

packages/build/src/local/trigger-release-draft.ts
addaleax: Yeah, I wasn’t referring to the tracking branch check specifically, sorry :) Maybe we could do some very crude regex check here, e.g. `/^master$|^main$|^v\d+\.\d+\.\d+/` (main because of https://jira.mongodb.org/browse/MONGOSH-245)?

rose-m: That's a good idea 👍 So we'd cover `master`/`main` and potential support release branches (which would equal to the designated version number)... ✅

addaleax: Hm yeah … and maybe also allow `[xyz0-9]` instead of `\d`, I don’t think we have any idea what the branch naming scheme will look like but using things like `v0.x.y` or `v0.8.x` would be common

addaleax (approved):

PR #670, opened 2021-02-18
chore(cli-repl): add Segment API integration test MONGOSH-363

addaleax:
Includes #669 because it’s needed here.

Add a test that verifies the actual API key bundled into the actual
release binary. In order to achieve this, add an internal test command
that performs a Segment API request and verifies that its callback
does not return an error.

Note that this is, currently, a test that with somewhat limited
meaningfullness, because the Segment API currently always gives
successful HTTP responses, regardless of the validitity of the API key.

(As such, I would also consider it okay to not merge this. It still tests
that we can reach the Segment API endpoint, though, i.e. the 
`analytics-node` package is properly configured.)

### DIFF
--- a/packages/cli-repl/src/arg-parser.ts
+++ b/packages/cli-repl/src/arg-parser.ts
@@ -45,6 +45,7 @@ const OPTIONS = {
   boolean: [
     'async',
     'help',
+    'internalTestCommands',
     'ipv6',
     'nodb',
     'norc',
@@ -87,6 +88,7 @@ function getLocale(args: string[], env: any): string {
   return lang ? lang.split('.')[0] : lang;
 }
 
+type AllCliOptions = CliOptions & { smokeTests: boolean, internalTestCommands: boolean };
 /**
  * Parses arguments into a JS object.
  *
@@ -94,7 +96,7 @@ function getLocale(args: string[], env: any): string {
  *
  * @returns {CliOptions} The arguments as cli options.
  */
-function parse(args: string[]): (CliOptions & { smokeTests: boolean }) {
+function parse(args: string[]): AllCliOptions {
   const programArgs = args.slice(2);
   i18n.setLocale(getLocale(programArgs, process.env));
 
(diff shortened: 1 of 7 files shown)


mmarcon:
> the Segment API currently always gives
successful HTTP responses, regardless of the validitity of the API key
😮

rose-m (commented):
That's where your tweet came from 😂 

Like the solution that it really test the analytics in a "smoke" fashion 👍 

On the other hand I have the tiny feeling that we are starting to make more and more exceptions and "tiny balcony" options here and there for tests, etc. Maybe I'm also just overreacting here but I just want to ensure we're not going down exception hell already 😅 
If I understood the ticket correctly it would maybe be sufficient to "just" run an API call from evergreen as we have other tests to verify that the analytics are properly generated/included [well, if we could see from the API call if it works at all].

addaleax:
> On the other hand I have the tiny feeling that we are starting to make more and more exceptions and "tiny balcony" options here and there for tests, etc. Maybe I'm also just overreacting here but I just want to ensure we're not going down exception hell already sweat_smile

I don’t really think we’re running into that problem – it’s still just a tiny fraction of our overall tests that makes this up.

> If I understood the ticket correctly it would maybe be sufficient to "just" run an API call from evergreen as we have other tests to verify that the analytics are properly generated/included [well, if we could see from the API call if it works at all].

Well … we have tests that verify that, and we have tests that verify that we can use analytics-node using a fake HTTP server – if we’re not testing that the actual functionality from the shell works, then I’m not sure that there’s a point in testing this at all :)

rose-m:
I understood it as such that the API key needs to be verified to be valid as we might generate a new one in the future / use a wrong expansion / etc. So that's what a "simple" GET at the right place would verify. Of course the smoke test solution verifies the whole chain 👍

addaleax:
Right, but as I mentioned, this currently doesn’t work with the segment API. Should we just not do this?

rose-m:
I'd leave it out for now, yes. So 🤦 that their API doesn't allow to check the keys...

addaleax:
Cool, I'll close this as wontfix then. We can always revisit if something changes in their API :)

PR #672, opened 2021-02-18
fix(build): update github release assets properly MONGOSH-606

rose-m:
Easier than thought - getting a release by tag already returns the uploaded assets with it (see https://docs.github.com/en/rest/reference/repos#get-a-release-by-tag-name - default response example).


addaleax (approved):
:crossed_fingers:

PR #674, opened 2021-02-19
fix(service-provider-core): validate --host option more strictly

addaleax:
This came up in a recent conversation in slack, where passing
multiple hosts through the `--host` option separated by commas
did not fail properly and instead led to connecting to the wrong
set of hosts.

### DIFF
--- a/packages/service-provider-core/src/uri-generator.ts
+++ b/packages/service-provider-core/src/uri-generator.ts
@@ -63,7 +63,7 @@ function validateConflicts(options: CliOptions): void {
  * @param {string} host - The value of the --host option.
  */
 function validateHost(host: string): void {
-  const invalidCharacter = host.match(/[^a-zA-Z0-9\.-:\[\]]/);
+  const invalidCharacter = host.match(/[^a-zA-Z0-9.:\[\]-]/);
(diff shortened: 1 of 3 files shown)


packages/service-provider-core/src/uri-generator.ts
```diff
@@ -41,17 +41,36 @@ const DEFAULT_PORT = '27017';
...
   }
 }
 
+/**
+ * Perform basic validation of the --host option.
+ *
+ * @param {string} host - The value of the --host option.
+ */
+function validateHost(host: string): void {
+  const invalidCharacter = host.match(/[^a-zA-Z0-9\.-:\[\]]/);
```
rose-m, line 66: Is the `-` mis-treated in the regex here? I think it should be at the end, right? And we can skip escaping of `.` inside the list.
```suggestion
  const invalidCharacter = host.match(/[^a-zA-Z0-9.:\[\]-]/);
```

addaleax: you are right, yes :) thank you!

rose-m (approved):
Check 😄 ✅

PR #678, opened 2021-02-23
chore(build): ensure to remove a GitHub asset before re-uploading

rose-m:
Updating an asset does not work with a PATCH call - only the
name of an asset and other properties can be updated that
way, not the binary itself.


gribnoysup (approved):
Nice fix! Weird that they don't allow a direct update, but 🤷

PR #680, opened 2021-02-24
chore(build): generate release notes on draft MONGOSH-325

rose-m:
This blatantly re-uses large parts of the code we have in Compass for generating the release notes there.

Noteworthy changes to that code are:
* Migrated to TypeScript
* Detects both `MONGOSH` and `COMPASS` references
* Code for detecting tags etc. was stripped as this happens elsewhere in the process


packages/build/src/run-draft.spec.ts
```diff
@@ -0,0 +1,190 @@
...
+
+      await ensureGithubReleaseExistsAndUpdateChangelogFn(
+        '0.8.1', 'v0.8.1', githubRepo, generateChangelog
+      );
+      expect(getPreviousReleaseTag).to.have.been.called;
+      expect(generateChangelog).to.have.been.called;
+      expect(updateDraftRelease).to.have.been.calledWith({
+        name: '0.8.1',
+        tag: 'v0.8.1',
+        notes: 'Release notes [in Jira](https://jira.mongodb.org/issues/?jql=project%20%3D%20MONGOSH%20AND%20fixVersion%20%3D%200.8.1)\n\nNice changelog'
```
mcasimir, line 167: shouldn't this be 'Nice changelog'?

rose-m: No, only the `changelog` function in the `git` package generates the pure commit stuff. The `ensure...` also adds the release notes link (it didn't make sense for me to include this part into `git`, as it is not git relevant).

packages/build/src/run-draft.spec.ts
```diff
@@ -0,0 +1,190 @@
...
+  });
+
+  describe('runDraft', () => {
+    let ensureGithubReleaseExistsAndUpdateChangelog: typeof ensureGithubReleaseExistsAndUpdateChangelogFn;
+
+    beforeEach(() => {
+      ensureGithubReleaseExistsAndUpdateChangelog = sinon.stub();
+    });
+
+    it('works as expected', async() => {
```
mcasimir, line 76: just a super silly thing, maybe this could be broken in a way that who reads understand what is the expected behavior behind 'works as expected', something like:

```
context('when there is a triggeringGitTag', () => {
   beforeEach(() => runDraft(...));
   
   it ('generates a changelog', () => {
      expect(ensureGithubReleaseExistsAndUpdateChangelog).to.have.been.calledWith(
        config.version, `v${config.version}`, githubRepo
      );
   })
   
   /// ....
})
```

packages/build/src/run-draft.spec.ts
mcasimir: What I meant is, do we keep the `Release notes [in Jira](https://jira.mongodb.org/issues/?jql=project%20%3D%20MONGOSH%20AND%20fixVersion%20%3D%200.8.1)` too?

Maybe we could move them something like in the end / make sure is not confusing what's what? I guess it doesn't make much of a difference though.

mcasimir (approved):
🎉

packages/build/src/run-draft.spec.ts
rose-m: Yeah good point, that makes it a lot more explicit and readable, like that.

packages/build/src/run-draft.spec.ts
rose-m: Ah understood... I rephrased it to: `See an overview of all solved issues in Jira` and append it instead of prepend it to the changelog.

PR #681, opened 2021-02-26
chore(build): fix detection of tasks for latest draft build

rose-m:
The Evergreen REST API returns all tasks that ever ran _based_ on the commit SHA. This includes multiple waterfall builds (e.g.  because these were triggered by tags), and also patch builds that are based on the given SHA.

We therefore have to filter based on the draft tag we are looking for.


mcasimir (approved):
👍

gribnoysup (approved):
Thanks for fixing this!

PR #682, opened 2021-02-26
chore(build): ensure packages are published for all distros MONGOSH-483

rose-m:
The issue was caused because it tried to publish the `-linux.tgz` files to the Ubuntu PPA instead of the `.deb` packages.

It is more or less "safe" to test for a release since publishing to Barque is the first step in the publishing stage. So in the worst case we will have to re-start publishing.

### DIFF
--- a/packages/build/src/barque.ts
+++ b/packages/build/src/barque.ts
@@ -1,12 +1,12 @@
 import childProcess from 'child_process';
+import fs from 'fs-extra';
 import gunzip from 'gunzip-maybe';
 import fetch from 'node-fetch';
-import tmp from 'tmp-promise';
-import stream from 'stream';
-import fs from 'fs-extra';
-import tar from 'tar-fs';
-import util from 'util';
 import path from 'path';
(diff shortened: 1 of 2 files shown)


packages/build/src/barque.ts
```diff
@@ -36,9 +37,8 @@ enum Distro {
  * This can be also moved to /config/build.conf.js in the future.
  */
 enum Arch {
-  Ubuntu = 'amd64',
-  Debian = 'amd64',
-  Redhat = 'x86_64',
+  AMD64 = 'amd64',
```
mcasimir, line 40: ❤️ ❤️ ❤️

packages/build/src/barque.ts
```diff
@@ -36,9 +37,8 @@ enum Distro {
...
+  X86_64 = 'x86_64'
```
mcasimir, line 41: super non-important, I think the convention for ts enums is camelcase

mcasimir (approved):
Nice!

PR #684, opened 2021-03-01
chore(build): document new release process MONGOSH-533

rose-m:
I've removed the super detailed explanation and only describe the bigger picture, otherwise I'm afraid the documentation will become immediately out of date. In the end for specific details looking at the code is probably not avoidable.

Note that the source for the SVG images is draw.io and the draw.io file is also checked in.


addaleax (approved):
Looks very cool and reads very nicely :)

PR #685, opened 2021-03-02
chore(build): wait until packages are really available in PPAs MONGOSH-619

rose-m:
This PR adds the following behavior:
* Publishing a distributable using Barque returns the URLs we expect the packages to be available at
* After all distributables are published we wait at most 5 minutes for them to become available checking every 10 seconds

Note: We are only doing a `HEAD` request instead of downloading the full package. Example to see with an existing package locally:
```
$ curl --head https://repo.mongodb.org/yum/redhat/8/mongodb-org/4.4/x86_64/RPMS/mongosh-0.8.2-x86_64.rpm
HTTP/2 200
...
```

**Open Question:** Should I calibrate timeout and interval to different values?

### DIFF
--- a/packages/build/src/barque.spec.ts
+++ b/packages/build/src/barque.spec.ts
@@ -49,17 +49,38 @@ describe('Barque', () => {
 
   describe('releaseToBarque', () => {
     context('platform is linux', () => {
-      it('execCurator function succeeds', async() => {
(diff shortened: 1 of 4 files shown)


addaleax (approved):
We can start with the current timeout and observe what the actual values look like when we do releases, and then maybe adjust if we want to, that seems fine to me

PR #695, opened 2021-03-03
chore(cli-repl): do not send telemetry for CI runs MONGOSH-611

addaleax:


rose-m (approved):

packages/cli-repl/src/cli-repl.ts
```diff
@@ -120,6 +124,9 @@ class CliRepl {
       this.bus,
       () => pino({ name: 'mongosh' }, logStream),
       () => {
+        if (process.env.IS_CI && !this.analyticsOptions?.alwaysEnable) {
```
mcasimir, line 127: Can this be picked also in user env, say by accident someone having an `IS_CI
`?

Also should we throw the error? Maybe here we can just do this:

const apiKey = process.env.WHATEVER_ENV_VAR_THAT_IS_REALLY_UNIQUE ? 'fake-key' :  options.apiKey

mcasimir: Then we need to set MONGOSH_SKIP_TELEMETRY also as a variable. For e2e tests should be fine.

We could also add another check that the version is not the placeholder one.

addaleax: > We could also add another check that the version is not the placeholder one.

I did that initially, but then I figured if we want to exclude CI, we probably also want to exclude release binaries from CI. No strong opinions, though.

> Can this be picked also in user env, say by accident someone having an `IS_CI `?

I mean, I guess? That user might also run into other problems if they do. I don't think we need to worry about this as a common case by any means.

> Also should we throw the error? Maybe here we can just do this:

Does it matter whether we throw an error or the segment package does? I'd rather be explicit here.

mcasimir: > we probably also want to exclude release binaries from CI

yeah we'd need to to so, also cause we do set the version quite early on in the CI. The idea to add an additional check for `dev` is to avoid cases where you have the api key on your computer (maybe one were trying the automation) and then run the shell, but is not so important. 

> I mean, I guess? That user might also run into other problems if they do.

I still didn't get would a user with `IS_CI` set get an error during normal execution from a binary?

If that's the case we should use a different name for that variable (maybe just add a prefix). Is common to run the shell in CI environments and `IS_CI` is quite a common variable name (although is still a matter of chances). Seems cheap enough for us to use a different and more unique name.

> Does it matter whether we throw an error or the segment package does? I'd rather be explicit here.

That is a good point, but why would you throw an error at all? We can just reset the api key to one that doesn't work or just a null one and let the segment package execute. After all that's what would happen anyway in the cli repl tests or in dev when you don't have the key file .. unless I'm missing something.

rose-m: Regarding the error: It's caught in the code when setting up telemetry and gets logged. Everything continues to work as usual as it will fall back to the `NoopAnalytics` stub that we have in place in case of any issues. So it's save for a user (no errors) and having an `IS_CI` set in your local env... what would the purpose be? I'd even see it as a benefit to not get telemetry from CI environments of our customers? But definitely no strong opinion.

mcasimir: > It's caught in the code when setting up telemetry and gets logged.

Nice that's all right then!

> having an IS_CI set in your local env... what would the purpose be

for your CI environment as we do. We definitely want the telemetry from the CI of our customers, that's a legit use case we want to collect data about. We just don't want ours.

addaleax: Hm yeah, that makes sense. I’ll use another env var then :+1:

rose-m (approved):
Now we should be safe 👍

PR #696, opened 2021-03-03
chore(build): add explicit E2E stage for all platforms

rose-m:
Introduces separate `e2e_tests_<platform>` tasks and runs them on every platform explicitly.

This PR also contains updated documentation - new picture preview:
image


.evergreen.yml
```diff
@@ -445,7 +479,58 @@ tasks:
         vars:
           node_js_version: "14.15.1"
 
+  ###
+  # E2E TESTS
+  ###
+  - name: e2e_tests_darwin
```
addaleax, line 485: Maybe this got lost when we switched to tag-based dependencies for tasks, but the draft publish task should also depend on these, right?

rose-m: Uh good point. Absolutely... I'll immediately add that. The arrows are in the picture but not in the YML :D :D

addaleax (approved):
:shipit:

PR #697, opened 2021-03-03
fix(build): add Windows .exe resource information MONGOSH-570

addaleax:
Pass information to boxednode, which is used to overwrite the Node.js
.rc file that is used for metadata (and the file icon) for the
generated executable.

Screenshot 2021-03-03 212719
Screenshot 2021-03-03 212800


rose-m (approved):
Nice 🎉

PR #700, opened 2021-03-04
chore(cli-repl): add SCRAM-SHA-1/256 integration tests

addaleax:
This deviates slightly from the tech design in that it doesn’t
spin up new servers, but instead specifies the required auth
mechanism when creating the users (and adding consistency checks
that verify that those auth mechanisms are actually required by
the server), in order to avoid the relatively costly spawning
of server instances and re-use more of the existing auth test code.


rose-m (approved):
I did not know you could limit it like that 😅 Nice!

PR #703, opened 2021-03-04
feat(cli-repl): add load() support MONGOSH-386

addaleax:
Paired on this with Michael because it seemed like a good ticket to
pair on and we’re going to be doing the loading-files epic soon.

### DIFF
--- a/packages/cli-repl/src/mongosh-repl.ts
+++ b/packages/cli-repl/src/mongosh-repl.ts
@@ -9,6 +9,7 @@ import type { MongoshBus, UserConfig } from '@mongosh/types';
 import askpassword from 'askpassword';
 import { Console } from 'console';
 import { once } from 'events';
+import path from 'path';
 import prettyRepl from 'pretty-repl';
 import { ReplOptions, REPLServer } from 'repl';
 import type { Readable, Writable } from 'stream';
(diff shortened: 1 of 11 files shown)


rose-m (approved):

packages/cli-repl/src/mongosh-repl.ts
```diff
@@ -24,18 +25,19 @@ export type MongoshCliOptions = ShellCliOptions & {
   redactInfo?: boolean;
 };
 
-export type MongoshConfigProvider = {
+export type MongoshIOProvider = {
```
mcasimir, line 28: just a non-important random observation: this type is getting to be a bit of a jack of all trades, did we consider having multiple interfaces that ties together only the methods that are naturally coupled?

mcasimir (approved):
This is so amazing that I don't even know what to say :) ❤️

packages/cli-repl/src/mongosh-repl.ts
addaleax: Yeah, so... we could definitely split it up, but I feel like this might be one of those things where it's okay to wait until we have a reason to do so (e.g. if we ever make MongoshRepl a public API, which seems possible)

No strong feelings though :)

mcasimir: Yeah, me neither, keeping it like that is super fine!

PR #712, opened 2021-03-09
chore(cli-repl): add E2E tests for AWS authentication MONGOSH-578

rose-m:
WIP due to the blocked AWS setup required for testing and specifying the required environment variables.

### DIFF
--- a/packages/cli-repl/test/test-shell.ts
+++ b/packages/cli-repl/test/test-shell.ts
@@ -5,7 +5,7 @@ import path from 'path';
 import stripAnsi from 'strip-ansi';
 import { eventually } from './helpers';
 
-
+export type TestShellStartupResult = { state: 'prompt' | 'exit'; exitCode?: number | undefined };
 type SignalType = ChildProcess extends { kill: (signal: infer T) => any } ? T : never;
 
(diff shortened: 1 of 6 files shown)


packages/cli-repl/test/test-shell.ts
```diff
@@ -7,6 +7,8 @@ import { eventually } from './helpers';
 
 
 type SignalType = ChildProcess extends { kill: (signal: infer T) => any } ? T : never;
+type ShellResultState = 'prompt' | 'exit';
+type ShellResult = { state: ShellResultState; exitCode?: number | undefined };
```
addaleax, line 11: Btw, we may want to pick another name here to avoid confusion with the "other" ShellResult: https://github.com/mongodb-js/mongosh/blob/c6eacd46ddcb9d0a3ad83754b667ba257f0bdc64/packages/shell-api/src/decorators.ts#L37-L53

rose-m: Thanks, didn't see that. I've renamed it to `TestShellStartupResult`.

addaleax (approved):
Not sure if this is ready for review yet, but LGTM :+1:

PR #717, opened 2021-03-10
feat(cli-repl): add tlsCertificateSelector support on Win32 MONGOSH-573

addaleax:
No tests so far (including no manual ones), and I think testing this automatically would have to involve stubbing out the `win-export-certificate-and-key` package.


packages/cli-repl/test/e2e-tls.spec.ts
```diff
@@ -181,7 +186,8 @@ describe('e2e TLS', () => {
       );
       const certUser = 'emailAddress=tester@example.com,CN=Wonderwoman,OU=DevTools Testers,O=MongoDB';
 
-      it('can connect with cert to create user', async() => {
+      before(async() => {
+        /* connect with cert to create user */
```
addaleax, line 190: ... took me half an hour to figure out why the server couldn't find the user :)

addaleax (commented):
This should be ready for review -- I've manually verified that this works with actual certificate/private key combinations from the Windows CA store.

packages/cli-repl/test/e2e-tls.spec.ts
rose-m: Sorrrryyy :D

packages/cli-repl/src/arg-parser.ts
```diff
@@ -106,9 +106,7 @@ const DEPRECATED_ARGS_WITH_REPLACEMENT: Record<string, string> = {
 const UNSUPPORTED_ARGS: Readonly<string[]> = [
   'eval',
   'sslFIPSMode',
-  'tlsFIPSMode',
-  'sslCertificateSelector',
```
rose-m, line 110: We should probably update the help text of the CLI to include the flags and also add a comment in there to say that it's currently only supported on Windows.

rose-m (approved):
Awesome :) I just think we should update the usage / help output and the README to document our newly supported flag :D

addaleax:
> Awesome :) I just think we should update the usage / help output and the README to document our newly supported flag :D

@rose-m I’ve updated the README/help output, but the flag was already documented – it should not have been, right?

rose-m:
Yeah it should have been removed by my other PR where I cleaned those up... anyway - now looks good ✅

PR #718, opened 2021-03-11
fix(cli-repl): add aws4 as dependency to cli-repl to enable AWS auth

rose-m:
Since the Homebrew formula uses the cli-repl package, the aws4
dependency has to be added to it in order for it to be available after
brewing it.

### DIFF
--- a/packages/service-provider-server/package-lock.json
+++ b/packages/service-provider-server/package-lock.json
@@ -60,6 +60,11 @@
 				"readable-stream": "^2.0.6"
 			}
 		},
+		"aws4": {
+			"version": "1.11.0",
+			"resolved": "https://registry.npmjs.org/aws4/-/aws4-1.11.0.tgz",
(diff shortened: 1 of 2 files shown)


packages/cli-repl/package.json
```diff
@@ -50,6 +50,7 @@
     "analytics-node": "^3.4.0-beta.1",
     "ansi-escape-sequences": "^5.1.2",
     "askpassword": "^1.2.3",
+    "aws4": "^1.11.0",
```
addaleax, line 53: This should be in service-provider-server if it's a peer dependency of the driver, which is also depended on there, right? (I think this currently would work by accident, because aws4 appears in this repository's top-level node_modules)

rose-m: Ah shoot... yeah it should go there then, absolutely! Interestingly it's not even a peer dependency of the driver, they only depend on `@types/aws4`.

addaleax (approved):

addaleax:
Side note: I really like how automatically this works: https://github.com/mongodb-js/mongosh/commit/ae4388554757d624bb49a57ccab289c4c62eb0ae

PR #722, opened 2021-03-17
fix(shell-api): follow-up findAndModify changes

addaleax:
Following up to the changes from b098674ea74 as suggested by @gribnoysup in https://github.com/mongodb-js/mongosh/pull/720#pullrequestreview-614289439


gribnoysup (approved):
Looks good!

PR #726, opened 2021-03-18
feat(cli-repl): support tlsCertificateSelector on macOS MONGOSH-574

rose-m:
Adds support for `--tlsCertificateSelector` on macOS using https://github.com/mongodb-js/macos-export-certificate-and-key.

The user will be prompted for their keychain password twice (which is default macOS behavior, same as with the old shell).

mongosh-574


packages/cli-repl/src/arg-mapper.ts
```diff
@@ -93,4 +91,27 @@ export function applyTlsCertificateSelector(
...
+      break;
+    case 'darwin':
+      packageName = 'macos-export-certificate-and-key';
+      break;
+    default:
+      return undefined;
+  }
+
+  try {
+    return require(packageName);
```
addaleax, line 112: Heads up: This is going to break bundling :) The `require()` call needs literals, unfortunately – I had run into this as well :)

addaleax (approved):

packages/cli-repl/src/arg-mapper.ts
rose-m: NOOOOO :D

PR #729, opened 2021-03-18
feat(cli-repl): allow loading files from command line MONGOSH-326

addaleax:


addaleax:
I’ve tacked on `--eval` support to this PR, as it’s very related and it probably makes sense to review them as one.

packages/cli-repl/src/cli-repl.ts
```diff
@@ -151,10 +151,41 @@ class CliRepl {
...
     await this.loadRcFiles();
     await this.mongoshRepl.startRepl(initialized);
   }
 
+  listCommandLineLoadFiles(): string[] {
+    const startIndex = this.cliOptions.nodb ? 0 : 1;
+    return (this.cliOptions._ ?? []).slice(startIndex);
+  }
+
+  async loadCommandLineFiles(files: string[]) {
```
rose-m, line 171: Nit: could we rename that to `loadCommandLineFilesAndEval`? Just from reading it seems more like a "side effect" that it also loads the eval code.

rose-m (approved):
Just a minor method name thing :)

packages/cli-repl/src/cli-repl.ts
addaleax: ah, sure :) will do that along with fixing the test that’s still borked up (which is due to nyc/coverage, it seems…)

rose-m (approved):

PR #738, opened 2021-03-22
MONGOSH-607: insert should not add id to users object

alenakhineika:
## Description

The legacy shell does not modify the object being inserted and does not add _id to this object. Keep the mongosh behavior compatible with the legacy shell behavior.

### Checklist
- [x] New tests and/or benchmarks are included
- [ ] Documentation is changed or added

## Motivation and Context

- [x] Bugfix
- [ ] New feature
- [ ] Dependency update
- [ ] Misc

## Types of changes

- [ ] *Backport Needed*
- [x] Patch (non-breaking change which fixes an issue)
- [ ] Minor (non-breaking change which adds functionality)
- [ ] Major (fix or feature that would cause existing functionality to change)

### DIFF
--- a/package.json
+++ b/package.json
@@ -127,6 +127,7 @@
     "mongodb": "4.0.0-beta.2",
     "mongodb-download-url": "^1.0.0",
     "mongodb-js-precommit": "^2.0.0",
+    "nock": "^13.0.11",
     "node-codesign": "^0.3.2",
     "node-fetch": "^2.6.1",
     "nyc": "^15.1.0",
--- a/packages/shell-api/src/collection.ts
+++ b/packages/shell-api/src/collection.ts
@@ -608,7 +608,7 @@ export default class Collection extends ShellApiClass {
       this._mongo._internalState.context.print
     );
     assertArgsDefined(docs);
-    const d: Document[] = Array.isArray(docs) ? docs : [docs];
+    const d: Document[] = Array.isArray(docs) ? docs.map((doc) => ({ ...doc })) : [{ ...docs }];
 
     this._emitCollectionApiCall('insert', { options });
     const result = await this._mongo._serviceProvider.insertMany(
@@ -641,12 +641,13 @@ export default class Collection extends ShellApiClass {
   @serverVersions(['3.2.0', ServerVersions.latest])
   async insertMany(docs: Document[], options: BulkWriteOptions = {}): Promise<InsertManyResult> {
     assertArgsDefined(docs);
+    const d: Document[] = Array.isArray(docs) ? docs.map((doc) => ({ ...doc })) : docs;
 
     this._emitCollectionApiCall('insertMany', { options });
     const result = await this._mongo._serviceProvider.insertMany(
       this._database._name,
       this._name,
-      docs,
+      d,
       { ...this._database._baseOptions, ...options }
     );
 
@@ -678,7 +679,7 @@ export default class Collection extends ShellApiClass {
     const result = await this._mongo._serviceProvider.insertOne(
       this._database._name,
       this._name,
-      doc,
+      { ...doc },
       { ...this._database._baseOptions, ...options }
     );
 
--- a/packages/shell-api/src/collection.ts
+++ b/packages/shell-api/src/collection.ts
@@ -608,13 +608,18 @@ export default class Collection extends ShellApiClass {
       this._mongo._internalState.context.print
     );
     assertArgsDefined(docs);
-    const d: Document[] = Array.isArray(docs) ? docs.map((doc) => ({ ...doc })) : [{ ...docs }];
+    // When inserting documents into MongoDB that do not contain the _id field,
+    // one will be added to each of the documents missing it by the Node driver,
+    // mutating the document. To prevent this behaviour we pass not the original document,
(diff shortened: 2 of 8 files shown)


packages/shell-api/src/collection.ts
```diff
@@ -641,12 +641,13 @@ export default class Collection extends ShellApiClass {
   @serverVersions(['3.2.0', ServerVersions.latest])
   async insertMany(docs: Document[], options: BulkWriteOptions = {}): Promise<InsertManyResult> {
     assertArgsDefined(docs);
+    const d: Document[] = Array.isArray(docs) ? docs.map((doc) => ({ ...doc })) : docs;
```
mcasimir, line 644: Isnt `: docs` also `[{ ...docs }]` ?

alenakhineika: Not really. I case of `insertMany` we expect only array as an input parameter, and if this is not an array but something else we want to throw the error that the docs parameter must be an array of documents, so we don't want to manually wrap it into the array.

package.json
```diff
@@ -127,6 +127,7 @@
     "mongodb": "4.0.0-beta.2",
     "mongodb-download-url": "^1.0.0",
     "mongodb-js-precommit": "^2.0.0",
+    "nock": "^13.0.11",
```
addaleax, line 130: Is this used somewhere?

packages/shell-api/src/integration.spec.ts
```diff
@@ -1061,7 +1104,8 @@ describe('Shell API (integration)', function() {
     });
 
     describe('aggregate', () => {
-      describe('up to server 4.4', () => {
+      describe('server from 4.2.2 till 4.4', () => {
+        skipIfServerVersion(testServer, '< 4.2.2');
```
addaleax, line 1108: We still want to test this on 4.0.x, I think.

```suggestion
      describe('server from 4.2.2 till 4.4 and 4.0.14 to 4.1', () => {
        skipIfServerVersion(testServer, '>= 4.2.0 < 4.2.2 || < 4.0.14');
```

addaleax (approved):

packages/shell-api/src/collection.ts
mcasimir: And it would not inject the _id if i pass an array?

packages/shell-api/src/collection.ts
```diff
@@ -608,7 +608,7 @@ export default class Collection extends ShellApiClass {
       this._mongo._internalState.context.print
     );
     assertArgsDefined(docs);
-    const d: Document[] = Array.isArray(docs) ? docs : [docs];
+    const d: Document[] = Array.isArray(docs) ? docs.map((doc) => ({ ...doc })) : [{ ...docs }];
```
mcasimir, line 611: why not keeping the variable name something more readable than 1 letter?

packages/shell-api/src/collection.ts
mcasimir: Oh sorry I got confused, you may want to write this in a bit more clear way: If we are assuming docs is already an array, then you can just do this:

``` js
const clonedDocs = docs.map(doc => {...doc});
```

mcasimir: this way is clear the intent, is hard to me to understand that we check fort a non legit condition and then we don't do anything about it, there is a lot of implicit going on.

package.json
alenakhineika: Yeah, here: https://github.com/mongodb-js/mongosh/blob/master/packages/build/src/barque.spec.ts#L3

packages/shell-api/src/collection.ts
alenakhineika: We could find a better name, i just updated the value.

mcasimir: true that, it was like that already :)

package.json
addaleax: Hm … but it’s in the `build` package’s `devDependencies`, so we shouldn’t have to add it here … but then again, I remember that somebody else was also running into this, maybe Irina? Maybe we could remove the one in the build package then, so that we don’t have to install it twice?

alenakhineika: It was throwing an error without adding it to package.json, like missing dependencies. I will try to remove it from build then.

packages/shell-api/src/integration.spec.ts
alenakhineika: Yes, wanted to add it and forgot after the weekend. Thanks a lot for pointing at it!

packages/shell-api/src/collection.ts
alenakhineika: Updated to make the name more clear.

packages/shell-api/src/collection.ts
alenakhineika: But in this case, if the user will pass not an array but an object for example the error will be that the map is not a function instead of server's error that parameter must be an array.

alenakhineika: So the idea here is that if the input parameter is an array, clone the input array to prevent mutation. If the input parameter is anything else but an array, pass the original value to the driver, it will throw the proper error and won't mutate the parameter because the command won't succeed.

PR #740, opened 2021-03-22
chore(shell-api): ensure backwards compatibility of old number classes

rose-m:
Ensures that the following classes accept both number and string arguments:
* `NumberDecimal`
* `NumberLong`
* `NumberInt`

Also shows a deprecation / precision warning:
```
> NumberDecimal(128)
Warning: NumberDecimal: specifying a number as argument is deprecated and may lead to loss of precision
```


addaleax (approved):
👍

PR #742, opened 2021-03-24
fix(shell-api): improve invalid input errors MONGOSH-600

rose-m:
Adds the affected method name to the InvalidInputErrors of the Shell API.


packages/shell-api/src/database.ts
```diff
@@ -234,7 +233,7 @@ export default class Database extends ShellApiClass {
    */
   @returnsPromise
   async runCommand(cmd: Document): Promise<Document> {
-    assertArgsDefined(cmd);
+    assertArgsDefinedType([cmd], [true], 'Database.runCommand');
```
gribnoysup, line 236: Not sure if we want to do something about it in the scope of this ticket, but we don't have any validation if input is a document and this leads to some very weird results when I'm providing non document input:

image

gribnoysup (approved):
Overall looks good to me! Definitely improves the experience of using shell. I noticed that there is `ReplicaSet.add`, and global `connect` and `load` methods in shell-api that are still using `assertArgsDefined`, should we also switch them to your new validation method?

If you don't mind, I think I will also open a ticket for the future to enhance this with some automated code generation as we discussed yesterday. I think this would be better for us in the long run if such validation is added to shell-api public interface (semi-)automatically

packages/shell-api/src/database.ts
rose-m: Yeah for these cases I did not start to change even more but I tried to just "replace" where I could immediately infer the logic. But in these cases it would probably make sense to check for `typeof cmd === 'object'`.

rose-m:
Huh I didn't see the other usages then... I'll cover those, too. And yep, sounds good to add a unit test for the `assertArgsDefinedType` and the logic there.

And yes, absolutely - ticket for the auto-generation would be awesome!!

packages/shell-api/src/helpers.spec.ts
```diff
@@ -23,6 +23,54 @@ describe('dataFormat', () => {
...
+    expect(() => assertArgsDefinedType([1, undefined], [true, [undefined, 'string']])).to.not.throw;
+    expect(() => assertArgsDefinedType([1, 'test'], [true, [undefined, 'string']])).to.not.throw;
+    try {
+      assertArgsDefinedType([1, 2], [true, [undefined, 'string']]);
+    } catch (e) {
+      return expect(e.message).to.contain('Argument at position 1 must be of type string');
+    }
+    expect.fail('Expected error');
+  });
+});
```
gribnoysup, line 72: Thanks for adding those, so easy to _assert_ how this new method works at a glance now!

gribnoysup (approved):
LGTM!

PR #743, opened 2021-03-24
MONGOSH-596: sh status should include not sharder dbs

alenakhineika:
The old `mongodb shell` returns sharded and not shared databases as a result of `sh.status()` command. Remove the filtering by `partitioned` value in order to ensure `mongosh` behavior compatible with the old shell.


rose-m (approved):
Nice, looks good to me! 😊

PR #745, opened 2021-03-30
chore: ensure AUTHORS are generated automatically MONGOSH-647

rose-m:


.github/workflows/cron-tasks.yml
```diff
@@ -0,0 +1,49 @@
...
+      - name: Commit changes
+        run: |
+          git config --local user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git config --local user.name "github-actions[bot]"
+          git add AUTHORS \*/AUTHORS
+          git commit --no-allow-empty -m "chore: update AUTHORS" || true
+      - name: Push updates
+        uses: ad-m/github-push-action@v0.6.0
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
```
addaleax, line 49: I think we’d want these in the same jobs, because otherwise the pushes race against each other and only one will come through (we’re already having this problem when we merge 2 PRs right after each other – one run for the new commits on master will fail because the other’s already pushed by the time it’s finished)

addaleax (approved):

.github/workflows/cron-tasks.yml
rose-m: Ah good point! Did not think of that. I'd then move it in the same job but keep it in separate commits?

addaleax: Yeah, sounds fine to me :)

PR #746, opened 2021-03-30
fix(cli-repl): spawn one mongocryptd per FLE-using shell MONGOSH-595

addaleax:
WIP because this doesn’t work on Windows yet


packages/cli-repl/src/cli-repl.ts
```diff
@@ -149,6 +157,15 @@ class CliRepl {
...
     }
 
+    if (driverOptions.autoEncryption) {
+      const extraOptions = {
+        ...(driverOptions.autoEncryption.extraOptions ?? {}),
+        ...(await this.startMongocryptd())
+      };
+
+      driverOptions.autoEncryption = { ...driverOptions.autoEncryption, extraOptions };
+    }
```
addaleax, line 167: This mirrors what we do in the `Mongo` class, and was missing before (i.e. it would always only try the system `mongocryptd` before).

I feel like this is a symptom of a bigger problem, and that – if possible – we should be using the `Mongo` constructor after initializing the REPL/before starting it to create the connection (as opposed to having the same code in two places).

packages/cli-repl/src/run.ts
```diff
@@ -23,14 +23,14 @@ import { generateUri } from '@mongosh/service-provider-server';
...
-      let mongocryptdSpawnPath = null;
+      let mongocryptdSpawnPaths = [['mongocryptd']];
       if (process.execPath === process.argv[1]) {
         // Remove the built-in Node.js listener that prints e.g. deprecation
         // warnings in single-binary release mode.
         process.removeAllListeners('warning');
         // Look for mongocryptd in the locations where our packaging would
         // have put it.
-        mongocryptdSpawnPath = await getMongocryptdPath();
+        mongocryptdSpawnPaths = await getMongocryptdPaths();
```
mcasimir, line 33: Can't we bury this path discovery mechanism inside `MongoCryptdManager.start` and hide it from the rest?

packages/cli-repl/src/cli-repl.ts
mcasimir: Makes sense to me, would you recommend doing this here / create a ticket?

mcasimir (approved):
Looks great!

packages/cli-repl/src/cli-repl.ts
addaleax: Sure, done: https://jira.mongodb.org/browse/MONGOSH-661

packages/cli-repl/src/run.ts
addaleax: So … we could, but I still feel like this makes sense in terms of separation of concerns + testability (it’s much easier to test this when we can pass the spawn paths to the MongocryptdManager/CliRepl) – we could probably make the result of `getMongocryptdPaths()` the default there and keep the option for overriding for testing, but I’m not sure if that’s really cleaner

PR #747, opened 2021-03-30
fix(async-rewriter2): improve error messages MONGOSH-656

addaleax:
Improve the error messages thrown at runtime by storing the
original source code next to the expression, and then using the
fact that literals are preserved in error messages by V8
in order to restore the original source code of the expression
at runtime.

### DIFF
--- a/packages/async-rewriter2/src/async-writer-babel.ts
+++ b/packages/async-rewriter2/src/async-writer-babel.ts
@@ -212,6 +212,7 @@ interface AsyncFunctionIdentifiers {
   markSyntheticPromise: babel.types.Identifier;
   isSyntheticPromise: babel.types.Identifier;
   syntheticPromiseSymbol: babel.types.Identifier;
+  demangleError: babel.types.Identifier;
 }
 /**
  * The second step that performs the heavy lifting of turning regular functions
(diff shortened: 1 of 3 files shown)


packages/async-rewriter2/src/async-writer-babel.ts
```diff
@@ -521,17 +566,46 @@ export const makeMaybeAsyncFunctionPlugin = ({ types: t }: { types: typeof Babel
...
+          // Turn `... catch (err) { ... }` into `... catch (err) { err = demangleError(err); ... }`
+          path.replaceWith(Object.assign(
+            t.catchClause(path.node.param,
+              t.blockStatement([
+                t.expressionStatement(
+                  t.assignmentExpression('=', path.node.param,
+                    t.callExpression(existingIdentifiers.demangleError, [path.node.param]))),
+                path.node.body
+              ])),
+            { [isGeneratedHelper]: 1 }
```
rose-m, line 606: Is there a reason to have `1` here instead of `true` as in the other places? :)

rose-m (approved):

packages/async-rewriter2/src/async-writer-babel.ts
addaleax: No, it’s just me being odd here :D Will fix

PR #753, opened 2021-04-01
fix(async-rewriter2): properly stringify functions MONGOSH-664

addaleax:
Add a copy of the original source code at the beginning of transformed
functions, so that `Function.prototype.toString` works properly
for them.


addaleax:
The integration test here should start passing after https://github.com/mongodb-js/mongosh/pull/751

PR #758, opened 2021-04-06
fix(shell-evaluator): only apply async-rewriter2 runtime support once

addaleax:
`hasAppliedAsyncWriterRuntimeSupport` was supposed to be a run-only
guard, it was just the actual setting of the variable that was missing
so far.

This fix makes the shell respond a lot faster when using the new async
rewriter. :)


rose-m (approved):
And I thought it had to be like this and wondered why 😅 🙈

mcasimir (approved):

addaleax:
Yeah, it’s actually surprising that even with this bug things were still relatively fast :joy:

PR #760, opened 2021-04-06
fix(service-provider-core): connection str options casing MONGOSH-669

addaleax:
When adding or changing keys, make sure not to ignore already-existing
but differently-cased keys.

### DIFF
--- a/packages/service-provider-core/src/uri-generator.spec.ts
+++ b/packages/service-provider-core/src/uri-generator.spec.ts
@@ -255,6 +255,24 @@ describe('uri-generator.generate-uri', () => {
           expect(generateUri(options)).to.equal('');
         });
       });
+
+      context('when providing explicit serverSelectionTimeoutMS', () => {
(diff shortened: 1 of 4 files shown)


packages/service-provider-core/src/uri-generator.spec.ts
```diff
@@ -255,6 +255,24 @@ describe('uri-generator.generate-uri', () => {
           expect(generateUri(options)).to.equal('');
         });
       });
+
+      context('when providing explicit serverSelectionTimeoutMS', () => {
+        const uri = 'mongodb://127.0.0.2/foo?serverSelectionTimeoutMS=10';
+        const options = { _: [uri] };
+
+        it('returns an empty string', () => {
```
rose-m, line 263: Nit: `returns an empty string` is probably copy-paste? 🤔

rose-m (approved):
Just a tiny copy-paste issue :)

packages/service-provider-core/src/uri-generator.spec.ts
addaleax: Yes :)

PR #764, opened 2021-04-06
fix(async-rewriter2): handle typeof + identifier properly

addaleax:
The source code comments hopefully explain well enough what’s
going on here.


rose-m (approved):
> This is annoying.

😂

PR #765, opened 2021-04-06
fix(async-rewriter2): treat for-in/of as assignments

addaleax:
Treat for-in/for-of with a non-variable-declarator left-hand side
like assignment expressions (which they effectively are).


rose-m (approved):
When I saw that tweet about it I was definitely like - WTF...

PR #774, opened 2021-04-08
fix(shell-api): never check $clusterTime in rs integration tests

addaleax:
See the referenced server issue, the problematic change has been
applied not only to 4.2.x but also to 4.0.x and 4.4.x, so let’s
skip this entirely for now.


addaleax:
(merging as it’s a trivial change to turn CI back green)

PR #780, opened 2021-04-09
feat: add global shell settings API MONGOSH-635

addaleax:


rose-m (approved):
I like the approach 👍 Would allow us to also provide different options / configs per environment (cli vs. compass) if I understood correctly 😄

addaleax:
> I like the approach +1 Would allow us to also provide different options / configs per environment (cli vs. compass) if I understood correctly smile

Yes, it would allow us to do that if we like :+1: 

Feel free to take another look, this has tests now and should be ready for review :)

rose-m (approved):
:shipit:

PR #782, opened 2021-04-13
feat: flip async-rewriter switch MONGOSH-651

addaleax:
After further manual testing, it seems to be a good idea to
flip the switch to enable the new async rewriter by default,
as doing so does improve the user experience significantly
and we do likely want to have some time between enabling it and
the shell 1.0 release.

(e.g.: Without this, running scripts like https://github.com/addaleax/mongosh-snippets/blob/main/snippets/analyze-schema/index.js would be virtually impossible)


rose-m (approved):
Gogogogogogo 🚀

PR #790, opened 2021-04-19
chore(cli-repl): only warn about missing AWS test env vars when used

addaleax:


rose-m (approved):
😅

addaleax:
Okay, not sure why this isn’t working, but I also don’t want to spend a ton of time investigating, so I’ll just close this, the warnings aren’t too terrible :)

PR #800, opened 2021-04-22
chore(cli-repl): work around Node.js REPL bug MONGOSH-589

addaleax:
In addition to the fix in c8cee80616162f, the test here is also flaky
because of a Node.js REPL bug where `repl.displayPrompt()` was called
twice after we reported an error, which confused the prompt detection
in the shell and made the test flaky depending on whether both prompts
where received in the same output data chunk or not.

For now, just wait until both parts of the double prompt are present
in the output before executing the next input line.


gribnoysup (approved):
Oh wow, what a find!

PR #801, opened 2021-04-22
chore(service-provider-server): fail fast for doomed connect MONGOSH-705

addaleax:
This was brought up as a nice way of erroring early in the driver sync
meeting today.

### DIFF
--- a/packages/cli-repl/test/e2e.spec.ts
+++ b/packages/cli-repl/test/e2e.spec.ts
@@ -878,5 +878,23 @@ describe('e2e', function() {
       });
     });
   });
+
+  describe('fail-fast connections', () => {
+    it('fails fast for ENOTFOUND errors', async() => {
+      const shell = TestShell.start({ args: [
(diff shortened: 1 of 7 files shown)


mmarcon:
This is awesome!

Can we (and should we) intercept this error message and customize it to make it more actionable for `mongosh` users?

```
MongoError: The apiVersion parameter is required, please configure your MongoClient's API version
```

It'd be nice if we could say something like:

```
MongoError: The apiVersion parameter is required, please run mongosh with the --apiVersion parameter.
```

I do see that we get the same error message also for `Mongo()` and `connect()`. Maybe we can come up with some copy that works in both cases. I believe that `configure your MongoClient` is confusing because `MongoClient` is not a thing for a `mongosh` user.

addaleax:
@mmarcon Good question. I know that previously, we said that we’re not in the business of rewriting driver error messages, but I see how that makes sense here. We can definitely re-write these errors if we want to.

PR #811, opened 2021-04-27
fix(shell-api): do not forward auth errors on completion MONGOSH-719

addaleax:


packages/cli-repl/src/cli-repl.spec.ts
```diff
@@ -799,6 +810,13 @@ describe('CliRepl', () => {
         }
       });
 
+      it('completes the db.version method', async() => {
+        output = '';
+        input.write('db.vers\u0009\u0009');
```
mcasimir, line 815: Just super random thought .. we could have `const TAB = "\u0009";` and `db.vers${TAB}${TAB}` or something .. not a life changing one :)

packages/shell-api/src/shell-internal-state.ts
```diff
@@ -308,7 +308,7 @@ export default class ShellInternalState {
           const collectionNames = await this.currentDb._getCollectionNamesForCompletion();
           return collectionNames.filter((name) => name.startsWith(collName));
         } catch (err) {
-          if (err.code === ShellApiErrors.NotConnected) {
+          if (err.code === ShellApiErrors.NotConnected || err.codeName === 'Unauthorized') {
```
mcasimir, line 311: Should `'Unauthorized'` be also part `ShellApiErrors` to be consistent with the other?

mcasimir (approved):

packages/shell-api/src/shell-internal-state.ts
addaleax: This is not a shell error, though, it’s a server error :)

packages/cli-repl/src/cli-repl.spec.ts
addaleax: Sounds good, we already do that for e.g. arrow-up in the other tests. I’ll do that right after merging this.

packages/shell-api/src/shell-internal-state.ts
mcasimir: Makes sense!

PR #812, opened 2021-04-27
fix(cli-repl): hide prompt if --eval throws async err MONGOSH-715

addaleax:


packages/cli-repl/src/mongosh-repl.ts
```diff
@@ -248,6 +252,7 @@ class MongoshNodeRepl implements EvaluationListener {
     this.lineByLineInput.start();
     repl.setPrompt(await this.getShellPrompt(internalState));
     repl.displayPrompt();
```
mcasimir, line 254: I think i miss lots of context here, but isn't this supposed to just `return` here since this.started is set on the line after?

addaleax: Ok, thanks for catching this, this should work now :+1:

mcasimir (approved):

PR #813, opened 2021-04-27
fix(cli-repl): do not print prompt while evaluating MONGOSH-700

addaleax:


mcasimir (approved):

addaleax:
@mcasimir Updated to account for the test failures. `async-repl.ts` is a better place for this, because that way we can reset the prompt before calling the REPL `eval` callback (not doing that caused the test failures), and because it’s the right place anyway (because `async-repl.ts` deals with everything related to the Node.js REPL not being able to handle async evaluation well).

PR #816, opened 2021-04-27
fix(shell-api): handle errors in cursor.map/.forEach MONGOSH-703

addaleax:
(Includes #809 for now, until that is merged.)

Handle exceptions from callbacks passed to `cursor.map()` and
`cursor.forEach()`.

Since `cursor.forEach()` returns a Promise, use that to propagate
the error to the user.

For `cursor.map()`, this is a bit trickier, since the method itself
does not return any results to the user. Therefore, we add checks
before and after accessing cursor methods that read data which forward
the exception if one has occurred.

Both of these feel like situations that the driver should take care of
explicitly, so NODE tickets have been opened for both cases.

(One alternative would have been to add async context/Node.js domain
propagation to these methods. However, I decided against that, because
that solution would be a) Node.js-specific, and b) because it would
still lead to undesirable results, like the Promise returned by
`.forEach()` never resolving.)

### DIFF
--- a/packages/shell-api/src/abstract-cursor.ts
+++ b/packages/shell-api/src/abstract-cursor.ts
@@ -24,10 +24,35 @@ export abstract class AbstractCursor extends ShellApiClass {
   abstract _cursor: ServiceProviderAggregationCursor | ServiceProviderCursor;
   _currentIterationResult: CursorIterationResult | null = null;
   _batchSize: number | null = null;
+  _mapError: Error | null = null;
 
   constructor(mongo: Mongo) {
     super();
     this._mongo = mongo;
+  }
+
+  // Wrap a function with checks before and after that verify whether a .map()
+  // callback has resulted in an exception. Such an error would otherwise result
+  // in an uncaught exception, bringing the whole process down.
+  // The downside to this is that errors will not actually be visible until
+  // the caller tries to interact with this cursor in a way that triggers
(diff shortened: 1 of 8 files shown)


packages/shell-api/src/abstract-cursor.ts
```diff
@@ -24,12 +24,37 @@ export abstract class AbstractCursor extends ShellApiClass {
...
+  // callback has resulted in an exception. Such an error would otherwise result
+  // in an uncaught exception, bringing the whole process down.
+  // The downside to this is that errors will not actually be visible until
+  // the caller tries to interact with this cursor in a way that triggers
+  // these checks. Since that is also the behavior for errors coming from the
+  // database server, it makes sense to match that.
+  // Ideally, this kind of code could be lifted into the driver (NODE-3231 and
+  // NODE-3232 are the tickets for that).
+  async _withCheckMapError<Ret>(fn: () => Ret): Promise<Ret> {
+    if (this._mapError) {
```
mcasimir, line 44: Nice trick! is also pretty clear, i thought it would have been more hacky

mcasimir (approved):

packages/shell-api/src/abstract-cursor.ts
addaleax: Yeah … would still be nice if the driver had something here :)

PR #819, opened 2021-04-27
feat(cli-repl): add showStackTraces option MONGOSH-601

addaleax:


mcasimir (approved):

mcasimir:
In the browser repl stack traces are always there but collapsed right?

addaleax:
> In the browser repl stack traces are always there but collapsed right?

Exactly, yes :+1:

PR #822, opened 2021-04-28
chore: update README to reflect current situation

addaleax:
Also update the main build script to actually set the right
distribution variant value in the config, which had accidentally
been broken (and so the example in the README wouldn’t work).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,12 +4,14 @@
 
 MongoDB welcomes community contributions! If you’re interested in making a contribution to the MongoDB Shell, please follow the steps below before you start writing any code:
 
-1. Sign the [contributor's agreement](http://www.mongodb.com/contributor). This will allow us to review and accept contributions.
+1. Sign the [contributor's agreement](https://www.mongodb.com/contributor). This will allow us to review and accept contributions.
 1. Fork the repository on GitHub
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -4,12 +4,14 @@
...
 MongoDB welcomes community contributions! If you’re interested in making a contribution to the MongoDB Shell, please follow the steps below before you start writing any code:
 
-1. Sign the [contributor's agreement](http://www.mongodb.com/contributor). This will allow us to review and accept contributions.
+1. Sign the [contributor's agreement](https://www.mongodb.com/contributor). This will allow us to review and accept contributions.
 1. Fork the repository on GitHub
 1. Create a branch with a name that briefly describes your feature
 1. Implement your feature or bug fix
 1. Add new test cases that verify your bug fix or make sure no one
    unintentionally breaks your feature in the future and run them with `npm test`
+   - You can use `it.only()` instead of `it()` in mocha tests to run only a subset of tests.
```
mcasimir, line 13: we have an eslint rule that catches it if someone forget it in the PR right?

mcasimir (approved):

CONTRIBUTING.md
addaleax: Yes :+1:

PR #829, opened 2021-04-29
feat(autocomplete): add show/use completions MONGOSH-563

addaleax:
Add support for autocompletion for shell commands, in particular,
`show` and `use`. Autocompletion for `use` works similar to
autocompletion for `db.<collection>`.

The implementation details for use/show autocompletion here have
been left to the shell-api package, so that changes to them do not
require any autocompletion changes.

### DIFF
--- a/packages/autocomplete/index.ts
+++ b/packages/autocomplete/index.ts
@@ -24,6 +24,7 @@ export interface AutocompleteParameters {
     server_version: string;
   },
   getCollectionCompletionsForCurrentDb: (collName: string) => string[] | Promise<string[]>;
+  getDatabaseCompletions: (dbName: string) => string[] | Promise<string[]>;
 }
 
 export const BASE_COMPLETIONS = EXPRESSION_OPERATORS.concat(
@@ -50,7 +51,7 @@ const GROUP = '$group';
  *
  * @returns {array} Matching Completions, Current User Input.
  */
-async function completer(params: AutocompleteParameters, line: string): Promise<[string[], string]> {
+async function completer(params: AutocompleteParameters, line: string): Promise<[string[], string, 'exclusive'] | [string[], string]> {
   const SHELL_COMPLETIONS = shellSignatures.ShellApi.attributes as TypeSignatureAttributes;
   const COLL_COMPLETIONS = shellSignatures.Collection.attributes as TypeSignatureAttributes;
   const DB_COMPLETIONS = shellSignatures.Database.attributes as TypeSignatureAttributes;
@@ -59,6 +60,26 @@ async function completer(params: AutocompleteParameters, line: string): Promise<
   const RS_COMPLETIONS = shellSignatures.ReplicaSet.attributes as TypeSignatureAttributes;
   const CONFIG_COMPLETIONS = shellSignatures.ShellConfig.attributes as TypeSignatureAttributes;
   const SHARD_COMPLETE = shellSignatures.Shard.attributes as TypeSignatureAttributes;
+
+  const splitLineWhitespace = line.split(' ');
+  const command = splitLineWhitespace[0];
+  if (SHELL_COMPLETIONS[command]?.isDirectShellCommand) {
+    // If we encounter a direct shell commmand, we know that we want completions
+    // specific to that command, so we set the 'exclusive' flag on the result.
+    // If the shell API provides us with a completer, use it.
+    const completer = SHELL_COMPLETIONS[command].shellCommandCompleter;
+    if (completer) {
+      if (splitLineWhitespace.length === 1) {
+        splitLineWhitespace.push(''); // Treat e.g. 'show' like 'show '.
+      }
+      const hits = await completer(params, splitLineWhitespace) || [];
+      // Adjust to full input, because `completer` only completed the last item
+      // in the line, e.g. ['profile'] -> ['show profile']
+      const fullLineHits = hits.map(hit => [...splitLineWhitespace.slice(0, -1), hit].join(' '));
+      return [fullLineHits, line, 'exclusive'];
+    }
+    return [[line], line, 'exclusive'];
+  }
 
   // keep initial line param intact to always return in return statement
   // check for contents of line with:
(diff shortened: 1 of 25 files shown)


packages/cli-repl/src/mongosh-repl.ts
```diff
@@ -137,11 +137,20 @@ class MongoshNodeRepl implements EvaluationListener {
         this.insideAutoComplete = true;
         try {
           // Merge the results from the repl completer and the mongosh completer.
-          const [ [replResults], [mongoshResults] ] = await Promise.all([
+          // eslint-disable-next-line @typescript-eslint/no-unused-vars
+          const [ [replResults], [mongoshResults, _, mongoshResultsExclusive] ] = await Promise.all([
```
mcasimir, line 141: Shouldn't `[mongoshResults,, mongoshResultsExclusive]` work too?

packages/autocomplete/index.ts
```diff
@@ -60,6 +61,26 @@ async function completer(params: AutocompleteParameters, line: string): Promise<
   const CONFIG_COMPLETIONS = shellSignatures.ShellConfig.attributes as TypeSignatureAttributes;
   const SHARD_COMPLETE = shellSignatures.Shard.attributes as TypeSignatureAttributes;
 
+  const splitLineWhitespace = line.split(' ');
+  const command = splitLineWhitespace[0];
+  if (SHELL_COMPLETIONS[command]?.isDirectShellCommand) {
+    // If we encounter a direct shell commmand, we know that we want completions
+    // specific to that command, so we set the 'exclusive' flag on the result.
```
mcasimir, line 68: Why is this necessary? Will this also work in the browser?

packages/cli-repl/src/mongosh-repl.ts
addaleax: @mcasimir Yes, I just didn’t think of that :)

packages/autocomplete/index.ts
addaleax: @mcasimir Are you referring to the `'exclusive'` flag here?

The mongosh-repl.ts changes have a somewhat more specific context here, but the gist of it is that we run both the Node.js default autocompleter and our own there and combine the results; but the Node.js REPL thinks that it should complete e.g. `use a` to `use assert`, because it just looks at the `a` and finds global variables with the same name.

> Will this also work in the browser? 

I don’t think the extra array entry should hurt, we’re only using the first entry anyway:

https://github.com/mongodb-js/mongosh/blob/831220a316738c0d28f4ce464c83ba56b8b8c75d/packages/browser-runtime-core/src/autocompleter/shell-api-autocompleter.ts#L22

packages/cli-repl/src/mongosh-repl.ts
```diff
@@ -137,11 +137,19 @@ class MongoshNodeRepl implements EvaluationListener {
         this.insideAutoComplete = true;
         try {
           // Merge the results from the repl completer and the mongosh completer.
-          const [ [replResults], [mongoshResults] ] = await Promise.all([
+          const [ [replResults], [mongoshResults,, mongoshResultsExclusive] ] = await Promise.all([
             (async() => await origReplCompleter(text) || [[]])(),
             (async() => await mongoshCompleter(text))()
           ]);
           this.bus.emit('mongosh:autocompletion-complete'); // For testing.
+
```
mcasimir, line 145: Ok, I think I've got a bit better the context. Makes sense.

packages/shell-api/src/database.spec.ts
```diff
@@ -84,7 +84,9 @@ describe('Database', () => {
         returnType: 'AggregationCursor',
         platforms: ALL_PLATFORMS,
         topologies: ALL_TOPOLOGIES,
-        serverVersions: ALL_SERVER_VERSIONS
+        serverVersions: ALL_SERVER_VERSIONS,
+        isDirectShellCommand: false,
```
mcasimir, line 88: This is a custom "autocompleter" for arguments? And this will work with a worker thread as well cause the function will be called inside the runtime right?

Slowly getting it :).

packages/shell-api/src/decorators.ts
```diff
@@ -142,6 +155,8 @@ export interface TypeSignature {
   deprecated?: boolean;
   returnType?: string | TypeSignature;
   attributes?: { [key: string]: TypeSignature };
+  isDirectShellCommand?: boolean;
```
mcasimir, line 158: Is this used only for autocompletion or also to set up the command in the various shells? Would be great if we have to have this for it to be the source of truth about that. Hope i've understood what this means.

I think a comment here could help, the concept of "direct command" may not be immediate without context.

mcasimir (approved):

packages/shell-api/src/database.spec.ts
addaleax: > This is a custom "autocompleter" for arguments?

The entry below is, yes.

> And this will work with a worker thread as well cause the function will be called inside the runtime right?

I *think* so, yes. The autocompleter in the browser also lives inside the same environment as e.g. the `ShellInternalState` object, and only the caller-facing `getCompletions()` method of the runtime is forwarded to the parent/rendering process, so this should work.

packages/shell-api/src/decorators.ts
addaleax: > Is this used only for autocompletion or also to set up the command in the various shells? Would be great if we have to have this for it to be the source of truth about that. Hope i've understood what this means.

The property on the function instance itself is what the shell evaluator uses to determine whether to allow bash-style calls to it, the property on the signature is what the autocompleter uses (like all other properties).

In either case, the source of truth is the `@directShellCommand` decorator.

> I think a comment here could help, the concept of "direct command" may not be immediate without context.

Good idea, done! :+1:

mcasimir: Actually now that you called it that way .. is "bash-style" command a name that's even more clear? Of course is fine as is.

addaleax: I mean, it’s not really bash-specific, it’s like that in most OS shells (even Windows cmd.exe) :)

PR #833, opened 2021-05-03
feat(ci): add s390x, ppc64le, arm64 binaries

addaleax:
See MONGOSH-693, MONGOSH-694, MONGOSH-695. This commit only includes
building binaries, not packaging them.


rose-m (approved):
Changes look good to me! 👍

PR #834, opened 2021-05-03
fix(shell-api): sharding methods account for 5.0 config.chunks change…

addaleax:
…s MONGOSH-740

### DIFF
--- a/packages/shell-api/src/helpers.ts
+++ b/packages/shell-api/src/helpers.ts
@@ -429,7 +429,8 @@ export async function getPrintableShardStatus(db: Database, verbose: boolean): P
           collRes.balancing = [ !coll.noBalance, { noBalance: coll.noBalance } ];
         }
         const chunksRes = [];
-        const chunksCollMatch = coll.uuid ? { $or: [ { uuid: coll.uuid, ns: coll._id } ] } : { ns: coll._id };
+        const chunksCollMatch =
+          coll.uuid ? { $or: [ { uuid: coll.uuid }, { ns: coll._id } ] } : { ns: coll._id };
         const chunks = await
         (await chunksColl.aggregate([
           { $match: chunksCollMatch },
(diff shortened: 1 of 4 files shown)


packages/shell-api/src/helpers.ts
```diff
@@ -403,8 +403,8 @@ export async function getPrintableShardStatus(db: Database, verbose: boolean): P
 
   // Special case the config db, since it doesn't have a record in config.databases.
   databases.push({ '_id': 'config', 'primary': 'config', 'partitioned': true });
-  databases.sort((a: any, b: any): any => {
-    return a._id > b._id;
+  databases.sort((a: any, b: any): number => {
+    return a._id.localeCompare(b._id);
```
addaleax, line 407: Just something I noticed, this would obviously not have worked before because the `.sort()` callback should not return a boolean.

packages/shell-api/src/helpers.ts
```diff
@@ -429,12 +429,14 @@ export async function getPrintableShardStatus(db: Database, verbose: boolean): P
...
         const chunks = await
-        (await chunksColl.aggregate({ $match: { ns: coll._id } },
+        (await chunksColl.aggregate([
+          { $match: chunksCollMatch },
           { $group: { _id: '$shard', cnt: { $sum: 1 } } },
           { $project: { _id: 0, shard: '$_id', nChunks: '$cnt' } },
-          { $sort: { shard: 1 } })
-        ).toArray();
+          { $sort: { shard: 1 } }
+        ])).toArray();
```
addaleax, line 439: The tests really should have discovered that `sh.status()` was also broken, I just noticed this because I was looking for other uses of `config.chunks` in the shell-api source. I’ve opened https://jira.mongodb.org/browse/MONGOSH-742 as a reminder to write proper tests here.

rose-m (approved):
Nice finds apart from the fixed 5.0 change itself :D

PR #837, opened 2021-05-03
fix(cli-repl): handle pre-4.4 logs for startup warnings MONGOSH-743

addaleax:


rose-m (approved):
What a nice way to parse a log line... 👏 😄

PR #840, opened 2021-05-04
chore(cli-repl): track arch in telemetry

addaleax:
With support for multiple architectures expanding, it seems to make
sense to track the system architecture along with the OS.


mmarcon (approved):
LGTM!

PR #841, opened 2021-05-04
chore(ci): use temporary Windows distro to help build team debug

addaleax:
See BUILD-12985.


addaleax:
I’ll merge this since it’s “only” a distro change and the build team would like us to try this out as much as we can :)

PR #844, opened 2021-05-04
fix(service-provider-core): make Long print evaluable MONGOSH-750

addaleax:


rose-m (approved):

addaleax:
Fwiw, I’m having an ongoing conversation with Max about whether this is the right thing to do in the end, but for now I’ll merge this so that we have something that works and is consistent with what we do for `Decimal128`.

PR #846, opened 2021-05-05
fix(cli-repl): bump pretty-repl to 3.0.0

addaleax:
Fixes MONGOSH-691 and MONGOSH-734.

### DIFF
--- a/packages/cli-repl/src/mongosh-repl.spec.ts
+++ b/packages/cli-repl/src/mongosh-repl.spec.ts
@@ -304,7 +304,7 @@ describe('MongoshNodeRepl', () => {
         input.write(`vers${tab}${tab}`);
         await tick();
         expect(output).to.include('version');
-        expect(output).to.not.match(/version\s+version/);
+        expect(output).to.not.match(/version[ \t]+version/);
       });
       it('autocompletes async shell api methods', async() => {
         input.write(`db.coll.find().${tab}${tab}`);
@@ -331,7 +331,7 @@ describe('MongoshNodeRepl', () => {
         input.write('\u0003'); // Ctrl+C for abort
         await tick();
         expect((mongoshRepl.runtimeState().repl as any)._prompt).to.equal('> ');
-        expect(stripAnsi(output)).to.equal('ddbdb.db.\tdb.\tfdb.\tfodb.\tfoo\r\nbbabar\r\n\r\n> ');
+        expect(stripAnsi(output)).to.equal('db.\tfoo\r\nbar\r\n\r\n> ');
       });
     });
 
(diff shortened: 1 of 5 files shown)


packages/cli-repl/src/mongosh-repl.spec.ts
```diff
@@ -566,19 +566,19 @@ describe('MongoshNodeRepl', () => {
...
       await tick();
-      expect(output).to.match(/const(\x1b\[.*m)+ cat = (\x1b\[.*m)+"Nori"(\x1b\[.*m)+/);
+      expect(output).to.match(/const(\x1b\[.*m)+ cat = (\x1b\[.*m)+"(\x1b\[.*m)+N(\x1b\[.*m)+o(\x1b\[.*m)+r(\x1b\[.*m)+i(\x1b\[.*m)+"(\x1b\[.*m)+/);
     });
 
     it('colorizes input function', async() => {
       input.write('function add (a, b) { return a + b }');
       await tick();
-      expect(output).to.match(/function(\x1b\[.*m)+ (\x1b\[.*m)+add(\x1b\[.*m)+ \(a, b\) \{ (\x1b\[.*m)+return(\x1b\[.*m)+ a \+ b/);
+      expect(output).to.match(/function(\x1b\[.*m)+ (\x1b\[.*m)+a(\x1b\[.*m)+d(\x1b\[.*m)+d(\x1b\[.*m)+ \(a, b\) \{ retur\x08+(\x1b\[.*m)+return(\x1b\[.*m)+ a \+ b/);
```
rose-m, line 575:
```suggestion
      expect(output).to.match(/function(\x1b\[.*m)+ (\x1b\[.*m)+a(\x1b\[.*m)+d(\x1b\[.*m)+d(\x1b\[.*m)+ \(a, b\) \{\x08+(\x1b\[.*m)+return(\x1b\[.*m)+ a \+ b/);
```
Some returns are mangled in there 🤔

rose-m (approved):
(Looking at the pretty-repl PR next)

packages/cli-repl/src/mongosh-repl.spec.ts
addaleax: Oh, this is on purpose – You enter `retur`, which is not a keyword, then you enter `n`, which *makes* it a keyword, so we go back a few times (\x08 is backspace), and overwrite it with the highlighted version :)

addaleax: Also, to be clear, this is the same visual behavior as before – the difference is that we can’t match the complete and fully highlighted string anymore, because we’re not printing the full line, only applying diffs to it as the user input is being entered

rose-m: 🤯

PR #847, opened 2021-05-05
chore(cli-repl): do not print MongoDB version when using --nodb

rose-m:


packages/cli-repl/src/mongosh-repl.spec.ts
```diff
@@ -679,7 +679,8 @@ describe('MongoshNodeRepl', () => {
           nodb: true
         };
         mongoshRepl = new MongoshNodeRepl(mongoshReplOptions);
-        await mongoshRepl.initialize(serviceProvider);
+        const initialized = await mongoshRepl.initialize(serviceProvider);
+        await mongoshRepl.startRepl(initialized);
```
addaleax, line 683: The warnings should also apply for e.g. `--eval` where `.startRepl()` is not called, is there a specific reason for adding it here? Is this just to make sure that `.startRepl()` also does not show the warnings?

addaleax (approved):

packages/cli-repl/src/mongosh-repl.spec.ts
rose-m: Ah good point... let me get rid of the start Repl!

PR #848, opened 2021-05-05
feat(shell-api): provide bulk.find().delete() instead of .remove() MONGOSH-752

addaleax:
This happens to account for the changes made in NODE-2978 (https://github.com/mongodb/node-mongodb-native/commit/c3a183938289).

The old names are kept as deprecated aliases.

MONGOSH-752

### DIFF
--- a/packages/i18n/src/locales/en_US.ts
+++ b/packages/i18n/src/locales/en_US.ts
@@ -1979,6 +1979,16 @@ const translations: Catalog = {
               description: 'Adds an removeOne to the bulk operation.',
               example: 'bulkOp.find(...).removeOne()',
             },
+            'delete': {
+              link: 'https://docs.mongodb.com/manual/reference/method/Bulk.find.delete/',
+              description: 'Adds an delete to the bulk operation.',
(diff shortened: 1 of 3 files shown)


packages/i18n/src/locales/en_US.ts
```diff
@@ -1979,6 +1979,16 @@ const translations: Catalog = {
               description: 'Adds an removeOne to the bulk operation.',
               example: 'bulkOp.find(...).removeOne()',
             },
+            'delete': {
+              link: 'https://docs.mongodb.com/manual/reference/method/Bulk.find.delete/',
```
rose-m, line 1983: Do we expect the links to work at some point in the future? Right now the `delete` and `deleteOne` links give a 404 error unfortunately (since they link to the old shell's API methods).

rose-m (approved):

packages/i18n/src/locales/en_US.ts
addaleax: Yes, I’d expect that they work once the documented for the linked ticket here is finished up

PR #849, opened 2021-05-06
fix(cli-repl): ctrl-c terminates running DB operations MONGOSH-640

rose-m:
Still a work in progress - but already implements a large deal of CTRL-C handling. The process is as follows:

1. CTRL-C is caught for an async execution
2. All MongoClients are closed
3. We print the "Error: Asynchronous execution was interrupted" message (no prompt)
3. We wait until the async execution finishes (since it is running in the background CTRL-C doesn't kill it) and silently catch any errors
4. We reconnect all MongoClients
5. The prompt returns and allows the user to enter new commands

The following needs to be completed:

* [x] Adapt the API decorators to check for interruption and immediately throw an internal error
* [x] Adapt the `async-writer` to ensure that the internal interrupt errors are not caught by user code

And here a quick demo how it looks like for a simple command:

mongosh-ctrlc

### DIFF
--- a/packages/cli-repl/src/async-repl.spec.ts
+++ b/packages/cli-repl/src/async-repl.spec.ts
@@ -68,10 +68,9 @@ describe('AsyncRepl', () => {
   it('allows sync interruption through SIGINT', async function() {
     if (process.platform === 'win32') {
       this.skip(); // No SIGINT on Windows.
-      return;
     }
 
-    const { input, output } = createDefaultAsyncRepl({ breakEvalOnSigint: true });
+    const { input, output } = createDefaultAsyncRepl({ onAsyncSigint: () => false });
 
     input.write('while (true) { process.kill(process.pid, "SIGINT"); }\n');
     await expectInStream(output, 'execution was interrupted');
@@ -80,15 +79,16 @@ describe('AsyncRepl', () => {
   it('allows async interruption through SIGINT', async function() {
     if (process.platform === 'win32') {
       this.skip(); // No SIGINT on Windows.
-      return;
     }
 
-    const { input, output } = createDefaultAsyncRepl({ breakEvalOnSigint: true });
+    const onAsyncSigint = sinon.stub().resolves(false);
+    const { input, output } = createDefaultAsyncRepl({ onAsyncSigint: onAsyncSigint });
 
     input.write('new Promise(oopsIdontResolve => 0)\n');
     await delay(100);
     process.kill(process.pid, 'SIGINT');
     await expectInStream(output, 'execution was interrupted');
+    expect(onAsyncSigint).to.have.been.calledOnce;
   });
 
   it('handles synchronous exceptions well', async() => {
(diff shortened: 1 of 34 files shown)


packages/shell-api/src/shell-internal-state.ts
```diff
@@ -330,6 +336,51 @@ export default class ShellInternalState {
...
+        await r.resume();
+        return true;
+      } catch (e) {
+        return false;
+      }
+    }) ?? [];
+    this.resumeMongosAfterInterrupt = undefined;
+
+    const result = await Promise.all(promises);
+    this.isInterrupted = true;
```
addaleax, line 380: Is this supposed to be `true` here? If yes, I think I need to read the code a bit more :)

addaleax (commented):
Looks great so far!

packages/shell-api/src/shell-internal-state.ts
rose-m: Ah no of course not! Thanks...

packages/cli-repl/src/async-repl.spec.ts
```diff
@@ -68,10 +68,9 @@ describe('AsyncRepl', () => {
   it('allows sync interruption through SIGINT', async function() {
     if (process.platform === 'win32') {
       this.skip(); // No SIGINT on Windows.
-      return;
```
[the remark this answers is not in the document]
rose-m: I just removed the return because the `this.skip()` has return type `never` and it was always highlighted in VSCode as an unreachable line. It still skips everything as it should.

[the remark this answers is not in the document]
rose-m: (I've made it a consistent `return this.skip()` in those cases 👍  easier to understand when reading and doesn't give a "visible warning")

packages/shell-api/src/decorators.ts
```diff
@@ -362,11 +406,20 @@ export function topologies(topologiesArray: Topologies[]): Function {
...
-  descriptor.value = markImplicitlyAwaited(descriptor.value);
-  if (orig.constructor.name !== 'AsyncFunction') {
-    nonAsyncFunctionsReturningPromises.push(orig.name);
+  const originalFunction = descriptor.value;
+  originalFunction.returnsPromise = true;
+
+  async function wrapper(this: any, ...args: any[]) {
+    const result = await originalFunction.call(this, ...args);
+    await new Promise(setImmediate);
+    return result;
```
addaleax, line 415: Maybe this, so that a) we always await the tick, even if it throws, and b) the java-shell is not affected by this?

```suggestion
    try {
      return await originalFunction.call(this, ...args);
    } finally {
       if (typeof setImmediate === 'function') {
         // Not all JS environments have setImmediate
         await new Promise(setImmediate);
       }
    }
```

(I’ll apply the suggestion, but I’ll leave the comment so you can see what I’ve changed here :))

addaleax (approved):

PR #855, opened 2021-05-07
fix(cli-repl): aggregate regular characters in LineByLineInput MONGOS…

addaleax:
…H-687

Instead of forwarding characters one-by-one to the REPL/readline
instance, group characters that do not have special meaning in a
REPL context together. This is relevant when pasting strings,
where multiple characters are read in rapid succession, and makes
the REPL behave a bit more closely to what e.g. the Node.js REPL
does in that case. There are two main advantages of this:

- This makes tab-indented code behave well, because the REPL does
  not trigger autocompletion when seeing `\tstring`, only when
  seeing `\t` and `string` separately.
- This makes input handling a bit faster, because e.g. the input
  highlighting needs to perform fewer operations for pasted code.


rose-m (approved):
Code LGTM 😊

PR #862, opened 2021-05-11
feat(service-provider-server): use OS DNS for SRV records MONGOSH-733

addaleax:
Obviously, the bulk of the work here is left to other packages :)

I’ll leave this as a draft until CI passes + we have confirmation that this actually works and wasn’t just a futile exercise in addon coding.


rose-m (approved):
Took a look at the other 2 modules, too :) I did understand it but... that native thing is 🧙 😄

addaleax:
@rose-m Yeah, the native thing is modeled quite closely on what the legacy shell did here … it’s been a while since I wrote DNS parsing code :D

PR #864, opened 2021-05-12
feat(async-rewriter2): add uncatchable exception support

addaleax:
(This is intended to be merged as part of https://github.com/mongodb-js/mongosh/pull/849)

### DIFF
--- a/packages/async-rewriter2/README.md
+++ b/packages/async-rewriter2/README.md
@@ -33,7 +33,7 @@ they reach their first `await` expression, and the fact that we can determine
 which `Promise`s need `await`ing by marking them as such using decorators
 on the API surface.
 
-The transformation takes place in two main steps.
+The transformation takes place in three main steps.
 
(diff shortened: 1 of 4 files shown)


packages/async-rewriter2/README.md
```diff
@@ -63,7 +63,62 @@ function foo() {
...
+
+is transformed into
+
+```js
+try {
+  foo3();
+} catch (_err) {
+  if (!err || !_err[Symbol.for('@@mongosh.uncatchable')]) {
+    bar3();
+  }
```
rose-m, line 89: I think for the sake of correctness there should be the throw in the else branch
```suggestion
  } else {
    throw _err;
  }
```

rose-m (approved):
That looks great! Never knew that... intended behavior on destructuring in catch, etc. :D

addaleax:
I’ll close this because @rose-m has seen this PR and that’s the purpose here :P

PR #865, opened 2021-05-12
chore(ci): match server OS/arch matrix MONGOSH-424

addaleax:


addaleax:
Okay, looks like we’ll need separate rpms for the different rpm distros due to mongocryptd linking differences. That’s going to make a few things more complicated.

rose-m (approved):
👍 👍 👍

PR #869, opened 2021-05-17
fix(build): do not utf8-read/write images MONGOSH-763

rose-m:
Fixes the issue of images not showing in the MSI installer.

### DIFF
--- a/packages/build/src/packaging/package/helpers.ts
+++ b/packages/build/src/packaging/package/helpers.ts
@@ -61,15 +61,19 @@ export async function generateDirFromTemplate(sourceDir: string, interpolations:
         await copyDirAndApplyTemplates(sourceFile, targetFile);
       } else {
         const sourceText = await fs.readFile(sourceFile, 'utf8');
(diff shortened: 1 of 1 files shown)


mcasimir (approved):

addaleax (approved):
This makes sense :)

Maybe, to avoid the filename check, we could just skip the interpolation for non-UTF-8 files? i.e.

```js
      } else {
        const sourceText = await fs.readFile(sourceFile, 'utf8');
        if (!sourceText.contains('\ufffd')) { // This is valid UTF-8, i.e. a text file
          const interpolatedText = sourceText.replace(
            /\{\{(\w+)\}\}/g,
            (_match, identifier) => {
              if (!(identifier in interpolations)) {
                throw new Error(`Need ${identifier} for replacement in ${sourceFile}`);
              }
              return interpolations[identifier];
            });
          await fs.writeFile(targetFile, interpolatedText);
        } else {
          await fs.copyFile(sourceFile, targetFile);
        }
      }
```

PR #872, opened 2021-05-17
feat(build): add macos arm64 support MONGOSH-442

addaleax:


packages/build/src/packaging/run-package.ts
```diff
@@ -12,10 +13,20 @@ export async function runPackage(
...
+  // TODO: add mongocryptd and E2E tests for darwin-arm64 once server builds
+  // are available for that platform.
+  if (distributionBuildVariant !== 'darwin-arm64') {
+    await fs.copyFile(
+      await downloadMongocrypt(distributionBuildVariant),
+      config.mongocryptdPath,
+      fsConstants.COPYFILE_FICLONE);
+  } else {
+    await fs.copyFile(
+      path.resolve(__dirname, '..', '..', '..', '..', 'scripts', 'no-mongocryptd.sh'),
```
mcasimir, line 26: How is this going to work? Will it affect something that mongosh does automatically or is rather there if users tries to run mongocryptd manually?

mcasimir (approved):
🎉

packages/build/src/packaging/run-package.ts
addaleax: Yeah, mongosh won’t be able to do automatic FLE for the time being. The exposed binary is still named `mongocryptd-mongosh`, so I don’t think it’s something the user would run by accident.

PR #878, opened 2021-05-18
feat: add stricter type validation for config settings MONGOSH-758

addaleax:
Created together with @rose-m :)


rose-m (approved):
:shipit:

PR #893, opened 2021-05-21
fix(errors): direct link to JIRA project

mmarcon:
If someone decides to create a bug when they find something
broken, we should point them directly to the JIRA project rather
than letting them figure out how to get to it.


rose-m (approved):
That totally makes sense 😊 👍

PR #895, opened 2021-05-21
fix(cli-repl): do not throw internal errors for invalid user code MONGOSH-777

rose-m:


addaleax (approved):
Thanks, this makes sense :+1:

rose-m:
Took me a while to find the reason since it showed `SyntaxError` 😅 But finally found it since the stack trace came from different packages...

PR #896, opened 2021-05-21
feat(shell-api): add reconfigForPSASet helper MONGOSH-702

addaleax:
This provides the same functionality as the newly introduced legacy
shell equivalent.

Notable differences are:
- This does some type cleanup around the replset code
- This is more verbose, and prints:
  - A warning if the new member is already in the existing config
    with a different _id but the same hostname
  - Indicators that retrying is still going on when it is
  - The second reconfig commmand, if it failed, so that the user
    can retry manually
- This performs exponential backoff while waiting for the second
  reconfig to succeed
- This uses a fixed number of attempts (12) instead of a fixed
  duration (90 seconds) for performing the second reconfig


rose-m (approved):
Nice, really understandable implementation 👍 😊

PR #898, opened 2021-05-25
MONGOSH-738: Show collection type in output

alenakhineika:
List collections with badges that identify collection types.


data/.mlaunch_startup
```diff
@@ -0,0 +1,54 @@
...
+"sslFIPSMode": false,
+"sslClientCertificate": null,
+"sslClientPEMKeyFile": null,
+"sslClientPEMKeyPassword": null
+},
+"unknown_args": [],
+"startup_info": {
+"27017": "\"mongod\"  --dbpath \"/Users/alenakhineika/www/mongosh/data/db\" --logpath \"/Users/alenakhineika/www/mongosh/data/mongod.log\" --port 27017 --fork   --wiredTigerCacheSizeGB 1 "
+}
+}
```
addaleax, line 54: I think this file might have been added by accident?

packages/browser-repl/src/components/types/show-collections-output.tsx
```diff
@@ -1,16 +1,30 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
+import textTable from 'text-table';
 
 interface ShowCollectionsOutputProps {
-  value: string[];
+  value: any;
```
addaleax, line 6: Isn’t this
```suggestion
  value: CollectionObject[];
```

?

packages/cli-repl/src/format-output.ts
```diff
@@ -118,8 +118,12 @@ function formatSimpleType(output: any, options: FormatOptions): any {
   return inspect(output, options);
 }
 
-function formatCollections(output: string[], options: FormatOptions): string {
-  return clr(output.join('\n'), 'bold', options);
+function formatCollections(output: any[], options: FormatOptions): string {
```
addaleax, line 121: Ditto here, I think we want to avoid `any` wherever we can… maybe you can export the type from `@mongosh/shell-api`, the same way we do it for `HelpProperties`?

packages/shell-api/src/database.ts
```diff
@@ -144,6 +144,17 @@ export default class Database extends ShellApiWithMongoClass {
...
 
+  async _getCollectionNamesWithTypes(options?: ListCollectionsOptions): Promise<Document[]> {
+    const collections = await this._listCollections({}, { ...options, nameOnly: true });
+    const typesToBages: any = {
+      collection: '',
+      timeseries: '[time-series]',
+      view: '[view]'
+    };
+
+    return collections.map((collection: any) => ({ name: collection.name, badge: typesToBages[collection.type] }));
```
addaleax, line 155: This should cache the collection names just like `_getCollectionNames` does, I would say, so that whenever we list collections, the cache contains the most up-to-date data.

We probably also need to handle the case that `collection.type` is something we’re not expecting yet (e.g. because it’s only added in a future version of the server), so that `badge` is actually always a string and not potentially `undefined`.

packages/cli-repl/src/format-output.spec.ts
```diff
@@ -156,12 +156,16 @@ for (const colors of [ false, true ]) {
...
+            { name: 'decimal128', badge: '' },
+            { name: 'coll', badge: '' },
+            { name: 'people_imported', badge: '[view]' },
+            { name: 'cats', badge: '[time-series]' }
           ],
           type: 'ShowCollectionsResult'
         }));
 
-        expect(output).to.contain('nested_documents\ndecimal128\ncoll\npeople_imported\ncats');
+        expect(output).to.contain('nested_documents\ndecimal128\ncoll\npeople_imported   [view]\ncats              [time-series]');
```
addaleax, line 168: Would it make sense to add an end-to-end test here that verifies that this works as expected when run against an actual server?

packages/shell-api/src/database.ts
alenakhineika: Hm, we use `_cachedCollectionNames` for autocomplete, so I can see the cache value here. But we use `_getCollectionNamesWithTypes` only for the showing collections command, where I would expect always up-to-date values. Which use case do you have in mind for this cache?

addaleax: Right, I’m not saying that the result of `_getCollectionNamesWithTypes` should be cached, only that the `_cachedCollectionNames` list should be updated here as well.

addaleax: i.e.:
```suggestion
    this._cachedCollectionNames = collection.map((collection: any) => collection.name);
    return collections.map((collection: any) => ({ name: collection.name, badge: typesToBages[collection.type] }));
```

alenakhineika: Ah, got it!

data/.mlaunch_startup
alenakhineika: Removed.

packages/browser-repl/src/components/types/show-collections-output.tsx
alenakhineika: Thx for pointinbg at it!

packages/cli-repl/src/format-output.spec.ts
alenakhineika: Added to e2e.ts

packages/cli-repl/test/e2e.spec.ts
```diff
@@ -962,5 +961,38 @@ describe('e2e', function() {
...
+        const dbName = `test-${Date.now()}`;
+
+        await shell.executeLine(`use ${dbName};`);
+        await shell.executeLine("db.coll2.insertOne({ some: 'field' });");
+        await shell.executeLine("db.createCollection('coll3', { timeseries: { timeField: 'time' } } );");
+
+        const result = await shell.executeLine('show collections');
+
+        expect(result).to.include('coll2');
+        expect(result).to.include('coll3');
```
addaleax, line 993: Maybe also this?

```suggestion
        expect(result).to.include('coll3');
        expect(result).to.include('[timeseries]');
```

packages/shell-api/src/database.ts
```diff
@@ -144,6 +149,21 @@ export default class Database extends ShellApiWithMongoClass {
...
+    const typesToBages: any = {
+      timeseries: '[time-series]',
+      view: '[view]'
+    };
+
+    this._cachedCollectionNames = collections.map((collection: any) => collection.name);
+
+    return collections.map((collection: any) => ({
+      name: collection.name,
+      badge: typesToBages[collection.type] ? typesToBages[collection.type] : ''
```
addaleax, line 163:
```suggestion
      badge: typesToBages[collection.type] ?? ''
```

(I think that’s just a bit more common in the mongosh source, at least by now)

data/.mlaunch_startup
addaleax: Looks like it’s still there, just so you know :)

addaleax (approved):
Looks good!

data/.mlaunch_startup
alenakhineika: It should be gone now :)

rose-m (approved):
Nice, looking forward to having that included 😊

PR #905, opened 2021-05-26
fix(shell-api): allow Explainable.aggregate to take individual stages MONGOSH-790

rose-m:
This PR includes some additional typing fixes.

### DIFF
--- a/packages/shell-api/src/collection.ts
+++ b/packages/shell-api/src/collection.ts
@@ -142,9 +142,12 @@ export default class Collection extends ShellApiWithMongoClass {
    *
    * @returns {Promise} The promise of aggregation results.
    */
+  async aggregate(pipeline: Document[], options: Document & { explain?: undefined | false }): Promise<AggregationCursor>
+  async aggregate(pipeline: Document[], options: Document & { explain: true }): Promise<Document>
+  async aggregate(...stages: Document[]): Promise<AggregationCursor>
   @returnsPromise
   @returnType('AggregationCursor')
-  async aggregate(...args: any[]): Promise<AggregationCursor> {
+  async aggregate(...args: any[]): Promise<any> {
     let options;
     let pipeline;
     if (args.length === 0 || Array.isArray(args[0])) {
--- a/packages/shell-api/src/collection.ts
+++ b/packages/shell-api/src/collection.ts
@@ -142,8 +142,8 @@ export default class Collection extends ShellApiWithMongoClass {
    *
    * @returns {Promise} The promise of aggregation results.
    */
-  async aggregate(pipeline: Document[], options: Document & { explain?: undefined | false }): Promise<AggregationCursor>
-  async aggregate(pipeline: Document[], options: Document & { explain: true }): Promise<Document>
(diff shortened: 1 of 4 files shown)


packages/shell-api/src/collection.ts
```diff
@@ -142,9 +142,12 @@ export default class Collection extends ShellApiWithMongoClass {
    *
    * @returns {Promise} The promise of aggregation results.
    */
+  async aggregate(pipeline: Document[], options: Document & { explain?: undefined | false }): Promise<AggregationCursor>
+  async aggregate(pipeline: Document[], options: Document & { explain: true }): Promise<Document>
```
addaleax, line 146: explain: false` is `explain: 'queryPlanner'`, I think that’s not what you are expecting here?

```suggestion
  async aggregate(pipeline: Document[], options: Document & { explain?: never }): Promise<AggregationCursor>
  async aggregate(pipeline: Document[], options: Document & { explain: ExplainVerbosityLike }): Promise<Document>`
```

packages/shell-api/src/collection.ts
```diff
@@ -350,8 +353,11 @@ export default class Collection extends ShellApiWithMongoClass {
    *
    * @returns {Array} The promise of the result.
    */
+  async distinct(field: string): Promise<Document>
+  async distinct(field: string, query: Document): Promise<Document>
+  async distinct(field: string, query: Document, options: DistinctOptions): Promise<Document>
   @returnsPromise
-  async distinct(field: string, query: Document, options: DistinctOptions = {}): Promise<Document> {
+  async distinct(field: string, query?: Document | undefined, options: DistinctOptions = {}): Promise<Document> {
```
addaleax, line 360:
```suggestion
  async distinct(field: string, query?: Document, options: DistinctOptions = {}): Promise<Document> {
```

because ? already implies `undefined`?

packages/shell-api/src/explainable.ts
```diff
@@ -90,9 +91,21 @@ export default class Explainable extends ShellApiWithMongoClass {
     return new ExplainableCursor(this._mongo, cursor, this._verbosity);
   }
 
+  async aggregate(pipeline: Document[], options: Document & { explain?: undefined | false }): Promise<AggregationCursor>
+  async aggregate(pipeline: Document[], options: Document & { explain: true }): Promise<Document>
```
addaleax, line 95: Ditto here, 

```suggestion
  async aggregate(pipeline: Document[], options: Document & { explain?: never }): Promise<AggregationCursor>
  async aggregate(pipeline: Document[], options: Document & { explain: ExplainVerbosityLike }): Promise<Document>
```

addaleax (approved):

packages/shell-api/src/collection.ts
rose-m: Yup, true... 👍

packages/shell-api/src/collection.ts
addaleax: I think that’s very surprising behavior, and not what we are implementing here…

rose-m: I see that the _server_ accepts the verbosity... I'll change it accordingly 👍 Sad that the docs don't show that 😞

PR #906, opened 2021-05-26
fix(service-provider-core): allow non-hostname chars in DB name MONGOSH-792

addaleax:
This also means that e.g. `host:port?option=value` is not allowed,
and instead any optional part after the port in the non-URL variant
needs to be separated by a slash (which was also the case in
the legacy shell).

### DIFF
--- a/packages/cli-repl/test/e2e-direct.spec.ts
+++ b/packages/cli-repl/test/e2e-direct.spec.ts
@@ -185,7 +185,7 @@ describe('e2e direct connection', () => {
         });
         it('when specifying multiple seeds through --host with a wrong replsetid', async() => {
           const hostlist = 'wrongreplset/' + await rs2.hostport() + ',' + await rs1.hostport() + ',' + await rs0.hostport();
-          const shell = TestShell.start({ args: ['--host', hostlist, 'admin?serverSelectionTimeoutMS=2000'] });
+          const shell = TestShell.start({ args: ['--host', hostlist, 'admin'] });
           await shell.waitForExit();
           shell.assertContainsOutput('MongoServerSelectionError');
         });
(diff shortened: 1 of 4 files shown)


packages/cli-repl/test/e2e-direct.spec.ts
```diff
@@ -185,7 +185,7 @@ describe('e2e direct connection', () => {
         });
         it('when specifying multiple seeds through --host with a wrong replsetid', async() => {
           const hostlist = 'wrongreplset/' + await rs2.hostport() + ',' + await rs1.hostport() + ',' + await rs0.hostport();
-          const shell = TestShell.start({ args: ['--host', hostlist, 'admin?serverSelectionTimeoutMS=2000'] });
+          const shell = TestShell.start({ args: ['--host', hostlist, 'admin'] });
```
addaleax, line 188: The legacy shell would have interpreted the previous string as an entire database name here (and it would be a valid one)

packages/service-provider-core/src/uri-generator.ts
```diff
@@ -129,8 +129,8 @@ function generateUri(options: CliOptions): string {
     return '';
   }
   const connectionString = generateUriNormalized(options);
-  if (connectionString.hosts.length === 1 &&
-      ['localhost', '127.0.0.1'].includes(connectionString.hosts[0].split(':')[0])) {
+  if (connectionString.hosts.every(host =>
+    ['localhost', '127.0.0.1'].includes(host.split(':')[0]))) {
```
addaleax, line 133: This is more of a drive-by change to account for the fact that the e2e-direct test cannot use serverSelectionTimeoutMS directly anymore, but I feel like it still just makes sense

PR #909, opened 2021-05-26
chore: remove compass shell

mcasimir:
Removing compass-shell from here, since is now in the compass monorepo.


addaleax (approved):
Nice. :)

rose-m (approved):
🚀

PR #912, opened 2021-05-27
chore: update node driver to 4.0.0-beta.5 MONGOSH-795

rose-m:
- Types that are now not visible anymore:
	- `Topology` - okay, we can just use `any` 
	- `ConnectionPool` / `Connection` - okay-ish but bad for our monkey patching type safety
- Members that are not visible anymore:
	- `MongoClient.topology` - have to rely on an any cast there
	- **breaking:** `ClientSession.clusterTime` / `ClientSession.operationTime` 
		- required for us to expose in `Session` object
		- also referenced for usage in code examples: [https://docs.mongodb.com/manual/core/read-isolation-consistency-recency/#std-label-sessions](https://docs.mongodb.com/manual/core/read-isolation-consistency-recency/#std-label-sessions)
		- workaround: can still use `any`


rose-m:
@nbbeeken @emadum In https://github.com/mongodb/node-mongodb-native/pull/2810 you restricted visibility of `ClientSession.clusterTime` / `ClientSession.operationTime` to internal. Would it be possible to make them public again? As far as I can see from the official docs (linked above, code sample) they seem to be part of regular API of other drivers.

packages/service-provider-server/src/mongodb-patches.ts
```diff
@@ -14,23 +14,23 @@ export function ensureMongoNodeNativePatchesAreApplied(): void {
   alreadyPatched = true;
 }
 
-const poolToConnections = new Map<ConnectionPool, Set<Connection>>();
+const poolToConnections = new Map<any, Set<any>>();
```
addaleax, line 17: I feel like this file might be a bit more readable with a `type ConnectionPool = any; type Connection = any;` at the top, just so we know which `any` is which

packages/shell-api/src/shell-internal-state.ts
```diff
@@ -258,7 +258,7 @@ export default class ShellInternalState {
       topology: () => {
         let topology: Topologies;
         const topologyDescription = this.currentServiceProvider.getTopology()?.description as TopologyDescription;
```
addaleax, line 260: It’s good that TopologyDescription is still public. :)

addaleax (approved):

packages/service-provider-server/src/mongodb-patches.ts
rose-m: Neat, yes absolutely 👍

PR #918, opened 2021-05-31
fix(cli-repl): set terminal window title MONGOSH-725

rose-m:


packages/cli-repl/src/run.ts
```diff
@@ -38,9 +39,12 @@ import { generateUri } from '@mongosh/service-provider-server';
         process.removeAllListeners('SIGINT');
       }
 
-      process.title = 'mongosh';
       const driverOptions = await mapCliToDriver(options);
       const driverUri = generateUri(options);
+
+      process.title = `mongosh ${redactCredentials(driverUri)}`;
+      setTerminalWindowTitle(process.title);
```
addaleax, line 46: Since `process.title` is a getter/setter that actually queries the current process title (which might differ from what we have just set it to, for example, because it was truncated or because setting `process.title` isn’t available on the current platform), I think it makes sense to store the title in a variable and then set it both with `setTerminalWindowTitle` and `process.title` separately.

packages/cli-repl/src/run.ts
```diff
@@ -63,3 +67,13 @@ import { generateUri } from '@mongosh/service-provider-server';
     process.exit(1);
   }
 })();
+
+function setTerminalWindowTitle(title: string): void {
+  // see: https://that.guru/blog/automatically-set-tmux-window-name/
```
addaleax, line 72:
```suggestion
  if (!process.stdout.isTTY) return;
  // see: https://that.guru/blog/automatically-set-tmux-window-name/
```

packages/cli-repl/src/run.ts
rose-m: O.o I didn't know that might actually change otherwise!

addaleax: Yeah … the problem is that on UNIX systems, this actually works by *overwriting* the original `argv` array that was passed to main; and so the maximum length of `process.title` is also the maximum length of the original command line:

```
$ node
rocesWelcome to Node.js v17.0.0-pre.
Type ".help" for more information.
> process.title = 'a'.repeat(100)
'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa'
> process.title.length
4
```

addaleax (approved):

PR #919, opened 2021-05-31
feat(cli-repl): only store non-default config keys on disk MONGOSH-794

addaleax:
For context, the original motivation here was that for snippets,
if we change the default index URL, users should receive that change
if  they have not explicitly changed the config themselves.

### DIFF
--- a/packages/cli-repl/src/cli-repl.spec.ts
+++ b/packages/cli-repl/src/cli-repl.spec.ts
@@ -111,6 +111,29 @@ describe('CliRepl', () => {
         expect(JSON.parse(content).enableTelemetry).to.be.false;
       });
 
+      it('does not store config options on disk that have not been changed', async() => {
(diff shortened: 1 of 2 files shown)


rose-m (approved):
👍 I somehow find the wording not _entirely_ correct as it's more a "only store explicitly set config options to disk". For the "non-default" I'd had expected it to only store it to disk when the _actual value_ diverges from the default setting.

addaleax:
> For the "non-default" I'd had expected it to only store it to disk when the _actual value_ diverges from the default setting.

Yeah, I’m also not particularly sure what the expected behavior there is – if you set it to (a value that happens to be) the default value, and then we change the default, what would the right value afterwards?

rose-m:
That's exactly what I was thinking about... but yeah I'm totally fine with the semantics of "if you set the value explicitely, it will stay like that".

PR #920, opened 2021-06-01
fix(cli-repl,browser-repl): sort collection names and mark system collections MONGOSH-782

rose-m:
image (1)

### DIFF
--- a/packages/cli-repl/src/format-output.ts
+++ b/packages/cli-repl/src/format-output.ts
@@ -119,9 +119,25 @@ function formatSimpleType(output: any, options: FormatOptions): any {
 }
 
 function formatCollections(output: CollectionNamesWithTypes[], options: FormatOptions): string {
-  const tableEntries = output.map(
-    (coll) => [clr(coll.name, 'bold', options), coll.badge]
-  );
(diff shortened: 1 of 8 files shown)


packages/cli-repl/src/format-output.ts
```diff
@@ -119,9 +119,25 @@ function formatSimpleType(output: any, options: FormatOptions): any {
...
+      otherCollections.push(coll);
+    }
+  });
+
+  const tableEntries = [
+    ...otherCollections.map(
+      coll => [clr(coll.name, 'bold', options), coll.badge]
+    ),
+    ...systemCollections.map(
+      coll => [`\u001b[2m${clr(coll.name, 'bold', options)}`, coll.badge]
```
addaleax, line 138: It’s too bad that ansi-escape-sequences doesn’t support this out of the box – but in any case, we want to only apply the `\u001b[2m` here if `options.colors` is set, I think

rose-m: Yeah I took a look but unfortunately they don't :/ But totally missed the options.color thing...

addaleax (approved):

PR #923, opened 2021-06-02
feat(service-provider-server): pass `promoteLongs: false` MONGOSH-809

addaleax:


packages/cli-repl/test/e2e-auth.spec.ts
```diff
@@ -269,7 +269,11 @@ describe('Auth e2e', function() {
             'db.dropAllUsers()'
           );
           await eventually(async() => {
-            shell.assertContainsOutput('{ n: 2, ok: 1 }');
+            try {
+              shell.assertContainsOutput('{ n: 2, ok: 1 }');
```
rose-m, line 273: Question: is that caused by a version difference that at some point the server returns `n` as a long now? 🤔

rose-m (approved):

packages/cli-repl/test/e2e-auth.spec.ts
addaleax: Yes, it does (on 5.0+) – good that you point this out, I’ll add a comment :+1:

PR #931, opened 2021-06-07
chore(build): publish new mongosh versions to homebrew/homebrew-core MONGOSH-807

rose-m:
The fork is also already created so we can create PRs to `homebrew/homebrew-core`.

### DIFF
--- a/packages/build/src/homebrew/generate-formula.spec.ts
+++ b/packages/build/src/homebrew/generate-formula.spec.ts
@@ -1,34 +1,109 @@
-import { expect } from 'chai';
-import { generateFormula } from './generate-formula';
+import chai, { expect } from 'chai';
+import sinon from 'ts-sinon';
+import { GithubRepo } from '../github-repo';
+import { generateUpdatedFormula } from './generate-formula';
 
+chai.use(require('sinon-chai'));
 
-describe('Homebrew generate-formula', () => {
-  it('returns the proper formula document', () => {
-    const expectedFormula = `require "language/node"
+const VALID_FORMULA = `require "language/node"
 
 class Mongosh < Formula
-  desc "The MongoDB Shell"
-
+  desc "MongoDB Shell to connect, configure, query, and work with your MongoDB database"
   homepage "https://github.com/mongodb-js/mongosh#readme"
-  url "https://registry.npmjs.org/@mongosh/cli-repl/-/cli-repl-1.0.0.tgz"
-  version "1.0.0"
+  url "https://registry.npmjs.org/@mongosh/cli-repl/-/cli-repl-0.14.0.tgz"
+  sha256 "7b5a140689b4460a8b87008e6b7e7cb19acbc6e6cd1ab713e1a8923f3a995ca8"
+  license "Apache-2.0"
 
-  # This is the checksum of the archive. Can be obtained with:
-  # curl -s https://registry.npmjs.org/@mongosh/cli-repl/-/cli-repl-1.0.0.tgz | shasum -a 256
-  sha256 "SHAHASH"
+  bottle do
+    sha256 arm64_big_sur: "5b4377a3429a66afcb34f8a14119112c045f86889c8a667f656282c2044c7ac1"
+    sha256 big_sur:       "b0b2c5bf6df51288ffc542c91394880689e7aa36fc5ab97d8b3e08d9d8fe145b"
+    sha256 catalina:      "099214e38e1c1ba6007719f3ff956e56b1a8c954280b3e6883bdc472d7677fe0"
+    sha256 mojave:        "9c46c746cf8f65a8418cb39237471cd6338135e398f2ace965a6c580e1165951"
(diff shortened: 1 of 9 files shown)


packages/build/src/homebrew/generate-formula.spec.ts
```diff
@@ -1,34 +1,109 @@
...
+    sha256 big_sur:       "b0b2c5bf6df51288ffc542c91394880689e7aa36fc5ab97d8b3e08d9d8fe145b"
+    sha256 catalina:      "099214e38e1c1ba6007719f3ff956e56b1a8c954280b3e6883bdc472d7677fe0"
+    sha256 mojave:        "9c46c746cf8f65a8418cb39237471cd6338135e398f2ace965a6c580e1165951"
+  end
+
+  depends_on "node@14"
+
+  def install
+    system "#{Formula["node@14"].bin}/npm", "install", *Language::Node.std_npm_install_args(libexec)
+    (bin/"mongosh").write_env_script libexec/"bin/mongosh", PATH: "#{Formula["node@14"].opt_bin}:$PATH"
```
addaleax, line 28: Do you know more about why we’re using `libexec` here?

packages/build/src/homebrew/generate-formula.ts
```diff
@@ -1,28 +1,35 @@
...
+  }
 
-  # This is the checksum of the archive. Can be obtained with:
-  # curl -s https://registry.npmjs.org/@mongosh/cli-repl/-/cli-repl-${context.version}.tgz | shasum -a 256
-  sha256 "${context.sha}"
+  const currentUrl = urlMatch[1];
+  const currentSha = shaMatch[1];
 
-  depends_on "node@14"
+  const newUrl = `https://registry.npmjs.org/@mongosh/cli-repl/-/cli-repl-${context.version}.tgz`;
```
addaleax, line 20: Do we maybe want to use the non-scoped `mongosh` npm package? (Also obviously unrelated to the existing code here)

addaleax (approved):

packages/build/src/homebrew/generate-formula.spec.ts
rose-m: Uh, no 😅 I just copied that from the old formula that was in there. I really have absolutely no idea. This basically is the content of the published formula: https://github.com/Homebrew/homebrew-core/blob/master/Formula/mongosh.rb

packages/build/src/homebrew/generate-formula.ts
rose-m: We can definitely do that... I'd probably suggest for this move that we provide a separate manual PR first to switch the formula structure in general and get it approved. Version bumps should really just be URL with version change and the hash change I assume...

nandahkrishna: Hello, Homebrew maintainer here. The automated pull request to update the formula (Homebrew/homebrew-core#79146) doesn't have a PR description/body, so that makes the `request-info` bot unhappy. I also see from this PR that you have a formula template instead of updating just the URL and checksum.

You could use a GitHub Action instead to update the formula, like [`action-homebrew-bump-formula`](https://github.com/dawidd6/action-homebrew-bump-formula) (the author, `dawidd6`, is a Homebrew maintainer). If you wouldn't want to use a 3rd party GitHub Action, just a simple workflow which does `brew bump-formula-pr` would be good too. (I'd be willing to submit a PR for this.)

rose-m: Thanks for the comment @nandahkrishna :) The formula template is only for the test - the code here essentially grabs the latest formula from `homebrew/homebrew-core` and then replaces the URL as well as the SHA by using two regular expressions.

Would that workflow be fine? Right now this is tied to a workflow that runs in the Evergreen CI.

An alternative we could do is that we set up a GitHub Action that is triggered by publishing a new release on GitHub (a non-draft release). This could then use the suggested action.

What do you think @addaleax? 😊

addaleax: > An alternative we could do is that we set up a GitHub Action that is triggered by publishing a new release on GitHub (a non-draft release). This could then use the suggested action.

I have to admit that it doesn’t feel *super* clean to me to mix two different CI systems in a way where they depend on each other…

I’m not sure, but it seems like the main concern here really is the missing PR description, I think just addressing that is fine – it should look the same from the homebrew maintainer’s perspective, right?

nandahkrishna: > Would that workflow be fine?

Thanks for the clarification, that would be fine. I incorrectly assumed that the template was being used to update the formula, in which case there would've been the risk of using an outdated template (in case there were updates in `homebrew/core` that weren't also updated here).

> it seems like the main concern here really is the missing PR description, I think just addressing that is fine – it should look the same from the homebrew maintainer’s perspective, right?

Yes, this should be good in my opinion.

PR #935, opened 2021-06-08
chore(cli-repl): improve error message for mongocryptd failure MONGOSH-799

rose-m:
image

### DIFF
--- a/packages/cli-repl/src/cli-repl.ts
+++ b/packages/cli-repl/src/cli-repl.ts
@@ -1,4 +1,4 @@
-import { MongoshInternalError, MongoshWarning } from '@mongosh/errors';
+import { MongoshInternalError, MongoshRuntimeError, MongoshWarning } from '@mongosh/errors';
 import { redactCredentials } from '@mongosh/history';
 import i18n from '@mongosh/i18n';
 import { bson, AutoEncryptionOptions } from '@mongosh/service-provider-core';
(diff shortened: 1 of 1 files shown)


packages/cli-repl/src/cli-repl.ts
```diff
@@ -428,7 +428,14 @@ class CliRepl {
   }
 
   async startMongocryptd(): Promise<AutoEncryptionOptions['extraOptions']> {
-    return await this.mongocryptdManager.start();
+    try {
+      return await this.mongocryptdManager.start();
+    } catch (e) {
+      if (e.code === 'ENOENT') {
+        throw new MongoshRuntimeError('Failed to launch mongocryptd - ensure your local installation works correctly. Please also refer to the documentation: https://docs.mongodb.com/manual/reference/security-client-side-encryption-appendix/');
```
addaleax, line 435:
```suggestion
        throw new MongoshRuntimeError('Could not find a working mongocryptd - ensure your local installation works correctly. Please also refer to the documentation: https://docs.mongodb.com/manual/reference/security-client-side-encryption-appendix/');
```

Otherwise this sounds like we actually found a binary to launch but did not succeed in doing so, which is probably a little different to debug.

Also, I don’t know if it’s worth mentioning, but there may be some valuable output in the log file.

addaleax (approved):

PR #938, opened 2021-06-09
fix: add --tlsCRLFile option to help

mmarcon:
The `--tlsCRLFile` option was missing in the output of `mongosh --help`. Now it's not anymore.

image


rose-m (approved):
Nice, thanks - sorry I missed that when adding the implementation 😅 

One nit: in the README we also have the full `--help` output - feel free to also update that there ;)

PR #941, opened 2021-06-09
feat: add strict versioned API autocomplete MONGOSH-770

addaleax:


packages/autocomplete/index.ts
```diff
@@ -217,14 +224,23 @@ function filterShellAPI(
     if (!c.startsWith(prefix)) return false;
     if (completions[c].deprecated) return false;
 
-    const serverVersion = params.connectionInfo()?.server_version;
-    if (!serverVersion) return true;
+    const apiVersionInfo = params.apiVersionInfo();
+    let isAcceptableVersion;
+    let acceptableApiVersions;
+    if (apiVersionInfo?.strict && (acceptableApiVersions = completions[c].apiVersions)) {
```
rose-m, line 230: I would strongly prefer not having the assignment as a side effect in there if you don't mind 😅 
```suggestion
    if (apiVersionInfo?.strict && completions[c].apiVersions) {
    	const acceptableApiVersions = completions[c].apiVersion;
```

rose-m (commented):
So nice how the decorators make it easy to implement that 👍

packages/autocomplete/index.ts
addaleax: Yeah, this is exactly what I had here before, character-by-character :) TypeScript doesn’t get that the first `completions[c].apiVersions` being truthy implies that `acceptableApiVersions` is also set, unfortunately…

rose-m: Ah crap... go ya... I'm fine with leaving it like that then 👍

rose-m (approved):

PR #943, opened 2021-06-10
chore: add apiStrict-only test to CI MONGOSH-756

addaleax:
Run the shell-api and cli-repl test suites with the `strict` and
`deprecationErrors` versioned API options, and skip tests where
this does not apply.

### DIFF
--- a/packages/cli-repl/test/e2e.spec.ts
+++ b/packages/cli-repl/test/e2e.spec.ts
@@ -336,7 +336,10 @@ describe('e2e', function() {
       shell.assertNoErrors();
     });
 
-    it('rewrites async properly for mapReduce', async() => {
+    it('rewrites async properly for mapReduce', async function() {
+      if (process.env.MONGOSH_TEST_FORCE_API_STRICT) {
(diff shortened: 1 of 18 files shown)


packages/cli-repl/test/e2e.spec.ts
```diff
@@ -336,7 +336,10 @@ describe('e2e', function() {
       shell.assertNoErrors();
     });
 
-    it('rewrites async properly for mapReduce', async() => {
+    it('rewrites async properly for mapReduce', async function() {
+      if (process.env.MONGOSH_TEST_FORCE_API_STRICT) {
+        return this.skip(); // mapReduce is unversioned
+      }
```
rose-m, line 342: We could replace these with `skipIfApiStrict()` but that's a nit IMHO :)

rose-m (approved):
:shipit:

packages/cli-repl/test/e2e.spec.ts
addaleax: Yeah, I just wanted to avoid the extra nesting here :)

PR #949, opened 2021-06-15
fix(service-provider-server): account for multiple hosts in fail-fast…

addaleax:
… MONGOSH-828

Only fail connection attempts early if they fail for all hosts,
not just a single host.

### DIFF
--- a/packages/cli-repl/test/e2e.spec.ts
+++ b/packages/cli-repl/test/e2e.spec.ts
@@ -939,9 +939,17 @@ describe('e2e', function() {
       expect(exitCode).to.equal(1);
     });
 
-    it('fails fast for ECONNREFUSED errors', async() => {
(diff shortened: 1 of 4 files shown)


packages/cli-repl/test/e2e.spec.ts
```diff
@@ -939,9 +939,17 @@ describe('e2e', function() {
       expect(exitCode).to.equal(1);
     });
 
-    it('fails fast for ECONNREFUSED errors', async() => {
+    it('fails fast for ECONNREFUSED errors to a single host', async() => {
       const shell = TestShell.start({ args: [
-        '--port', '0'
+        '--port', '65535'
```
addaleax, line 944: Using 0 actually made this fail with a MongoParseError, without trying to connect at all

rose-m (approved):

PR #974, opened 2021-06-30
chore(shell-api): deprecate collection.mapReduce for 5.0+ MONGOSH-854

rose-m:
Changes the `@deprecate` decorator to take a parameter to allow to specify a version `since` that method is deprecated and include a message that should be printed as deprecation message.

Using `@deprecate()` will be equivalent to the old behavior (always deprecated) and will _not_ automatically print a deprecation message.

Deprecated methods will not be included in autocompletion (depending on the `since` value compared to server version).


rose-m:
See #976

PR #976, opened 2021-06-30
chore(shell-api): deprecate Collection.mapReduce MONGOSH-854

rose-m:


packages/shell-api/src/decorators.ts
```diff
@@ -459,7 +505,15 @@ export function topologies(topologiesArray: Topologies[]): Function {
     descriptor.value.topologies = topologiesArray;
   };
 }
+
 export const nonAsyncFunctionsReturningPromises: string[] = []; // For testing.
+/**
+ * Marks the decorated method as having a synthetic promise return value that needs to be implicitly
+ * await by the async rewriter.
```
addaleax, line 512:
```suggestion
 * awaited by the async rewriter.
```

addaleax (approved):

PR #980, opened 2021-07-01
chore(ci): test s390x builds on RHEL 8.0 MONGOSH-768

addaleax:
The platform support matrix for the 5.0 server now indicates that
RHEL 8 will be supported in addition to, not instead of, RHEL 7
on the zSeries/s390x platform.

This is the optimistic approach, where we build the same executable
for both targets. If this fails in CI, we may need to start
building separate binaries.

### DIFF
--- a/.evergreen.yml
+++ b/.evergreen.yml
@@ -1457,8 +1457,8 @@ buildvariants:
     tasks:
       - name: compile_artifact
   - name: linux_s390x_build
-    display_name: "RHEL 8.0 s390x (build)"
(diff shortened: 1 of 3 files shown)


rose-m (approved):
🤞

addaleax:
Not sure why evergreen isn’t creating an automatic patch for this, I started https://spruce.mongodb.com/version/60dd9cfc2a60ed2f9c2c336c/tasks as a manual patch

PR #984, opened 2021-07-02
chore(shell-api): add tests for dots and dollars in fields MONGOSH-846

addaleax:


rose-m (approved):
Looks good - however the `apiStrict` version does not like it 😅

addaleax:
@rose-m That might be a server bug… I’ll try to figure out

rose-m (approved):

PR #992, opened 2021-07-06
fix(shell-api): configurable batch size option applies only to display MONGOSH-877

rose-m:


packages/shell-api/src/change-stream-cursor.ts
```diff
@@ -41,7 +41,7 @@ export default class ChangeStreamCursor extends ShellApiWithMongoClass {
       throw new MongoshRuntimeError('ChangeStreamCursor is closed');
     }
     const result = this._currentIterationResult = new CursorIterationResult();
-    return iterate(result, this, this._batchSize ?? await this._mongo._batchSize());
+    return iterate(result, this, this._batchSize || await this._mongo._displayBatchSize());
```
addaleax, line 44: I’m not sure I understand why we treat this differently than the other cursors? Is there a particular reason?

addaleax (approved):

packages/shell-api/src/change-stream-cursor.ts
rose-m: The `.batchSize()` method for the ChangeStreamCursor does not have an effect apart from the visual batching. The other cursor batch sizes do affect the `getMore()` calls that the driver's cursor internally makes. Hence I figured to treat it differently here as well as in the help for the `batchSize` method.

addaleax: Right … I think it’s very confusing that the `.batchSize()` methods do entirely different things now? I’d probably just remove the method then.

rose-m: Yeah good point... I'll kick it out. And in light of the discussion and NODE-3411 I'd also remove the default batch size options we introduced completely.

packages/shell-api/src/abstract-cursor.ts
```diff
@@ -63,7 +63,7 @@ export abstract class AbstractCursor extends ShellApiWithMongoClass {
 
   async _it(): Promise<CursorIterationResult> {
     const results = this._currentIterationResult = new CursorIterationResult();
-    await iterate(results, this, this._batchSize ?? await this._mongo._batchSize());
+    await iterate(results, this, await this._mongo._displayBatchSize());
```
addaleax, line 66: This means that the `batchSize` method here and the `_batchSize` property on the abstract cursor can be removed?

rose-m: Ah good spot - I think we can also refactor that a bit since we only have `AggregationCursor` and `Cursor` as the subclass which are essentially wrappers for driver cursors.

addaleax (approved):

PR #1002, opened 2021-07-09
fix(cli-repl): json-clone objects before passing them to pino MONGOSH…

addaleax:
…-884

### DIFF
--- a/packages/cli-repl/src/setup-logger-and-telemetry.ts
+++ b/packages/cli-repl/src/setup-logger-and-telemetry.ts
@@ -32,6 +32,7 @@ import type {
   SpResolveSrvSucceededEvent,
   SpMissingOptionalDependencyEvent
 } from '@mongosh/types';
+import { inspect } from 'util';
 
 interface MongoshAnalytics {
   identify(message: {
@@ -47,6 +48,22 @@ interface MongoshAnalytics {
       [key: string]: any;
     }
   }): void;
+}
+
+// The default serializer for pino replaces circular properties it finds
+// anywhere in the object graph. In extreme cases, e.g. when a mongosh cursor
+// is passed to it, it follows the graph all the way to the network socket and
+// replaces random properties of its internal state with the string
+// '"[Circular"]':
+//   https://github.com/davidmarkclements/fast-safe-stringify/blob/0e011f068962e8f8974133a47afcabfc003f2183/index.js#L36
+// As a workaround, we do a JSON-based clone for any input that can contain
+// arbitrary values.
+function jsonClone<T>(input: T): T | { _inspected: string } {
+  try {
+    return JSON.parse(JSON.stringify(input));
+  } catch (error) {
+    return { _inspected: inspect(input) };
+  }
 }
 
 // set up a noop, in case we are not able to connect to segment.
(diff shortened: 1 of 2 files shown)


packages/cli-repl/src/setup-logger-and-telemetry.ts
```diff
@@ -355,7 +372,10 @@ export default function setupLoggerAndTelemetry(
   });
 
   bus.on('mongosh-sp:connect-heartbeat-failure', function(ev: SpConnectHeartbeatFailureEvent) {
-    log.info('mongosh-sp:connect-heartbeat-failure', ev);
+    log.info('mongosh-sp:connect-heartbeat-failure', {
+      ...ev,
+      failure: ev.failure?.message
+    });
```
addaleax, line 378: This is more of a drive-by fix, it matches what we do below. Otherwise, `Error` instances are always just serialized to `{}` (regardless of the rest of this patch).

packages/cli-repl/src/setup-logger-and-telemetry.ts
```diff
@@ -186,11 +203,11 @@ export default function setupLoggerAndTelemetry(
   });
 
   bus.on('mongosh:setCtx', function(args: ApiEvent) {
-    log.info('mongosh:setCtx', args);
+    log.info('mongosh:setCtx', jsonClone(args));
   });
 
   bus.on('mongosh:api-call', function(args: ApiEvent) {
-    log.info('mongosh:api-call', redactInfo(args));
+    log.info('mongosh:api-call', redactInfo(jsonClone(args)));
```
mcasimir, line 210: is it the other way around? or `redactInfo` also modifies the object?

mcasimir (approved):

rose-m (approved):

packages/cli-repl/src/setup-logger-and-telemetry.ts
addaleax: @mcasimir `redactInfo` does not modify its input, so the order does not matter here

mcasimir: Ah ok! What happens if `jsonClone` gets the `_inspected` version?

addaleax: `redactInfo` works on strings – if it gets an object, it traverses that object and looks for strings to replace. So I think this way it’s fine, or probably even better, because this way `redactInfo` will handle circular input properly as well, which it didn’t before:

```
> c = {}; c.c = c; require('mongodb-redact')(c)
Uncaught RangeError: Maximum call stack size exceeded
    at Object.toString (<anonymous>)
    at objectToString (/home/addaleax/src/mongosh/packages/cli-repl/node_modules/lodash/lodash.js:6598:35)
    at baseGetTag (/home/addaleax/src/mongosh/packages/cli-repl/node_modules/lodash/lodash.js:3111:11)
    at Function.isPlainObject (/home/addaleax/src/mongosh/packages/cli-repl/node_modules/lodash/lodash.js:12105:35)
    at redact (/home/addaleax/src/mongosh/packages/cli-repl/node_modules/mongodb-redact/lib/index.js:5:9)
    at /home/addaleax/src/mongosh/packages/cli-repl/node_modules/lodash/lodash.js:13469:38
```

PR #1009, opened 2021-07-14
fix(ci): fix vscode integration test

addaleax:
Account for the fact that some things apparently don’t work with the
git version of the driver.


addaleax:
I’ll merge since it un-breaks CI

PR #1012, opened 2021-07-15
feat: add --build-info cli flag MONGOSH-896

addaleax:
As part of this:

- Rename the analytics config writer part of the build package
  and add information other than the Segment API key to it
- Clean up some functions in the build package to just take
  a config argument instead of a list of arguments with names
  that differ between call site and function definition (this
  was super confusing before, and still is in a lot of places)
- Make sure that the compiled binary identifies itself as such
  in the evergreen script


packages/build/src/compile/generate-bundle.ts
```diff
@@ -1,29 +1,28 @@
...
+  console.info('mongosh: creating bundle:', config.execInput);
 
   // Parcel is the saviour here since it was the only bundling
   // tool that could figure out how to handle everything in a
   // complex lerna project with cyclic dependencies everywhere.
-  const bundler = new Bundler(entryFile, {
-    outDir: path.dirname(bundleOutputFile),
-    outFile: path.basename(bundleOutputFile),
+  const bundler = new Bundler(config.input, {
+    outDir: path.dirname(config.execInput),
```
rose-m, line 24: Yeah I remember being confused by the name `execInput` here as the `outFile` for the bundle. Maybe we can rename `execInput` to `bundleFile` inside config? (no need to do in that PR)

rose-m (approved):

packages/build/src/compile/generate-bundle.ts
addaleax: Yes, I’m very much up for renaming a number of config options :)

PR #1022, opened 2021-07-19
fix(cli-repl): do not require password for GSSAPI authentication MONGOSH-899

rose-m:
Also verified to test with Kerberos on Windows.

### DIFF
--- a/packages/cli-repl/src/cli-repl.ts
+++ b/packages/cli-repl/src/cli-repl.ts
@@ -127,6 +127,7 @@ class CliRepl {
     if (this.isPasswordMissing(driverOptions)) {
       await this.requirePassword(driverUri, driverOptions);
     }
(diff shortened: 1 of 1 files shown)


addaleax (approved):
Is there any chance we could add a test for this, either here or on the Node.js side? An option having different behavior between being `undefined` vs not being present at all seems like the kind of thing that a driver change could accidentally break very easily

rose-m:
Yeah I'm not sure how to best test it... The Kerberos connectivity tests already work without a password as the `askpassword` handles that with an empty password it seems. The only thing I could do is to verify in there that it even does not ask for a password anymore...?

rose-m:
@addaleax Yeah I'll try add a test case to the driver for that 👍 Would then merge this PR.

addaleax:
@rose-m Yes, please feel free to do so :)

PR #1024, opened 2021-07-20
chore(docs): add execution flow of CLI REPL eval

rose-m:
Added a basic doc description of how execution flows through the shell when user input is evaluated. Basically includes also the following diagram:

cli-code-evaluation

Diagram source is a draw.io file and is checked in.


addaleax (approved):
nice :)

PR #1026, opened 2021-07-20
chore(ci): cache compiled TS result MONGOSH-870

addaleax:


rose-m (approved):
Looks great! That should save quite some execution time for the individual tasks... somewhere around 5mins it seems after doing a quick comparison 🚀

PR #1037, opened 2021-07-26
fix(shell-api): prevent max listeners warning MONGOSH-919

rose-m:


packages/shell-api/src/interruptor.ts
```diff
@@ -17,6 +17,12 @@ export class InterruptFlag {
   private interrupted = false;
   private onInterrupt = new EventEmitter();
 
+  constructor() {
+    // since we might have user code in a very high async fashion
+    // we don't know beforehand how many listeners we have to register
+    this.onInterrupt.setMaxListeners(Number.POSITIVE_INFINITY);
```
addaleax, line 23:
```suggestion
    this.onInterrupt.setMaxListeners(Infinity);
```

for consistency?

```
$ git grep -w POSITIVE_INFINITY .|wc -l
0
$ git grep -w Infinity .|wc -l
23
```

packages/shell-api/src/shell-internal-state.ts
```diff
@@ -209,8 +209,8 @@ export default class ShellInternalState {
       contextObject.console = {};
     }
     for (const key of ['log', 'warn', 'info', 'error']) {
-      contextObject.console[key] = async(...args: any[]): Promise<void> => {
-        return await contextObject.print(...args);
+      contextObject.console[key] = (...args: any[]): any => {
```
addaleax, line 212:
```suggestion
      contextObject.console[key] = (...args: any[]): Promise<void> => {
```

would still be okay and accurate, right?

addaleax (approved):

packages/shell-api/src/shell-internal-state.ts
rose-m: Ah yeah, removed that just as a first while testing...

packages/shell-api/src/interruptor.ts
rose-m: SGTM!

PR #1048, opened 2021-07-30
fix(service-provider-core): allow specifying a host with _ characters MONGOSH-933

rose-m:


addaleax (approved):
Can we add a connectivity test (e.g. docker with a host with underscore in /etc/hosts) or at least verify that if we use `mongosh --host foo_bar`, it fails with a dns error, i.e. it is actually passed down to the OS?

This seems like something that could very easily regress because it *is* invalid and so regexps that do match valid hostnames will reject this (for example, this could also be something the driver validates).

rose-m:
@addaleax Yeah makes a lot of sense 😅 I've added a test to check we really get a `MongoNetworkError`.

packages/cli-repl/test/e2e.spec.ts
```diff
@@ -10,6 +10,7 @@ import rimraf from 'rimraf';
 import path from 'path';
 import { readReplLogfile } from './repl-helpers';
 import { bson } from '@mongosh/service-provider-core';
+import { SIGKILL } from 'constants';
```
addaleax, line 13: `constants` is deprecated: https://nodejs.org/api/deprecations.html#DEP0008

`os.constants` is probably the better choice here

addaleax (approved):

packages/cli-repl/test/e2e.spec.ts
rose-m: Thanks 😅 👍

PR #1062, opened 2021-08-05
fix(shell-api): implement own map/forEach for cursors MONGOSH-936

addaleax:
Because the driver expects `.map()` and `.forEach()` callbacks to work
synchronously, but mongosh users can use pseudo-synchronous async
functions, the driver cannot properly handle the functions we pass
to it.
In order to fix this, if we don’t jump through a lot of extra hoops,
it’s easiest to just implement `.map()` and `.forEach()` ourselves.

### DIFF
--- a/packages/cli-repl/test/e2e.spec.ts
+++ b/packages/cli-repl/test/e2e.spec.ts
@@ -446,6 +446,59 @@ describe('e2e', function() {
         });
       });
     });
+
+    describe('cursor transform operations', () => {
+      beforeEach(async() => {
+        await shell.executeLine(`use ${dbName}`);
+        await shell.executeLine('for (let i = 0; i < 3; i++) db.coll.insertOne({i})');
+      });
+
+      it('works with .map() with immediate .toArray() iteration', async() => {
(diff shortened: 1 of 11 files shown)


rose-m (approved):
Nice 😊 

Just something I noticed while looking at the code: the `Cursor` instance is not state-less in regards to the `map` function (something I kind of had expected due to the similarity with `Array.map` but what makes sense overall). The docs for `cursor.map()` currently state: _cursor.map() returns a Cursor object_. I think we should make sure that the docs clarify that `cursor.map()` modifies the _behavior_ of the cursor instance and _does not_ return a new instance but the _modified_ instance for further chaining.

addaleax:
Fwiw, in response to https://github.com/mongodb/docs/pull/5646#pullrequestreview-719582554 they added:

> Note that ``.map()`` only converts the type, it does not create a new cursor.

I guess that’s as good as it gets?

rose-m:
Ah yeah that sounds good to me!

addaleax:
Rebased to include #1065

PR #1066, opened 2021-08-10
feat(cli-repl): use mongodb log format for log files MONGOSH-944

addaleax:
Switch the log format to a mongod-style log format, and split out
the implementation parts that could be shared by another application
(specifically, Compass).

This uses log message ids starting from 1000000 to avoid conflicts
with mongod log ids.


rose-m (approved):

mcasimir (approved):
This was super quick! Is there any way to make the whole `id` situation transparent? I am a bit afraid it could become challenging to keep track of those ids if log calls are spread around the codebase.. but probably i'm thinking too much of Compass here.

addaleax:
@rose-m suggested trying to do the same approach we use for errors, which scans the codebase for error codes, to catalogue the log ids that we’re using and make sure that there aren’t any duplicates

addaleax:
Okay, I’ve decided against doing something typescript-based, at least for now, because a main point here is integrating with Compass and we won’t have the TS API available there initially. I have added a `mongoLogId()` wrapper for the ids that must be used, so that we can add tooling to verify the uniqueness of ids at some point.

addaleax:
Quoting https://github.com/mongodb/mongo/blob/master/src/mongo/logv2/README.md#basic-usage:

> The ID is a signed 32bit integer in the same number space as the error code
numbers. It is used to uniquely identify a log statement. If changing existing
code, using a new ID is strongly advised to avoid any parsing ambiguity. When
selecting ID during work on JIRA ticket `SERVER-ABCDE` you can use the JIRA
ticket number to avoid ID collisions with other engineers by taking ID from the
range `ABCDE00` - `ABCDE99`.

so I’ve gone ahead and bumped the base index for mongosh to 1e9.

PR #1074, opened 2021-08-24
feat(browser-runtime-core): remove preprocessing MONGOSH-931

addaleax:
Remove the preprocessing inside the browser package, as it mismatches
with the preprocessing we do inside the async rewriter. Since none
of the steps taken here were actually browser-specific in any way,
we should keep further adjustments to the async-rewriter package
itself.

Some of the existing tests for the preprocessor were ported over to
be applied directly to the `OpenContextRuntime` class.


rose-m (approved):
That line diff... 👍

PR #1077, opened 2021-08-24
feat(JMongosh): don't convert object from Java to JS and back if results are returned directly to user

kornilova203:
A simple fix but it needs to access private fields `Cursor._cursor` and `Cursor._transform`

It sets `resultsUsedInScript` flag to `false` right before cursor is returned to user.
When `resultsUsedInScript` is `false`, internal Java cursor doesn't convert Java objects to JS

### DIFF
--- a/packages/java-shell/src/test/resources/collection/insertOne.expected.txt
+++ b/packages/java-shell/src/test/resources/collection/insertOne.expected.txt
@@ -1,3 +1,3 @@
 { "acknowledged": true, "insertedId": "UNKNOWN" }
 true
-[ { "_id": <ObjectID>, "a": 1, "objectId": <ObjectID>, "maxKey": {"$maxKey": 1}, "minKey": {"$minKey": 1}, "binData": {"$binary": {"base64": "MTIzNA==", "subType": "10"}}, "date": {"$date": {"$numberLong": "1355875200000"}}, "isoDate": {"$date": {"$numberLong": "1355875200000"}}, "numberInt": 24, "timestamp": {"$timestamp": {"t": 100, "i": 0}}, "undefined": null, "null": null, "uuid": <UUID> } ]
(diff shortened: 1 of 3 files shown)


packages/java-shell/src/test/resources/collection/insertOne.expected.txt
```diff
@@ -1,3 +1,3 @@
 { "acknowledged": true, "insertedId": "UNKNOWN" }
 true
-[ { "_id": <ObjectID>, "a": 1, "objectId": <ObjectID>, "maxKey": {"$maxKey": 1}, "minKey": {"$minKey": 1}, "binData": {"$binary": {"base64": "MTIzNA==", "subType": "10"}}, "date": {"$date": {"$numberLong": "1355875200000"}}, "isoDate": {"$date": {"$numberLong": "1355875200000"}}, "numberInt": 24, "timestamp": {"$timestamp": {"t": 100, "i": 0}}, "undefined": null, "null": null, "uuid": <UUID> } ]
+[ { "_id": <ObjectID>, "a": 1, "objectId": <ObjectID>, "maxKey": {"$maxKey": 1}, "minKey": {"$minKey": 1}, "binData": {"$binary": {"base64": "MTIzNA==", "subType": "10"}}, "date": {"$date": {"$numberLong": "1355875200000"}}, "isoDate": {"$date": {"$numberLong": "1355875200000"}}, "numberInt": 24, "timestamp": Timestamp{value=429496729600, seconds=100, inc=0}, "undefined": null, "null": null, "uuid": <UUID> } ]
```
addaleax, line 3: This is a bit curious – isn’t the output supposed to be EJSON (just basing off `$maxKey`, `$minKey`, `$binary` etc.)?

addaleax (approved):

packages/java-shell/src/test/resources/collection/insertOne.expected.txt
kornilova203: Actually yes, it should.
I can fix it but then I'll need to find a better way to check if values were converted to JS or not.
I'll do it but later

PR #1085, opened 2021-09-02
chore(shell-api): add internal API documentation MONGOSH-948

addaleax:
Add internal API documentation to the parts of the shell API
that are not user-facing, i.e. exist as supporting infrastructure.

As part of this:
- Rename `ShellInternalState` to `ShellInstanceState`
- Remove the unnecessary `ShellApiInterface` interface
- Clean up the interrupt handling code a bit, in particular,
  remove the dependency on `EventEmitter` as the shell-api
  package should not depend on Node.js-specific concepts


packages/shell-api/src/decorators.ts
```diff
@@ -123,16 +168,27 @@ export async function toShellResult(rawValue: any): Promise<ShellResult> {
...
+ *
+ * This applies exclusively to classes which use the {@link addSourceToResults}
+ * decorator. These classes need to implement a {@link namespaceInfo} getter.
+ *
+ * Specifically, for classes like Collection, it can be useful to attach
+ * information to the result about the original data source, so that downstream
+ * consumers of the shell can e.g. figure out how to edit a document returned
+ * from the shell.
+ *
+ * This helper also attach the `shellApiType` property to the
```
lerouxb, line 183: probably attaches, not attach.

packages/shell-api/src/interruptor.ts
```diff
@@ -13,28 +15,30 @@ export class MongoshInterruptedError extends MongoshBaseError {
   }
 }
 
+/**
+ * Contains the interruption state for a given shell instance and
+ * informs exposes ways to listen to changes of that state.
```
lerouxb, line 20: informs exposes? :)

lerouxb (approved):

PR #1088, opened 2021-09-02
fix(cli-repl): disable short-option-groups args parser option MONGOSH…

addaleax:
…-974

This enables things like `-tls` to work like they did in the legacy shell.

### DIFF
--- a/packages/cli-repl/src/arg-parser.spec.ts
+++ b/packages/cli-repl/src/arg-parser.spec.ts
@@ -417,7 +417,7 @@ describe('arg-parser', () => {
         });
 
         context('when providing -tlsCertificateKeyFile (single dash)', () => {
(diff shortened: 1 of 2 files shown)


packages/cli-repl/src/arg-parser.spec.ts
```diff
@@ -404,6 +416,18 @@ describe('arg-parser', () => {
           });
         });
 
+        context('when providing -tlsCertificateKeyFile (single dash)', () => {
+          const argv = [ ...baseArgv, uri, '--tlsCertificateKeyFile', 'test' ];
```
rose-m, line 420:
```suggestion
          const argv = [ ...baseArgv, uri, '-tlsCertificateKeyFile', 'test' ];
```
You would want to test the single dash version, right? 😅

rose-m (changes requested):

rose-m (approved):

PR #1095, opened 2021-09-09
feat: add db.sql helper for ADL $sql MONGOSH-900

mcasimir:
Add a `db.sql` helper that runs `$sql` aggregations in ADL. 

NOTE: Also fixes the integration of eslint with VSCode.

### DIFF
--- a/packages/shell-api/src/database.ts
+++ b/packages/shell-api/src/database.ts
@@ -1427,7 +1427,7 @@ export default class Database extends ShellApiWithMongoClass {
     await this._mongo.setSecondaryOk();
   }
 
-  @serverVersions(['3.1.0', ServerVersions.latest])
+  @serverVersions(['4.4.0', ServerVersions.latest])
   @topologies([Topologies.ReplSet, Topologies.Sharded])
   @apiVersions([1])
   @returnsPromise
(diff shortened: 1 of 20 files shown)


packages/shell-api/src/database.ts
```diff
@@ -1427,7 +1427,7 @@ export default class Database extends ShellApiWithMongoClass {
     await this._mongo.setSecondaryOk();
   }
 
-  @serverVersions(['3.1.0', ServerVersions.latest])
+  @serverVersions(['4.4.0', ServerVersions.latest])
```
addaleax, line 1430: This isn’t accurate, is it? https://docs.mongodb.com/v4.0/reference/method/db.watch/

packages/shell-api/src/database.ts
```diff
@@ -1445,4 +1445,18 @@ export default class Database extends ShellApiWithMongoClass {
     this._mongo._instanceState.currentCursor = cursor;
     return cursor;
   }
+
+  @returnsPromise
+  @returnType('AggregationCursor')
+  async sql(sqlString: string, options?: Document): Promise<AggregationCursor> {
```
addaleax, line 1451: I think you’ll need an entry for this in the i18n package as well (but CI will tell you for sure :))

addaleax (approved):

PR #1097, opened 2021-09-10
MONGOSH-956: Open external editor

alenakhineika:
Open an external editor from mongosh.


packages/types/src/index.ts
```diff
@@ -332,6 +353,13 @@ export interface MongoshBusEventsMap {
   'mongosh-sp:reset-connection-options': () => void;
   /** Signals that an optional dependency of the mongodb package is missing. */
   'mongosh-sp:missing-optional-dependency': (ev: SpMissingOptionalDependencyEvent) => void;
+
+  /** Signals that a run edit command has started executing. */
+  'mongosh-editor:run-edit-command': (ev: EditorRunEditCommandEvent) => void;
+  /** Signals that a run edit command has failed. */
+  'mongosh-editor:read-vscode-extensions-done': (ev: EditorReadVscodeExtensionsDoneEvent) => void;
+  /** Signals that reading vscode extensions from disc failed. */
+  'mongosh-editor:read-vscode-extensions-failed': (ev: EditorReadVscodeExtensionsFailedEvent) => void;
```
addaleax, line 363: I think we want to add these to `setup-logger-and-telemetry` in the cli-repl package so that we get log events from this :)

packages/editor/src/editor.ts
```diff
@@ -0,0 +1,187 @@
...
+  npmArgv: string[];
+
+  constructor({ input, vscodeDir, tmpDir, internalState, makeMultilineJSIntoSingleLine }: EditorOptions) {
+    const { load, config, print, require } = internalState.context;
+    this._input = input;
+    this._vscodeDir = vscodeDir;
+    this._tmpDir = tmpDir;
+    this._tmpDocName = `edit-${uuidv4()}`;
+    this._tmpDoc = '';
+    this._internalState = internalState;
```
addaleax, line 40: This was renamed to `instanceState` in https://github.com/mongodb-js/mongosh/pull/1085 :)

packages/editor/src/editor.ts
```diff
@@ -0,0 +1,187 @@
...
+  print: (...args: any[]) => Promise<void>;
+  npmArgv: string[];
+
+  constructor({ input, vscodeDir, tmpDir, internalState, makeMultilineJSIntoSingleLine }: EditorOptions) {
+    const { load, config, print, require } = internalState.context;
+    this._input = input;
+    this._vscodeDir = vscodeDir;
+    this._tmpDir = tmpDir;
+    this._tmpDocName = `edit-${uuidv4()}`;
+    this._tmpDoc = '';
```
addaleax, line 39: I think we don’t need to store these on the instance, and instead can generate a new filename every time we open the editor?

packages/editor/src/editor.ts
```diff
@@ -0,0 +1,187 @@
...
+
+  activate(): void {
+    this._internalState.registerPlugin(this);
+  }
+
+  // In the case of using VSCode as an external editor,
+  // we should detect whether the MongoDB extension is installed and open a .mongodb file.
+  // In all other cases we should open a .js file.
+  async getExtension(cmd: string): Promise<string> {
+    if (!cmd.includes('code')) {
```
addaleax, line 72: We will need a stricter check here, because this would also match things like `xcode` or `codeblocks` :)

packages/editor/.eslintrc.js
```diff
@@ -0,0 +1 @@
+module.exports = require('../../config/eslintrc.base');
```
addaleax, line 1: Just leaving a comment here so that we don’t forget to update this the same way we did in https://github.com/mongodb-js/mongosh/commit/c4372550cc02878738319f7767e9ff896491fed4 :)

packages/editor/src/editor.ts
alenakhineika: I kept it to prevent passing around lots of arguments, but I moved it to the `_editorSession` to kind of keep these values altogether.

packages/editor/src/editor.ts
```diff
@@ -0,0 +1,236 @@
...
+    try {
+      return !!regex.exec(cmd);
+    } catch (error) {
+      this.messageBus.emit('mongosh-editor:is-vscode-app-regex-failed', { regex, cmd });
+      return false;
+    }
+  }
+
+  _isStatement(code: string): boolean {
+    const regex = /\b[^()]+\((.*)\)$/;
```
addaleax, line 142: This tries to check if the input is a function call, no? I think we need to be more strict here, and check the inverse, i.e. whether the input is an identifier or a series of identifiers connected by `.` (e.g.: `4 + 2`, `class A {}`, `"string"` etc. should all not be handled as identifiers)

packages/editor/src/editor.ts
```diff
@@ -0,0 +1,236 @@
...
+
+    try {
+      return !!regex.exec(code);
+    } catch (error) {
```
addaleax, line 146: I might be wrong, but I don’t think `regex.exec()` can throw (and `regex.test` might be a bit nicer because that already returns a boolean :))

packages/editor/src/editor.ts
```diff
@@ -0,0 +1,236 @@
...
+  print: (...args: any[]) => Promise<void>;
+
+  constructor({ input, vscodeDir, tmpDir, instanceState, makeMultilineJSIntoSingleLine, loadExternalCode }: EditorOptions) {
+    this._input = input;
+    this._vscodeDir = vscodeDir;
+    this._tmpDir = tmpDir;
+    this._instanceState = instanceState;
+    this._makeMultilineJSIntoSingleLine = makeMultilineJSIntoSingleLine;
+    this._loadExternalCode = loadExternalCode;
+    this._editorSession = { cmd: '', args: [], content: '', code: '', tmpDoc: '' };
```
addaleax, line 48: Is there a reason to store the state here, instead of locally in the functions where it is used? Since the state (other than maybe the content from the of the last `edit` command) does not span across multiple `runEditCommand()` calls, that would feel a bit cleaner, I think

packages/editor/src/editor.ts
```diff
@@ -0,0 +1,236 @@
...
+  }
+
+  async _readTempFile(tmpDoc: string): Promise<string> {
+    const content = await fse.readFile(tmpDoc, 'utf8');
+    await fse.unlink(tmpDoc);
+    return this._makeMultilineJSIntoSingleLine(content);
+  }
+
+  _isVscodeApp(cmd: string): boolean {
+    const regex = /^((.*[\/\\])*)?[cC]ode(.exe)?$/;
```
addaleax, line 131: fwiw, this variant of the regexp reduces to:

```suggestion
    const regex = /^(.*)[\/\\]?[cC]ode(.exe)?$/;
```

packages/editor/src/editor.ts
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    return editor;
+  }
+
+  async _createTempFile(): Promise<void> {
+    const ext = await this._getExtension(this._editorSession.cmd);
+    const content = await this._getEditorContent();
+
+    this._editorSession.tmpDoc = path.join(this._tmpDir, `edit-${uuidv4()}.${ext}`);
```
addaleax, line 117: Regarding `uuidv4()` – in mongosh, we generally use `ObjectId()`s to generate unique filenames, including for the logs (which has a few other advantages, like being able to tell the creation time or identify which files come from the same process). If there’s a reason to use UUID v4 specifically, then the `bson` package also provides an implementation of that, so I think using that would be a bit more convenient given that we already use `bson` across the board :)

packages/editor/src/editor.ts
```diff
@@ -0,0 +1,236 @@
...
+    return this._makeMultilineJSIntoSingleLine(content);
+  }
+
+  _isVscodeApp(cmd: string): boolean {
+    const regex = /^((.*[\/\\])*)?[cC]ode(.exe)?$/;
+
+    try {
+      return !!regex.exec(cmd);
+    } catch (error) {
+      this.messageBus.emit('mongosh-editor:is-vscode-app-regex-failed', { regex, cmd });
```
addaleax, line 136: (same as below)

packages/editor/src/editor.ts
```diff
@@ -0,0 +1,236 @@
...
+  async _getEditorContent(): Promise<string> {
+    if (!this._editorSession.code) {
+      return this._editorSession.content;
+    }
+
+    const evalResult = await this._loadExternalCode(this._editorSession.code, '@(shell eval)');
+    return this._isStatement(this._editorSession.code) ? this._editorSession.code : evalResult.toString();
+  }
+
+  async runEditCommand([ code, ...args ]: string[]): Promise<void> {
```
addaleax, line 161: We also need to be able to run things like `edit function foo() { return 'a  b'; }`.

The arguments here would currently be split at whitespace – that’s not useful for us, because it loses information (e.g. about neighboring whitespaces). We’ll need to be able to pass a single string from the shell-evaluator package to this function.

packages/editor/src/editor.ts
```diff
@@ -0,0 +1,236 @@
...
+    await this.print('Opening an editor...');
+    const editor: string|null = await this._getEditor();
+
+    // If none of the above configurations are found return an error.
+    if (!editor) {
+      throw new Error(`Command failed: ${[code, ...args].join(' ')} with an error: please define an external editor`);
+    }
+
+    const [ cmd, ...cmdArgs ] = editor.split(' ');
```
addaleax, line 170: I think this is going to be problematic on Windows, where the path to the editor can contain spaces as well.

(i.e.: There is effectively no good way for us to tell where in `C:\Program Files\Editor\editor.exe --flag` the path to the binary itself ends and where other arguments begin, and we’ll need the system shell to handle that.)

packages/editor/src/editor.ts
```diff
@@ -0,0 +1,236 @@
...
+    });
+
+    await this._createTempFile();
+    await this._runInChildProcess();
+  }
+
+  async _runInChildProcess(): Promise<void> {
+    const { cmd, tmpDoc, args, code } = this._editorSession;
+    const proc = spawn(cmd, [tmpDoc, ...args], {
+      env: { ...process.env, MONGOSH_RUN_NODE_SCRIPT: '1' },
```
addaleax, line 189: I don’t think we’d want to let people specify `mongosh` itself as part of the editor with modified behavior:
```suggestion
```

packages/editor/src/editor.ts
```diff
@@ -0,0 +1,236 @@
...
+
+    let stdout = '';
+    let stderr = '';
+
+    if (proc.stdout) {
+      proc.stdout.on('data', (chunk) => { stdout += chunk; });
+    }
+    if (proc.stderr) {
+      proc.stderr.on('data', (chunk) => { stderr += chunk; });
+    }
```
addaleax, line 201:
```suggestion
    (proc.stdout as NodeJS.ReadableStream).setEncoding('utf8').on('data', (chunk) => { stdout += chunk; });
    (proc.stderr as NodeJS.ReadableStream).setEncoding('utf8').on('data', (chunk) => { stderr += chunk; });
```

matches what snippet-manager does. (The `.setEncoding('utf8')` call is missing in any case.)

packages/editor/src/editor.ts
```diff
@@ -0,0 +1,236 @@
...
+
+    this._input.pause();
+
+    try {
+      const [ exitCode ] = await once(proc, 'close');
```
addaleax, line 206: Wouldn’t the `exit` event be more relevant to us?

packages/editor/src/editor.ts
```diff
@@ -0,0 +1,236 @@
...
+    }
+  }
+
+  get messageBus(): MongoshBus {
+    return this._instanceState.messageBus;
+  }
+
+  transformError(err: Error): Error {
+    return err;
+  }
```
addaleax, line 235:
```suggestion
```

packages/editor/src/editor.ts
addaleax: You could still pass the things that are now grouped into `_editorSession` as a single argument, though?

alenakhineika: Got it!

packages/editor/.eslintrc.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+  ...require('../../config/eslintrc.base'),
+  root: true,
+  parserOptions: {
+    tsconfigRootDir: __dirname,
+    project: './tsconfig.lint.json'
+  }
+};
```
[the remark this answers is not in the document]
alenakhineika: 🙈

packages/editor/src/editor.ts
```diff
@@ -0,0 +1,236 @@
...
+
+  async _createTempFile(): Promise<void> {
+    const ext = await this._getExtension(this._editorSession.cmd);
+    const content = await this._getEditorContent();
+
+    this._editorSession.tmpDoc = path.join(this._tmpDir, `edit-${uuidv4()}.${ext}`);
+
+    // Create a temp file to store content that is being edited.
+    await fse.ensureFile(this._editorSession.tmpDoc);
```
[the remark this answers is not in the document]
alenakhineika: Nice! I didn't know about 0o600, will update my code this way. And this is exactly why I use `fs-extra` because I like the `ensureFile` method :)

packages/editor/src/editor.ts
```diff
@@ -0,0 +1,236 @@
...
+  loadExternalCode: any;
+}
+
+export class Editor implements ShellPlugin {
+  _input: Readable;
+  _vscodeDir: string;
+  _tmpDir: string;
+  _instanceState: ShellInstanceState;
+  _makeMultilineJSIntoSingleLine: any;
+  _loadExternalCode: any;
```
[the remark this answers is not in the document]
alenakhineika: Yeah, I was going to do that at some point. The PR is still draft and not open ;)

packages/editor/src/editor.ts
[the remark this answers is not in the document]
addaleax: Well … the thing about `ensureFile` is that it’s almost never necessary, and a great way to introduce potential TOCTOU problems: https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use

packages/editor/src/editor.ts
alenakhineika: I think it can, but I will replace it with `test` because you are right it is already a boolean then :)

addaleax: There’s nothing in the spec that I could see throwing inside `.exec()`, unless you call it on a non-RegExp `this` value or with an argument that cannot be converted to string (e.g. `{ toString() { throw new Error }}`) … also, if `.exec()` could throw, tons of code that use it would be broken :)

packages/editor/src/editor.ts
```diff
@@ -0,0 +1,207 @@
...
+  async _getEditorContent(code: string): Promise<string> {
+    if (!code) {
+      return this._content;
+    }
+
+    if (!this._isIdentifier(code)) {
+      return code;
+    }
+
+    const evalResult = await this._loadExternalCode(code, '@(shell eval)');
```
addaleax, line 141:
```suggestion
    const evalResult = await this._loadExternalCode(code, '@(editor)');
```

packages/editor/src/editor.ts
```diff
@@ -0,0 +1,207 @@
...
+  print: (...args: any[]) => Promise<void>;
+
+  constructor({ input, vscodeDir, tmpDir, instanceState, makeMultilineJSIntoSingleLine, loadExternalCode }: EditorOptions) {
+    this._input = input;
+    this._vscodeDir = vscodeDir;
+    this._tmpDir = tmpDir;
+    this._instanceState = instanceState;
+    this._makeMultilineJSIntoSingleLine = makeMultilineJSIntoSingleLine;
+    this._loadExternalCode = loadExternalCode;
+    this._content = '';
```
addaleax, line 41: Can we name this something like `_lastContent` to make it a bit more obvious that it’s not the current, but the most recent content?

packages/editor/src/editor.ts
```diff
@@ -0,0 +1,207 @@
...
+    // Create a temp file to store content that is being edited.
+    await fs.mkdir(path.dirname(tmpDoc), { recursive: true, mode: 0o700 });
+    await fs.writeFile(tmpDoc, beautify(content));
+
+    return tmpDoc;
+  }
+
+  async _readTempFile(tmpDoc: string): Promise<string> {
+    const content = await fs.readFile(tmpDoc, 'utf8');
+    await fs.unlink(tmpDoc);
```
addaleax, line 118:
```suggestion
    await fs.unlink(tmpDoc);
    this._content = content;
```

(right? this was missing before, wasn’t it?)

packages/editor/src/editor.ts
```diff
@@ -0,0 +1,207 @@
...
+    const ext = await this._getExtension(editor);
+    const tmpDoc = await this._createTempFile({ content, ext });
+
+    this.messageBus.emit('mongosh-editor:run-edit-command', {
+      tmpDoc,
+      editor,
+      code
+    });
+
+    const proc = spawn(editor, [tmpDoc], { stdio: 'inherit' });
```
addaleax, line 166: I think this is definitely better than the previous state, but now there’s the reverse problem, because I guess it could still be reasonable for somebody to specify something in `editor` that *does* contain multiple parts (for example, I use `EDITOR='nano -cA'` locally). We can defer that to another subtask, but I think we should definitely still address it.

I’m pretty sure that we’ll need to spawn the editor using a shell by default – there is an option for that in the Node.js child_process for sure, maybe that’s even enough, maybe not.

(So we should support:

1.`EDITOR=code`
2. `EDITOR=C:\Program Files\Editor\editor.exe`
3. `EDITOR=nano -cA`

and since we *cannot* easily distinguish between the semantics of the spaces between cases 2 and 3, we will need the OS shell to do that for us.)

packages/editor/src/editor.ts
```diff
@@ -0,0 +1,207 @@
...
+        this._input.unshift(content);
+        return;
+      }
+
+      // Allow exit code 1 if stderr is empty, i.e. no error occurred, because
+      // that is how commands like `npm outdated` report their result.
+      if (exitCode === 1 && stderr === '' && stdout) {
+        stderr = stdout;
+      }
+
```
addaleax, line 193:
```suggestion
```

since this was specifically about `npm outdated`, and we’re not running that here :)

packages/editor/src/editor.ts
```diff
@@ -0,0 +1,207 @@
...
+    const proc = spawn(editor, [tmpDoc], { stdio: 'inherit' });
+    let stdout = '';
+    let stderr = '';
+
+    if (proc.stdout) {
+      (proc.stdout as NodeJS.ReadableStream).setEncoding('utf8').on('data', (chunk) => { stdout += chunk; });
+    }
+    if (proc.stderr) {
+      (proc.stdout as NodeJS.ReadableStream).setEncoding('utf8').on('data', (chunk) => { stderr += chunk; });
+    }
```
addaleax, line 175: Sorry I didn’t noticed it during my first pass, but since we’re not actually gathering stdout/stderr (because of `stdio: 'inherit'`), we can drop this

packages/editor/src/editor.ts
alenakhineika: Missed this after refactoring, was here before and will bring it back)

packages/editor/src/editor.ts
alenakhineika: Created a ticket for that: https://jira.mongodb.org/browse/MONGOSH-996

packages/editor/src/editor.ts
```diff
@@ -0,0 +1,202 @@
...
+
+    return editor;
+  }
+
+  async _createTempFile({ content, ext }: { content: string; ext: string }): Promise<string> {
+    const tmpDoc = path.join(this._tmpDir, `edit-${new bson.ObjectId()}.${ext}`);
+
+    // Create a temp file to store a content that is being edited.
+    await fs.mkdir(path.dirname(tmpDoc), { recursive: true, mode: 0o700 });
+    await fs.writeFile(tmpDoc, beautify(content));
```
addaleax, line 113:
```suggestion
    await fs.writeFile(tmpDoc, beautify(content), { mode: 0o600 });
```

packages/shell-api/src/database.spec.ts
```diff
@@ -20,6 +20,10 @@ import ChangeStreamCursor from './change-stream-cursor';
 import { CommonErrors, MongoshDeprecatedError, MongoshInvalidInputError, MongoshRuntimeError, MongoshUnimplementedError } from '@mongosh/errors';
 chai.use(sinonChai);
 
+type ShellApiErrorsWrapper = Error & {
+  codeName: string
+};
```
addaleax, line 25: If we do need to do this (why?) then can we share a single type with this name across the `shell-api` package, instead of multiple slightly divergent ones?

alenakhineika: Removed all ShellApiErrorsWrappers.

addaleax (approved):

packages/shell-api/src/database.spec.ts
```diff
@@ -2078,7 +2079,7 @@ describe('Database', () => {
       });
       it('returns warning if throws with auth error', async() => {
         const expectedError = new Error();
-        (expectedError as any).codeName = 'Unauthorized';
+        expectedError.codeName = 'Unauthorized';
```
addaleax, line 2082: I think you’ll need to undo this for CI :)

addaleax (approved):

PR #1100, opened 2021-09-10
feat: add db.sql helper for ADL $sql MONGOSH-900

mcasimir:


packages/i18n/src/locales/en_US.ts
```diff
@@ -946,6 +946,11 @@ const translations: Catalog = {
         help: {
           description: 'Database Class',
           attributes: {
+            sql: {
+              link: 'https://docs.mongodb.com/manual/reference/method/db.sql',
+              description: 'Queries an Altas Data Lake database using SQL',
```
mcasimir, line 951:
```suggestion
              description: '(Experimental) Queries an Altas Data Lake database using SQL. Note: this is an experimental feature and may be subject to change in future releases.',
```

mcasimir: @mmarcon does this looks right?

mmarcon: `(Experimental) Runs a SQL query against Atlas Data Lake. Note: this is an experimental feature that may be subject to change in future releases.`

A couple of tweaks and note that `Atlas` was misspelled.

mcasimir: Great, thanks!

addaleax (approved):

PR #1111, opened 2021-09-16
chore: split setup-logger-and-telemetry into separate package 

addaleax:
COMPASS-5035

This is done so that the Compass shell can also use the same code
to hook its own logging to the shell’s message bus.


alenakhineika (approved):
🚀

PR #1113, opened 2021-09-21
chore(logging): provide driver version for every connection MONGOSH-1000

addaleax:
TS requested that we provide a driver version for connection
attempts from the embedded shell in Compass. This moves the
driver version reporting to the service-provider-server package,
which is a more natural place for it anyway.


gribnoysup (approved):
LGTM!

PR #1116, opened 2021-09-22
chore(logging): always explicitly log URI for conn attempt

addaleax:


packages/logging/src/setup-logger-and-telemetry.ts
```diff
@@ -371,7 +371,10 @@ export function setupLoggerAndTelemetry(
   });
 
   bus.on('mongosh-sp:connect-attempt-initialized', function(ev: SpConnectAttemptInitializedEvent) {
-    log.info('MONGOSH-SP', mongoLogId(1_000_000_042), 'connect', 'Initiating connection attempt', ev);
+    log.info('MONGOSH-SP', mongoLogId(1_000_000_042), 'connect', 'Initiating connection attempt', {
+      ...ev,
+      uri: redactURICredentials(ev.uri)
```
mcasimir, line 376: Is this smart enough to redact also AWS credentials and stuff that are inside the authMechanismProperties?

mcasimir (approved):
👍

packages/logging/src/setup-logger-and-telemetry.ts
addaleax: We’re currently doing this for username + password + AWS session token: https://github.com/mongodb-js/mongosh/blob/10de823ca630b77521a20fa8f688b29bdc53bb8b/packages/history/src/redact-credentials.ts

PR #1118, opened 2021-09-24
test(editor): use a node script as an external editor MONGOSH-992

alenakhineika:
Check spawing processes by replacing an external editor with nodejs script.

### DIFF
--- a/packages/cli-repl/test/e2e.spec.ts
+++ b/packages/cli-repl/test/e2e.spec.ts
@@ -9,7 +9,7 @@ import { promisify } from 'util';
 import rimraf from 'rimraf';
 import path from 'path';
 import os from 'os';
-import { readReplLogfile, fakeExternalEditor, setTemporaryHomeDirectory } from './repl-helpers';
+import { readReplLogfile, setTemporaryHomeDirectory } from './repl-helpers';
 import { bson } from '@mongosh/service-provider-core';
 const { EJSON } = bson;
 
@@ -790,8 +790,6 @@ describe('e2e', function() {
       homedir = homeInfo.homedir;
       env = homeInfo.env;
       if (process.platform === 'win32') {
-        env.LOCALAPPDATA = path.join(homedir, 'local');
-        env.APPDATA = path.join(homedir, 'roaming');
         logBasePath = path.resolve(homedir, 'local', 'mongodb', 'mongosh');
         configPath = path.resolve(homedir, 'roaming', 'mongodb', 'mongosh', 'config');
         historyPath = path.resolve(homedir, 'roaming', 'mongodb', 'mongosh', 'mongosh_repl_history');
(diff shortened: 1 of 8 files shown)


packages/editor/src/editor.spec.ts
```diff
@@ -199,6 +222,54 @@ describe('Editor', () => {
...
-  // TODO: Write a node script that acts as an external editor and write e2e tests
+  context('runEditCommand', () => {
+    let externalEditor: string;
+
+    beforeEach(async() => {
+      externalEditor = await setupExternalEditor(base);
+    });
+
+    afterEach(async() => {
+      await fs.unlink(externalEditor);
```
addaleax, line 232: You could also make the `fs.rmdir` call above remove `base` instead of `tmpDir`, then we don’t need to worry about that here :)

packages/editor/src/editor.spec.ts
```diff
@@ -12,6 +12,29 @@ import { eventually } from '../../../testing/eventually';
...
 chai.use(sinonChai);
 
+const setupExternalEditor = async(base: string): Promise<string> => {
+  const tmpDoc = path.join(base, 'editor-script.js');
+
+  await fs.mkdir(path.dirname(tmpDoc), { recursive: true, mode: 0o700 });
+  await fs.writeFile(tmpDoc, '', { mode: 0o600 });
+
+  return tmpDoc;
+};
```
addaleax, line 22: Can we remove this function? It seems a bit like the only purpose it serves is to do `path.join(base, 'editor-script.js')` (and the `mkdir` call, but that can just go back before the `writeFile` call again)

packages/editor/src/editor.spec.ts
```diff
@@ -199,6 +221,54 @@ describe('Editor', () => {
...
+      externalEditor = await setupExternalEditor(base);
+    });
+
+    afterEach(async() => {
+      await fs.unlink(externalEditor);
+    });
+
+    it('returns an error if editor is not defined', async() => {
+      cmd = null;
+      delete process.env.EDITOR;
```
addaleax, line 237: Anything with side effects should probably go into before/after hooks, in general

alenakhineika:
Thx a lot for taking a look so quick ❤️

packages/editor/src/editor.spec.ts
```diff
@@ -83,6 +98,10 @@ describe('Editor', () => {
     });
   });
 
+  after(async() => {
+    await fs.rmdir(base, { recursive: true });
```
addaleax, line 102: Btw, my suggestion was about the `rmdir` call above, not about adding new `rmdir` calls :)

packages/editor/src/editor.spec.ts
```diff
@@ -199,6 +218,50 @@ describe('Editor', () => {
     expect(content).to.be.equal('1 + 1');
   });
 
-  // TODO: e2e or mock tests for _getExtension
-  // TODO: Write a node script that acts as an external editor and write e2e tests
+  context('runEditCommand', () => {
+    context('when editor is not defined', () => {
+      before(() => {
+        cmd = null;
+        delete process.env.EDITOR;
```
addaleax, line 225: You can drop this, it’s already being taken care of by the same line in the outer `beforeEach` hook

packages/editor/src/editor.spec.ts
alenakhineika: Yeah, I was experimenting, it is still a draft pr ;) I think now it should be what you expect. Please let me know if it isn't.

packages/shell-evaluator/src/shell-evaluator.ts
```diff
@@ -43,7 +43,7 @@ class ShellEvaluator<EvaluationResultType = ShellResult> {
       shellApi[cmd]?.acceptsRawInput &&
       !(argv[0] ?? '').startsWith('(')
     ) {
-      return shellApi[cmd](input);
+      return shellApi[cmd](input.trim().split(' ').slice(1).join(' '));
```
addaleax, line 46: This is breaking when the whitespace after the command name is not a plain space character – it’s probably best to use the same regexp as above for consistency (just without the `g` flag):

```suggestion
      return shellApi[cmd](input.trim().replace(/\s+/, ''));
```

packages/editor/src/editor.ts
```diff
@@ -127,7 +127,16 @@ export class Editor {
...
+    if (this._isNumeric(code)) {
+      return false;
+    }
+
+    const regex = /^([^!"#%&'()*+,\-/\[\]\\^`{|}~]+)$/;
+    return regex.test(code);
+  }
+
+  _isNumeric(code: string) {
+    const regex = /^-?\d+$/;
```
addaleax, line 139: This matches things like `-42`, but not `-42.0` – I’m not sure if we need to be precise about this, we’re not going to be able to catch every case here with regexes anyway

packages/editor/src/editor.spec.ts
```diff
@@ -4,14 +4,32 @@ import { Duplex, PassThrough } from 'stream';
...
 
+const fakeExternalEditor = async(
+  { base, name, flags = '', output }: { base: string, name: string, flags?: string, output: string }
+): Promise<string> => {
+  const tmpDoc = path.join(base, name);
+  const script = `(async () => {
+    const tmpDoc = process.argv[process.argv.length - 1];
+    const { promises: { writeFile } } = require('fs');
+
+    await writeFile(tmpDoc, \`${output}\`, { mode: 0o600 });
```
addaleax, line 24:
```suggestion
    await writeFile(tmpDoc, ${JSON.stringify(output)}, { mode: 0o600 });
```

So that this doesn’t break if `output` contains a backtick or escape sequences?

packages/cli-repl/test/e2e.spec.ts
```diff
@@ -785,11 +786,9 @@ describe('e2e', function() {
...
 
     beforeEach(() => {
-      homedir = path.resolve(
-        __dirname, '..', '..', '..', 'tmp', `cli-repl-home-${Date.now()}-${Math.random()}`);
-      env = {
-        ...process.env, HOME: homedir, USERPROFILE: homedir
-      };
+      const homeInfo = setTemporaryHomeDirectory();
+      homedir = homeInfo.homedir;
+      env = homeInfo.env;
```
addaleax, line 791:
```suggestion
      { homedir, env } = setTemporaryHomeDirectory();
```

packages/cli-repl/test/e2e.spec.ts
```diff
@@ -785,11 +786,9 @@ describe('e2e', function() {
...
       if (process.platform === 'win32') {
         env.LOCALAPPDATA = path.join(homedir, 'local');
         env.APPDATA = path.join(homedir, 'roaming');
```
addaleax, line 794:
```suggestion
```

packages/cli-repl/test/e2e.spec.ts
```diff
@@ -1143,5 +1142,120 @@ describe('e2e', function() {
...
+      const editor = await fakeExternalEditor(output);
+      const result = await shell.executeLine(`config.set("editor", "${editor}");`);
+
+      expect(result).to.include('"editor" has been changed');
+      shell.writeInputLine('edit');
+      await eventually(() => {
+        shell.assertContainsOutput(output);
+      });
+    });
+  });
```
addaleax, line 1259: We put the snippets tests into its own file, if you like you can do that here as well (especially since the editor tests here don’t connect to a database)

packages/cli-repl/test/e2e.spec.ts
```diff
@@ -1143,5 +1142,120 @@ describe('e2e', function() {
...
+  describe('external editor', () => {
+    let db;
+    let client;
+    let shell: TestShell;
+    let dbName;
+    let homedir: string;
+    let env: Record<string, string>;
+
+    beforeEach(async() => {
+      const homeInfo = setTemporaryHomeDirectory();
```
addaleax, line 1155:
```suggestion
      { homeInfo, env } = setTemporaryHomeDirectory();
```

packages/cli-repl/test/e2e.spec.ts
```diff
@@ -1143,5 +1142,120 @@ describe('e2e', function() {
...
+      const connectionString = await testServer.connectionString();
+
+      dbName = `test-${Date.now()}`;
+      homedir = homeInfo.homedir;
+      env = homeInfo.env;
```
addaleax, line 1160:
```suggestion
```

packages/cli-repl/test/e2e.spec.ts
```diff
@@ -1143,5 +1142,120 @@ describe('e2e', function() {
...
+
+      shell = TestShell.start({
+        args: [ connectionString ],
+        forceTerminal: true,
+        env
+      });
+
+      client = await MongoClient.connect(connectionString, {});
+      db = client.db(dbName);
```
addaleax, line 1169: Do we need `dbName`, `db`, `client`, `connectionString` etc.? I’m not seeing any database calls here (and I don’t think we need any, either), so this could be a test that uses `--nodb`

packages/editor/src/editor.spec.ts
alenakhineika: Well, I don't have these cases in my tests, so it doesn't :) If we want to cover such cases in tests I can replace it with JSON.stringify.

packages/cli-repl/test/e2e.spec.ts
alenakhineika: I also tried like this first, but typescript doesn't like ut :) `Left side of comma operator is unused and has no side effects.`

alenakhineika: Moved destructing to describe itself. It doesn't look like tests change these values, we can do it once instead of before each test.

addaleax: The problem with that is that then it always runs, even if none of the tests in the file are run, and they won’t be cleaned up by the afterEach hook anymore

packages/shell-evaluator/src/shell-evaluator.ts
```diff
@@ -43,7 +43,7 @@ class ShellEvaluator<EvaluationResultType = ShellResult> {
       shellApi[cmd]?.acceptsRawInput &&
       !(argv[0] ?? '').startsWith('(')
     ) {
-      return shellApi[cmd](input);
+      return shellApi[cmd](input.trim().split(/\s+/).slice(1).join(' '));
```
addaleax, line 46: This is still a bit buggy, now multiple adjacent whitespaces in the command are converted into a single whitespace

addaleax (approved):

packages/shell-evaluator/src/shell-evaluator.ts
```diff
@@ -35,15 +35,17 @@ class ShellEvaluator<EvaluationResultType = ShellResult> {
...
+    const argv = trimmedInput.replace(/;$/, '').split(/\s+/g);
     const cmd = argv.shift() as keyof typeof shellApi;
 
     if (
       shellApi[cmd]?.isDirectShellCommand &&
       shellApi[cmd]?.acceptsRawInput &&
       !(argv[0] ?? '').startsWith('(')
     ) {
-      return shellApi[cmd](input);
+      const code = /^(\S+)$/.test(trimmedInput) ? '' : trimmedInput.replace(/^\S+\s+/, '');
```
addaleax, line 47: This is the same as

```suggestion
      const code = trimmedInput.replace(/^\S+\s*/, '');
```

right?

Also, I wouldn't call this `code`, it's code when it refers to the `edit` command but this is fairly generic, I would call it something like `rawArgument` or `rawInput` (also because that's consistent with `acceptsRawInput` elsewhere)

PR #1123, opened 2021-09-29
chore: increase coverage for newly added packages

addaleax:
Bring `js-multiline-to-singleline` and `logging` to full line
coverage, and fix a few minor issues in the process. These
packages were recently added and not yet counted for coverage/run
in CI, see https://github.com/mongodb-js/mongosh/pull/1121.

### DIFF
--- a/packages/editor/src/editor.ts
+++ b/packages/editor/src/editor.ts
@@ -83,8 +83,8 @@ export class Editor {
       return hasMongodbExtension ? 'mongodb' : 'js';
     } catch (error) {
       this.messageBus.emit('mongosh-editor:read-vscode-extensions-failed', {
-        action: 'mongosh-editor:read-vscode-extensions-failed',
-        error: (error as Error).message
+        vscodeDir: this._vscodeDir,
+        error: error as Error
       });
 
       return 'js';
@@ -187,7 +187,7 @@ export class Editor {
         return;
       }
 
-      throw new Error(`Command '${code}' failed with an exit code ${exitCode}`);
+      throw new Error(`Command '${editor} ${path.basename(tmpDoc)}' failed with an exit code ${exitCode}`);
     } finally {
       // Resume the parent readable stream to recive data events.
       this._input.resume();
--- a/packages/logging/src/setup-logger-and-telemetry.ts
+++ b/packages/logging/src/setup-logger-and-telemetry.ts
@@ -273,7 +273,10 @@ export function setupLoggerAndTelemetry(
   });
 
   bus.on('mongosh:mongocryptd-error', function(ev: MongocryptdErrorEvent) {
-    log.warn('MONGOCRYPTD', mongoLogId(1_000_000_017), 'mongocryptd', 'Error running mongocryptd', ev);
+    log.warn('MONGOCRYPTD', mongoLogId(1_000_000_017), 'mongocryptd', 'Error running mongocryptd', {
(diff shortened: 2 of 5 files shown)


packages/editor/src/editor.ts
```diff
@@ -187,7 +187,7 @@ export class Editor {
         return;
       }
 
-      throw new Error(`Command '${code}' failed with an exit code ${exitCode}`);
+      throw new Error(`Command '${editor} ${path.basename(tmpDoc)}' failed with an exit code ${exitCode}`);
```
addaleax, line 190: This was probably not quite intended :)

packages/editor/src/editor.ts
```diff
@@ -83,8 +83,8 @@ export class Editor {
       return hasMongodbExtension ? 'mongodb' : 'js';
     } catch (error) {
       this.messageBus.emit('mongosh-editor:read-vscode-extensions-failed', {
-        action: 'mongosh-editor:read-vscode-extensions-failed',
-        error: (error as Error).message
+        vscodeDir: this._vscodeDir,
```
addaleax, line 86: This is probably the relevant information here, and matches the success case

packages/logging/src/setup-logger-and-telemetry.ts
```diff
@@ -273,7 +273,10 @@ export function setupLoggerAndTelemetry(
   });
 
   bus.on('mongosh:mongocryptd-error', function(ev: MongocryptdErrorEvent) {
-    log.warn('MONGOCRYPTD', mongoLogId(1_000_000_017), 'mongocryptd', 'Error running mongocryptd', ev);
+    log.warn('MONGOCRYPTD', mongoLogId(1_000_000_017), 'mongocryptd', 'Error running mongocryptd', {
+      ...ev,
+      error: ev.error?.message
```
addaleax, line 278: This was previously buggy, so yay for coverage catching it :)

packages/logging/src/setup-logger-and-telemetry.ts
```diff
@@ -423,14 +426,17 @@ export function setupLoggerAndTelemetry(
...
   });
 
   bus.on('mongosh-editor:read-vscode-extensions-done', function(ev: EditorReadVscodeExtensionsDoneEvent) {
-    log.error('MONGOSH-EDITOR', mongoLogId(1_000_000_043), 'editor', 'Reading vscode extensions from disc succeeded', ev);
+    log.error('MONGOSH-EDITOR', mongoLogId(1_000_000_043), 'editor', 'Reading vscode extensions from file system succeeded', ev);
   });
 
   bus.on('mongosh-editor:read-vscode-extensions-failed', function(ev: EditorReadVscodeExtensionsFailedEvent) {
-    log.error('MONGOSH-EDITOR', mongoLogId(1_000_000_044), 'editor', 'Reading vscode extensions from disc failed', ev);
+    log.error('MONGOSH-EDITOR', mongoLogId(1_000_000_044), 'editor', 'Reading vscode extensions from file system failed', {
```
addaleax, line 437: I felt like `disc` would be odd here because usually when I saw something referring to the file system, it would use `disk` with a k instead, then I discovered that it’s a whole [rabbit hole](https://en.wikipedia.org/wiki/Spelling_of_disc) and decided to just make this a bit more precise :)

alenakhineika (approved):

PR #1131, opened 2021-10-04
test: fake vscode as an external editor MONGOSH-1011

alenakhineika:
Test a use case when code is selected as an external editor.

### DIFF
--- a/packages/cli-repl/test/e2e-editor.spec.ts
+++ b/packages/cli-repl/test/e2e-editor.spec.ts
@@ -1,15 +1,17 @@
 import { expect } from 'chai';
+import fs, { promises as fsPromises } from 'fs';
 import path from 'path';
-import { promises as fs } from 'fs';
 import { promisify } from 'util';
 import rimraf from 'rimraf';
 
 import { eventually } from '../../../testing/eventually';
-import { useTmpdir, fakeExternalEditor, setTemporaryHomeDirectory } from './repl-helpers';
 import { TestShell } from './test-shell';
+import { useTmpdir, fakeExternalEditor, setTemporaryHomeDirectory } from './repl-helpers';
 
 describe('external editor e2e', () => {
   const tmpdir = useTmpdir();
+  let tmpfilesWatcher: fs.FSWatcher;
+  let tmpEditorFile: string;
   let homedir: string;
   let env: Record<string, string>;
   let shell: TestShell;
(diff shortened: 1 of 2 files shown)


packages/cli-repl/test/repl-helpers.ts
```diff
@@ -77,29 +77,32 @@ async function readReplLogfile(logPath: string) {
...
   let script: string;
 
   if (typeof output === 'string') {
-    script = `(async () => {
+    script = `#!/usr/bin/env node
+    (async () => {
       const tmpDoc = process.argv[process.argv.length - 1];
       const { promises: { writeFile } } = require("fs");
       await writeFile(tmpDoc, ${JSON.stringify(output)}, { mode: 0o600 });
     })()`;
```
addaleax, line 92:
```suggestion
    })().catch(err => { process.nextTick(() => { throw err; }); });`;
```

packages/editor/src/editor.ts
```diff
@@ -122,7 +122,7 @@ export class Editor {
   }
 
   _isVscodeApp(cmd: string): boolean {
-    const regex = /^(.*)[\/\\]?[cC]ode(.exe)?$/;
+    const regex = /^(.*)[\/\\]?[cC]ode(.exe)?((\s(.*))?)$/;
```
addaleax, line 125:
```suggestion
    const regex = /^(.*)[\/\\]?[cC]ode(.exe)?(\s(.*))?$/;
```

packages/cli-repl/test/repl-helpers.ts
```diff
@@ -77,29 +77,40 @@ async function readReplLogfile(logPath: string) {
 }
 
 
-const fakeExternalEditor = async(output?: string) => {
-  const base = path.resolve(__dirname, '..', '..', '..', 'tmp', 'test', `${Date.now()}`, `${Math.random()}`);
-  const tmpDoc = path.join(base, 'editor-script.js');
+const fakeExternalEditor = async(
+  { output, expectedExtension, tmpdir }: { output?: string, expectedExtension?: string, tmpdir: string }
+) => {
+  const editor = (expectedExtension === '.mongodb') ? 'code' : 'editor';
```
addaleax, line 83: I liked the approach where the filename is passed explicitly that you had before – this looks somewhat surprising, because it’s not obvious from the test itself that this is what makes the editor have `code` as its filename. And additionally, this would break a test where we’d test that we still use `.js` for `code` when the extension is not installed (which we should have).

packages/cli-repl/test/repl-helpers.ts
```diff
@@ -77,29 +77,40 @@ async function readReplLogfile(logPath: string) {
...
       const { promises: { writeFile } } = require("fs");
+      const assert = require("assert");
+      const path = require("path");
+
       await writeFile(tmpDoc, ${JSON.stringify(output)}, { mode: 0o600 });
-    })()`;
+
+      if (${JSON.stringify(expectedExtension ?? '')}) {
+        assert.strictEqual(path.extname(tmpDoc), ${JSON.stringify(expectedExtension)});
+      }
```
addaleax, line 99: - Can we do this before the `writeFile()` call so that writing to the doc doesn’t accidentally make a test pass at some point?
- Can we also similarly verify that the `--wait` flag was actually passed to the script?

addaleax (approved):
Only one comment, feel free to take it or leave it :)

packages/cli-repl/test/e2e-editor.spec.ts
```diff
@@ -45,76 +45,224 @@ describe('external editor e2e', () => {
...
+      });
     });
   });
 
-  it('returns a modified identifier for a.b.c', async() => {
-    const shellOriginalInput = "const myObj = { field: { child: 'string value' } }; edit('myObj')";
-    const editorOutput = `const myObj = {
-      field: {
-        child: 'new   value'
+  context('when editor is path', () => {
```
addaleax, line 166:
```suggestion
  context('when editor is executable file', () => {
```

addaleax (approved):
Just one tiny comment that could clarify the test description a bit, again, feel free to ignore if you like :)

PR #1143, opened 2021-10-26
chore: skip fetching snippet index if not entering interactive mode 

addaleax:
MONGOSH-1030

### DIFF
--- a/packages/cli-repl/src/cli-repl.ts
+++ b/packages/cli-repl/src/cli-repl.ts
@@ -222,11 +222,16 @@ class CliRepl {
     const initialServiceProvider = await this.connect(driverUri, driverOptions);
     const initialized = await this.mongoshRepl.initialize(initialServiceProvider);
 
+    const commandLineLoadFiles = this.cliOptions.fileNames ?? [];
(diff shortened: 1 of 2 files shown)


Anemy (approved):
lgtm! We can always switch to loading the command line files/scripts before snippets if users expect it but I think this is good. We could add a test in https://github.com/mongodb-js/mongosh/blob/main/packages/cli-repl/test/e2e-snippet.spec.ts but I think it's fine as is.

addaleax:
> We can always switch to loading the command line files/scripts before snippets if users expect it but I think this is good.

For context, this does make a difference in that this way, the command line scripts can use `snippet()`, e.g. you could run `mongosh --nodb --eval "snippet install analyze-schema"` or similar commands.

> We could add a test in https://github.com/mongodb-js/mongosh/blob/main/packages/cli-repl/test/e2e-snippet.spec.ts

I’m honestly not quite sure how to best do that – this should be more or less transparent, and only affect whether the cache is pre-populated or not (i.e. no visible difference in behavior except for timing differences).

addaleax:
I’ll merge this but I’m happy to adjust this if you think we should :)

PR #1147, opened 2021-11-01
fix(cli-repl): work around Windows libuv TTY bug MONGOSH-998

addaleax:


gribnoysup (approved):
Thank you for such an in-depth description of the cause and how the workaround should solve it. This sounds good to me and even though I couldn't test it, seems like it works for reporters in the JIRA ticket

PR #959, opened 2021-06-21
chore: add depcheck and cleanup dependencies MONGOSH-836

rose-m:

PR #1149, opened 2021-11-05
fix(cli-repl): re-add aws4 as dependency to cli-repl to enable AWS auth

denise-sanders:
When I used mongosh with aws roles I got the below error in the screenshot attached.

I saw that this package was added it this earlier pr
https://github.com/mongodb-js/mongosh/pull/718/files
and removed in this one
https://github.com/mongodb-js/mongosh/pull/959/files

Screen Shot 2021-11-05 at 10 53 26 AM

### DIFF
--- a/packages/service-provider-server/package-lock.json
+++ b/packages/service-provider-server/package-lock.json
@@ -64,6 +64,11 @@
 				"readable-stream": "^2.0.6"
 			}
 		},
+		"aws4": {
+			"version": "1.11.0",
+			"resolved": "https://registry.npmjs.org/aws4/-/aws4-1.11.0.tgz",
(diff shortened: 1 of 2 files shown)


addaleax (approved):
Thank you! Out of curiousity, what environment are you running this on? I assume it’s not the mongosh binary downloaded from https://www.mongodb.com/try/download/shell, right?

addaleax:
@denise-sanders If you haven’t done so already, could you sign the [MongoDB contributor’s agreement](https://www.mongodb.com/contributor)? I don’t think this type of PR would strictly require it but it might still be good to have to be sure.

denise-sanders:
Hi @addaleax !

I did sign the agreement but i used a different email (same github username though)

I installed mongosh through brew, mms is a script I have that builds a connection string based on cli params (I connect to multiple clusters) then just calls mongosh with that string

addaleax:
@denise-sanders Thanks, that makes sense then! I’ll merge this once CI is green.

addaleax:
@denise-sanders Just fyi, this is included in mongosh 1.1.2 which has just been released -- Thank you again for the PR! 🙂

PR #1150, opened 2021-11-05
chore(service-provider-core): remove tlsCertificateKeyFile option copying

addaleax:
MONGOSH-1047

Not necessary anymore since we bumped the Node.js driver to 4.1.2.


Anemy (approved):
👍  (sorry for the late review)

PR #1157, opened 2021-11-18
Upgrade react-ace to 9.5.0

lerouxb:
Some migration documentation: https://github.com/securingsincity/react-ace/blob/master/docs/Migrate-v7-to-v8.md

### DIFF
--- a/packages/browser-repl/src/components/editor.tsx
+++ b/packages/browser-repl/src/components/editor.tsx
@@ -3,12 +3,13 @@ import AceEditor from 'react-ace';
 import { Autocompleter } from '@mongosh/browser-runtime-core';
 import { AceAutocompleterAdapter } from './ace-autocompleter-adapter';
 
-import 'brace/ext/language_tools';
-import 'brace/mode/javascript';
+import 'ace-builds/src-noconflict/ext-language_tools';
+import 'ace-builds/src-noconflict/mode-javascript';
 import './ace-theme';
 
-import ace from 'brace';
-const tools = ace.acequire('ace/ext/language_tools');
+import ace from 'ace-builds';
+// TODO: can we not just import this?
+const tools = ace.require('ace/ext/language_tools');
 
 const noop = (): void => {};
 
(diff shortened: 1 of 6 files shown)


packages/browser-repl/src/components/editor.tsx
```diff
@@ -3,12 +3,13 @@ import AceEditor from 'react-ace';
...
-import 'brace/mode/javascript';
+import 'ace-builds/src-noconflict/ext-language_tools';
+import 'ace-builds/src-noconflict/mode-javascript';
 import './ace-theme';
 
-import ace from 'brace';
-const tools = ace.acequire('ace/ext/language_tools');
+import ace from 'ace-builds';
+// TODO: can we not just import this?
+const tools = ace.require('ace/ext/language_tools');
```
lerouxb, line 12: To be determined. No idea why this did ace.acequire before? Can we not just `import tools from 'ace-builds/src-noconflict/ext-language_tools';` above? And if so why didn't we always do that?

packages/browser-repl/src/components/editor.tsx
```diff
@@ -22,7 +23,7 @@ interface EditorProps {
   onClearCommand(): void | Promise<void>;
   onSigInt(): Promise<boolean>;
   operationInProgress: boolean;
-  setInputRef?(ref: { editor?: HTMLElement }): void;
+  setInputRef?(ref: { container?: HTMLElement }): void;
```
lerouxb, line 26: I'm not entirely sure what's going on here. I basically just made it compile. Looking at the types the only html element is container, not editor. But so far I haven't found evidence in history somewhere to see where that changed and if it did.

addaleax: It’s totally possible that the type here should have been `Editor` from ace – that’s also something with a `.focus()` method

lerouxb: Yup. Figured that out eventually. :)

lerouxb:
Screenshot 2021-11-19 at 11 28 48

packages/browser-repl/src/components/editor.tsx
```diff
@@ -1,15 +1,14 @@
 import React, { Component } from 'react';
 import AceEditor from 'react-ace';
+import { IAceEditor } from 'react-ace/lib/types';
```
addaleax, line 3:
```suggestion
import type { IAceEditor } from 'react-ace/lib/types';
```

addaleax (approved):

PR #1163, opened 2021-11-25
chore: no need for webpack-resolver, set useWorker: false COMPASS-4528

lerouxb:
some reference https://github.com/securingsincity/react-ace/issues/725

See https://jira.mongodb.org/browse/COMPASS-5328 for some future follow-up work

### DIFF
--- a/packages/browser-repl/src/components/editor.tsx
+++ b/packages/browser-repl/src/components/editor.tsx
@@ -1,8 +1,10 @@
 import React, { Component } from 'react';
 import AceEditor from 'react-ace';
 import type { IAceEditor } from 'react-ace/lib/types';
-import 'ace-builds';
-import 'ace-builds/webpack-resolver';
+import ace from 'ace-builds';
+// @ts-ignore
+import jsWorkerUrl from "file-loader!ace-builds/src-noconflict/worker-javascript";
+ace.config.setModuleUrl("ace/mode/javascript_worker", jsWorkerUrl);
 import tools from 'ace-builds/src-noconflict/ext-language_tools';
 import 'ace-builds/src-noconflict/mode-javascript';
 import { Autocompleter } from '@mongosh/browser-runtime-core';
--- a/packages/browser-repl/src/components/editor.tsx
+++ b/packages/browser-repl/src/components/editor.tsx
@@ -1,13 +1,9 @@
 import React, { Component } from 'react';
-import AceEditor from 'react-ace';
-import type { IAceEditor } from 'react-ace/lib/types';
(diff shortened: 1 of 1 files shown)


packages/browser-repl/src/components/editor.tsx
```diff
@@ -93,7 +92,8 @@ export class Editor extends Component<EditorProps> {
         enableLiveAutocompletion: !!this.props.autocompleter,
         enableSnippets: false,
         showLineNumbers: false,
-        tabSize: 2
+        tabSize: 2,
+        useWorker: false
```
lerouxb, line 96: Compass is already using this. This side-steps the webpack build problem now. Workers are separate entry points so are separate files that you have to build and serve up. That requires extra config to get working.

How did this work before? Ace seems to fall back if the webworker errors. In storybook and in tests everything would work and pass it would just also log the 404 as it tries to load the worker and it fails.

packages/browser-repl/src/components/editor.tsx
```diff
@@ -1,10 +1,9 @@
 import React, { Component } from 'react';
-import AceEditor from 'react-ace';
-import type { IAceEditor } from 'react-ace/lib/types';
 import 'ace-builds';
-import 'ace-builds/webpack-resolver';
 import tools from 'ace-builds/src-noconflict/ext-language_tools';
 import 'ace-builds/src-noconflict/mode-javascript';
+import type { IAceEditor } from 'react-ace/lib/types';
+import AceEditor from 'react-ace';
```
lerouxb, line 6: Everything I've experienced and read in the issues on their repo points at it being better to only import react-ace _after_ ace-builds.

packages/browser-repl/src/components/editor.tsx
```diff
@@ -1,10 +1,9 @@
 import React, { Component } from 'react';
-import AceEditor from 'react-ace';
-import type { IAceEditor } from 'react-ace/lib/types';
 import 'ace-builds';
-import 'ace-builds/webpack-resolver';
```
lerouxb, line 5: I was barking up the wrong tree - webpack-resolver is a bit of a sledgehammer here and what compass does is it uses AceEditor with setOptions useWorker: false so loading the worker files isn't an issue in that case anyway.

packages/browser-repl/src/components/editor.tsx
lerouxb: https://jira.mongodb.org/browse/COMPASS-5328

addaleax (approved):

PR #1164, opened 2021-11-29
Revert "chore(deps): upgrade react-ace to 9.5.0 COMPASS-4528 (#1157)"

lerouxb:
This reverts commit b1ab401e04fba998b4d86001573814b5e92c23eb.

Not feeling comfortable leaving this in until we know how we're going to solve this properly.

### DIFF
--- a/packages/browser-repl/package-lock.json
+++ b/packages/browser-repl/package-lock.json
@@ -941,6 +941,24 @@
 				"@babel/helper-plugin-utils": "^7.8.3"
 			}
 		},
+		"@babel/polyfill": {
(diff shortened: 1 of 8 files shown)


lerouxb:
Thinking about it some more..

lerouxb:
Nevermind. New plan: Do `useWorker: false` like we do in compass which sidesteps all the issues with loading the worker just like there and I'll file an issue to make ace a peer dep so we can use the same version in browser-repl as in compass (or whatever the top level dependency might be. cloud, perhaps?). In that task we can just fix the webpack config in general.

PR #1166, opened 2021-11-30
fix: remove sunsetted bintray repo

durran:
Bintray was sunsetted several months ago and seems they finally pulled the plug completely. Kotlin resources are now in Maven Central.


addaleax (approved):
Thank you for noticing this :)

PR #1167, opened 2021-12-01
fix(async-rewriter): ensure finally blocks always run MONGOSH-1076

addaleax:
Ensure that `finally` blocks are run even if the `try` block did
not result in an exception.

(The README change probably gives the best idea of what exactly is being changed in the generated code here)


lerouxb (approved):
ouch

PR #1190, opened 2022-01-12
fix(cli-repl): persist custom appname MONGOSH-1015

alenakhineika:
Allow users to overwrite appName in a connection string. Keep the application name via driverInfo metadata.

### DIFF
--- a/packages/cli-repl/src/arg-mapper.spec.ts
+++ b/packages/cli-repl/src/arg-mapper.spec.ts
@@ -3,8 +3,10 @@ import chai, { expect } from 'chai';
 import path from 'path';
 import sinonChai from 'sinon-chai';
 import sinon from 'ts-sinon';
-import mapCliToDriver, { applyTlsCertificateSelector } from './arg-mapper';
+import mapCliToDriver, { getTlsCertificateSelector } from './arg-mapper';
 chai.use(sinonChai);
+
+const packageJSON = require('../package.json');
 
 describe('arg-mapper.mapCliToDriver', () => {
   context('when cli args have authenticationDatabase', () => {
@@ -15,8 +17,7 @@ describe('arg-mapper.mapCliToDriver', () => {
         authSource: 'authDb',
         driverInfo: {
           name: 'mongosh',
-          platform: 'darwin',
-          version: '0.0.0-dev.0'
+          version: packageJSON.version
         }
       });
     });
@@ -30,8 +31,7 @@ describe('arg-mapper.mapCliToDriver', () => {
         authMechanism: 'SCRAM-SHA-1',
         driverInfo: {
           name: 'mongosh',
-          platform: 'darwin',
-          version: '0.0.0-dev.0'
+          version: packageJSON.version
         }
       });
     });
@@ -45,8 +45,7 @@ describe('arg-mapper.mapCliToDriver', () => {
         loggerLevel: 'error',
         driverInfo: {
           name: 'mongosh',
-          platform: 'darwin',
-          version: '0.0.0-dev.0'
+          version: packageJSON.version
         }
       });
     });
@@ -60,8 +59,7 @@ describe('arg-mapper.mapCliToDriver', () => {
         loggerLevel: 'debug',
         driverInfo: {
           name: 'mongosh',
-          platform: 'darwin',
-          version: '0.0.0-dev.0'
+          version: packageJSON.version
         }
       });
     });
@@ -77,8 +75,7 @@ describe('arg-mapper.mapCliToDriver', () => {
         },
         driverInfo: {
           name: 'mongosh',
-          platform: 'darwin',
-          version: '0.0.0-dev.0'
+          version: packageJSON.version
         }
       });
     });
@@ -94,8 +91,7 @@ describe('arg-mapper.mapCliToDriver', () => {
         },
         driverInfo: {
           name: 'mongosh',
-          platform: 'darwin',
-          version: '0.0.0-dev.0'
+          version: packageJSON.version
         }
       });
     });
@@ -112,8 +108,7 @@ describe('arg-mapper.mapCliToDriver', () => {
         },
         driverInfo: {
           name: 'mongosh',
-          platform: 'darwin',
-          version: '0.0.0-dev.0'
+          version: packageJSON.version
         }
       });
     });
@@ -127,8 +122,7 @@ describe('arg-mapper.mapCliToDriver', () => {
         retryWrites: true,
         driverInfo: {
           name: 'mongosh',
-          platform: 'darwin',
-          version: '0.0.0-dev.0'
+          version: packageJSON.version
         }
       });
     });
@@ -142,8 +136,7 @@ describe('arg-mapper.mapCliToDriver', () => {
         tls: true,
         driverInfo: {
           name: 'mongosh',
-          platform: 'darwin',
-          version: '0.0.0-dev.0'
+          version: packageJSON.version
         }
       });
     });
@@ -157,8 +150,7 @@ describe('arg-mapper.mapCliToDriver', () => {
         tlsAllowInvalidCertificates: true,
         driverInfo: {
           name: 'mongosh',
-          platform: 'darwin',
-          version: '0.0.0-dev.0'
+          version: packageJSON.version
         }
       });
     });
@@ -172,8 +164,7 @@ describe('arg-mapper.mapCliToDriver', () => {
         tlsAllowInvalidHostnames: true,
         driverInfo: {
           name: 'mongosh',
-          platform: 'darwin',
-          version: '0.0.0-dev.0'
+          version: packageJSON.version
         }
       });
     });
@@ -187,8 +178,7 @@ describe('arg-mapper.mapCliToDriver', () => {
         tlsCAFile: 'ca',
         driverInfo: {
           name: 'mongosh',
-          platform: 'darwin',
-          version: '0.0.0-dev.0'
+          version: packageJSON.version
         }
       });
     });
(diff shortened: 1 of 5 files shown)


packages/cli-repl/src/arg-mapper.ts
```diff
@@ -103,6 +104,25 @@ export function applyTlsCertificateSelector(
...
+   * @returns {DriverInfo} The driverInfo metadata.
+   */
+export function applyDriverInfo(
+  nodeOptions: MongoClientOptions
+): void {
+  const { version } = require('../package.json');
+  nodeOptions.driverInfo = {
+    name: 'mongosh',
+    version,
+    platform: process.platform
```
addaleax, line 122: This is not quite the idea behind `platform` – it doesn’t provide the OS, it provides the runtime that the user is connecting from.

This is what a connection attempt currently looks like in the server logs:

```js
$ cat mongod.log | tail -n1 | json_pp
{
   "attr" : {
      "client" : "conn40",
      "doc" : {
         "application" : {
            "name" : "mongosh 1.1.2" // <-- from appName
         },
         "driver" : {
            "name" : "nodejs", // <-- will become "nodejs|mongosh" with this patch
            "version" : "4.1.4" // <-- will become e.g. "4.3.0|1.1.9" with this patch
         },
         "os" : {
            "architecture" : "x64",
            "name" : "linux",
            "type" : "Linux",
            "version" : "5.11.0-46-generic"
         },
         "platform" : "Node.js v14.18.0, LE (unified)|Node.js v14.18.0, LE (unified)"
      },
      "remote" : "127.0.0.1:35364"
   },
   "c" : "NETWORK",
   "ctx" : "conn40",
   "id" : 51800,
   "msg" : "client metadata",
   "s" : "I",
   "t" : {
      "$date" : "2022-01-13T10:47:01.451+01:00"
   }
}
```

packages/cli-repl/src/arg-mapper.spec.ts
```diff
@@ -315,6 +436,11 @@ describe('arg-mapper.mapCliToDriver', () => {
           strict: true,
           deprecationErrors: true,
           version: '1'
+        },
+        driverInfo: {
+          name: 'mongosh',
+          platform: 'darwin',
+          version: '0.0.0-dev.0'
```
addaleax, line 443: Won’t this assertion fail when we try to do a release, where the version will be set to something different?

packages/cli-repl/src/arg-mapper.ts
```diff
@@ -103,6 +104,25 @@ export function applyTlsCertificateSelector(
   }
 }
 
+/**
+   * Applies driverInfo to track usage of mongosh,
+   * because appName can be overwritten by a user in their connection string.
+   *
+   * @param {MongoClientOptions} nodeOptions - The MongoClient options.
+   *
+   * @returns {DriverInfo} The driverInfo metadata.
```
addaleax, line 113: I think it would be a bit cleaner if the PR followed this line, i.e. instead of modifying a passed-in object of options, we only return the `DriverInfo` as an object

packages/cli-repl/src/cli-repl.ts
```diff
@@ -163,6 +163,10 @@ class CliRepl {
           cs.password = await this.requirePassword();
           driverUri = cs.href;
         }
+        if (this.isAppNameMissingURI(cs)) {
+          cs.searchParams.set('appName', `mongosh ${version}`);
+          driverUri = cs.href;
+        }
```
addaleax, line 169: If this `if` remains here, if wouldn’t be triggered in the cases in which we prompt the user for password (line 159 above, happens with e.g. `mongosh --host localhost --username user`).

packages/cli-repl/src/cli-repl.ts
```diff
@@ -454,6 +458,17 @@ class CliRepl {
...
 
+  /**
+   * Is the appName missing from the connection string?
+   *
+   * @param {ConnectionString} cs - The existing connection string.
+   *
+   * @returns {boolean} If the appName is missing.
+   */
+  isAppNameMissingURI(cs: ConnectionString): boolean {
+    return !cs.searchParams.get('appName');
```
addaleax, line 469: I think this can be inlined into the source – `!cs.searchParams.get('appName')` essentialls means “is the appName missing from the URI” :)

packages/cli-repl/src/arg-mapper.spec.ts
alenakhineika: Yes, it will, but I thought the idea is to not change this value. I took the idea from what Maurizio added to vscode: https://github.com/mongodb-js/vscode/pull/315/files#diff-a884cbf73a5998822710b6ce02a6bdb680fb24427451911940539004b83e3dbeR515

What would be a better approach here?

packages/cli-repl/src/arg-mapper.ts
alenakhineika: I aligned the behavior with the applyTlsCertificateSelector: https://github.com/mongodb-js/mongosh/pull/1190/files#diff-44d5c1b1de68d646516652eae9e4eaab41715e8afb81936829928aa1135b5e75R72-R73

I get that applyTlsCertificateSelector does more stuff, but the use cases feel very similar to me.

packages/cli-repl/src/arg-mapper.ts
alenakhineika: I think I can omit the platform then since it is not a required property.

packages/cli-repl/src/arg-mapper.spec.ts
addaleax: > I thought the idea is to not change this value.

Well, CI changes that value when we release and before we run the tests, so that’s not really going to work

> What would be a better approach here?

You can also use `require('../package.json').version` here. I know if feels a bit silly to use the same code in the test and in the source code, but I don’t really have better ideas :woman_shrugging: It’s also not entirely without precedent: https://github.com/mongodb-js/mongosh/blob/e6e75b455a65fad7b78c42571d170bc5cc1891ef/packages/cli-repl/test/e2e.spec.ts#L232

alenakhineika: I didn't know that ci changes it, yeah in this case it won't work. I will require package.json then.

packages/cli-repl/src/arg-mapper.ts
addaleax: That’s a good point – I think we should also be doing this differently in `applyTlsCertificateSelector` :slightly_smiling_face:  If it’s alright with you, you can fix that, or let me do that after this PR is merged (it could return `MongoClientOptions` instead of taking it as an argument).

alenakhineika: Sure I will fix it! Just wanted to make sure I don't miss any reason why they might be different.

packages/cli-repl/src/arg-mapper.ts
```diff
@@ -69,15 +69,23 @@ function mapCliToDriver(options: CliOptions): MongoClientOptions {
...
+  const tlsCertificateSelector = getTlsCertificateSelector(options.tlsCertificateSelector);
+  if (tlsCertificateSelector) {
+    nodeOptions.passphrase = tlsCertificateSelector.passphrase;
+    nodeOptions.pfx = tlsCertificateSelector.pfx;
+  }
+
+  const { version } = require('../package.json');
+  nodeOptions.driverInfo = { name: 'mongosh', version };
+
   return nodeOptions;
```
addaleax, line 82: Just a suggestion to simplify this a bit, feel free to take it or leave it (but I think it’s nice that it avoids spelling out `passphrase` and `pfx` explicitly):

```suggestion
  const { version } = require('../package.json');

  return {
    ...nodeOptions,
    ...getTlsCertificateSelector(options.tlsCertificateSelector),
    driverInfo: { name: 'mongosh', version }
  };
```

packages/cli-repl/src/cli-repl.ts
```diff
@@ -155,15 +155,25 @@ class CliRepl {
     await this.verifyNodeVersion();
 
     if (!this.cliOptions.nodb) {
+      let cs;
+      if (driverUri !== '') {
```
addaleax, line 159: I don’t think this condition is possible to reach – this is the return value of `generateUri()` in `run.ts`, which only returns an empty string if `nodb` is set:

https://github.com/mongodb-js/mongosh/blob/e6e75b455a65fad7b78c42571d170bc5cc1891ef/packages/service-provider-core/src/uri-generator.ts#L158-L171

So you can drop the conditional and make `cs` const (and make `driverUri = cs.href` the last line in this block, no need to duplicate the line anymore) :)

addaleax (approved):

packages/cli-repl/src/cli-repl.ts
alenakhineika: Yeah, it will look much better! I didn't like to do `driverUri = cs.href` several times :) but I didn't know about the fact that `driverUri` is always a string and just moved this check to the beginning of the function: https://github.com/mongodb-js/mongosh/pull/1190/files#diff-d998d2f6015503912abc1073f3bdb46da4c12651a7ba3d2e6170711edb19b45aL160

I will get rid of it then! Thanks for sharing the place in code when the empty string is being returned!

addaleax (approved):

packages/cli-repl/test/e2e.spec.ts
```diff
@@ -203,6 +203,44 @@ describe('e2e', function() {
...
+        shell = TestShell.start({ args: [`mongodb://${await testServer.hostport()}/`] });
+        await shell.waitForPrompt();
+        shell.assertNoErrors();
+      });
+      it('appName set correctly', async function() {
+        if (process.env.MONGOSH_TEST_FORCE_API_STRICT) {
+          return this.skip(); // $currentOp is unversioned
+        }
+        const currentOp = await shell.executeLine('db.currentOp()');
+        expect(currentOp).to.include("appName: 'mongosh 0.0.0-dev.0'");
```
addaleax, line 219: @alenakhineika This is also breaking release CI right now -- just commenting so we don't forget about it :)

PR #1192, opened 2022-01-17
[DEBUG] dummy PR for building win32 kerberos debugging binaries

addaleax:


addaleax:
I think we got out of this what we needed to know, closing this now :)

PR #1196, opened 2022-01-27
chore(ci): use macos notarization service MONGOSH-994

addaleax:
We’ll need to get proper credentials into the evergreen config, but this seems to be working so far. Yay!


.evergreen/package-and-upload-artifact.sh
```diff
@@ -48,10 +45,30 @@ else
   fi
 
   npm run evergreen-release package
+  ls -lh dist/
 
   if [ "$(uname)" == Darwin ]; then
+    # download macnotary client
+    curl -LO https://macos-notary-1628249594.s3.amazonaws.com/releases/client/v3.1.0/darwin_amd64.zip
```
Anemy, line 52: A question for my own understand if that's alright, where does this link and the link below, `https://dev.macos-notary.build.10gen.cc/api` come from? 
Nice job on this, looks a like a solid solution and it's a lot simpler to follow now.

addaleax: It’s taken from the resources linked in https://wiki.corp.mongodb.com/display/BUILD/How+to+use+MacOS+notary+service + a short Slack conversation … which are absolutely not the most discoverable things in the world :slightly_smiling_face: I’ve added a comment that links to the wiki page here.

Anemy (approved):
🔥

PR #1197, opened 2022-01-28
feat(shell-api): allow db.stats(options) MONGOSH-1108

addaleax:
Starting with 5.0.6, the server allows additional `db.stats()`
options. We extend the shell to also accept an options object.

### DIFF
--- a/packages/shell-api/src/database.ts
+++ b/packages/shell-api/src/database.ts
@@ -941,12 +941,17 @@ export default class Database extends ShellApiWithMongoClass {
 
   @returnsPromise
   @apiVersions([])
-  async stats(scale = 1): Promise<Document> {
-    this._emitDatabaseApiCall('stats', { scale: scale });
+  async stats(scaleOrOptions: number | Document = 1): Promise<Document> {
+    assertArgsDefinedType([scaleOrOptions], [['number', 'object']], 'Database.stats');
+    if (typeof scaleOrOptions === 'number') {
+      scaleOrOptions = { scale: scaleOrOptions };
+    }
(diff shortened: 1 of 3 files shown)


packages/shell-api/src/database.ts
```diff
@@ -941,12 +941,17 @@ export default class Database extends ShellApiWithMongoClass {
 
   @returnsPromise
   @apiVersions([])
-  async stats(scale = 1): Promise<Document> {
-    this._emitDatabaseApiCall('stats', { scale: scale });
+  async stats(scaleOrOptions: number | Document = 1): Promise<Document> {
+    assertArgsDefinedType([scaleOrOptions], [['number', 'object']], 'Database.stats');
+    if (typeof scaleOrOptions === 'number') {
+      scaleOrOptions = { scale: scaleOrOptions };
```
Anemy, line 947: Does the behavior of db.stats('not a number or object') change with these changes? Does the handling for that matter? Would it expand the string below into keys in the object?

Anemy (approved):

packages/shell-api/src/database.ts
addaleax: > Does the behavior of db.stats('not a number or object') change with these changes? Does the handling for that matter?

I don’t think it changes enough to be of concern – previously, we would send an invalid `scale` value in that case and the server would reject it with a `MongoServerError`, not the type checking here would just outright reject it with a `MongoshInvalidInputError`:

```js
// before
test> db.stats('foo')
MongoServerError: scale has to be a number > 0

// after
test> db.stats('foo')
MongoshInvalidInputError: [COMMON-10001] Argument at position 0 must be of type number or object, got string instead (Database.stats)
```

PR #1201, opened 2022-02-08
feat: expand config support MONGOSH-{1063,1055,959}

addaleax:
Please review on a commit-by-commit basis :)

##### feat(cli-repl): add support for global config file MONGOSH-1063

This adds support for a global YAML config file.

mongosh loads one global config file, attempting:
- `.\mongosh.cfg` on Windows, relative to the mongosh.exe binary
- `/usr/local/etc/mongosh.conf`, `/opt/homebrew/etc/mongosh.conf`
  and `/etc/mongosh.conf` on Darwin
- `/etc/mongosh.conf` on all other platforms

On Windows, this file is only loaded if mongosh is distributed
as a single-executable binary, since otherwise there is no reasonable
starting point for searching the global config file.

The global config file uses YAML, with options namespaced under
the `mongosh` key, instead of only supporting plain EJSON, since that
matches the mongod global configuration file.

##### chore(logging): move telemetry disable/enable tracking to cli-repl

Instead of manually keeping track of whether telemetry is
enabled/disabled inside the logging package, always generate
telemetry events there and let the consumer determine whether
the events should be sent out or not.

The logging package provides a utility class that lets
consumers do this easily, and adds support for queueing up events.

This has a few advantages:
- It allows consumers to pause telemetry until they have determined
  whether telemetry should be enabled. This determination should not
  need to be made at the same time as instantiating logging (which
  is as early as posssible, generally speaking).
- It simplifies the logging code to remove repetitive `if (telemetry)`
  statements by localizing this decision to a single piece of code.
- It gets rid of the unnecessary `makeAnalytics()` redirection
  in the `setupLoggerAndTelemetry` arguments.

The last part needs to come with a change in Compass as well
(mongodb-js/compass@45518410d0e900629ac85e4558b48f38fab3f094).

##### feat(cli-repl): add global forceDisableTelemetry option MONGOSH-1055

- Add a global `forceDisableTelemetry` config variable. This
  can not be set in the user config.
- Test that when the CLI starts in telemetry-disabled mode,
  no telemetry data is emitted.
- Ensure that the relative timing of command-line scripts
  and telemetry data is correct (in particular, that command-line
  scripts can disable telemetr without starting to send any).

##### feat(shell-api): add config.reset() method MONGOSH-959

This allows resetting configuration values to their default
states without having to know which value that is.

### DIFF
--- a/packages/cli-repl/src/cli-repl.spec.ts
+++ b/packages/cli-repl/src/cli-repl.spec.ts
@@ -830,7 +830,8 @@ describe('CliRepl', () => {
         it('posts analytics data', async() => {
           await cliRepl.start(await testServer.connectionString(), {});
           if (requests.length < 1) {
-            await once(srv, 'request');
+            const [, res] = await once(srv, 'request');
+            await once(res, 'close'); // Wait until HTTP response is written
           }
           expect(requests[0].req.headers.authorization)
             .to.include(Buffer.from('🔑:').toString('base64'));
@@ -895,6 +896,54 @@ describe('CliRepl', () => {
           const flushEntry = (await readReplLogfile(logFilePath)).find(entry => entry.id === 1_000_000_045);
           expect(flushEntry.attr.flushError).to.equal(null);
           expect(flushEntry.attr.flushDuration).to.be.a('number');
+          expect(requests).to.have.lengthOf(2);
+        });
+
+        it('sends out telemetry data for command line scripts', async() => {
+          cliReplOptions.shellCliOptions.eval = 'db.hello()';
+          cliRepl = new CliRepl(cliReplOptions);
+          await startWithExpectedImmediateExit(cliRepl, await testServer.connectionString());
+          expect(requests).to.have.lengthOf(2);
+        });
+
+        it('does not send out telemetry if the user starts with a no-telemetry config', async() => {
+          await fs.writeFile(path.join(tmpdir.path, 'config'), EJSON.stringify({ enableTelemetry: false }));
+          await cliRepl.start(await testServer.connectionString(), {});
+          input.write('db.hello()\n');
+          input.write('exit\n');
+          await waitBus(cliRepl.bus, 'mongosh:closed');
+          expect(requests).to.have.lengthOf(0);
+        });
+
+        it('does not send out telemetry if the user starts with global force-disable-telemetry config', async() => {
+          const globalConfigFile = path.join(tmpdir.path, 'globalconfig.conf');
+          await fs.writeFile(globalConfigFile, 'mongosh:\n  forceDisableTelemetry: true');
+
+          cliReplOptions.globalConfigPaths = [ globalConfigFile ];
+          cliRepl = new CliRepl(cliReplOptions);
+          await cliRepl.start(await testServer.connectionString(), {});
+          input.write('db.hello()\n');
+          input.write('exit\n');
+          await waitBus(cliRepl.bus, 'mongosh:closed');
+          expect(requests).to.have.lengthOf(0);
+        });
+
+        it('does not send out telemetry if the user only runs a script for disabling telemetry', async() => {
+          cliReplOptions.shellCliOptions.eval = 'disableTelemetry()';
+          cliRepl = new CliRepl(cliReplOptions);
+          await startWithExpectedImmediateExit(cliRepl, await testServer.connectionString());
+          expect(requests).to.have.lengthOf(0);
+        });
+
+        it('does not send out telemetry if the user runs a script for disabling telemetry and drops into the shell', async() => {
+          cliReplOptions.shellCliOptions.eval = 'disableTelemetry()';
+          cliReplOptions.shellCliOptions.shell = true;
+          cliRepl = new CliRepl(cliReplOptions);
+          await cliRepl.start(await testServer.connectionString(), {});
+          input.write('db.hello()\n');
+          input.write('exit\n');
+          await waitBus(cliRepl.bus, 'mongosh:closed');
+          expect(requests).to.have.lengthOf(0);
         });
 
         context('with a 5.0+ server', () => {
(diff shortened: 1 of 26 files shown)


packages/cli-repl/src/cli-repl.spec.ts
```diff
@@ -485,6 +485,60 @@ describe('CliRepl', () => {
...
+          await fs.writeFile(globalConfigFile, 'mongosh:\n  redactHistory: meow');
+
+          cliReplOptions.globalConfigPaths = [ globalConfigFile ];
+          cliRepl = new CliRepl(cliReplOptions);
+          await cliRepl.start('', {});
+
+          expect(output).to.include('Warning: Ignoring config option "redactHistory" from');
+          expect(output).to.include("redactHistory must be one of 'keep', 'remove', or 'remove-redact'");
+        });
+      });
```
Anemy, line 557: Do we have a test that ensures the options from the global conf are overiden by the user's conf?

packages/cli-repl/src/cli-repl.ts
```diff
@@ -344,6 +353,40 @@ class CliRepl {
...
+      if (fileContents.trim().startsWith('{')) {
+        config = bson.EJSON.parse(fileContents) as any;
+      } else {
+        config = (yaml.load(fileContents) as any)?.mongosh ?? {};
+      }
+      for (const [key, value] of Object.entries(config) as [keyof CliUserConfig, any][]) {
+        const validationResult = await CliUserConfigValidator.validate(key, value);
+        if (validationResult) {
+          const msg = `Warning: Ignoring config option "${key}" from ${filename}: ${validationResult}\n`;
+          this.output.write(this.clr(msg, ['bold', 'yellow']));
```
Anemy, line 417: A bit unrelated, so maybe we do nothing for now - We have these `bold', 'yellow'` warning settings in a few places in mongosh, would it make sense to have them shared as a value somewhere? Maybe in `format-output.ts`?

packages/cli-repl/src/cli-repl.ts
```diff
@@ -383,7 +426,9 @@ class CliRepl {
    */
   // eslint-disable-next-line @typescript-eslint/require-await
   async getConfig<K extends keyof CliUserConfig>(key: K): Promise<CliUserConfig[K]> {
```
Anemy, line 468: Maybe a bit of a silly question, but should the CliRepl class implement the `ConfigProvider` interface?

Anemy (commented):
Changes look good, couple questions for my understanding.
I like that were removing a bit of roundabout here. 
Will look a bit more in the morning at last two commits.

packages/cli-repl/src/cli-repl.ts
addaleax: Not a silly question at all – It should implement `MongoshIOProvider`, which includes a number of properties from `ConfigProvider`. :slightly_smiling_face: Typescript should already error out if it doesn’t do so, but I also don’t see any reason not to make the `implements` relationship explicit – done in 96d760033adfad588fb60f7f0307688692b4649c :)

packages/cli-repl/src/cli-repl.ts
addaleax: Yeah, I really love the idea. We already have centralized color handling in `clr.ts`, I’ve put something there (ac52225702e1d127f212941895cf652ea90273f8) :)

packages/cli-repl/src/cli-repl.spec.ts
addaleax: No, thank you for catching that! Added this in 8052b5f07bad4b1f4484643f19f902a7ef3544fc

packages/cli-repl/src/cli-repl.ts
```diff
@@ -263,10 +283,39 @@ class CliRepl {
...
+    // build-info.json is created as a part of the release process
+    const apiKey = this.analyticsOptions?.apiKey ?? require('./build-info.json').segmentApiKey;
+    this.segmentAnalytics = new Analytics(
+      apiKey,
+      this.analyticsOptions);
+    this.toggleableAnalytics = new ToggleableAnalytics(this.segmentAnalytics);
+  }
+
+  setTelemetryEnabled(enabled: boolean): void {
+    if (this.globalConfig === null) {
```
lerouxb, line 306: Maybe a bit nitpicky, but could you get here via user action? ie. a user wants to explicitly enable/disable telemetry and then it silently ignores that request.

packages/cli-repl/src/cli-repl.ts
```diff
@@ -340,7 +389,41 @@ class CliRepl {
...
+  }
+
+  async loadGlobalConfigFile(): Promise<Partial<CliUserConfig>> {
+    let fileContents = '';
+    let filename = '';
+    for (filename of this.globalConfigPaths) {
+      try {
+        fileContents = await fs.readFile(filename, 'utf8');
+        break;
+      } catch { /* ignore error */ }
```
lerouxb, line 403: I'm wondering if all errors should be ignored. Like.. what if one of them exists but there's some permissions error - should it not warn in that case? Maybe too much of an edge case to worry about.

lerouxb (approved):

packages/cli-repl/src/cli-repl.ts
addaleax: No, the global config is loaded pretty early – I think this line here would technically not even be necessary if the global config is loaded before the local one, but I’d also argue that this here is the safer behavior, because it prevents the user from enabling telemetry when that should not be allowed according to the global config

packages/cli-repl/src/cli-repl.ts
addaleax: Right, I also wondered about that because we do handle this a bit better for the user’s home directory files. Printing a warning might be annoying though, because users would frequently not be able to _act_ on it.

I’ve pushed an update to make it so that we log errors other than ENOENT. I think that’s reasonable as a debugging mechanism if this does happen.

packages/cli-repl/src/cli-repl.ts
Anemy: Ah cool makes sense. Nice add to make it explicit.

Anemy (approved):
lgtm! Nice feature add, and code quality improvements at the same time.

PR #1210, opened 2022-02-22
chore(deps): bump kerberos to 2.0.0

addaleax:
This also resolves MONGOSH-1059.


Anemy (approved):
lgtm! Nice work on making that kerberos 2.0 happen.

PR #1218, opened 2022-02-28
fix(ci): use Rocky Linux 8 instead of CentOS 7 for packaging

addaleax:
Rocky Linux 8 comes with a newer version of rpm, which also
provides SHA256 rpm signing, which is being requested in
HELP-31642.


mcasimir (approved):
👍

PR #1219, opened 2022-03-01
feat(shell-api): Support boolean "true" with currentOp MONGOSH-1131

leorossi:


packages/shell-api/src/database.ts
```diff
@@ -787,8 +787,11 @@ export default class Database extends ShellApiWithMongoClass {
 
   @returnsPromise
   @apiVersions([])
-  async currentOp(opts: Document = {}): Promise<Document> {
+  async currentOp(opts: Document | boolean = {}): Promise<Document> {
     this._emitDatabaseApiCall('currentOp', { opts: opts });
+    if (opts === true) {
+      opts = { $all: true };
```
addaleax, line 793: I think you want something like

```suggestion
    if (typeof opts === 'boolean') {
      opts = { $all: opts };
```

even though it probably won’t be used with `false` as an argument much?

leorossi: By [documentation](https://docs.mongodb.com/manual/reference/method/db.currentOp/) it seems that only `true` is an accepted value. It's not clear what happens using `false` as value.

addaleax: I mean, passing `false` is the same as passing no argument, but the docs do explicitly say “Can pass either a boolean or a document”.

(The legacy shell just ignores the argument entirely if it is false: https://github.com/mongodb/mongo/blob/8c85963c5e8e6cec0ba996de80e3cc7aa5a6b39d/src/mongo/shell/db.js#L835)

leorossi: ok makes sense. Added this case and related test case.

addaleax: I think you might have forgotten to `git push`? :)

leorossi: Yeah, exactly, now the ocde is updated

addaleax (approved):

PR #1225, opened 2022-03-10
fix: use user id and anonymous id as segment identity MONGOSH-1143

alenakhineika:
In `mongosh`, we don't have MongoDB user identifiers. We randomly generate an ID the first time the product is run and then we use that as a user id. The `userId` should not be a random UUID, it should store only AUID. The randomly generated UUID should be sent as `anonymousId`.

To not break statistics for existing users:
- We should leave all current users with the tracking they have.
- For new users, we should use the `anonymousID` field and fill it with a random UUID.

That being said:
- If `userId` exists we send it to segment. We also copy userId to `anonymousId` and send it to Segment as well.
- If `userId` does not exist we do not create a new one, we only generate a random UUID and send it to Segment as `anonymousId`.

### DIFF
--- a/packages/cli-repl/README.md
+++ b/packages/cli-repl/README.md
@@ -99,8 +99,8 @@ bus.emit('mongosh:connect', {
 })
 ```
 
-### bus.on('mongosh:new-user', userID, enableTelemetry)
-Where `userID` is a [BSON ObjectID][object-id] and `enableTelemetry` is a boolean flag.
+### bus.on('mongosh:new-user', telemetryAnonymousId, enableTelemetry)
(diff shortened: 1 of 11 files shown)


addaleax (approved):
Looks good, might be nice to also have somebody take a look who reviewed the vscode/compass variants of this tho

addaleax:
Also, I think the test failure in the `types` package might be related:

```
[2022/03/10 12:15:38.129]   1) config validation
[2022/03/10 12:15:38.129]        validates config option values:
[2022/03/10 12:15:38.129]      AssertionError: expected null to equal undefined
[2022/03/10 12:15:38.129]       at Context.<anonymous> (src/index.spec.ts:7:48)
```

mcasimir (approved):
👍

PR #1228, opened 2022-03-14
fix(shell-api): never pass schemaMap to ClientEncryption ctor 

addaleax:
MONGOSH-1158


Anemy (approved):
nice regression test add 👍

PR #1229, opened 2022-03-14
fix(logging): keep userId for old and new users MONGOSH-1143

alenakhineika:
We need to keep `userId` for old and new users as well for backward compatibility with the older `mongosh` versions. If a user installed a newer version first, with code that is currently in `main` only `anonymousId` will be saved to a new config file. If the user then downgrades the version and tries to run `mongosh`, they will get the `AssertionError: You must pass either an "anonymousId" or a "userId".` error, because the config file stores only `anonymousId`, but the older version of `mongosh` doesn't know what it is and needs `userId` for telemetry.


addaleax (approved):
Looks good! (with the changes to the package-lock.json files undone)

alenakhineika:
Yeah already on it, wrong node/npm locally.

PR #1238, opened 2022-03-17
fix: updates help message for --retryWrites option MONGOSH-1150

leorossi:


packages/cli-repl/README.md
```diff
@@ -21,7 +21,7 @@ CLI interface for [MongoDB Shell][mongosh], an extension to Node.js REPL with Mo
         --nodb                                 Don't connect to mongod on startup - no 'db address' [arg] expected
         --norc                                 Will not run the '.mongoshrc.js' file on start up
         --eval [arg]                           Evaluate javascript
-        --retryWrites                          Automatically retry write operations upon transient network errors
+        --retryWrites [true|fa;se]             Automatically retry write operations upon transient network errors (Default: true)
```
addaleax, line 24:
```suggestion
        --retryWrites [true|false]             Automatically retry write operations upon transient network errors (Default: true)
```

I would *slightly* prefer us using 

```suggestion
        --retryWrites=[true|false]             Automatically retry write operations upon transient network errors (Default: true)
```

here, even though yargs can parse both, because if we ever switch out the parser then I think we can typically guarantee that the former format works, but not necessarily the `=`-less one

addaleax (approved):

addaleax (approved):

PR #1241, opened 2022-03-18
feat: translate CLI flags to connection string parameters MONGOSH-1127

addaleax:
- Split out and group together code for generating connection
  information from CLI options
  - Move `CliOptions`, `generateUri()` from `service-provider-core`
    to `arg-parser` (`service-provider-core` should not contain
    CLI-specific code anyway)
  - Remove default-db setting code from `generateUri()`
  - Move `mapCliToDriver()` to `arg-parser`
- Refactor `mapCliToDriver()`:
  - Instead of using `lodash.set()`, use TypeScript-safe methods
    for modifying connection info
  - Store options that map to connection string options in the
    connection string
- Adjust consuming code in the cli-repl and in the Mongo
  constructor for these changes


gribnoysup (approved):
Does look good to me, I like that we are moving all this logic to one place, that makes a lot of sense. I will note that I definitely might've missed something while reviewing as I'm not extremely familiar with how driver options map to server arguments, but the fact that e2e tests . Seems like tests are failing due to `_apiOptions` being asserted in one of the tests in `shell-api`, but this prop is removed in this PR

PR #1242, opened 2022-03-21
Fix problem caused by GraalJS limitation on Proxy methods invocation

kornilova203:
I'm trying to update GraalJS library.
Currently tests with new GraalJS version don't pass because of two problems, this is one of them.
The problem existed before but it just didn't show up because the assertion in GraalJS used to be written incorrectly.

### DIFF
--- a/packages/shell-api/src/database.ts
+++ b/packages/shell-api/src/database.ts
@@ -90,6 +90,18 @@ export default class Database extends ShellApiWithMongoClass {
         }
 
         return collections[prop];
+      },
+      getOwnPropertyDescriptor: (target, prop): PropertyDescriptor | undefined => {
+        // Workaround for Java Shell.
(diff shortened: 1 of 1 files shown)


addaleax (commented):
A few things:

- If this needs to be done, then it probably needs to be done at the same time as the java-shell changes, because otherwise this patch seems like something that should be reverted
- The linked github ticket seems to be about accessing properties via `.get()`, not `.invokeMember()` – but either way, it seems like a bug that GraalJS is using `getOwnPropertyDescriptor()` instead of `get()`
- Would it be possible to use `context.eval("db.getName()")` rather than `currentDb.invokeMember("getName").asString()` in the java-shell code? That seems like it would work and avoid this bug altogether

kornilova203:
@addaleax, the last suggestion works!
I didn't think about it, thanks!
I'll create a new PR

kornilova203:
#1243

PR #1245, opened 2022-03-23
chore(deps): bump to bson@4.6.2

leorossi:


addaleax:
@leorossi This looks like it was made with npm@7.x?

leorossi:
@addaleax yes, what version should I use instead?

addaleax:
@leorossi We’re still on npm 6 in mongosh until we do MONGOSH-975, unfortunately :confused:

leorossi:
@addaleax ran again with npm 6

addaleax:
@leorossi This seems good, but I think you want to bump it for the `bson` libraries inside the individual packages as well, right?

addaleax (approved):

PR #1246, opened 2022-03-24
feat: detect CSFLE library path and pass it to driver MONGOSH-1118

addaleax:
- Add `--csfleLibrarySearchPath` and `csfleLibraryPath` command line
  flags to control CSFLE library search path behavior
  - This is necessary for us for testing, but also likely for users,
    because simply adding an entry to the `PATH` environment variable
    will not be enough to locate the library if it is in a non-standard
    location (unlike mongocryptd).
  - Use these options in our e2e/smoke tests
- Download the shared library when required for testing
- Add CSFLE path detection logic, which replaces our mongocryptd
  path detection logic
  - Pass the resulting AutoEncryptionExtraOptions down to the driver.

---

This fully working depends on upstream work in libmongocrypt and the Node.js driver. I’ve locally verified that this *does* work when those changes are in, up to the point of running into [SERVER-64859](https://jira.mongodb.org/browse/SERVER-64859) which is not something that we can easily work around.

No evergreen patch because it’s pointless, compiling TS fails without the Node.js driver typings for this.


packages/cli-repl/src/csfle-library-paths.ts
```diff
@@ -0,0 +1,92 @@
...
+}
+
+async function ensureMatchingPermissions(filename: string, execPathStat: { uid: number, gid: number }): Promise<null | object> {
+  await fs.access(filename, fsConstants.R_OK);
+  const stat = await fs.stat(filename);
+  // On UNIX systems, only load shared libraries if they are coming
+  // from a user we can consider trusted (current user or the one who owns
+  // the mongosh binary to begin with) and they are not writable by other
+  // users. On Windows, no such check is feasible.
+  if (process.platform !== 'win32' &&
```
mcasimir, line 81: NIT: just a silly thing, I got a bit puzzled by the fact that the function  "returns" `null`, in a good case and an object otherwise. Basically, I understood that we were reporting permission issues Infos only when I got to the usage.

Probably a comment that explains that we return details useful for debugging in case of a mismatch would have helped me.

  
 maybe you could early return here for `win32`

mcasimir (approved):

packages/cli-repl/src/csfle-library-paths.ts
addaleax: Yup, all of that makes sense, done! :+1:

PR #1249, opened 2022-03-30
chore(java-shell): minify the file produced by browserify

kornilova203:
I'm not sure if I'm supposed to commit `package-lock.json` too?


addaleax (changes requested):
It seems like you ran `npm install` here with npm@7.x or npm@8.x rather than npm@6.x. That’s going to fail in CI, unfortunately.

We’re planning to upgrade npm for this project fairly soon (https://jira.mongodb.org/browse/MONGOSH-975), but unfortunately, fairly soon is not yet today :slightly_smiling_face: 

> I'm not sure if I'm supposed to commit package-lock.json too?

To be clear: Yes, you should :slightly_smiling_face: Just needs to be the right package-lock version.

kornilova203:
@addaleax, is it okay if I just remove `package-lock.json` from the commit?
I want to avoid spending time now on figuring out node/npm combination and how to output old version of `package-lock.json`

addaleax:
@kornilova203 CI will need a package-lock.json that matches package.json, unfortunately. However, the good news is that I can push this change for you :slightly_smiling_face:

addaleax (approved):

PR #1250, opened 2022-03-30
feat(JMongosh): allow setting MongoClient after MongoShell was created

kornilova203:
This is needed to improve JDBC driver performance - we'll create MongoShell instances beforehand


packages/java-shell/settings.gradle
```diff
@@ -4,5 +4,5 @@ pluginManagement {
         maven { url 'https://plugins.gradle.org/m2/' }
     }
 }
-rootProject.name = 'java-shell'
+rootProject.name = 'JMongosh'
```
addaleax, line 7: Just a note, our PR titles refer to the individual directories inside the `packages/` folder, so if you want to start using `feat(jmongosh):`, you will also want to rename the directory (and then run `npm run update-evergreen-config` in the top-level directory of this repo)

addaleax (approved):

PR #1261, opened 2022-04-18
feat(shell-api): Mark `collection.reIndex()` deprecated MONGOSH-1182

Anemy:
MONGOSH-1182

This PR marks the collection `reIndex` function as deprecated.

## Open questions
Should we print a deprecation warning? 
```ts
await this._instanceState.printDeprecationWarning(
  'Collection.reIndex() is deprecated and not recommended for use.'
);
```

### DIFF
--- a/packages/shell-api/src/collection.ts
+++ b/packages/shell-api/src/collection.ts
@@ -1265,6 +1265,7 @@ export default class Collection extends ShellApiWithMongoClass {
    * @return {Promise}
    */
   @returnsPromise
+  @deprecated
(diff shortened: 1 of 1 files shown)


addaleax (approved):
> Should we print a deprecation warning?

I think it’s fine either way – this probably isn’t a very frequently-used method to begin with.

Anemy:
Sounds good, ci failures look like some flakes and notary fails so I'm merging - if I was wrong feel free to revert!

PR #1263, opened 2022-04-21
feat(java-shell): fix findAndModify with `new` option

kornilova203:
https://youtrack.jetbrains.com/issue/DBE-15415

### DIFF
--- a/packages/shell-api/src/helpers.ts
+++ b/packages/shell-api/src/helpers.ts
@@ -600,12 +600,14 @@ export function processFindAndModifyOptions<BaseOptions extends FindOneAndReplac
     options.returnDocument = options.returnNewDocument ? 'after' : 'before';
     delete options.returnOriginal;
     delete options.returnNewDocument;
+    delete options.new;
     return options;
   }
   if ('new' in options) {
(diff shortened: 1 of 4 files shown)


packages/shell-api/src/helpers.ts
```diff
@@ -605,7 +605,7 @@ export function processFindAndModifyOptions<BaseOptions extends FindOneAndReplac
   if ('new' in options) {
     options.returnDocument = options.new ? 'after' : 'before';
     delete options.returnOriginal;
-    delete options.returnNewDocument;
+    delete options.new;
```
addaleax, line 608: I think we’ll want to remove all three of `returnOriginal, returnNewDocument, new` in all of the conditionals here, right?

kornilova203: @addaleax, at that point there cannot be `returnNewDocument` in the options. But I can add it for clarity and readability. See my new commit

addaleax: Ah yeah, you’re right. Thank you either way! :slightly_smiling_face:

PR #1268, opened 2022-04-25
fix(shell-api): pass through collection.drop() options MONGOSH-1199

addaleax:
Passing through options is a) documented and b) something we also do for other methods.


Anemy (approved):
👍

PR #1271, opened 2022-04-29
numeral rather than pretty bytes MONGOSH-1124

lerouxb:
This treats 1KB as 1024 bytes, has no space between the number and unit and uses KB, MB, etc all just like legacy shell. BUT I don't know about all the nuances of 0.1GB vs 100MB, for example. How much do we care about matching that level of detail? Also unclear about things like rounding up and down.

### DIFF
--- a/packages/browser-repl/src/components/types/show-dbs-output.tsx
+++ b/packages/browser-repl/src/components/types/show-dbs-output.tsx
@@ -13,8 +13,8 @@ type DatabaseObject = {
 };
 
 function formatBytes(value: number): string {
-  const precision = value <= 1000 ? '0' : '0.0';
-  return numeral(value).format(precision + 'b');
+  const precision = value <= 1000 ? '0' : '0.00';
+  return numeral(value).format(precision + ' ib');
 }
 
 export class ShowDbsOutput extends Component<ShowDbsOutputProps> {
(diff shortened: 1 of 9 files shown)


packages/browser-repl/src/components/types/show-dbs-output.tsx
```diff
@@ -12,14 +12,19 @@ type DatabaseObject = {
   map: any;
 };
 
+function formatBytes(value: number): string {
+  const precision = value <= 1000 ? '0' : '0.0';
+  return numeral(value).format(precision + 'b');
```
addaleax, line 17:
```suggestion
  const precision = value <= 1000 ? '0' : '0.00';
  return numeral(value).format(precision + ' ib');
```

would make this closer to what we want/had previously and actually use 1024-based units, I think

packages/cli-repl/src/format-output.spec.ts
```diff
@@ -172,7 +178,7 @@ for (const colors of [ false, true ]) {
           type: 'ShowCollectionsResult'
         }));
 
-        expect(output).to.contain('nested_documents\ndecimal128\ncoll\npeople_imported   [view]\ncats              [time-series]');
+        expect(output).to.equal('nested_documents\ndecimal128\ncoll\npeople_imported   [view]\ncats              [time-series]');
```
addaleax, line 181: :+1:

addaleax (approved):

packages/cli-repl/src/format-output.spec.ts
lerouxb: accident. meant to change the ones I was changing. But if it works I suppose there's no harm ;)

addaleax:
nb, I force-pushed because we also use conventional-commits in this repo

PR #1277, opened 2022-05-09
feat: drop collection using encrypted fields from collection list MONGOSH-1208

alenakhineika:
If a collection was not listed in the Mongo() `encryptedFieldsMap` and no explicit `encryptedFields` option was passed to `collection.drop(options)`, we always look up the `encryptedFields` from `listCollections` and pass that instead.

### DIFF
--- a/packages/cli-repl/test/e2e-fle.spec.ts
+++ b/packages/cli-repl/test/e2e-fle.spec.ts
@@ -8,10 +8,9 @@ import { once } from 'events';
 import { serialize } from 'v8';
 import { inspect } from 'util';
 import path from 'path';
-import stripAnsi from 'strip-ansi';
 
 describe('FLE tests', () => {
-  const testServer = startTestServer('shared');
+  const testServer = startTestServer('not-shared', '--replicaset', '--nodes', '1');
   skipIfServerVersion(testServer, '< 4.2'); // FLE only available on 4.2+
   skipIfCommunityServer(testServer); // FLE is enterprise-only
   useBinaryPath(testServer); // Get mongocryptd in the PATH for this test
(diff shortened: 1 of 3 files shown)


packages/cli-repl/test/e2e-fle.spec.ts
```diff
@@ -237,6 +237,53 @@ describe('FLE tests', () => {
     expect(plainMongoResult).to.not.include("phoneNumber: '+12874627836445'");
   });
 
+  it('works when a encryptedFieldsMap option has been passed', async() => {
```
addaleax, line 240: tiny typo:
```suggestion
  it('works when an encryptedFieldsMap option has been passed', async() => {
```

packages/shell-api/src/collection.spec.ts
```diff
@@ -1873,6 +1874,72 @@ describe('Collection', () => {
...
+      it('passes encryptedFields through options when collection is not in encryptedFieldsMap', async() => {
+        serviceProvider.listCollections.resolves([{
+          options: { encryptedFields: { fields: [ { path: 'phoneNumber', keyId, bsonType: 'string' } ] } }
+        }]);
+        serviceProvider.dropCollection.resolves();
+        await mongo2.getDB('db1').getCollection('collfle2').drop();
+        expect(serviceProvider.dropCollection).to.have.been.calledWith(
+          'db1', 'collfle2', { encryptedFields: { fields: [ { path: 'phoneNumber', keyId, bsonType: 'string' } ] } }
+        );
+      });
```
addaleax, line 1940: Just wondering, do we want to have an integration or e2e test for this as well?

addaleax (approved):

packages/cli-repl/test/e2e-fle.spec.ts
```diff
@@ -237,6 +236,72 @@ describe('FLE tests', () => {
...
+
+      // Drivers will create the auxilliary FLE2 collections only when explicitly creating collections
+      // via the createCollection() command.
+      await shell.executeLine(`autoMongo.getDB('${dbname}').createCollection('collfle2');`);
+      await shell.executeLine(`autoMongo.getDB('${dbname}').collfle2.insertOne({ \
+        phoneNumber: '+12874627836445' \
+      });`);
+
+      const autoMongoResult = await shell.executeLine(`autoMongo.getDB('${dbname}').collfle2.find()`);
+      expect(autoMongoResult).to.include("phoneNumber: '+12874627836445'");
```
addaleax, line 284: Since this is currently the only test for FLE2 support, do you want to also verify that this is returned as encrypted when using `plainMongo`, similar to the test above?

addaleax (approved):

addaleax (approved):

PR #1285, opened 2022-05-12
feat: bump Node.js to 16.x MONGOSH-1219

addaleax:


alenakhineika (approved):
✨

packages/service-provider-core/package.json
```diff
@@ -24,7 +24,7 @@
     "access": "public"
   },
   "engines": {
-    "node": ">=12.4.0"
+    "node": ">=14.15.1"
```
Anemy, line 27: For my own understanding, is there a reason we aren't going to 16 everywhere? I see we're keeping some 14s in the evergreen config and skipping the node check. Changes look good.

addaleax: Yeah, 16.x+ is for the CLI itself and modules that cannot be used by the CLI. We’re keeping the rest on 14.x+ because we’re supporting Compass as well, which runs Node.js 14. :)

Anemy (approved):
🔥

PR #1289, opened 2022-05-17
feat: add support for `show freeMonitoring` and friends MONGOSH-1185

addaleax:
Add support for the following and run them at startup:
- `show startupWarnings`
- `show freeMonitoring`
- `show automationNotices`

This matches the behavior of these calls in the legacy shell.

The `startupWarnings` code was extracted from its current existing
location in the `cli-repl` package, with the log entry parsing
helper being moved as well.


Anemy (commented):
Will try out a bit tomorrow, looks good, nice tests too. Two smaller questions

packages/browser-repl/src/components/shell-output-line.tsx
```diff
@@ -67,14 +68,18 @@ export class ShellOutputLine extends Component<ShellOutputLineProps> {
       return <StatsResultOutput value={value} />;
     }
 
-    if (type === 'ListCommandsResult)') {
+    if (type === 'ListCommandsResult') {
```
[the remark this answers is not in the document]
addaleax: Yes. I can’t see a visual difference, though. :)

packages/cli-repl/src/format-output.ts
```diff
@@ -147,6 +151,22 @@ function formatCollections(output: CollectionNamesWithTypes[], options: FormatOp
   return textTable(tableEntries, { align: ['l', 'l'] });
 }
 
+function formatBanner(output: null | { header?: string, content: string }, options: FormatOptions): string {
```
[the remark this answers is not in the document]
addaleax: Mh yeah, good point. There’s tests for the output being present in this PR, but not the exact format of it. Added one :+1:

packages/shell-api/src/helpers.ts
```diff
@@ -794,3 +794,15 @@ export function getBadge(collections: Document[], index: number): string {
...
+export const FREE_MONITORING_BANNER = `\
+Enable MongoDB's free cloud-based monitoring service, which will then receive and display
+metrics about your deployment (disk utilization, CPU, operation statistics, etc).
+
+The monitoring data will be available on a MongoDB website with a unique URL accessible to you
+and anyone you share the URL with. MongoDB may use this information to make product
+improvements and to suggest MongoDB products and deployment options to you.
+
+To enable free monitoring, run the following command: db.enableFreeMonitoring()
+To permanently disable this reminder, run the following command: db.disableFreeMonitoring()
```
Anemy, line 807: Can't think of anything to improve, but this might be a good time to check with Max or product if we want to update/improve this message https://github.com/mongodb/mongo/blob/2cba27684ff70e38b79bc01d7b570b0af6f4a068/src/mongo/shell/utils.js#L1153-L1167
Nothing to change, something I wanted to mention in case anyone thinks of ways to improve it. (Feel free to resolve).

testing/integration-testing-hooks.ts
```diff
@@ -507,6 +507,21 @@ export function skipIfCommunityServer(server: MongodSetup): void {
...
+/**
+ * Skip tests in the suite if the test server is an enterprise server.
+ *
+ * describe('...', () => {
+ *   e.g. skipIfEnterpriseServer(testServer)
+ * });
+ */
+ export function skipIfEnterpriseServer(server: MongodSetup): void {
+  before(async function() {
+    if (!await server.isCommunityServer()) {
```
Anemy, line 519: Definitely a small not really doing anything suggestion, so feel free to ignore. Maybe we could add a helper to be more explicit about this `community` vs `enterprise` check on the [mongod setup class](https://github.com/mongodb-js/mongosh/blob/011ee10a0d505b7ec2e9af63c16f365e2e2c1373/testing/integration-testing-hooks.ts#L210). I'm thinking way down the line there could be something besides community that isn't enterprise and having it more explicitly set there would make that check more transparently updated.
```suggestion
    if (await server.isEnterpriseServer()) {
```

Anemy (approved):
Nice, left a couple small comments, fine to merge as is, the comments are just some light questions/thoughts.

packages/shell-api/src/helpers.ts
addaleax: Yeah, pinged Max about this, he said we’ll leave it like this for now and can get more input from product on this later

packages/browser-repl/src/components/types/show-banner-result-output.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+export class ShowBannerResultOutput extends Component<ShowBannerResultOutputProps> {
+  static propTypes = {
+    value: PropTypes.any
+  };
+
+  render(): JSX.Element {
+    return (<>
+      {this.props.value?.header && <h3>{this.props.value.header}</h3>}
+      {this.props.value?.content && <pre>{this.props.value.content}</pre>}
```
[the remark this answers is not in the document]
addaleax: Sure – done!

testing/integration-testing-hooks.ts
addaleax: We could do that, but I think if/when that happens we’ll need to revisit all our `skipIfEnterpriseServer()`/`skipIfCommunityServer()` calls anyway, so I think it’s okay to just wait rather than over-planning for the future?

Anemy: sgtm!

addaleax: Fwiw, I had to remove this function anyway because this check requires the server to be running, but the `--setParameter` flags in the test prevent it from starting in the first place if it’s an enterprise server.

PR #1291, opened 2022-05-20
feat(ci): create shared-openssl binaries MONGOSH-1223

addaleax:


.evergreen/compile-artifact.sh
```diff
@@ -13,9 +14,52 @@ if uname -a | grep -q 'Linux.*x86_64'; then
...
+
+  popd # openssl-*
+  popd # /tmp/m
+
+  export NODE_JS_CONFIGURE_ARGS='[
+    "--shared-openssl",
+    "--shared-openssl-includes=/tmp/m/opt/include",
+    "--shared-openssl-libpath=/tmp/m/opt/lib",
+    "--shared-openssl-libname='"$MONGOSH_OPENSSL_LIBNAME"'",
+    "--shared-zlib"
```
addaleax, line 56: I don’t know why specifically this is necessary, but it’s definitely related to the Node.js build system being a bit wonky around zlib support (because Node.js itself doesn’t bundle the original zlib, and instead uses an API-compatible fork from Chromium). I guess the --shared-openssl setting has some influence on the zlib build. In any case, this should be fairly unproblematic since zlib is almost universally available on Linux.

mcasimir (approved):
awesome!

PR #1292, opened 2022-05-20
feat(cli-repl): add support for the --tlsFIPSMode flag MONGOSH-1221

addaleax:
Proper testing of the functionality is left to MONGOSH-1222.

This does not add support for `--sslFIPSMode` or spelling variants
of `--tlsFIPSMode` that would be recognized by yargs, in order to
keep the setup code as straightforward as possible.

### DIFF
--- a/packages/cli-repl/test/e2e.spec.ts
+++ b/packages/cli-repl/test/e2e.spec.ts
@@ -156,6 +156,20 @@ describe('e2e', function() {
         shell.assertContainsOutput('233');
       });
     });
+    it('accepts a --tlsFIPSMode argument', async() => {
+      shell = TestShell.start({
+        args: [ '--nodb', '--tlsFIPSMode' ]
+      });
+      const result = await shell.waitForPromptOrExit();
+      // Whether this worked depends on the environment the test is running in.
+      // We check both possibilities.
+      if (result.state === 'exit') {
+        shell.assertContainsOutput('Could not enable FIPS mode');
+        expect(result.exitCode).to.equal(1);
+      } else {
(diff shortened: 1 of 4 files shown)


packages/cli-repl/test/e2e.spec.ts
```diff
@@ -156,6 +156,20 @@ describe('e2e', function() {
         shell.assertContainsOutput('233');
       });
     });
+    it('accepts a --tlsFIPSMode argument', async() => {
+      shell = TestShell.start({
+        args: [ '--nodb', '--tlsFIPSMode' ]
+      });
+      const result = await shell.waitForPromptOrExit();
+      // Whether this worked depends on the environment the test is running in.
```
Anemy, line 164: Would it help to test here that the expected result is one which is expected for the environment? It might be extra coupling with the environment, is there a way we can do it without reproducing all of that?

Anemy (approved):
lgtm, one question on possible test improvement, not a blocker.

packages/cli-repl/test/e2e.spec.ts
addaleax: Well how do we know which result is the expected one? :) We could try to check it by enabling FIPS in the test as well and seeing what happens, but that feels like it would just be redundant compared to the code itself. (We could also check that this doesn’t work on macOS or Windows, which are systems where it is _very_ unlikely that we’re running this on a FIPS-enabled Node.js version, but even there that’s not really guaranteed)

[MONGOSH-1222](https://jira.mongodb.org/projects/MONGOSH/issues/MONGOSH-1222) will have an actual e2e test for FIPS mode, including makign sure that in those environments, we won’t get an error with `--tlsFIPSMode`.

PR #1294, opened 2022-05-24
chore(ci,build): perform dry run of publishing in draft task MONGOSH-1227

addaleax:


packages/build/src/homebrew/generate-formula.ts
```diff
@@ -24,7 +24,7 @@ export async function generateUpdatedFormula(
   }
 
   const currentVersion = /cli-repl-(\d+\.\d+\.\d+)\.tgz/.exec(currentUrl)?.[1];
-  if (currentVersion && semver.compare(currentVersion, context.version, { includePrerelease: true }) !== -1) {
+  if (currentVersion && semver.compare(currentVersion, context.version, { includePrerelease: true }) !== -1 && !context.sha.includes('dryRun')) {
```
mcasimir, line 27: NIT: I like how we added `dryRun` for `publishNpmPackages`, making these `dryRun` booleans stand out consistently on interfaces without a default. It will remind us that we have deal with the `dryRun` flow.

Would it be possible to add an explicit `{dryRun}` argument also to `generateUpdatedFormula`?

mcasimir (approved):

PR #1297, opened 2022-05-30
chore: replace references to csfle shared library MONGOSH-1237

addaleax:


packages/cli-repl/src/arg-parser.ts
```diff
@@ -24,6 +24,7 @@ const OPTIONS = {
     'awsSessionToken',
     'awsIamSessionToken',
     'csfleLibraryPath',
+    'cryptLibraryPath',
```
addaleax, line 27: I think I’ll need to wait for confirmation on the new driver option name (if it is being changed) here

Anemy (approved):
lgtm! https://github.com/mongodb-js/devtools-connect/pull/14 looks good too.

PR #1299, opened 2022-06-01
feat(build): add shared-openssl packages to releases MONGOSH-1231

addaleax:
- Move the package matrix definition from the evergreen template
  to a separate file in config/, and add data that would otherwise
  be computed in the build package (e.g. human-readable descriptions)
  there.
- Consistently use “package variant” instead of
  “distribution build variant”. The latter term is a leftover
  from when package artifacts were tightly coupled with evergreen
  build variants, but now these are completely independent
  and it’s probably better to be consistent in referring to
  these as “packages”.
- Make the `packageInformation` property in the build config
  a function that returns the package information for a given
  package variant – this is necessary so that the tasks that
  operate on all packaging variants at once (i.e. the publishing
  tasks) understand that the package information actually varies
  by packaging variant in the case of static OpenSSL/
  shared OpenSSL 1.1/shared OpenSSL 3.
- Merge all .rpm and all .deb package variants into a single
  entity each. Since we started using the crypt_shared library
  (formerly known as CSFLE shared library), our artifacts no longer
  differ between Linux distributions.
  This simplifies picking the correct package from a user perspective
  and reduces the publishing step complexity (i.e. counters the
  increase in the number of package variants coming from the
  shared-openssl additions).
- Adjust build package tests to account for all of the above.

### DIFF
--- a/.evergreen.yml
+++ b/.evergreen.yml
@@ -4926,11 +4926,11 @@ functions:
   #
   # package_and_upload_artifact expects the following arguments:
   # - distro_id
-  # - distribution_build_variant
+  # - package_variant
   # - executable_os_id
   #
   # get_artifact_url expects the following arguments:
-  # - distribution_build_variant
+  # - package_variant
   ###
   package_and_upload_artifact:
     - command: expansions.write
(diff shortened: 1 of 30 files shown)


mcasimir (approved):
Nice! Love the renaming from `build_variant` to `package_variant`

PR #1300, opened 2022-06-01
feat(shell-api): add new key management helpers MONGOSH-1206

addaleax:
This also adds the aliases suggested in MONGOSH-1239, since
doing so is straightforward and touches the same code.


Anemy (approved):
lgtm! Let's be sure to open this for docs so the https://docs.mongodb.com/manual/reference/method/KeyVault.rewrapManyDataKey/#KeyVault.rewrapManyDataKey link resolves 👍

PR #1305, opened 2022-06-02
chore(service-provider-server): clean up after resolved upstream tickets

addaleax:


packages/service-provider-server/src/cli-service-provider.ts
```diff
@@ -258,27 +229,9 @@ class CliServiceProvider extends ServiceProviderCore implements ServiceProvider
...
 
   async getConnectionInfo(): Promise<ConnectionInfo> {
-    // buildInfo try/catch can be removed after MONGOCRYPT-308
-    let buildInfo;
-    try {
-      buildInfo = await this.runCommandWithCheck('admin', {
-        buildInfo: 1
-      }, this.baseCmdOptions);
-    } catch (e: unknown) {
-      if ((e as Error)?.message.includes('not supported for auto encryption')) {
```
Anemy, line 268: Should we have been re-throwing this error when it didn't match before?

Anemy (approved):
🔥

packages/service-provider-server/src/cli-service-provider.ts
addaleax: I think so, yes, otherwise this would have ended up breaking the calls inside `getConnectInfo()`. I think this is a good way to resolve this tho :)

addaleax: Ah wait, this is failing CI because now `buildInfo` does fail in API-strict mode. I’ll take a look in more detail, but we may just want to add it to the other calls in the `Promise.all()` below and handle it the same way

PR #1306, opened 2022-06-09
chore: bump driver to 4.7.0 MONGOSH-1173

lerouxb:
I just bumped the driver to 4.7.0 and hacked at the typescript errors, so this is probably not mergeable.

This was just so I can check that the new change stream features work:

```
db.createCollection('changes', { changeStreamPreAndPostImages: { enabled: true }} );

while (!watchCursor.isClosed()) { let next = watchCursor.tryNext(); while (next !== null) { printjson(next); next = watchCursor.tryNext(); } }
```

then in another shell:
```
db.changes.insert({})
{
  acknowledged: true,
  insertedIds: { '0': ObjectId("62a1cfaaa1fe03bb51bf2649") }
}

db.changes.updateOne({_id:  ObjectId("62a1cfaaa1fe03bb51bf2649")}, {$set: { foo: '1'}})
{
  acknowledged: true,
  insertedId: null,
  matchedCount: 1,
  modifiedCount: 1,
  upsertedCount: 0
}
```

then the first one prints:

```
{
  _id: {
    _data: '8262A1CFAA000000012B022C0100296E5A10048BFF2FA8A3684E41A99B4C1CB6E05E5446645F6964006462A1CFAAA1FE03BB51BF26490004'
  },
  operationType: 'insert',
  clusterTime: Timestamp({ t: 1654771626, i: 1 }),
  collectionUUID: UUID("8bff2fa8-a368-4e41-a99b-4c1cb6e05e54"),
  wallTime: ISODate("2022-06-09T10:47:06.053Z"),
  fullDocument: { _id: ObjectId("62a1cfaaa1fe03bb51bf2649") },
  ns: { db: 'test', coll: 'changes' },
  documentKey: { _id: ObjectId("62a1cfaaa1fe03bb51bf2649") }
}
{
  _id: {
    _data: '8262A1D01D000000012B022C0100296E5A10048BFF2FA8A3684E41A99B4C1CB6E05E5446645F6964006462A1CFAAA1FE03BB51BF26490004'
  },
  operationType: 'update',
  clusterTime: Timestamp({ t: 1654771741, i: 1 }),
  collectionUUID: UUID("8bff2fa8-a368-4e41-a99b-4c1cb6e05e54"),
  wallTime: ISODate("2022-06-09T10:49:01.944Z"),
  ns: { db: 'test', coll: 'changes' },
  documentKey: { _id: ObjectId("62a1cfaaa1fe03bb51bf2649") },
  updateDescription: {
    updatedFields: { foo: '1' },
    removedFields: [],
    truncatedArrays: []
  },
  fullDocumentBeforeChange: { _id: ObjectId("62a1cfaaa1fe03bb51bf2649") }
}
```

Seems to be working.

Should I be making any changes to test this? I can't find any specific / explicitly stated options in mongosh around change feed / watch options here - it all seems to just be passed to the driver which is doing all the work required.

So maybe this PR is not needed other than that someone should bump the driver at some stage.

### DIFF
--- a/packages/arg-parser/src/arg-mapper.ts
+++ b/packages/arg-parser/src/arg-mapper.ts
@@ -105,9 +105,7 @@ const MAPPINGS: {
   awsSecretAccessKey: (i, v) => setAWSKMS(i, 'secretAccessKey', v),
   awsSessionToken: (i, v) => setAWSKMS(i, 'sessionToken', v),
   awsIamSessionToken: (i, v) => setAuthMechProp(i, 'AWS_SESSION_TOKEN', v),
-  // @ts-expect-error AutoEncryption options have been renamed, drop the (outer) csflePath call after the next release
   csfleLibraryPath: (i, v) => setAutoEncryptExtra(i, 'cryptSharedLibPath', v),
-  // @ts-expect-error AutoEncryption options have been renamed, drop the (outer) csflePath call after the next release
   cryptSharedLibPath: (i, v) => setAutoEncryptExtra(i, 'cryptSharedLibPath', v),
   gssapiServiceName: (i, v) => setAuthMechProp(i, 'SERVICE_NAME', v),
   sspiRealmOverride: (i, v) => setAuthMechProp(i, 'SERVICE_REALM', v),
--- a/packages/shell-api/src/session.ts
+++ b/packages/shell-api/src/session.ts
@@ -106,6 +106,8 @@ export default class Session extends ShellApiWithMongoClass {
     assertArgsDefinedType([fn, options], ['function', [undefined, 'object']]);
     // The driver doesn't automatically ensure that fn is an async
     // function/convert its return type to a Promise, so we do that here.
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-expect-error
     return await this._session.withTransaction(async() => await fn(), options);
   }
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -53,7 +53,7 @@
         "karma-typescript": "^5.5.1",
         "lerna": "^4.0.0",
         "mocha": "^7.1.2",
-        "mongodb": "^4.6.0",
+        "mongodb": "^4.7.0",
         "mongodb-download-url": "^1.2.2",
         "mongodb-js-precommit": "^2.0.0",
         "nock": "^13.0.11",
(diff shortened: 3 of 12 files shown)


packages/arg-parser/src/arg-mapper.ts
```diff
@@ -105,9 +105,7 @@ const MAPPINGS: {
   awsSecretAccessKey: (i, v) => setAWSKMS(i, 'secretAccessKey', v),
   awsSessionToken: (i, v) => setAWSKMS(i, 'sessionToken', v),
   awsIamSessionToken: (i, v) => setAuthMechProp(i, 'AWS_SESSION_TOKEN', v),
-  // @ts-expect-error AutoEncryption options have been renamed, drop the (outer) csflePath call after the next release
```
lerouxb, line 108: Don't know what these comments meant (drop what where?), but I removed the comments so it will compile.

packages/shell-api/src/session.ts
```diff
@@ -106,6 +106,8 @@ export default class Session extends ShellApiWithMongoClass {
     assertArgsDefinedType([fn, options], ['function', [undefined, 'object']]);
     // The driver doesn't automatically ensure that fn is an async
     // function/convert its return type to a Promise, so we do that here.
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-expect-error
```
lerouxb, line 110: Something's up here. I just hacked it to build for now:

```
src/session.ts(109,5): error TS2322: Type 'Document | undefined' is not assignable to type 'ReturnType<T>'.
  Type 'undefined' is not assignable to type 'ReturnType<T>'.
```

packages/arg-parser/src/arg-mapper.ts
addaleax: The things that should be dropped have already been removed, I forgot to update the comment – all good here :)

packages/shell-api/src/session.ts
addaleax: https://jira.mongodb.org/browse/NODE-4108 – they changed the return type to be more accurate. We should do the same and update the return type of our wrapper to `Document | undefined` as well.

lerouxb: OK I updated it now. The rest of the body of that function should remain unchanged, right?

addaleax: Yes, that is also my understanding :)

addaleax (approved):

PR #1307, opened 2022-06-09
chore(cli-repl): add explicit QE e2e test

addaleax:
Was writing a (very similar) script as a bug repro anyway,
so I figured I might as well add it as an e2e test. :)


alenakhineika (approved):
✨

PR #1331, opened 2022-08-10
chore(deps): bump boxednode to 1.12.0 MONGOSH-1258

addaleax:
Opening as draft for now because it should be possible to update the docker images a bit to account for this, but testing that locally requires new .rpm/.deb packages and the easiest way to get those is to open a draft PR and let CI do its thing. :slightly_smiling_face:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,7 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "mongosh",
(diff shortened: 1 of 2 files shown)


addaleax:
Okay, not making changes to the docker containers since that also requires https://github.com/nodejs/node/pull/42799, expected to be included in the next Node.js 16.x release.

PR #1333, opened 2022-08-12
chore(shell-api): make single-doc KeyVault methods return Documents MONGOSH-1266

addaleax:
Return a Proxy that merges the single returned `Document` from a cursor
with the cursor itself (or a copy of it) to migrate users to a more
driver-key-management-like API.


alenakhineika (approved):
👍

PR #1334, opened 2022-08-19
feat(shell-api): added a new shell helper 'getShardedDataInformation' MONGOSH-1254

PolPinol:
Add a shell helper for sharded data distribution info api.

Part of the project [PM-2934](https://jira.mongodb.org/browse/PM-2934).

** This PR wil not be merged until [SERVER-67891](https://jira.mongodb.org/browse/SERVER-67891) is committed.

### DIFF
--- a/packages/shell-api/src/shard.ts
+++ b/packages/shell-api/src/shard.ts
@@ -434,7 +434,7 @@ export default class Shard extends ShellApiWithMongoClass {
   async getShardedDataDistribution(options = {}): Promise<AggregationCursor> {
     this._emitShardApiCall('getShardedDataDistribution', {});
 
-    const cursor = await this._database.aggregate([{ $shardedDataDistribution: options }]);
+    const cursor = await this._database.getSiblingDB('admin').aggregate([{ $shardedDataDistribution: options }]);
 
     try {
       await cursor.hasNext();
--- a/packages/shell-api/src/shard.ts
+++ b/packages/shell-api/src/shard.ts
@@ -431,10 +431,10 @@ export default class Shard extends ShellApiWithMongoClass {
   @apiVersions([])
   @serverVersions(['6.1.0', ServerVersions.latest])
   @returnType('AggregationCursor')
-  async getShardedDataDistribution(): Promise<AggregationCursor> {
+  async getShardedDataDistribution(options = {}): Promise<AggregationCursor> {
     this._emitShardApiCall('getShardedDataDistribution', {});
 
-    const cursor = await this._database.aggregate([{ $shardedDataDistribution: {} }]);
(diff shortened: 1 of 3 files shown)


packages/shell-api/src/shard.ts
```diff
@@ -424,4 +426,29 @@ export default class Shard extends ShellApiWithMongoClass {
...
   }
+
+  @returnsPromise
+  @apiVersions([])
+  @serverVersions(['6.1.0', ServerVersions.latest])
+  @returnType('AggregationCursor')
+  async getShardedDataDistribution(): Promise<AggregationCursor> {
+    this._emitShardApiCall('getShardedDataDistribution', {});
+
+    const cursor = await this._database.aggregate([{ $shardedDataDistribution: {} }]);
```
addaleax, line 437: I assume it doesn’t matter which specific database the aggregation is being run against?

addaleax (approved):
looks great, thank you!

packages/shell-api/src/shard.ts
addaleax, line 437: Actually, looking at this again I have one idea:

```suggestion
  async getShardedDataDistribution(options = {}): Promise<AggregationCursor> {
    this._emitShardApiCall('getShardedDataDistribution', {});

    const cursor = await this._database.aggregate([{ $shardedDataDistribution: options }]);
```

This way, if `$shardedDataDistrubition` ever starts taking arguments, all mongosh versions that provide this helper can pass them along. How does that sound?

PolPinol: Sounds great! Thanks

PolPinol: With new approach, it must be run in the admin database!

packages/shell-api/src/shard.ts
```diff
@@ -424,4 +426,29 @@ export default class Shard extends ShellApiWithMongoClass {
...
   }
+
+  @returnsPromise
+  @apiVersions([])
+  @serverVersions(['6.1.0', ServerVersions.latest])
+  @returnType('AggregationCursor')
+  async getShardedDataDistribution(options = {}): Promise<AggregationCursor> {
+    this._emitShardApiCall('getShardedDataDistribution', {});
+
+    const cursor = await this._database.aggregate([{ $shardedDataDistribution: options }]);
```
addaleax, line 437: Let’s make this 
```suggestion
    const cursor = await this._database.getSiblingDB('admin').aggregate([{ $shardedDataDistribution: options }]);
```

if it’s important to run it against the admin database (that matches the behavior of the methods above which use `_runAdminCommand` instead of `_runCommand ` to transparently select the right db)

PolPinol: Nice 😄

addaleax (approved):

PR #1340, opened 2022-08-30
chore(ci): use MacOS 11.00 in unit tests MONGOSH-1058

addaleax:
This distro does not exhibit the Python problem that macos-1014 has.
Moving over the unit tests seems like a reasonable workaround at this
point, given that all other attempts at solving this issue have failed,
and we still keep end-to-end test coverage on macos-1014 around.


addaleax:
No objection in standup, merging

PR #1342, opened 2022-09-05
feat(cli-repl): add --json CLI flag MONGOSH-1249

addaleax:
Add a `--json` CLI flag with possible `--json=canonical` and
`--json=relaxed` flags, the default being `canonical`.
When used, mongosh will print Extended JSON as output. This
flag can only be used in conjunction with `--eval`, and is
mutually exclusive with `--shell` and providing files on
the command line.

Errors and regular output are not distinguished based on
the output alone, only in the exit code of mongosh.
Repeated errors while trying to serialize the output to
Extended JSON will not be serialized.

### DIFF
--- a/packages/cli-repl/src/cli-repl.ts
+++ b/packages/cli-repl/src/cli-repl.ts
@@ -19,6 +19,7 @@ import type { StyleDefinition } from './clr';
 import { ConfigManager, ShellHomeDirectory, ShellHomePaths } from './config-directory';
 import { CliReplErrors } from './error-codes';
 import type { CryptLibraryPathResult } from './crypt-library-paths';
+import { formatForJSONOutput } from './format-json';
 import { MongoLogManager, MongoLogWriter, mongoLogId } from 'mongodb-log-writer';
 import MongoshNodeRepl, { MongoshNodeReplOptions, MongoshIOProvider } from './mongosh-repl';
 import { setupLoggerAndTelemetry, ToggleableAnalytics } from '@mongosh/logging';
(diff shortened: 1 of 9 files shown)


packages/cli-repl/src/cli-repl.ts
```diff
@@ -309,7 +314,11 @@ class CliRepl implements MongoshIOProvider {
     if (willExecuteCommandLineScripts) {
       this.mongoshRepl.setIsInteractive(willEnterInteractiveMode);
       this.bus.emit('mongosh:start-loading-cli-scripts', { usesShellOption: !!this.cliOptions.shell });
-      await this.loadCommandLineFilesAndEval(commandLineLoadFiles, evalScripts);
+      const exitCode = await this.loadCommandLineFilesAndEval(commandLineLoadFiles, evalScripts);
+      if (exitCode !== 0) {
+        await this.exit(exitCode);
```
Anemy, line 319: I'm not too familiar with this code, would exiting earlier here have any side effects that didn't happen before?

Anemy (approved):
lgtm! One question to help me get a bit of context if that's alright.

packages/cli-repl/src/cli-repl.ts
addaleax: So … in practice, `this.exit()` translates into doing some cleanup (e.g. flushing logs and telemetry), then `process.exit()`, so it should not really have any side effects beyond that. But maybe I’m not quite understanding the question here?

Anemy: All good, gave it a bit more of a read, I misunderstood.

PR #1350, opened 2022-09-23
feat: add db.collection.analyze() helper MONGOSH-1091

alenakhineika:
Add a `db.collection.analyze()` shell helper [not implemented by the server yet].

### DIFF
--- a/packages/i18n/src/locales/en_US.ts
+++ b/packages/i18n/src/locales/en_US.ts
@@ -683,6 +683,11 @@ const translations: Catalog = {
               link: 'https://docs.mongodb.com/manual/reference/method/db.collection.getShardDistribution',
               description: 'Prints the data distribution statistics for a sharded collection.',
               example: 'db.coll.getShardDistribution()'
+            },
+            analyze: {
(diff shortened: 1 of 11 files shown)


packages/i18n/src/locales/en_US.ts
```diff
@@ -683,6 +683,11 @@ const translations: Catalog = {
               link: 'https://docs.mongodb.com/manual/reference/method/db.collection.getShardDistribution',
               description: 'Prints the data distribution statistics for a sharded collection.',
               example: 'db.coll.getShardDistribution()'
+            },
+            analyze: {
+              link: '',
```
addaleax, line 688: I think we've added links here in the past, even if they 404'd, knowing that they would all follow the same pattern for collection methods.

addaleax (changes requested):
The code looks good! However, as you pointed out, there's a notable discrepancy in the server's planning docs here (scope doc says mongosh should add a helper, design doc says it's internal), so I'd hold off on merging until we've followed up on what exactly to do here.

PR #1364, opened 2022-11-29
feat(shell-api): add convertShardKeyToHashed() MONGOSH-1050

addaleax:


gribnoysup (approved):
Seems like the tests are failing due to missing translations, but otherwise looks good to me

PR #1366, opened 2022-12-01
fix(cli-repl): drop nanobus in production usage MONGOSH-1330

addaleax:
nanobus inherently leaks memory, which means that it can make mongosh crash in certain scenarios.

Since we do not actually use the features of nanobus (mostly being able to listen for a generic `'*'` event), just replace it with the standard `EventEmitter` to fix this.


Anemy (approved):

Anemy (approved):
Nice find!

PR #1369, opened 2022-12-05
fix(cli-repl): flush output before exiting MONGOSH-1300

addaleax:


packages/cli-repl/src/cli-repl.ts
```diff
@@ -680,6 +680,20 @@ class CliRepl implements MongoshIOProvider {
...
+      if (this.output.writableEnded) {
+        // .end() has been called but not finished; 'close' will be emitted in that case.
+        // (This should not typically happen in the context of mongosh, but there's also
+        // no reason not to handle this case properly.)
+        try {
+          await once(this.output, 'close');
+        } catch { /* ignore */ }
+      } else {
+        // .end() has not been called; write an empty chunk and wait for it to be fully written.
+        await new Promise(resolve => this.output.write('', resolve));
```
alenakhineika, line 694: Why do we write this empty chunk? To verify it is still writable?

alenakhineika (approved):

packages/cli-repl/src/cli-repl.ts
addaleax: As the comment says, the goal here is to wait for the chunk to be fully written; because when the new chunk has been written, that implies that all previously pending data has also been written to the underlying stream.

PR #1376, opened 2022-12-12
chore(shell-api): update error message for getShardedDataDistribution MONGOSH-1301

Anemy:
MONGOSH-1301

Two questions for reviewers:
- Are there specific cases for when we show `MongoshRuntimeError`? I wasn't sure if that should be used here, in these changes we are going back to showing `MongoServerError`.
- Should we make this do a full check for the server version before showing the server version part of the error message? We can do something like https://github.com/mongodb-js/mongosh/blob/bed1edfe73b4b374d4ba6089d8c48d01fa56f93f/packages/shell-api/src/database.ts#L620 in that case.

```ts
async _improveErrorMessageForGetShardedDataDistribution(err: Error): Promise<void> {
  try {
    const serverVersion = await this.version();
    if (serverVersion && +serverVersion.split('.')[0] < 6) {
      err.message = `Your server version is ${serverVersion}, which does not support the sh.getShardedDataDistribution command [Original Error: ${err.message}]`;
      return;
    }
  } catch { /* ignore error from fetching server version */ }

  err.message = `sh.getShardedDataDistribution only works on mongos [Original Error: ${err.message}]`;
}
```

Screen Shot 2022-12-12 at 12 45 01 PM

### DIFF
--- a/packages/shell-api/src/shard.spec.ts
+++ b/packages/shell-api/src/shard.spec.ts
@@ -1194,9 +1194,9 @@ describe('Shard', () => {
       it('throws if not mongos', async() => {
         const serviceProviderCursor = stubInterface<ServiceProviderAggCursor>();
         serviceProvider.aggregateDb.returns(serviceProviderCursor as any);
-        serviceProviderCursor.hasNext.throws(Object.assign(new Error(), { code: 40324 }));
+        serviceProviderCursor.hasNext.throws(Object.assign(new Error(), { code: 40324, message: "Unrecognized pipeline stage name: '$shardedDataDistribution'" }));
         const error: any = await shard.getShardedDataDistribution().catch(err => err);
-        expect(error.message).to.match(/sh\.getShardedDataDistribution only works on mongos/);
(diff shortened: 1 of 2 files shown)


addaleax (approved):
> Are there specific cases for when we show `MongoshRuntimeError`? I wasn't sure if that should be used here, in these changes we are going back to showing `MongoServerError`.

I think that’s fine, it’s ultimately a server-side error and it should be okay to report it as that.

> Should we make this do a full check for the server version before showing the server version part of the error message?

Feel free to add that if you want, I wouldn’t mind but I also don’t think we need to overthink this too much?

Anemy:
> Feel free to add that if you want, I wouldn’t mind but I also don’t think we need to overthink this too much?

👍 going to merge this then as is. Thanks!

Wernfried:
In the last status mail I got, it states:

```
Change By:	Rhys Howell 

Component/s:	Shell API 
Documentation Changes:	**Not Needed** 
Fix Version/s:	1.6.2 
Resolution:	Fixed 
Status:	In Code Review Closed
```

Please note, the function is not documented at https://www.mongodb.com/docs/manual/reference/method/js-sharding/ , so a documentation change might be needed.
I know about this function only from output of ```sh.help()```

Kind Regards
Wernfried

PR #1379, opened 2023-01-05
fix(shell-api): don't pass options into cursor.close() MONGOSH-1356

baileympearson:
Here's a patch from the driver running against the changes in this branch: 

https://spruce.mongodb.com/version/63b7157961837d672d740cca/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

### DIFF
--- a/packages/shell-api/src/abstract-cursor.ts
+++ b/packages/shell-api/src/abstract-cursor.ts
@@ -53,8 +53,8 @@ export abstract class AbstractCursor<CursorType extends ServiceProviderAggregati
   }
 
   @returnsPromise
-  async close(options: Document = {}): Promise<void> {
-    await this._cursor.close(options);
+  async close(): Promise<void> {
+    await this._cursor.close();
   }
 
   @returnsPromise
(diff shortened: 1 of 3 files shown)


packages/shell-api/src/abstract-cursor.ts
```diff
@@ -53,8 +53,8 @@ export abstract class AbstractCursor<CursorType extends ServiceProviderAggregati
   }
 
   @returnsPromise
-  async close(options: Document = {}): Promise<void> {
-    await this._cursor.close(options);
+  async close(): Promise<void> {
```
baileympearson, line 56: I wasn't sure if this was okay (I'm not sure if this method gets exposed).  If we need to preserve the method signature, I can skip the "no unused vars" lint rule for this line.

addaleax: Mh yeah … I guess this would be an issue only if we ever would decide to reintroduce an argument for this function (with possibly different semantics), but that seems very unlikely? I think this is fine as it is, tbh.

addaleax (approved):

PR #1384, opened 2023-01-11
feat(shell-api): use collStats aggregation stage instead of command for collection.stats() MONGOSH-1157

Anemy:
MONGOSH-1157

This PR updates our `db.collection.stats()` shell api method to use the [aggregation stage `$collStats`](https://www.mongodb.com/docs/manual/reference/operator/aggregation/collStats/) for fetching collection statistics instead of the [`db.runCommand({ collStats: collectionName })`](https://www.mongodb.com/docs/manual/reference/command/collStats/) form. The run command form is to be deprecated by 7.0 (PM-2362)

The [`$collStats`](https://www.mongodb.com/docs/manual/reference/operator/aggregation/collStats/) stage returns a document for each individual shard's statistics. A lot of the code of this pr involves aggregating those documents into one document to be returned. I initially was trying to do this all in one aggregation pipeline, however I found it was going to be a bit harder to test and maintain. Things like new keys like `numOrphanDocs` added in 6.0 make an aggregation a lot more complex. If curious a fairly non-optimized implementation (without `timeseries` support) can be found on this gist: https://gist.github.com/Anemy/1e3b38f8f4f07f055a8c18d4b6ab443b

The [`$collStats`](https://www.mongodb.com/docs/manual/reference/operator/aggregation/collStats/) aggregation stage fails on sharded timeseries collections so we fallback to the old command form when that error code is encountered.

### Notes

- These changes alter the output format of `db.collection.stats()`
     Using the aggregation $collStats stage does not return the same data
     as the soon to be deprecated collStats command. Namely:
       - `primary` is not given now.
       - `nchunks` is not given now.
    These fields already aren't listed in our documentation: https://www.mongodb.com/docs/manual/reference/command/collStats/ but I'm curious to hear if folks feel we should work to re-add them or have more insights.

### DIFF
--- a/packages/shell-api/src/collection.ts
+++ b/packages/shell-api/src/collection.ts
@@ -1571,7 +1571,7 @@ export default class Collection extends ShellApiWithMongoClass {
             }
           }
         } else if (
-          // NOTE: `numOrphanDocs` is new in 6.0.
+          // NOTE: `numOrphanDocs` is new in 6.0. `totalSize` is new in 4.4.
           ['count', 'size', 'storageSize', 'totalIndexSize', 'totalSize', 'numOrphanDocs'].includes(fieldName)
         ) {
           if (counts[fieldName] === undefined) {
@@ -1654,7 +1654,10 @@ export default class Collection extends ShellApiWithMongoClass {
     }
     result.ns = ns;
     result.nindexes = nindexes;
-    result.scaleFactor = scale;
+    if (collStats[0].storageStats.scaleFactor !== undefined) {
+      // The `scaleFactor` property started being returned in 4.2.
+      result.scaleFactor = scale;
+    }
     result.ok = 1;
 
     return result;
--- a/packages/shell-api/src/collection.ts
+++ b/packages/shell-api/src/collection.ts
@@ -1503,58 +1503,346 @@ export default class Collection extends ShellApiWithMongoClass {
     options.indexDetails = options.indexDetails || false;
 
     this._emitCollectionApiCall('stats', { options });
-    // TODO(MONGOSH-1157): Adjust along the lines of the mongos code in
-    // https://github.com/mongodb/mongo/blob/master/src/mongo/s/commands/cluster_coll_stats_cmd.cpp
-    const result = await this._database._runCommand(
-      {
-        collStats: this._name, scale: options.scale
+
+    const collStatsAggOptions: {
+      storageStats: {
+        scale?: number;
       }
+    } = {
+      storageStats: {}
+    };
+
+    if (options.scale !== undefined) {
+      collStatsAggOptions.storageStats.scale = options.scale;
+    }
+
+    const cursor = await this.aggregate(
+      [
+        {
+          $collStats: collStatsAggOptions,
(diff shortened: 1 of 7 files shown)


Anemy:
Ah looks like a few fields were added in 4.2 and 4.4 so integration tests are failing on those versions. Updating to support those versions.

packages/shell-api/src/collection.ts
```diff
@@ -1475,6 +1476,231 @@ export default class Collection extends ShellApiWithMongoClass {
...
+        if (
+          [
+            'userFlags', 'capped', 'max', 'paddingFactorNote', 'indexDetails', 'wiredTiger'
+          ].includes(fieldName)
+        ) {
+          // Fields that are copied from the first shard only, because they need to
+          // match across shards.
+          if (result[fieldName] === undefined) {
+            result[fieldName] = shardStorageStats[fieldName];
+          }
```
addaleax, line 1555: (Feel free to ignore) If you want to simplify this a bit:

```suggestion
          result[fieldName] ??= shardStorageStats[fieldName];
```

packages/shell-api/src/collection.ts
```diff
@@ -1475,6 +1476,231 @@ export default class Collection extends ShellApiWithMongoClass {
...
+
+      const aggregatedCollStats = await this._aggregateAndScaleCollStats(collStats, scale);
+      return aggregatedCollStats;
+    } catch (e: any) {
+      if (e?.code === 13388) {
+        // Fallback to the deprecated way of fetching that folks can still
+        // fetch the stats of sharded timeseries collections. SERVER-72686
+        try {
+          const legacyCollStatsResult = await this._getLegacyCollStats(scale);
+          return legacyCollStatsResult;
```
addaleax, line 1694:
```suggestion
          return await this._getLegacyCollStats(scale);
```

packages/shell-api/src/collection.ts
```diff
@@ -1475,6 +1476,231 @@ export default class Collection extends ShellApiWithMongoClass {
...
+
+      if (!collStats || collStats[0] === undefined) {
+        throw new MongoshRuntimeError(
+          `Error running $collStats aggregation stage on ${this.getFullName()}`,
+          CommonErrors.CommandFailed
+        );
+      }
+
+      const aggregatedCollStats = await this._aggregateAndScaleCollStats(collStats, scale);
+      return aggregatedCollStats;
```
addaleax, line 1687:
```suggestion
      return await this._aggregateAndScaleCollStats(collStats, scale);
```

addaleax (commented):
> Is the way we're checking for sharding okay? Is there a better way? Does the current way require special permissions?

Yeah, I think it does require permissions that not all users might have… we could maybe fall back to doing something like the (somewhat inaccurate) `collStats.length > 1` if we cannot query `config.collections`?

packages/shell-api/src/collection.ts
Anemy: Ah nice! I somehow often forget this operator exists.

addaleax (approved):

PR #1391, opened 2023-01-18
feat(shell-api): introduces public Mongo.getURI

himanshusinghs:
MONGOSH-545

This PR introduces a public method `getURI` on Mongo which returns connection string of current active connection.


addaleax (approved):
Looks good! Only catch is that in our monorepos (Compass and mongosh) we put the package name in parentheses after the commit type, i.e. you probably want to make this `feat(shell-api): …` :slightly_smiling_face:

PR #1395, opened 2023-01-30
feat(shell-api): use $currentOp agg. stage for db.currentOp

himanshusinghs:
This PR updates the implementation of `db.currentOp` to make use of `$currentOp` aggregation stage instead of the command form. There is no change in the output format of the helper. The old arguments are also left intact and are appropriately mapped to relevant options for `$currentOp` stage (reference taken from the [old shell](https://github.com/10gen/mongo/blob/master/src/mongo/shell/db.js#L817-L835))

Closes:
- MONGOSH-1367
- MONGOSH-1157

### DIFF
--- a/packages/i18n/src/locales/en_US.ts
+++ b/packages/i18n/src/locales/en_US.ts
@@ -1134,7 +1134,7 @@ const translations: Catalog = {
             },
             currentOp: {
               link: 'https://docs.mongodb.com/manual/reference/method/db.currentOp',
-              description: 'Runs an aggregation using $currentOp operator. Returns an aggregated document that contains information on in-progress operations for the database instance. For further information, see $currentOp.',
+              description: 'Runs an aggregation using $currentOp operator. Returns a document that contains information on in-progress operations for the database instance. For further information, see $currentOp.',
               example: 'db.currentOp()'
             },
             killOp: {
--- a/packages/i18n/src/locales/en_US.ts
+++ b/packages/i18n/src/locales/en_US.ts
@@ -1134,7 +1134,7 @@ const translations: Catalog = {
             },
             currentOp: {
               link: 'https://docs.mongodb.com/manual/reference/method/db.currentOp',
-              description: 'Calls the currentOp command. Returns a document that contains information on in-progress operations for the database instance. The db.currentOp() method wraps the database command currentOp.',
+              description: 'Runs an aggregation using $currentOp operator. Returns an aggregated document that contains information on in-progress operations for the database instance. For further information, see $currentOp.',
               example: 'db.currentOp()'
             },
             killOp: {
--- a/packages/shell-api/src/database.ts
+++ b/packages/shell-api/src/database.ts
@@ -824,8 +824,7 @@ export default class Database extends ShellApiWithMongoClass {
 
     if (typeof opts === 'object') {
       const matchingFilters: Document = {};
-      for (const filtername in opts) {
-        if (Object.prototype.hasOwnProperty.call(opts, filtername)) {
+      for (const filtername of Object.keys(opts)) {
           if (filtername !== '$ownOps' && filtername !== '$all' && filtername !== '$truncateOps') {
             matchingFilters[filtername] = opts[filtername];
           }
--- a/packages/shell-api/src/database.ts
+++ b/packages/shell-api/src/database.ts
@@ -850,7 +850,7 @@ export default class Database extends ShellApiWithMongoClass {
     this._emitDatabaseApiCall('currentOp', { opts: opts });
     const currentOps = await this._getCurrentOperations(opts);
     return {
-      inprog: currentOps.length > 0 ? currentOps : [],
+      inprog: currentOps,
       ok: 1
     };
   }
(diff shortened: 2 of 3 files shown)


packages/i18n/src/locales/en_US.ts
```diff
@@ -1134,7 +1134,7 @@ const translations: Catalog = {
             },
             currentOp: {
               link: 'https://docs.mongodb.com/manual/reference/method/db.currentOp',
-              description: 'Calls the currentOp command. Returns a document that contains information on in-progress operations for the database instance. The db.currentOp() method wraps the database command currentOp.',
+              description: 'Runs an aggregation using $currentOp operator. Returns an aggregated document that contains information on in-progress operations for the database instance. For further information, see $currentOp.',
```
addaleax, line 1137: The return value isn’t really aggregated (in the English sense of the word), right? I think “Returns a document” was good wording here

packages/shell-api/src/database.ts
```diff
@@ -809,19 +809,50 @@ export default class Database extends ShellApiWithMongoClass {
...
-    }
-    return await this._runAdminCommand(
-      {
-        currentOp: 1,
-        ...opts
-      }
-    );
+    const currentOps = await this._getCurrentOperations(opts);
+    return {
+      inprog: currentOps.length > 0 ? currentOps : [],
```
addaleax, line 853: Isn’t this the same as this?

```suggestion
      inprog: currentOps,
```

packages/shell-api/src/database.ts
```diff
@@ -809,19 +809,50 @@ export default class Database extends ShellApiWithMongoClass {
...
+        'allUsers': !legacyCurrentOpOptions.$ownOps,
+        'idleConnections': legacyCurrentOpOptions.$all,
+        'truncateOps': false,
+      }
+    }];
+
+    if (typeof opts === 'object') {
+      const matchingFilters: Document = {};
+      for (const filtername in opts) {
+        if (Object.prototype.hasOwnProperty.call(opts, filtername)) {
```
addaleax, line 828: This can be simplified a bit:

```suggestion
      for (const filtername of Object.keys(opts)) {
```

packages/shell-api/src/database.ts
```diff
@@ -809,19 +809,50 @@ export default class Database extends ShellApiWithMongoClass {
     );
   }
 
+  async _getCurrentOperations(opts: Document | boolean) {
```
addaleax, line 812:
```suggestion
  async _getCurrentOperations(opts: Document | boolean): Promise<Document[]> {
```

packages/shell-api/src/database.ts
```diff
@@ -809,19 +809,50 @@ export default class Database extends ShellApiWithMongoClass {
...
+    const legacyCurrentOpOptions = typeof opts === 'boolean'
+      ? ({ '$all': opts, '$ownOps': false })
+      : ({ '$all': !!opts.$all, '$ownOps': !!opts.$ownOps });
+
+    const pipeline: Document[] = [{
+      $currentOp: {
+        'allUsers': !legacyCurrentOpOptions.$ownOps,
```
himanshusinghs, line 819: This might look a bit cryptic but that would only be because from the documentation -
- $ownOps - when true, will report current user's operations
- allUsers - when false, will report current user's operation

packages/shell-api/src/database.ts
```diff
@@ -809,19 +809,50 @@ export default class Database extends ShellApiWithMongoClass {
...
+        'idleConnections': legacyCurrentOpOptions.$all,
+        'truncateOps': false,
```
himanshusinghs, line 821: This one is has been derived from old shell's implementation

packages/shell-api/src/database.ts
```diff
@@ -809,19 +809,50 @@ export default class Database extends ShellApiWithMongoClass {
...
+      }
+    }];
+
+    if (typeof opts === 'object') {
+      const matchingFilters: Document = {};
+      for (const filtername in opts) {
+        if (Object.prototype.hasOwnProperty.call(opts, filtername)) {
+          if (filtername !== '$ownOps' && filtername !== '$all' && filtername !== '$truncateOps') {
```
himanshusinghs, line 829: Same goes for excluding `$truncateOps` here - was pulled from old shell's implementation

packages/shell-api/src/database.ts
```diff
@@ -809,19 +809,50 @@ export default class Database extends ShellApiWithMongoClass {
...
-      {
-        currentOp: 1,
-        ...opts
-      }
-    );
+    const currentOps = await this._getCurrentOperations(opts);
+    return {
+      inprog: currentOps.length > 0 ? currentOps : [],
+      ok: 1
+    };
```
himanshusinghs, line 855: Over here I have tried keeping the return format similar to the old one but in the old shell we also use the same format to signify an error which I am clearly not doing here as I assumed that we simply surface the error up to the shell.

packages/i18n/src/locales/en_US.ts
himanshusinghs: I think I was more focused on the process but you're right the return value itself is not. Will change that.

addaleax (approved):

addaleax (approved):

PR #1399, opened 2023-02-01
chore(deps): bump `mongodb-client-encryption` to 2.4.0 & related updates MONGOSH-1370

addaleax:
This PR is adjusting our dependencies, fixing our mock so that the updated libmongocrypt does not crash, and adds a new test for the Node.js bindings now being able to pick up AWS KMS credentials from environment variables if that is desired.

##### chore(deps): bump `mongodb-client-encryption` to 2.4.0

Also add `@aws-sdk/credential-prociders` as an explicit dependency.

The `mongodb-client-encryption` package can use this as an optional
peer dependency, and the `mongodb` package currently depends on this
as a peer dependency but will switch to making it an optional peer
dependency as well in the v5.0.0 release.

Installing it as an explicit dependency ensures its presence regardless
of optionality (instead of just relying on the package-lock.json file
for keeping it available).

##### chore: fix mock AWS KMS server

The latest version of mongodb-client-encryption crashes when
receiving invalid base64 from the AWS server; work around that
by never returning invalid base64 from the mock.

##### chore(cli-repl): add e2e test for automatic AWS KMS credentials

MONGOSH-1370

### DIFF
--- a/packages/cli-repl/test/e2e-fle.spec.ts
+++ b/packages/cli-repl/test/e2e-fle.spec.ts
@@ -44,86 +44,122 @@ describe('FLE tests', () => {
   });
   afterEach(TestShell.cleanup);
 
-  for (const useApiStrict of [ false, true ]) {
-    for (const withSessionToken of [ false, true ]) {
-      // eslint-disable-next-line no-loop-func
-      context(`with AWS KMS (apiStrict=${useApiStrict}, ${withSessionToken ? 'with' : 'without'} sessionToken)`, () => {
-        if (useApiStrict) {
-          skipIfServerVersion(testServer, '< 5.0');
+  function* awsTestCases() {
+    for (const useApiStrict of [ false, true ]) {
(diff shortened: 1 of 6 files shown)


packages/cli-repl/test/e2e-fle.spec.ts
```diff
@@ -44,86 +44,122 @@ describe('FLE tests', () => {
...
+            withSessionToken,
+            withEnvVarCredentials,
+            testDescription: `with AWS KMS (apiStrict=${useApiStrict}, ` +
+              `${withSessionToken ? 'with' : 'without'} sessionToken, ` +
+              `${withEnvVarCredentials ? 'with' : 'without'} credentials in env vars)`
+          };
         }
+      }
+    }
+  }
```
addaleax, line 62: This diff is much easier to view with whitespace – using a generator here to avoid increasing the indentation for the entire test more and more :)

gribnoysup (approved):

PR #1400, opened 2023-02-02
feat: add support for QE range indexes MONGOSH-1371

addaleax:
Add `ClientEncryption.encryptExpression` as an analogue to `ClientEncryption.encrypt` and add e2e tests for QE range functionality.

### DIFF
--- a/config/eslintrc.base.js
+++ b/config/eslintrc.base.js
@@ -38,7 +38,8 @@ module.exports = {
     'no-console': [1, { allow: ['warn', 'error', 'info'] }],
     'no-shadow': 0,
     'no-use-before-define': 0,
-    'no-cond-assign': [2, 'except-parens']
+    'no-cond-assign': [2, 'except-parens'],
+    'no-multi-str': 0
   },
   overrides: [{
     files: ['**/*.js'],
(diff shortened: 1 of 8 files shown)


packages/build/src/packaging/download-crypt-library.ts
```diff
@@ -13,8 +13,8 @@ export async function downloadCryptLibrary(variant: PackageVariant | 'host'): Pr
   console.info('mongosh: downloading latest crypt shared library for inclusion in package:', JSON.stringify(opts));
 
   const cryptTmpTargetDir = path.resolve(__dirname, '..', '..', '..', '..', 'tmp', 'crypt-store', variant);
-  // Download mongodb for latest server version.
-  const libdir = await downloadMongoDb(cryptTmpTargetDir, 'stable', opts);
+  // Download mongodb for latest server version, including rapid releases.
+  const libdir = await downloadMongoDb(cryptTmpTargetDir, 'continuous', opts);
```
addaleax, line 17: Asked on Slack to check whether this seemed like a reasonable thing to do, will wait for an answer before moving this out of draft state

config/eslintrc.base.js
```diff
@@ -38,7 +38,8 @@ module.exports = {
     'no-console': [1, { allow: ['warn', 'error', 'info'] }],
     'no-shadow': 0,
     'no-use-before-define': 0,
-    'no-cond-assign': [2, 'except-parens']
+    'no-cond-assign': [2, 'except-parens'],
+    'no-multi-str': 0
```
alenakhineika, line 42: Surprised that it is 1 by default.

alenakhineika (approved):

config/eslintrc.base.js
addaleax: Yeah, it’s referring to old browsers which don’t support it (pre-ES5), so … I think this is very safe to disable :)

PR #1402, opened 2023-02-06
feat(shell-api, cli-repl): on-boot warnings for fake MongoDBs

himanshusinghs:
This PR adds an on-boot on invokable warning when connected to a fake mongodb deployment (DocumentDB for example).

Closes: MONGOSH-57

### DIFF
--- a/packages/service-provider-server/src/cli-service-provider.ts
+++ b/packages/service-provider-server/src/cli-service-provider.ts
@@ -236,9 +236,19 @@ class CliServiceProvider extends ServiceProviderCore implements ServiceProvider 
   async getConnectionInfo(): Promise<ConnectionInfo> {
     const topology = this.getTopology();
     const { version } = require('../package.json');
-    const [buildInfo = null, cmdLineOpts = null, atlasVersion = null, fcv = null] = await Promise.all([
+    const [buildInfo = null, cmdLineOptsOrServerError = null, atlasVersion = null, fcv = null] = await Promise.all([
       this.runCommandWithCheck('admin', { buildInfo: 1 }, this.baseCmdOptions).catch(() => {}),
-      this.runCommandWithCheck('admin', { getCmdLineOpts: 1 }, this.baseCmdOptions).catch(() => {}),
+      this.runCommandWithCheck('admin', { getCmdLineOpts: 1 }, this.baseCmdOptions).catch((e) => {
+        // mongodb-build-info.getGenuineMongoDB expects either
+        // the successful or failure response from server
+        // Ref: https://github.com/mongodb-js/mongodb-build-info/blob/main/index.js#L89
+        return {
+          ok: e.ok,
+          code: e.code,
+          errmsg: e.message,
+          operationTime: e.operationTime
+        };
+      }),
       this.runCommandWithCheck('admin', { atlasVersion: 1 }, this.baseCmdOptions).catch(() => {}),
       this.runCommandWithCheck('admin', { getParameter: 1, featureCompatibilityVersion: 1 }, this.baseCmdOptions).catch(() => {})
     ]);
@@ -247,7 +257,7 @@ class CliServiceProvider extends ServiceProviderCore implements ServiceProvider 
       this.uri?.toString() ?? '',
       version,
       buildInfo,
-      cmdLineOpts,
+      cmdLineOptsOrServerError,
       atlasVersion,
       topology
     );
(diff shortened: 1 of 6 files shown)


packages/shell-api/src/mongo.ts
```diff
@@ -421,6 +421,37 @@ export default class Mongo extends ShellApiClass {
         }
         return new CommandResult('ShowBannerResult', null);
       }
+      case 'nonGenuineMongoDBCheck': {
+        const nonGenuineWarning = new CommandResult('ShowBannerResult', {
+          header: 'Warning: Non-Genuine MongoDB Detected',
+          content: [
+            'This server or service appears to be an emulation of MongoDB rather than an official MongoDB product.',
+            'Some documented MongoDB features may work differently, be entirely missing or incomplete, or have unexpected performance characteristics.',
+            'To learn more please visit: https://dochub.mongodb.org/core/non-genuine-mongodb-server-warning.',
```
himanshusinghs, line 430: @mmarcon is reviewing the content and the link

himanshusinghs:
Well I was looking at telemetry info for genuine check and realised that the work to determine wether a deployment is genuine or not is already done. I will probably make changes just to use that.

packages/shell-api/src/mongo.ts
```diff
@@ -421,6 +421,37 @@ export default class Mongo extends ShellApiClass {
...
+        }
+
+        try {
+          await db.serverCmdLineOpts();
+        } catch (error: any) {
+          const matchesKnownImposterSignature = error.message.indexOf('not supported') !== -1;
+          if (matchesKnownImposterSignature) {
+            return nonGenuineWarning;
+          }
+        }
```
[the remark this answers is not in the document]
himanshusinghs: yea this is exactly what I stumbled upon while I was looking into telemetry for is_genuine, though there is a problem with the detection - our package `mongodb-build-info` expect `cmdLineOpts` to be of the form of legacy options but here in mongosh we throw when we get an error which is why the non-genuine mongodb's are never detected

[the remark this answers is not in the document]
himanshusinghs: thank you anna! I addressed the changes in a recent commit

addaleax (approved):

mmarcon:
If it's not too much trouble, I'd wait for the ACK from the docs team on the URL before merging. If we are not planning a release soon I am also happy to merge and change the URL in a separate PR.

packages/service-provider-server/src/cli-service-provider.ts
```diff
@@ -236,9 +236,19 @@ class CliServiceProvider extends ServiceProviderCore implements ServiceProvider
...
     const topology = this.getTopology();
     const { version } = require('../package.json');
-    const [buildInfo = null, cmdLineOpts = null, atlasVersion = null, fcv = null] = await Promise.all([
+    const [buildInfo = null, cmdLineOptsOrServerError = null, atlasVersion = null, fcv = null] = await Promise.all([
       this.runCommandWithCheck('admin', { buildInfo: 1 }, this.baseCmdOptions).catch(() => {}),
-      this.runCommandWithCheck('admin', { getCmdLineOpts: 1 }, this.baseCmdOptions).catch(() => {}),
+      this.runCommandWithCheck('admin', { getCmdLineOpts: 1 }, this.baseCmdOptions).catch((e) => {
+        // mongodb-build-info.getGenuineMongoDB expects either
+        // the successful or failure response from server
+        // Ref: https://github.com/mongodb-js/mongodb-build-info/blob/main/index.js#L89
```
addaleax, line 244: I’d make this a permalink since the line numbers might shift around on a given branch:

```suggestion
        // Ref: https://github.com/mongodb-js/mongodb-build-info/blob/9247eeba730a905397ad09fe5a377067edc49b34/index.js#L89
```

addaleax:
@mmarcon Yeah, I’d like to do a release ideally this week, otherwise early next week … the docs link here is a dochub link, my understanding is that the idea here is that these serve as permanent link targets for external sources (like our products) with the intention that the actual location that they redirect to can be changed later as we need it, so we should be fine here?

@davemungo @ianf-mongodb maybe you can confirm?

packages/service-provider-server/src/cli-service-provider.ts
himanshusinghs: oh that's right

ianf-mongodb:
@mmarcon @addaleax that is correct currently the dochub entry is pointing at: https://www.mongodb.com/docs/manual/release-notes/4.2/#mongo-shell but can be updated if needed.

PR #1404, opened 2023-02-08
chore(shell-api): simplify key-mangement helpers by using driver helpers

himanshusinghs:
This PR updates the below mentioned methods to start using helpers provided by [libmongocrypt](https://github.com/mongodb/libmongocrypt/blob/f38c549/bindings/node/lib/clientEncryption.js)
- KeyVault.getKeys
- KeyVault.deleteKey
- KeyVault.addKeyAlternateName
- KeyVault.removeKeyAlternateName

Closes: MONGOSH-1205

### DIFF
--- a/packages/shell-api/src/field-level-encryption.spec.ts
+++ b/packages/shell-api/src/field-level-encryption.spec.ts
@@ -11,7 +11,6 @@ import { signatures, toShellResult } from './decorators';
 import { ALL_PLATFORMS, ALL_SERVER_VERSIONS, ALL_TOPOLOGIES, ALL_API_VERSIONS } from './enums';
 import { ClientEncryption, ClientSideFieldLevelEncryptionOptions, ClientSideFieldLevelEncryptionKmsProvider as KMSProvider, KeyVault } from './field-level-encryption';
 import Mongo from './mongo';
-import { DeleteResult } from './result';
 import ShellInstanceState from './shell-instance-state';
 import { CliServiceProvider } from '../../service-provider-server';
 import { startTestServer } from '../../../testing/integration-testing-hooks';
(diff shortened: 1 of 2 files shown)


packages/shell-api/src/field-level-encryption.spec.ts
```diff
@@ -366,75 +365,45 @@ describe('Field Level Encryption', () => {
...
-          DB,
-          COLL,
-          { _id: KEY_ID, keyAltNames: { $size: 0 } },
-          { $unset: { 'keyAltNames': '' }, $currentDate: { 'updateDate': true } },
-          { returnDocument: 'before' }
-        ]);
-        expect(result).to.deep.equal(r2.value);
+        expect(result).to.deep.equal(r.value);
       });
       it('reads keyAltNames and keyMaterial from DataKeyEncryptionKeyOptions', async() => {
```
himanshusinghs, line 408: Not related - but I don't think this test tests `removeKeyAlternateName` implementation, or?

packages/shell-api/src/field-level-encryption.ts
```diff
@@ -350,47 +350,36 @@ export class KeyVault extends ShellApiWithMongoClass {
     return await makeSingleDocReturnValue(() => this._keyColl.find({ 'keyAltNames': keyAltName }), 'KeyVault.getKeyByAltName', this._instanceState);
   }
 
+  // eslint-disable-next-line @typescript-eslint/require-await
   @returnType('Cursor')
   @apiVersions([1])
   @returnsPromise
   async getKeys(): Promise<Cursor> {
-    return this._keyColl.find({});
+    return new Cursor(
```
himanshusinghs, line 358: `libmongocrypt.getKeys` returns a `FindCursor` which is why I am wrapping that up in a `Cursor` here. I had doubts about `returning FindCursor instead of Cursor` being an API change. Let me know.

addaleax: Yes, this should be using `new Cursor` to return a shell cursor :+1:

packages/shell-api/src/field-level-encryption.spec.ts
addaleax: Right, these should not be grouped together :+1:

addaleax (approved):

PR #1406, opened 2023-02-09
fix(shell-api): adjust .drop() return value for server changes

addaleax:
See SERVER-43894, the server now returns no error when dropping non-existent collections.


Anemy (approved):

himanshusinghs:
Thanks a bunch for this - I was just looking into this as well!

PR #1412, opened 2023-02-20
fix(autocomplete): fixes autocomplete not working for db.aggregate

himanshusinghs:
This PR fixes the problem with autocomplete that was earlier not working for db level aggregate (`db.aggregate`)

### DIFF
--- a/packages/autocomplete/src/index.ts
+++ b/packages/autocomplete/src/index.ts
@@ -51,7 +51,7 @@ export const MATCH_COMPLETIONS = ([] as AnyCompletions).concat(
 );
 
 const DB_AGGREGATE_COMPLETIONS = STAGE_OPERATORS.filter(({ namespaces }) => {
-  return namespaces.length === 1 && (namespaces as readonly string[]).includes(DATABASE);
+  return namespaces.length === 1 && namespaces[0] === DATABASE;
 });
 
 /**
--- a/packages/autocomplete/src/index.ts
+++ b/packages/autocomplete/src/index.ts
@@ -11,7 +11,8 @@ import {
   BSON_TYPES,
   ATLAS,
   ADL,
-  ON_PREM
+  ON_PREM,
+  DATABASE
 } from '@mongodb-js/mongodb-constants';
 
 type TypeSignatureAttributes = { [key: string]: TypeSignature };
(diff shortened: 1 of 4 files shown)


himanshusinghs:
Another issues that came to light:
- We don't exclude `first only` and `database namespace only` stages while providing autocomplete for collection aggregation

packages/autocomplete/src/index.ts
```diff
@@ -49,6 +50,10 @@ export const MATCH_COMPLETIONS = ([] as AnyCompletions).concat(
   BSON_TYPES
 );
 
+const DB_AGGREGATE_COMPLETIONS = STAGE_OPERATORS.filter(({ namespaces }) => {
+  return namespaces.length === 1 && (namespaces as readonly string[]).includes(DATABASE);
```
addaleax, line 54: You’re already checking that `namespaces` only contains one value here, so it should be possible to avoid the cast here without too much trouble:

```suggestion
  return namespaces.length === 1 && namespaces.[0] === DATABASE;
```

It might also be worth adding a comment about this not being all `db.aggregate` completions as the name might imply, but only first-stage completions?

packages/autocomplete/src/index.ts
```diff
@@ -116,6 +121,22 @@ async function completer(params: AutocompleteParameters, line: string): Promise<
...
+      const prefix = splitQuery.pop()?.trim() || '';
+      const command: string = prefix ? line.split(prefix).shift() as string : line;
+      const suggestFirstStage = splitQuery.length <= 2;
+
+      const expressions = suggestFirstStage
+        ? DB_AGGREGATE_COMPLETIONS
+        : ([] as AnyCompletions).concat(
+          BASE_COMPLETIONS,
+          getStageAccumulators(params, elToComplete)
+        );
```
addaleax, line 135: TS should also be fine with simplifying this (and this probably applies elsewhere in the file as well):

```suggestion
        : [
          ...BASE_COMPLETIONS,
          ...getStageAccumulators(params, elToComplete)
        ];
```

addaleax (approved):

packages/autocomplete/src/index.ts
himanshusinghs: Oh that's right. Thanks for pointing that out.

himanshusinghs: About the comment regarding `DB_AGGREGATE_COMPLETIONS` variable itself. I made the usage of it more explicit which conveys that we are only providing operators that are first stage only.

PR #1416, opened 2023-02-23
feat(shell-api): adds createEncryptedCollection on Database and ClientEncryption

himanshusinghs:
Closes
--
MONGOSH-1316

Description:
--
This PR exposes a helper method called `createEncryptedCollection` provided by `libmongocrypt` to create an encrypted collection without explicitly specifying the data encryption keys for every encrypted fields.

The heavy lifting is already done by `libmongocrypt` and here we only provide an interface for users to interact with this functionality.

Usage:
--
With this change, users will be able to create an encrypted collection using any of the following forms:

- `db.createEncryptedCollection(collectionName, createCollectionOptions)`, or
- `db.getMongo().getClientEncryption().createEncryptedCollection(dbName, collectionName, createCollectionOptions)`

Assuming that the underlying connection was initiated with relevant FLE options.

Returns:
--
Originally `libmongocrypt's createEncryptedCollection` returns the native Collection object alongside a list of encrypted fields but here we augment the return type and instead we return the mongosh Collection object alongside the provided list of encrypted fields. This was done to make sure that users will be able to tail expected mongosh collection methods on the returned collection.

### DIFF
--- a/packages/cli-repl/test/e2e-fle.spec.ts
+++ b/packages/cli-repl/test/e2e-fle.spec.ts
@@ -471,9 +471,10 @@ describe('FLE tests', () => {
       const uri = JSON.stringify(await testServer.connectionString());
       await shell.waitForPrompt();
       await shell.executeLine('local = { key: BinData(0, "kh4Gv2N8qopZQMQYMEtww/AkPsIrXNmEMxTrs3tUoTQZbZu4msdRUaR8U5fXD7A7QXYHcEvuu4WctJLoT+NvvV3eeIg3MD+K8H9SR794m/safgRHdIfy6PD+rFpvmFbY") }');
-      await shell.executeLine(`keyMongo = Mongo(${uri}, { \
-        keyVaultNamespace: '${dbname}.keyVault', \
-        kmsProviders: { local } \
+      await shell.executeLine(`keyMongo = Mongo(${uri}, {
+        keyVaultNamespace: '${dbname}.keyVault',
+        kmsProviders: { local },
+        explicitEncryptionOnly: true
       });`);
       await shell.executeLine(`secretDB = keyMongo.getDB('${dbname}')`);
       await shell.executeLine(`var { collection, encryptedFields } = secretDB.createEncryptedCollection('secretCollection', {
--- a/packages/cli-repl/test/e2e-fle.spec.ts
+++ b/packages/cli-repl/test/e2e-fle.spec.ts
@@ -476,18 +476,18 @@ describe('FLE tests', () => {
         kmsProviders: { local } \
       });`);
       await shell.executeLine(`secretDB = keyMongo.getDB('${dbname}')`);
-      await shell.executeLine('var { collection, encryptedFields } = secretDB.createEncryptedCollection(\'secretCollection\', { \
-        provider: \'local\', \
-        createCollectionOptions: { \
-          encryptedFields: { \
-            fields: [{ \
-              keyId: null, \
-              path: \'secretField\', \
-              bsonType: \'string\' \
-            }] \
-          } \
-        } \
-      });');
+      await shell.executeLine(`var { collection, encryptedFields } = secretDB.createEncryptedCollection('secretCollection', {
+        provider: 'local',
+        createCollectionOptions: {
+          encryptedFields: {
+            fields: [{
+              keyId: null,
+              path: 'secretField',
+              bsonType: 'string'
+            }]
+          }
+        }
+      });`);
 
       await shell.executeLine(`plainMongo = Mongo(${uri});`);
       const collections = await shell.executeLine(`plainMongo.getDB('${dbname}').getCollectionNames()`);
@@ -497,7 +497,7 @@ describe('FLE tests', () => {
       expect(collections).to.include('secretCollection');
 
       const dekCount = await shell.executeLine(`plainMongo.getDB('${dbname}').getCollection('keyVault').countDocuments()`);
-      // Since there is only field to be encrypted hence there would only be one DEK in our keyvault collection
+      // Since there is only one field to be encrypted hence there would only be one DEK in our keyvault collection
       expect(parseInt(dekCount.trim(), 10)).to.equal(1);
     });
   });
--- a/packages/cli-repl/test/e2e-fle.spec.ts
+++ b/packages/cli-repl/test/e2e-fle.spec.ts
@@ -465,6 +465,41 @@ describe('FLE tests', () => {
       const compactResult = await shell.executeLine(`autoMongo.getDB('${dbname}').test.compactStructuredEncryptionData()`);
       expect(compactResult).to.include('ok: 1');
     });
+
+    it('creates an encrypted collection and generates data encryption keys automatically per encrypted fields', async() => {
+      const shell = TestShell.start({ args: ['--nodb', `--cryptSharedLibPath=${cryptLibrary}`] });
+      const uri = JSON.stringify(await testServer.connectionString());
+      await shell.waitForPrompt();
+      await shell.executeLine('local = { key: BinData(0, "kh4Gv2N8qopZQMQYMEtww/AkPsIrXNmEMxTrs3tUoTQZbZu4msdRUaR8U5fXD7A7QXYHcEvuu4WctJLoT+NvvV3eeIg3MD+K8H9SR794m/safgRHdIfy6PD+rFpvmFbY") }');
+      await shell.executeLine(`keyMongo = Mongo(${uri}, { \
+        keyVaultNamespace: '${dbname}.keyVault', \
+        kmsProviders: { local } \
+      });`);
+      await shell.executeLine(`secretDB = keyMongo.getDB('${dbname}')`);
+      await shell.executeLine('var { collection, encryptedFields } = secretDB.createEncryptedCollection(\'secretCollection\', { \
+        provider: \'local\', \
+        createCollectionOptions: { \
+          encryptedFields: { \
+            fields: [{ \
+              keyId: null, \
+              path: \'secretField\', \
+              bsonType: \'string\' \
+            }] \
+          } \
+        } \
+      });');
+
+      await shell.executeLine(`plainMongo = Mongo(${uri});`);
+      const collections = await shell.executeLine(`plainMongo.getDB('${dbname}').getCollectionNames()`);
+      expect(collections).to.include('enxcol_.secretCollection.ecc');
+      expect(collections).to.include('enxcol_.secretCollection.esc');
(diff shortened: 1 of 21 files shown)


packages/shell-api/src/database.ts
```diff
@@ -617,6 +618,21 @@ export default class Database extends ShellApiWithMongoClass {
     }
   }
 
+  @returnsPromise
+  @apiVersions([])
```
himanshusinghs, line 622: Since this method was not mentioned in stable api [supported commands](https://www.mongodb.com/docs/manual/reference/stable-api-changelog/#std-label-stable-api-changelog), I skipped adding anything to this. Let me know if this is not the case.

packages/shell-api/src/database.ts
```diff
@@ -617,6 +618,21 @@ export default class Database extends ShellApiWithMongoClass {
...
+  async createEncryptedCollection(name: string, options: CreateEncryptedCollectionOptions): Promise<{ collection: Collection, encryptedFields: Document }> {
+    assertArgsDefinedType([name], ['string'], 'ClientEncryption.createEncryptedCollection');
+    assertArgsDefinedType([options], ['object'], 'ClientEncryption.createEncryptedCollection');
+    assertKeysDefined(options, ['provider', 'createCollectionOptions']);
```
himanshusinghs, line 626: Maybe asserting args are fine but am I being too itchy going in and asserting keys? I have a feeling that it would be fine to leave it to libmongocrypt and throw when required params are missing.

packages/shell-api/src/field-level-encryption.ts
```diff
@@ -208,10 +215,34 @@ export class ClientEncryption extends ShellApiWithMongoClass {
...
+    collName: string,
+    options: CreateEncryptedCollectionOptions
+  ): Promise<{ collection: Collection, encryptedFields: Document }> {
+    assertArgsDefinedType([dbName], ['string'], 'ClientEncryption.createEncryptedCollection');
+    assertArgsDefinedType([collName], ['string'], 'ClientEncryption.createEncryptedCollection');
+    assertArgsDefinedType([options], ['object'], 'ClientEncryption.createEncryptedCollection');
+    assertKeysDefined(options, ['provider', 'createCollectionOptions']);
+
+    const nativeClient = this._mongo._serviceProvider.getRawClient();
+    const db = nativeClient.db(dbName);
```
himanshusinghs, line 235: Is there a better way to get a hold of native driver's database object?

packages/shell-api/src/field-level-encryption.ts
```diff
@@ -208,10 +215,34 @@ export class ClientEncryption extends ShellApiWithMongoClass {
...
+    const { encryptedFields } = await this._libmongocrypt.createEncryptedCollection(
+      db,
+      collName,
+      options
+    );
+    return {
+      collection: new Collection(this._mongo, this._mongo.getDB(dbName), collName),
```
himanshusinghs, line 242: Also how do we feel about this? I am returning our Collection object instead of native Collection object just to make sure users are able to chain mongosh Collection's methods on top of it.

packages/shell-api/src/database.ts
addaleax: This method doesn’t do anything beyond `create` and `insert` in terms of command usage (I think), so it should be supported … I think you can wait and see how `test_apistrict` does with the e2e test for this one

packages/shell-api/src/database.ts
addaleax: This seems fine either way to me. I don’t think you really need to validate here, though, since you’re already validating arguments for the `ClientEncryption` method that this wraps?

packages/cli-repl/test/e2e-fle.spec.ts
```diff
@@ -465,6 +465,41 @@ describe('FLE tests', () => {
...
+      const uri = JSON.stringify(await testServer.connectionString());
+      await shell.waitForPrompt();
+      await shell.executeLine('local = { key: BinData(0, "kh4Gv2N8qopZQMQYMEtww/AkPsIrXNmEMxTrs3tUoTQZbZu4msdRUaR8U5fXD7A7QXYHcEvuu4WctJLoT+NvvV3eeIg3MD+K8H9SR794m/safgRHdIfy6PD+rFpvmFbY") }');
+      await shell.executeLine(`keyMongo = Mongo(${uri}, { \
+        keyVaultNamespace: '${dbname}.keyVault', \
+        kmsProviders: { local } \
+      });`);
+      await shell.executeLine(`secretDB = keyMongo.getDB('${dbname}')`);
+      await shell.executeLine('var { collection, encryptedFields } = secretDB.createEncryptedCollection(\'secretCollection\', { \
+        provider: \'local\', \
```
addaleax, line 480: You don’t need to do all of this string escaping if you use template strings, like a few lines above (and you can probably also drop the line continuations at the end of each line then)

packages/cli-repl/test/e2e-fle.spec.ts
```diff
@@ -465,6 +465,41 @@ describe('FLE tests', () => {
...
+
+      await shell.executeLine(`plainMongo = Mongo(${uri});`);
+      const collections = await shell.executeLine(`plainMongo.getDB('${dbname}').getCollectionNames()`);
+      expect(collections).to.include('enxcol_.secretCollection.ecc');
+      expect(collections).to.include('enxcol_.secretCollection.esc');
+      expect(collections).to.include('enxcol_.secretCollection.ecoc');
+      expect(collections).to.include('secretCollection');
+
+      const dekCount = await shell.executeLine(`plainMongo.getDB('${dbname}').getCollection('keyVault').countDocuments()`);
+      // Since there is only field to be encrypted hence there would only be one DEK in our keyvault collection
```
addaleax, line 500:
```suggestion
      // Since there is only one field to be encrypted hence there would only be one DEK in our keyvault collection
```

Probably?

packages/service-provider-server/src/cli-service-provider.ts
```diff
@@ -84,6 +85,7 @@ import type { MongoshBus } from '@mongosh/types';
 import { forceCloseMongoClient } from './mongodb-patches';
 import ConnectionString from 'mongodb-connection-string-url';
 import { EventEmitter } from 'events';
+import { CreateEncryptedCollectionOptions } from '@mongosh/service-provider-core/lib/admin';
```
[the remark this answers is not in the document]
himanshusinghs: Got it.

packages/cli-repl/test/e2e-fle.spec.ts
himanshusinghs: I actually started with template strings but my eslint went havoc on me and won't let me by without this autofix. Weird but could you confirm if that is actually the case?

himanshusinghs: Working fine now - somehow 😮

packages/shell-api/src/database.ts
himanshusinghs: Perfect, thanks :)

packages/cli-repl/test/e2e-fle.spec.ts
```diff
@@ -465,6 +465,41 @@ describe('FLE tests', () => {
       const compactResult = await shell.executeLine(`autoMongo.getDB('${dbname}').test.compactStructuredEncryptionData()`);
       expect(compactResult).to.include('ok: 1');
     });
+
+    it('creates an encrypted collection and generates data encryption keys automatically per encrypted fields', async() => {
+      const shell = TestShell.start({ args: ['--nodb', `--cryptSharedLibPath=${cryptLibrary}`] });
```
addaleax, line 470: This shouldn’t require automatic encryption. We usually omit `--cryptSharedLibPath=...` in those cases, to make sure that this still works even if there is no crypt shared library available (which should only ever be required for automatic encryption):

```suggestion
      const shell = TestShell.start({ args: ['--nodb'] });
```

packages/shell-api/src/field-level-encryption.ts
```diff
@@ -16,18 +16,19 @@ import {
...
 import Cursor from './cursor';
 import { DeleteResult } from './result';
-import { assertArgsDefinedType } from './helpers';
+import { assertArgsDefinedType, assertKeysDefined } from './helpers';
 import { asPrintable, shellApiType } from './enums';
 import { redactURICredentials } from '@mongosh/history';
 import type Mongo from './mongo';
 import { CommonErrors, MongoshInvalidInputError, MongoshRuntimeError } from '@mongosh/errors';
 import ShellInstanceState from './shell-instance-state';
+import { CreateEncryptedCollectionOptions } from '@mongosh/service-provider-core';
```
addaleax, line 31:
```suggestion
import type { CreateEncryptedCollectionOptions } from '@mongosh/service-provider-core';
```

packages/service-provider-core/src/admin.ts
```diff
@@ -116,4 +129,14 @@ export default interface Admin {
...
   getFleOptions?: () => AutoEncryptionOptions | undefined;
+
+  /**
+   * The helper method to correctly access FLE implementation's createEncryptedCollection
+   */
+  createEncryptedCollection?(
+    dbName: string,
+    collName: string,
+    options: CreateEncryptedCollectionOptions,
+    libmongocrypt: MongoCryptClientEncryption
```
addaleax, line 140: `libmongocrypt` is the same as `this.fle` in the service provider, so I would probably not pass it back to a method on the server provider? If they’re different, that would be a sign that something’s very broken

himanshusinghs: Oh but this is an instance of libmongocrypt's ClientEncryption and what ServiceProviders have is just the imported class itself. From what I understood, the instance is created only by `Mongo.getClientEncryption` and also cached there on the Mongo instance. Since an instance was already in scope I decided to use it otherwise I would be doing something similar to `Mongo.getClientEncryption` in the service providers as well.

packages/cli-repl/test/e2e-fle.spec.ts
addaleax: Mh no, this should really work without the shared library present… Maybe you need to set `explicitEncryptionOnly: true` (or `bypassQueryAnalysis: true`, but that would be less desirable because it leaves more features enabled) when creating `keyMongo`, similar to other tests here which don’t use the shared library?

himanshusinghs: Ahh you're right. It works fine with `bypassQueryAnalysis`. I guess I missed a few readings - gonna go back to docs now. Thanks :)

addaleax (approved):

PR #1418, opened 2023-02-23
fix(shell-api): fix getCollections for Object.prototype keys

addaleax:
Otherwise:

```js
> db.getCollection('hasOwnProperty')
[Function: hasOwnProperty]
```


gribnoysup (approved):
Wow, good catch!

PR #1421, opened 2023-03-07
fix(shell-api): removes the retry behaviour from rs.reconfig - MONGOSH-1304

himanshusinghs:
This PR removes the retry behaviour from `rs.reconfig` method which instead has now been moved to ReplicaSet tests and only optionally used to reduce flakes during cleanup.

### DIFF
--- a/packages/shell-api/src/replica-set.spec.ts
+++ b/packages/shell-api/src/replica-set.spec.ts
@@ -986,7 +986,7 @@ describe('ReplicaSet', () => {
       const addArbWithRetry = createRetriableFunc(rs, 'addArb');
       /**
        * Small hack warning:
-       * rs.reconfigForPSASet internally use rs.reconfig twice with different configs
+       * rs.reconfigForPSASet internally uses rs.reconfig twice with different configs
        * rs.reconfig itself could lead to a pseudo test failure because of delay in propogation
        * of new config. This small hack here helps us reduce flakiness in our test runs
        */
--- a/packages/shell-api/src/replica-set.spec.ts
+++ b/packages/shell-api/src/replica-set.spec.ts
@@ -780,7 +780,7 @@ describe('ReplicaSet', () => {
     let instanceState: ShellInstanceState;
     let db: Database;
     let rs: ReplicaSet;
-    let reconfigWithRetry: RetriableFn;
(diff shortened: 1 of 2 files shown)


packages/shell-api/src/replica-set.spec.ts
```diff
@@ -26,6 +26,41 @@ function deepClone<T>(value: T): T {
   return JSON.parse(JSON.stringify(value));
 }
 
+
+type RetriableFn = (...fnArgs: any) => Promise<any>;
+
+function createRetriableFunc(rs: ReplicaSet, fn: keyof ReplicaSet, totalRetries = 12): RetriableFn {
```
addaleax, line 32: You can probably make this a bit more precise by using:

```suggestion
function createRetriableFunc<F extends keyof ReplicaSet>(rs: ReplicaSet, fn: F, totalRetries = 12): ReplicaSet[F] {
```

instead of having a `RetriableFn` which is equivalent to just “any function”.

You’d need to use `let reconfigWithRetry: ReplicaSet['reconfig'];` below then

himanshusinghs: This is definitely what I was looking for. Thanks 👍

packages/shell-api/src/replica-set.spec.ts
```diff
@@ -1058,6 +983,14 @@ describe('ReplicaSet', () => {
       const instanceState = new ShellInstanceState(serviceProvider);
       const db = instanceState.currentDb;
       const rs = new ReplicaSet(db);
+      const addArbWithRetry = createRetriableFunc(rs, 'addArb');
+      /**
+       * Small hack warning:
+       * rs.reconfigForPSASet internally use rs.reconfig twice with different configs
```
addaleax, line 989:
```suggestion
       * rs.reconfigForPSASet internally uses rs.reconfig twice with different configs
```

addaleax (approved):

addaleax:
Just fyi, there’s a typo in the ticket number in the PR title :slightly_smiling_face:

himanshusinghs:
> Just fyi, there’s a typo in the ticket number in the PR title 🙂

Ohh wow, also in the branch name 🤦‍♂️ I will create a new PR and link this one.

PR #1427, opened 2023-03-10
chore: add npm workspaces MONGOSH-990

YakoYakoYokuYoku:
Basically this adds a `workspaces` entry in `package.json` that reflects the one from `lerna.json` and in the latter the `useWorkspaces` entry is set to `true`.

This was tested with the following commands.

```
npm install
npm run compile-cli
node packages/cli-repl/bin/mongosh.js
```

With this PR there's now the possibility to hoist `devDependencies`, deprecate usages of `lerna bootstrap` and to update `lerna` to a much newer version.


addaleax:
Hi @YakoYakoYokuYoku! In order to accept external contributions, we need a signed [Contributor Agreement](https://www.mongodb.com/legal/contributor-agreement). Could you do that?

YakoYakoYokuYoku:
> Hi @YakoYakoYokuYoku! In order to accept external contributions, we need a signed Contributor Agreement. Could you do that?

Done, now this is ready for reviews. And thanks for reminding me to sign the CA.

addaleax:
@YakoYakoYokuYoku Thank you! I’ve updated/removed the `package-lock` files here, that’s required in order for this to pass CI. We’ll see how that goes :slightly_smiling_face:

YakoYakoYokuYoku:
> @YakoYakoYokuYoku Thank you! I’ve updated/removed the `package-lock` files here, that’s required in order for this to pass CI. We’ll see how that goes :slightly_smiling_face:

This surprises me, I thought that only `package.json` and `lerna.json` needed to be modified, but it seems that everything was moved to the root `package-lock.json`.
Regardless, I have faith that this is going to work.
Also if `lerna bootstrap` doesn't work then `npm install` should do the trick.

addaleax:
@YakoYakoYokuYoku Yeah, using workspaces means that dependencies are all hoisted to the top level (by default?), so the package-lock.json becomes a global one, not a per-package one, and correspondingly in order for `npm ci` to work successfully the package-lock.json file needs to be updated. (To be clear, this is a good thing! :slightly_smiling_face:)

YakoYakoYokuYoku:
There's mongodb-js/boxednode#41 that addresses an issue in the e2e test as it doesn't display a clear stack trace.

I've also raised `browser-repl`'s webpack version to `^5.75.0` as that version processes `node:module` much more correctly.

Though I theorize that if all of `devDependencies` are hoisted then many of these CI failures will disappear.

addaleax:
There’s a lot of CI failures because a nodejs.org website outage it seems … :disappointed: will try to restart CI later today or tomorrow.

YakoYakoYokuYoku:
At least the number of failures is going down each time a fix is pushed, I can wait and we are clearly closer to succeed.

addaleax:
Alright, the only remaining failure here is the VSCode extension integration tests, which I think is expeted because that’s still on the 4.x driver. I’ll merge this :+1: This won’t resolve MONGOSH-990 because for that, we’ll likely also want to move the dependencies from the top-level package.json to the individual packages’ devDependencies lists.

PR #1446, opened 2023-04-20
fix(cli-repl,shell-api): skip/fix tests for latest server alpha

addaleax:
- Skip FLE tests on 7.x until MONGOSH-1410
- Skip sharding auto-splitter tests entirely after the removal in 7.0
- Make the code in an `after` hook pass if the before hook never ran
- Fix the `collStats` compatibility code in the timeseries case (since now the agg pipeline variant works on timeseries collections in 7.1+)


lerouxb (approved):
👏

PR #1452, opened 2023-04-25
fix(browser-repl): account for node-runtime-worker-thread-specific result rendering MONGOSH-1323

addaleax:
d1278ecd5d7d attempted to address MONGOSH-1323 but failed to do so properly because node-runtime-worker-thread adjusts behavior so that most values result in reporting a `string` result back to the frontend.


gribnoysup (approved):
That was a quick fix! And thanks for adding a test for that!

PR #1453, opened 2023-04-25
fix(ci): update Python version for Windows CI

addaleax:
This broke today because the build hosts were updated to a more recent Python version.

https://mongodb.slack.com/archives/C0V896UV8/p1682441001192679


Anemy (approved):
Lgtm, will keep an eye out, if this happens often it's maybe something we can automate or dynamically set

PR #1458, opened 2023-05-05
feat: implement OIDC support MONGOSH-1384

addaleax:


packages/browser-runtime-electron/README.md
```diff
@@ -17,7 +17,7 @@ const runtime = new ElectronRuntime(serviceProvider);
 import { Shell, ElectronRuntime } from 'mongosh-browser-repl';
 
 const runtime = new ElectronRuntime(
-  CompassServiceProvider.fromDataService(dataService)
+  await CompassServiceProvider.connect(...)
```
alenakhineika, line 20: Do we want maybe to have some simple example of what can be used as the function arg instead of `...?

packages/cli-repl/src/cli-repl.ts
```diff
@@ -287,7 +287,22 @@ export class CliRepl implements MongoshIOProvider {
       delete driverOptions.autoEncryption;
     }
 
-    const initialServiceProvider = await this.connect(driverUri, driverOptions);
+    await this.prepareOIDCOptions(driverOptions);
```
alenakhineika, line 290: For me, it would read more clearly if the function returned the value, rather than updated `driverOptions` under the hood. Is there some best practice depending on the use case?

alenakhineika (approved):

packages/browser-runtime-electron/README.md
addaleax: I'm ... leaning towards no? This example isn't meant to be even close to runnable, it's intended to demonstrate how the different pieces of architecture (ElectronRuntime, ServiceProvider, browser-repl) work together.

packages/cli-repl/src/cli-repl.ts
addaleax: (updated, although github isn't showing the update right now...)

PR #1461, opened 2023-05-15
feat(shell-api): add autoMerger helpers MONGOSH-1375

lerouxb:


packages/shell-api/src/shard.ts
```diff
@@ -464,4 +464,73 @@ export default class Shard extends ShellApiWithMongoClass {
...
+      { _id: 'automerge' },
+      { $set: { enabled: false } },
+      { upsert: true, writeConcern: { w: 'majority', wtimeout: 30000 } }
+    ) as UpdateResult;
+  }
+
+  @returnsPromise
+  @apiVersions([])
+  @serverVersions(['7.0.0', ServerVersions.latest])
+  async shouldAutoMerge(): Promise<boolean> {
```
addaleax, line 497: Did the sharding team come up with this name? I’m wondering if something like “sh.isAutoMergerEnabled()” is a better name, also considering that “sh.isBalancerRunning()” is a thing

packages/shell-api/src/shard.spec.ts
```diff
@@ -1217,6 +1217,275 @@ describe('Shard', () => {
...
+      });
+
+      it('throws if not mongos', async() => {
+        const expectedResult = { ok: 1 } as any;
+        serviceProvider.runCommandWithCheck.resolves({ ok: 1, msg: 'not dbgrid' });
+        serviceProvider.updateOne.resolves(expectedResult);
+        await shard.enableAutoMerger('ns');
+        expect(warnSpy.calledOnce).to.equal(true);
+      });
+    });
```
addaleax, line 1488: To be clear, what I meant on Slack was that I had assumed that there were tests in the `integration` section below :)

lerouxb: Will check. I thought that I copied all the code for the things I copy/pasted from :)

packages/shell-api/src/shard.ts
lerouxb: Yes I copied from where the ticket said I should copy from: https://github.com/10gen/mongo/pull/11291/files

addaleax: Yeah, that’s the nice thing about the legacy shell not shipping anymore … we can do whatever we think is best in terms of UX.

addaleax (approved):

lerouxb:
I think these errors are all things we're seeing elsewhere, so should be safe to merge.

PR #1468, opened 2023-05-23
feat(shell-api): add db.collection.analyzeShardKey() and db.collection.configureQueryAnalyzer() MONGOSH-1442

lerouxb:


packages/shell-api/src/replica-set.spec.ts
```diff
@@ -946,6 +946,53 @@ describe('ReplicaSet', () => {
         expect(status.members.length).to.equal(3);
       });
     });
+
+    describe('analyzeShardKey()', () => {
```
lerouxb, line 950: Is this the right place to put the integration tests that makes sure these work without sharding just on a replicaset?

packages/shell-api/src/collection.ts
```diff
@@ -2033,4 +2034,29 @@ export default class Collection extends ShellApiWithMongoClass {
...
+    return await this._database._runAdminCommand({
+      analyzeShardKey: this.getFullName(),
+      key
+    });
+  }
+
+  @returnsPromise
+  @topologies([Topologies.ReplSet, Topologies.Sharded])
+  @apiVersions([])
+  async configureQueryAnalyzer({ mode, sampleRate }: { mode: 'full' | 'off', sampleRate?: number }): Promise<Document> {
```
lerouxb, line 2053: sampleRate is only for mode: 'full' in which case it is required. Not sure how best to represent that in typescript without excessive branching below.

addaleax: I think that would be 

```suggestion
  async configureQueryAnalyzer({ mode, sampleRate }: { mode: 'full', sampleRate: number } | { mode: 'off', sampleRate?: never }): Promise<Document> {
```

but I really think we don’t need to worry about those details here.

packages/shell-api/src/collection.ts
```diff
@@ -2033,4 +2034,29 @@ export default class Collection extends ShellApiWithMongoClass {
...
+    assertArgsDefinedType([mode, sampleRate], ['string', [undefined, 'number']], 'Collection.configureQueryAnalyzer');
+    this._emitCollectionApiCall('configureQueryAnalyzer', { mode, sampleRate });
+    return await this._database._runAdminCommand({
+      configureQueryAnalyzer: this.getFullName(),
+      mode,
+      sampleRate: mode === 'full' && sampleRate ? new bson.Double(sampleRate) : undefined
+    });
```
addaleax, line 2060: Is there a specific reason for now passing through the entire options object, i.e. any extra keys that we don’t know about? That might be nice for forward compatibility with options we don’t know about yet.

addaleax (approved):

PR #1476, opened 2023-05-31
feat: stop sending truncateOps to ADF MONGOSH-1431

alenakhineika:
Re-try `currentOp` without `truncateOps` if we encounter a `FailedToParse` error.

### DIFF
--- a/packages/shell-api/src/database.ts
+++ b/packages/shell-api/src/database.ts
@@ -819,23 +819,19 @@ export default class Database extends ShellApiWithMongoClass {
 
     const adminDb = this.getSiblingDB('admin');
     const aggregateOptions = { $readPreference: { mode: 'primaryPreferred' } };
-    let result;
 
     try {
       const cursor = await adminDb.aggregate(pipeline, aggregateOptions);
-      result = await cursor.toArray();
+      return await cursor.toArray();
     } catch (error) {
-      if ((error as any).codeName === 'FailedToParse') {
+      if ((error as any)?.codeName === 'FailedToParse') {
         delete pipeline[0].$currentOp.truncateOps;
 
         const cursor = await adminDb.aggregate(pipeline, aggregateOptions);
-        result = await cursor.toArray();
-      } else {
-        throw error;
+        return await cursor.toArray();
       }
+      throw error;
     }
-
-    return result;
   }
 
   @returnsPromise


packages/shell-api/src/database.ts
```diff
@@ -817,13 +817,36 @@ export default class Database extends ShellApiWithMongoClass {
...
+        .aggregate(pipeline, {
+          $readPreference: {
+            mode: 'primaryPreferred'
+          }
+        });
+      result = await cursor.toArray();
+    } catch (error) {
+      if ((error as any).codeName === 'FailedToParse') {
+        const deepCopyPipeline = JSON.parse(JSON.stringify(pipeline));
+        delete deepCopyPipeline[0].$currentOp.truncateOps;
```
addaleax, line 834: Why are we cloning the pipeline here? I think that would make sense if the pipeline was passed in from the user, but we create it here inside this function, so it should be fine to just keep re-using it

(I’m bringing this up because the deep clone makes it seem like this could be any user-controlled input, but if it is user-controlled input, it could contain arbitrary BSON, and using JSON to clone that wouldn’t handle BSON values well, so this looks “wrong” to some degree without actually being wrong of course)

packages/shell-api/src/database.ts
```diff
@@ -817,13 +817,36 @@ export default class Database extends ShellApiWithMongoClass {
...
+
+        const cursor = await this
+          .getSiblingDB('admin')
+          .aggregate(deepCopyPipeline, {
+            $readPreference: {
+              mode: 'primaryPreferred'
+            }
+          });
```
addaleax, line 842: Maybe `{ $readPreference: { mode: 'primaryPreferred' } }` could be stored in some `aggregateOptions` variable and then passed to both calls? Similarly `this.getSiblingDB('admin')` could be stored in e.g. `adminDb` to deduplicate them?

packages/shell-api/src/database.ts
alenakhineika: I started from this, but it affects tests when truncateOps disappears not only for the second but also for the first call. And I think we want to track in tests that there are actually 2 calls that one rejects but the second resolves. What would be a better solution here?

alenakhineika: Maybe I also can assign $currentOp to a variable 🤔

addaleax: Yeah, thanks for the context. I think that’s something that the test should handle, though – you’re already modifying `serviceProvider.aggregateDb` with `.returns()`, you could also make that `.callsFake((...args) => ...)` and then store (a copy of) `args` each time? It’s partially re-implementing a bit of sinon functionality, yes, but I think that’s okay here

packages/shell-api/src/database.ts
```diff
@@ -817,13 +817,25 @@ export default class Database extends ShellApiWithMongoClass {
...
-          'mode': 'primaryPreferred'
-        }
-      })).toArray();
+    const adminDb = this.getSiblingDB('admin');
+    const aggregateOptions = { $readPreference: { mode: 'primaryPreferred' } };
+    let result;
+
+    try {
+      const cursor = await adminDb.aggregate(pipeline, aggregateOptions);
+      result = await cursor.toArray();
```
addaleax, line 826:
```suggestion
      return await cursor.toArray();
```

packages/shell-api/src/database.ts
```diff
@@ -817,13 +817,25 @@ export default class Database extends ShellApiWithMongoClass {
...
+    } catch (error) {
+      if ((error as any).codeName === 'FailedToParse') {
+        delete pipeline[0].$currentOp.truncateOps;
+
+        const cursor = await adminDb.aggregate(pipeline, aggregateOptions);
+        result = await cursor.toArray();
```
addaleax, line 832:
```suggestion
        return await cursor.toArray();
```

packages/shell-api/src/database.ts
```diff
@@ -817,13 +817,25 @@ export default class Database extends ShellApiWithMongoClass {
...
-      })).toArray();
+    const adminDb = this.getSiblingDB('admin');
+    const aggregateOptions = { $readPreference: { mode: 'primaryPreferred' } };
+    let result;
+
+    try {
+      const cursor = await adminDb.aggregate(pipeline, aggregateOptions);
+      result = await cursor.toArray();
+    } catch (error) {
+      if ((error as any).codeName === 'FailedToParse') {
```
addaleax, line 828:
```suggestion
      if ((error as any)?.codeName === 'FailedToParse') {
```

(I think we do this elsewhere too, probably doesn’t matter much in practice but I think it’s a good standard to stick to)

packages/shell-api/src/database.ts
```diff
@@ -817,13 +817,25 @@ export default class Database extends ShellApiWithMongoClass {
...
-    return (await this
-      .getSiblingDB('admin')
-      .aggregate(pipeline, {
-        '$readPreference': {
-          'mode': 'primaryPreferred'
-        }
-      })).toArray();
+    const adminDb = this.getSiblingDB('admin');
+    const aggregateOptions = { $readPreference: { mode: 'primaryPreferred' } };
+    let result;
```
addaleax, line 822:
```suggestion
```

packages/shell-api/src/database.ts
```diff
@@ -817,13 +817,25 @@ export default class Database extends ShellApiWithMongoClass {
...
+        delete pipeline[0].$currentOp.truncateOps;
+
+        const cursor = await adminDb.aggregate(pipeline, aggregateOptions);
+        result = await cursor.toArray();
+      } else {
+        throw error;
+      }
+    }
+
+    return result;
```
addaleax, line 838:
```suggestion
```

addaleax (approved):

addaleax (approved):

PR #1477, opened 2023-05-31
feat(shell-api): search index management helpers MONGOSH-1455 MONGOSH-1456

lerouxb:
I added getSearchIndexes() which, like getIndexes() returns an array. I made the service provider method listSearchIndexes which returns the cursor. So we could add easily add listSearchIndexes() later that returns the cursor if we want to. ie. by convention get could return an array and list could return a cursor. We'd probably want to do the same for normal indexes if we go that route, though.

### DIFF
--- a/packages/autocomplete/src/index.spec.ts
+++ b/packages/autocomplete/src/index.spec.ts
@@ -280,7 +280,9 @@ describe('completer.completer', () => {
       const adjusted = collComplete
         .filter(c =>
           ![
-            'count', 'update', 'remove', 'insert', 'save', 'findAndModify', 'reIndex', 'mapReduce'
+            'count', 'update', 'remove', 'insert', 'save', 'findAndModify', 'reIndex', 'mapReduce',
+            // search index helpers are 6.0+
+            'getSearchIndexes', 'createSearchIndex', 'createSearchIndexes', 'dropSearchIndex', 'updateSearchIndex'
           ].includes(c)
         )
         .map(c => `${i}${c}`);


packages/service-provider-server/src/cli-service-provider.ts
```diff
@@ -1208,7 +1210,15 @@ class CliServiceProvider extends ServiceProviderCore implements ServiceProvider
...
-  // TODO(MONGOSH-1456): getSearchIndexes
+  listSearchIndexes(
+    database: string,
+    collection: string,
+    options: ListSearchIndexesOptions,
+    dbOptions?: DbOptions): ListSearchIndexesCursor {
+    return this.db(database, dbOptions)
+      .collection(collection)
+      .listSearchIndexes(options);
+  }
```
addaleax, line 1221: If we’re not exposing `listSearchIndexes`, I’d probably also align this with `getIndexes` here in the SP interface. Passing an array across the interface boundary is probably easier to implement in alternative SP implementations like the java-shell one than passing a (new type of?) Cursor.

packages/service-provider-server/src/cli-service-provider.spec.ts
```diff
@@ -862,6 +862,40 @@ describe('CliServiceProvider', () => {
     });
   });
 
+  describe('#getSearchIndexes', () => {
```
addaleax, line 865: You may want to align the name here with the actual method name :)

packages/service-provider-server/src/cli-service-provider.ts
lerouxb: I just thought we'd ultimately end up with a larger interface? I don't feel too strongly about it, though

packages/shell-api/src/collection.ts
```diff
@@ -2056,4 +2056,66 @@ export default class Collection extends ShellApiWithMongoClass {
...
+      this._name,
+      [description]
+    );
+    return results[0];
+  }
+
+  @serverVersions(['6.0.0', ServerVersions.latest])
+  @returnsPromise
+  @apiVersions([])
+  async createSearchIndexes(descriptions: any[]): Promise<string[]> {
```
addaleax, line 2088:
```suggestion
  async createSearchIndexes(descriptions: Document[]): Promise<string[]> {
```

packages/shell-api/src/collection.ts
```diff
@@ -2056,4 +2056,66 @@ export default class Collection extends ShellApiWithMongoClass {
...
+      this._database._name,
+      this._name,
+      indexName
+    );
+  }
+
+  @serverVersions(['6.0.0', ServerVersions.latest])
+  @returnsPromise
+  @apiVersions([])
+  async updateSearchIndex(indexName: string, description: any): Promise<void> {
```
addaleax, line 2112:
```suggestion
  async updateSearchIndex(indexName: string, description: Document): Promise<void> {
```

packages/shell-api/src/collection.ts
```diff
@@ -2056,4 +2056,66 @@ export default class Collection extends ShellApiWithMongoClass {
...
+      this._database._name,
+      this._name,
+      { ...await this._database._baseOptions(), ...options }
+    );
+  }
+
+  @serverVersions(['6.0.0', ServerVersions.latest])
+  @returnsPromise
+  @apiVersions([])
+  async createSearchIndex(description: any): Promise<string> {
```
addaleax, line 2075:
```suggestion
  async createSearchIndex(description: Document): Promise<string> {
```

packages/shell-api/src/collection.ts
```diff
@@ -2056,4 +2056,66 @@ export default class Collection extends ShellApiWithMongoClass {
       ...options
     });
   }
+
+  @serverVersions(['6.0.0', ServerVersions.latest])
+  @returnsPromise
+  @apiVersions([])
+  async getSearchIndexes(options?: any): Promise<Document[]> {
```
addaleax, line 2063:
```suggestion
  async getSearchIndexes(options?: Document | string): Promise<Document[]> {
```

(I guess? And I guess we’re not doing `(name, options?)` then?)

packages/shell-api/src/collection.ts
lerouxb: I had it as Document, but then deliberately changed it to any because it felt like that's more likely to stand out as something we have to type properly once we can.

lerouxb: See https://github.com/mongodb-js/mongosh/pull/1477/commits/de6038f8582c56198e30dbec9606a8a97739e9d5

addaleax: We’re using `Document` quite extensively here in the shell-api package, and that’s probably a good thing, because these methods are user-facing and it’s not like we could tell the user to really always give us input that exactly matches the driver types

addaleax (approved):

PR #1483, opened 2023-06-06
feat(shell-api): align search index helper syntax with the syntax document

lerouxb:
I only discovered this document _after_ I added the search index helpers https://docs.google.com/document/d/1rCU7vLmMEn-87SuSZ3xnnSpnkwoWpekyOl15NPIqTJk/edit#heading=h.lf3td4t22dg7

### DIFF
--- a/packages/service-provider-server/src/cli-service-provider.ts
+++ b/packages/service-provider-server/src/cli-service-provider.ts
@@ -1226,14 +1226,10 @@ class CliServiceProvider extends ServiceProviderCore implements ServiceProvider 
     // TODO(MONGOSH-1471): use SearchIndexDescription[] once available
     specs: {name: string, definition: Document}[],
     dbOptions?: DbOptions): Promise<string[]> {
-    const driverSpecs = specs.map(({ name, definition }) => {
-      // NOTE: Unlike in our docs or in the command, the driver calls it description, not definition
-      return { name, description: { ...definition } };
-    });
     return this.db(database, dbOptions)
       .collection(collection)
       // @ts-expect-error still @internal
-      .createSearchIndexes(driverSpecs);
+      .createSearchIndexes(specs);
   }
 
   dropSearchIndex(
--- a/packages/service-provider-server/src/cli-service-provider.spec.ts
+++ b/packages/service-provider-server/src/cli-service-provider.spec.ts
@@ -920,8 +920,8 @@ describe('CliServiceProvider', () => {
 
     beforeEach(() => {
       descriptions = [
-        { name: 'foo', description: {} },
-        { name: 'bar', description: {} }
+        { name: 'foo', definition: {} },
+        { name: 'bar', definition: {} }
       ];
 
       nativeMethodResult = [


packages/shell-api/src/collection.ts
```diff
@@ -85,6 +85,14 @@ type CollStatsShellOptions = CollStatsOptions & {
   indexDetailsName?: string;
 };
 
+function optionalFirstString(arg1?: string|Document, arg2?: Document): [arg1: null|string, arg2: null|Document] {
```
lerouxb, line 88: Is this gross? I can't find any helpers where we have the first param optional. ie. the `function(name, options)` OR `function(options)` pattern. The driver does it extensively, but our linting rules really don't like the way they do it.

lerouxb: Also is there a better place to put this kind of helper or is it OK here until we need it elsewhere?

addaleax: https://github.com/mongodb-js/mongosh/blob/9b3d27adac4d425d2b892b79e00e0988c2b11fb2/packages/shell-api/src/shard.ts#L116-L119 is the typical way to do this in JS, and I’d also inline this into the API methods for this case too

packages/shell-api/src/collection.ts
```diff
@@ -2074,12 +2080,17 @@ export default class Collection extends ShellApiWithMongoClass {
...
+      description = indexName;
+      indexName = undefined;
+    }
+
+    this._emitCollectionApiCall('createSearchIndex', { indexName, description });
     const results = await this._mongo._serviceProvider.createSearchIndexes(
       this._database._name,
       this._name,
-      [description]
+      [{ name: (indexName as string|undefined) ?? 'default', description: { ...description } }]
```
lerouxb, line 2093: The doc says that with no name it should go with `'default'`. Not clear if we should do it, the driver should do it or the server should do it. I can't see the driver doing it and since there's no way for me to test what the server does I'm just filling in the default here.

packages/service-provider-server/src/cli-service-provider.ts
```diff
@@ -1210,25 +1210,30 @@ class CliServiceProvider extends ServiceProviderCore implements ServiceProvider
...
 
   createSearchIndexes(
     database: string,
     collection: string,
     // TODO(MONGOSH-1471): use SearchIndexDescription[] once available
-    descriptions: any[],
+    specs: {name: string, definition: Document}[],
     dbOptions?: DbOptions): Promise<string[]> {
+    const driverSpecs = specs.map(({ name, definition }) => {
+      // NOTE: Unlike in our docs or in the command, the driver calls it description, not definition
```
lerouxb, line 1230: https://github.com/mongodb/node-mongodb-native/pull/3672/files#diff-7ce3e491e925915c35caff51955ce54f08a866b6a7c51e7c3daa2c0c6b0a848dR15

lerouxb: Actually I'm fairly sure that's a bug on the driver. Will ping them.

addaleax: Yeah, it’s `definition`: https://github.com/10gen/mongo-enterprise-modules/blob/adaa6e278c12045b70c70d53a3b6bb957472f2ab/src/search/search_index_commands.idl#L57

I think in the driver that doesn’t really affect anything, though, it’s just variable naming

addaleax (approved):

PR #1489, opened 2023-06-08
feat(shell-api): add shell helpers for checkMetadataConsistency MONGOSH-1432

lerouxb:


packages/shell-api/src/abstract-cursor.ts
```diff
@@ -149,19 +150,19 @@ export abstract class AbstractCursor<CursorType extends ServiceProviderAggregati
 
   @returnType('this')
   projection(spec: Document): this {
-    this._cursor.project(spec);
+    (this._cursor as any).project(spec);
```
addaleax, line 153: I’d strongly suggest to do this in proper typing, i.e.

```ts
export abstract class AbstractCursor<CursorType extends ServiceProviderAggregationCursor | ServiceProviderCursor | ServiceProviderRunCommandCursor> extends ShellApiWithMongoClass {
}
...
export abstract class AggregateOrFindCursor<CursorType extends ServiceProviderAggregationCursor | ServiceProviderCursor> extends AbstractCursor<CursorType> {
}
...
export default class Cursor extends AggregateOrFindCursor<ServiceProviderCursor> {
}
...
export default class AggregationCursor extends AggregateOrFindCursor<ServiceProviderAggregationCursor> {
}
```

`as any` is meant to be a promise that the typing is actually correct and Typescript just doesn’t know/can’t tell, but here Typescript is absolutely right to complain, because users _can’t_ call `.projection()` on this type of cursor

addaleax (approved):

PR #1491, opened 2023-06-15
fix(deps): do not emit max listener count warning when reconnecting MONGOSH-1479

addaleax:


Anemy (approved):
lgtm! Nice regression test.

PR #1502, opened 2023-06-30
chore: add depcheckrcs, move dependencies to packages MONGOSH-1484

addaleax:


package.json
```diff
@@ -91,72 +91,41 @@
...
-    "browserify": "^16.5.0",
     "chai": "^4.2.0",
-    "command-exists": "^1.2.9",
     "cross-env": "^6.0.3",
-    "depcheck": "^1.4.1",
-    "download": "^8.0.0",
     "duplexpair": "^1.0.1",
     "eslint": "^7.31.0",
     "eslint-config-mongodb-js": "^5.0.3",
     "eslint-plugin-mocha": "^9.0.0",
```
addaleax, line 108: I didn’t really want to touch the linting config here to avoid conflicts with @lerouxb’s ongoing work, we can follow up afterwards

package.json
```diff
@@ -91,72 +91,41 @@
...
-    "js-yaml": "^3.13.1",
-    "karma": "^6.3.4",
-    "karma-chrome-launcher": "^3.1.0",
-    "karma-cli": "^2.0.0",
-    "karma-mocha": "^2.0.1",
-    "karma-mocha-reporter": "^2.2.5",
-    "karma-typescript": "^5.5.1",
     "lerna": "^4.0.0",
     "mocha": "^10.2.0",
     "mongodb": "^5.6.0",
```
addaleax, line 112: This is used by the scripts in `testing/`, we may want to turn this into a proper package (but preferably wait with that until the mongodb-runner work is done?)

packages/cli-repl/.depcheckrc
```diff
@@ -1 +1,10 @@
-ignores: ['lodash', 'moment']
+ignores:
+ - '@typescript-eslint/parser'
+ - '@typescript-eslint/eslint-plugin'
+ - chai
+ - sinon-chai
+ - sinon
+ - ts-sinon
```
addaleax, line 7: Opened https://jira.mongodb.org/browse/MONGOSH-1501 to at least deduplicate these two

packages/cli-repl/package.json
```diff
@@ -95,9 +96,14 @@
...
     "@types/yargs-parser": "^15.0.0",
     "chai-as-promised": "^7.1.1",
+    "depcheck": "^1.4.3",
     "lodash": "^4.17.21",
     "moment": "^2.29.1",
-    "mongodb-crypt-library-dummy": "^1.0.2"
+    "mongodb": "^5.6.0",
+    "mongodb-crypt-library-dummy": "^1.0.2",
+    "node-fetch": "^2.6.1",
+    "rimraf": "^5.0.1",
```
addaleax, line 105: Opened https://jira.mongodb.org/browse/MONGOSH-1500 to get rid of `rimraf` where we can use `fs.rm` instead

packages/shell-api/src/bulk.spec.ts
```diff
@@ -73,7 +72,7 @@ describe('Bulk API', () => {
           const bulkWriteResult = {
             ok: 1,
             nInserted: 1,
-            insertedIds: { 0: new ObjectId() },
+            insertedIds: { 0: new bson.ObjectId() },
```
addaleax, line 75: I was hoping to avoid source changes entirely, I hope these don’t conflict with the linter update :crossed_fingers:

lerouxb (approved):

PR #1504, opened 2023-07-05
chore: use shared precommit hook and prettier MONGOSH-1482

lerouxb:
* Just like in compass we'll now run prettier on all the changed files when committing. Via a precommit hook provided by husky.
* The shared prettier config is now used everywhere.
* lint runs eslint and prettier check
* reformat runs prettier and eslint fix
* prettier reformatted all our code 😬


addaleax (approved):
:rocket:

Not reviewing the second commit here, I think that's okay :)

lerouxb:
I messed up something in the package-lock.json and could really do this better, so I'm opening a new PR in place.

PR #1506, opened 2023-07-07
chore: use depalign MONGOSH-1485

lerouxb:


.depalignrc.json
```diff
@@ -0,0 +1,13 @@
+{
+  "ignore": {
+    "@babel/core": [
```
lerouxb, line 3: I just assumed these are intentional for now. Should these match?

Anemy: Good to match things and reduce this when we can I think. Not sure on if they were intentional though, maybe someone else knows.

Anemy (approved):

PR #1507, opened 2023-07-10
chore: use @mongodb-js/eslint-config-devtools@0.9.7, fix eslint command args MONGOSH-1505

lerouxb:
* Bump @mongodb-js/eslint-config-devtools
* use common.jsRules and common.jsxRules
* ignore all the noconsole and eqeqeq errors
* fix the typescript test overrides (it was accidentally also applied to .js files)
* noticed that the eslint command wasn't running in the right folder. broke it at the last minute last time when I fixed it for cygwyn. oops. also fixed all the errors this hilighted
* now properly fixed linting under cygwin in our CI
* split the eslint and ts configs out into packages
* aligned our per-package lint config more closely with compass


packages/arg-parser/package.json
```diff
@@ -14,7 +14,7 @@
     "test": "mocha -r \"../../scripts/import-expansions.js\" --timeout 60000 -r ts-node/register \"./{src,lib}/**/*.spec.ts\"",
     "test-ci": "node ../../scripts/run-if-package-requested.js npm test",
     "eslint": "eslint",
-    "lint": "npm run eslint && npm run prettier -- --check .",
+    "lint": "npm run eslint . && npm run prettier -- --check .",
```
lerouxb, line 17: oops

packages/errors/tsconfig-lint.json
```diff
@@ -1,5 +1,5 @@
 {
   "extends": "./tsconfig.json",
-  "include": ["./src/**/*", "./test/**/*"],
+  "include": ["./src/**/*", "./scripts/**/*", "./test/**/*"],
```
lerouxb, line 3: This kind of thing meant there were lint errors at the top of those files. Might also have been the real problem behind what I saw on windows in CI. Had to include a few folders here and there to fix it.

packages/node-runtime-worker-thread/.eslintignore
```diff
@@ -1,2 +1,3 @@
 node_modules
-lib/
\ No newline at end of file
+lib/
+dist/
```
lerouxb, line 3: I have no idea how things worked without this before.

addaleax (approved):
Looks good!

packages/build/src/homebrew/generate-formula.ts
```diff
@@ -33,6 +33,8 @@ export async function generateUpdatedFormula(
   if (
     currentVersion &&
     semver.compare(currentVersion, context.version, {
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore 'includePrerelease' does not exist in type 'Options'.
```
lerouxb, line 37: So.. I move the eslint config to a package and then suddenly this ts compile error appears.

PR #1508, opened 2023-07-11
chore: bump typescript, @mongodb-js/eslint-config-devtools MONGOSH-1503 MONGOSH-1516

lerouxb:
* Use the same version of typescript as in compass
* Use the same version of eslint-config-devtools (and eslint plugins) as in compass.
* This also (finally) gets rid of the eslint plugin deps at the top level


packages/cli-repl/src/format-output.ts
```diff
@@ -298,7 +298,9 @@ export function formatError(error: Error, options: FormatOptions): string {
 
 function removeUndefinedValues<T>(obj: T) {
```
addaleax, line 299: Unrelated, but should this be:

```suggestion
function removeUndefinedValues<T>(obj: T): Partial<T> {
```

?

addaleax (approved):

packages/cli-repl/src/format-output.ts
lerouxb: Screenshot 2023-07-11 at 19 28 23

packages/cli-repl/src/async-repl.ts
```diff
@@ -83,7 +83,8 @@ export function start(opts: AsyncREPLOptions): REPLServer {
   const repl = (opts.start ?? originalStart)(opts);
   const originalEval = promisify(
     wrapPauseInput(repl.input, wrapNoSyncDomainError(repl.eval.bind(repl)))
-  );
+    // string, Context, string is not string, any, string
+  ) as (arg1: string, arg2: any, arg3: string) => Promise<any>;
```
lerouxb, line 87: Difference between our test ts config and commong ts config?

packages/cli-repl/src/format-output.ts
lerouxb: I suppose I can fix that in turn..

PR #1509, opened 2023-07-12
chore: locks semver 5.7.x to 5.7.2, 6.3.x to 6.3.1 and 7.5.x to 7.5.4 - MONGOSH-1494

himanshusinghs:
This PR addresses a reported Snyk security vulnerability ([SNYK-JS-SEMVER-3247795](https://security.snyk.io/vuln/SNYK-JS-SEMVER-3247795)) introduced by semver < 7.5.2

semver backported the [fix applied in 7.5.2](https://github.com/npm/node-semver/pull/564) for the above mentioned vulnerability also to v5.x (in [5.7.2](https://github.com/npm/node-semver/releases/tag/v5.7.2)) and to v6.x (in [6.3.1](https://github.com/npm/node-semver/releases/tag/v6.3.1)). Since these versions were released <30 hours ago, the snyk advisory hasn't catch up to the updated info which is why we are disabling this particular vulnerability in our snyk policy for a month.

PR #1510, opened 2023-07-12
chore: update semver and babel MONGOSH-1494

mcasimir:


addaleax:
https://github.com/mongodb-js/mongosh/pull/1509 seems to cover part of this as well

mcasimir:
If possible it would be nice to update, not urgent though. This version of babel doesn't depend on semver at all, and we can start to move to latest semver everywhere.

FYI: @himanshusinghs

PR #1513, opened 2023-07-12
chore: migrate towards mongodb-runner MONGOSH-1490

addaleax:


packages/service-provider-server/src/cli-service-provider.integration.spec.ts
```diff
@@ -798,7 +798,7 @@ describe('CliServiceProvider [integration]', function () {
 
   describe('#getURI', function () {
     it('returns the current URI', function () {
-      expect(serviceProvider.getURI()).to.equal(connectionString + '/');
+      expect(serviceProvider.getURI()).to.equal(connectionString);
```
himanshusinghs, line 801: There was no change in ServiceProvider, will this not have a trailing slash anymore?

packages/shell-api/src/field-level-encryption.spec.ts
```diff
@@ -799,7 +799,7 @@ describe('Field Level Encryption', function () {
 
     beforeEach(async function () {
       dbname = `test_fle_${Date.now()}`;
-      uri = `${await testServer.connectionString()}/${dbname}`;
+      uri = await testServer.connectionString();
```
himanshusinghs, line 802:
```suggestion
      uri = await testServer.connectionString({}, { pathname: `/${dbname}` });
```
Should it have dbname as pathname here, like in other places?

packages/service-provider-server/src/cli-service-provider.integration.spec.ts
addaleax: It’s not the `ServiceProvider` method that changed here, it’s that `connectionString` variable in the test itself now includes the trailing slash (which I think is technically required by the connection string spec, so I’d consider it a bug fix either way)

packages/shell-api/src/field-level-encryption.spec.ts
addaleax: We’re not making use of the fact that `dbname` is used here, so I dropped it. No strong feelings, happy to add it back if you prefer. (It’s added in the e2e tests because for the CLI, the db name after the slash indicates the database that the shell is initially connected to.)

packages/service-provider-server/src/cli-service-provider.integration.spec.ts
himanshusinghs: Ahh got it. Thanks for explaining 👍

himanshusinghs (approved):
The new mongodb runner setup is amazing!🌟

addaleax:
Going to merge this since the coverage problem should be more broadly addressed by #1514

PR #1516, opened 2023-07-19
chore: fixes build failures for s390x and ppc64le archs

himanshusinghs:
Somehow in the [following commit](https://github.com/mongodb-js/mongosh/pull/1503/commits/74b4834086712754bbe08e276eccbfeb6e6c968e#diff-3f8569b068ef11ea8f1fc5b5682d8ba0c6f494982ca211e549ae2156805f3561R72-R75), mongodb-client-encryption as well as kerberos ended up being `devDependencies` in addition to being `optionalDependencies`.

I looked back in commits before this one and concluded that these have always been `optionalDependencies` so this might have been just a mistake. It can be seen from the [base build of this failed build](https://spruce.mongodb.com/version/64a41f0d5623435ea6d2a7b0/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC&variant=linux_s390x_build) that we always had [the same failure](https://parsley.mongodb.com/evergreen/mongosh_linux_s390x_build_compile_artifact_817467053c0cb21f08e4bd62cd7794be51ae00f0_23_07_03_15_12_54/0/task?bookmarks=0,30277&shareLine=16909) for installing `mongodb-client-encryption` on s390x and ppc64le archs but since they were `optionalDependencies`, our bootstrap did not fail until this change (which marked them non-optional).

### DIFF
--- a/.evergreen/install-node.sh
+++ b/.evergreen/install-node.sh
@@ -43,7 +43,9 @@ else
   fi
 
   npm cache clear --force || true # Try to work around `Cannot read property 'pickAlgorithm' of null` errors in CI
-  npm i -g npm@8.x
+  # Started observing CI failures on RHEL 7.2 (s390x) for installing npm, all
+  # related to network issues hence adding a retry with backoff here.
+  bash "$BASEDIR/retry-with-backoff.sh" npm i -g npm@8.x
 fi
 
 . "$BASEDIR/setup-env.sh"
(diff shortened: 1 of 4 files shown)


lerouxb:
Not 100% sure if this is a good idea, because `packages/service-provider-core/src/all-fle-types.ts` gets imported regardless and that imports 'mongodb-client-encryption'. In both cases as types. So we probably _do_ need them as dev deps? I'm no expert here :)

PR #1519, opened 2023-07-20
feat: add atlas streams support MONGOSH-1514

shaketbaby:
This PR adds Atlas Stream Processing support to the mongosh:
- adds a new flag `is_stream` to the `ConnectInfo` type
- changes default prompt to `AtlasStreamProcessing> ` when `is_stream` is `true`
- adds a new global object `sp` for doing streams related operations

### DIFF
--- a/packages/service-provider-core/src/connect-info.ts
+++ b/packages/service-provider-core/src/connect-info.ts
@@ -55,22 +55,10 @@ export default function getConnectInfo(
     is_enterprise: getBuildInfo.isEnterprise(buildInfo),
     auth_type,
     is_data_federation,
-    is_stream: isStream(uri),
+    is_stream: getBuildInfo.isAtlasStream(uri),
     dl_version,
     atlas_version: atlasVersion?.atlasVersion ?? null,
     is_genuine,
     non_genuine_server_name,
   };
-}
-
-function isStream(uri: string): boolean {
-  try {
-    return (
-      getBuildInfo.isAtlas(uri) &&
-      !!/^atlas-stream-.+/.exec(new URL(uri).hostname)
-    );
-  } catch {
-    // invalid uri
-  }
-  return false;
 }
--- a/packages/service-provider-core/src/connect-info.ts
+++ b/packages/service-provider-core/src/connect-info.ts
@@ -13,6 +13,7 @@ export interface ConnectInfo {
   is_enterprise: boolean;
   auth_type?: string;
   is_data_federation: boolean;
+  is_stream: boolean;
   dl_version?: string;
   atlas_version?: string;
   is_genuine: boolean;
@@ -54,9 +55,22 @@ export default function getConnectInfo(
     is_enterprise: getBuildInfo.isEnterprise(buildInfo),
     auth_type,
     is_data_federation,
+    is_stream: isStream(uri),
     dl_version,
     atlas_version: atlasVersion?.atlasVersion ?? null,
     is_genuine,
     non_genuine_server_name,
   };
+}
+
+function isStream(uri: string): boolean {
+  try {
+    return (
+      getBuildInfo.isAtlas(uri) &&
+      !!/^atlas-stream-.+/.exec(new URL(uri).hostname)
+    );
+  } catch {
+    // invalid uri
+  }
+  return false;
 }
--- a/packages/shell-api/src/shell-instance-state.ts
+++ b/packages/shell-api/src/shell-instance-state.ts
@@ -33,6 +33,7 @@ import { TransformMongoErrorPlugin } from './mongo-errors';
 import NoDatabase from './no-db';
 import type { ShellBson } from './shell-bson';
 import constructShellBson from './shell-bson';
+import { Streams } from './streams';
 
 /**
  * The subset of CLI options that is relevant for the shell API's behavior itself.
@@ -299,6 +300,11 @@ export default class ShellInstanceState {
         configurable: true,
         set: setFunc,
         get: () => this.currentDb,
+      });
+      // add the global "sp" for stream processing
+      const streams = Streams.newInstance(this);
+      Object.defineProperty(contextObject, 'sp', {
+        get: () => streams,
       });
     }
 
--- a/packages/shell-api/src/shell-instance-state.ts
+++ b/packages/shell-api/src/shell-instance-state.ts
@@ -227,6 +227,7 @@ export default class ShellInstanceState {
     this.currentDb = newDb;
     this.context.rs = new ReplicaSet(this.currentDb);
     this.context.sh = new Shard(this.currentDb);
+    this.context.sp = Streams.newInstance(this.currentDb);
     this.fetchConnectionInfo().catch((err) =>
       this.messageBus.emit('mongosh:error', err, 'shell-api')
     );
@@ -282,6 +283,7 @@ export default class ShellInstanceState {
 
     contextObject.rs = new ReplicaSet(this.currentDb);
     contextObject.sh = new Shard(this.currentDb);
+    contextObject.sp = Streams.newInstance(this.currentDb);
 
     const setFunc = (newDb: any): Database => {
       if (getShellApiType(newDb) !== 'Database') {
@@ -300,11 +302,6 @@ export default class ShellInstanceState {
         configurable: true,
         set: setFunc,
         get: () => this.currentDb,
-      });
-      // add the global "sp" for stream processing
-      const streams = Streams.newInstance(this);
-      Object.defineProperty(contextObject, 'sp', {
-        get: () => streams,
       });
     }
 
--- a/packages/shell-api/src/shell-instance-state.ts
+++ b/packages/shell-api/src/shell-instance-state.ts
@@ -481,6 +481,10 @@ export default class ShellInstanceState {
 
   // eslint-disable-next-line @typescript-eslint/require-await
   async getDefaultPrompt(): Promise<string> {
+    if (this.connectionInfo?.extraInfo?.is_stream) {
+      return 'AtlasStreamProcessing> ';
+    }
+
     const prefix = this.getDefaultPromptPrefix();
     const topologyInfo = this.getTopologySpecificPrompt();
     let dbname = '';
--- a/packages/shell-api/src/shell-instance-state.ts
+++ b/packages/shell-api/src/shell-instance-state.ts
@@ -437,7 +437,7 @@ export default class ShellInstanceState {
   }
 
   async onInterruptExecution(): Promise<boolean> {
-    this.interrupted.set();
+    await this.interrupted.set();
     this.currentCursor = null;
 
     this.resumeMongosAfterInterrupt = await Promise.all(
--- a/packages/shell-api/src/stream-processor.ts
+++ b/packages/shell-api/src/stream-processor.ts
@@ -71,7 +71,7 @@ export default class StreamProcessor extends ShellApiWithMongoClass {
   async _sampleFrom(cursorId: number) {
     let currentCursorId = cursorId;
     // keep pulling until end of stream
-    while (currentCursorId !== 0) {
+    while (String(currentCursorId) !== '0') {
       const res = await this._streams._runStreamCommand({
         getMoreSampleStreamProcessor: this.name,
         cursorId: currentCursorId,
--- a/packages/shell-api/src/stream-processor.ts
+++ b/packages/shell-api/src/stream-processor.ts
@@ -1,5 +1,3 @@
-import { promisify } from 'util';
-
 import type { Document } from '@mongosh/service-provider-core';
 
 import type Mongo from './mongo';
@@ -11,8 +9,6 @@ import {
 } from './decorators';
 
 import type { Streams } from './streams';
-
-const sleep = promisify(setTimeout);
 
 @shellApiClassDefault
 export default class StreamProcessor extends ShellApiWithMongoClass {
@@ -30,31 +26,28 @@ export default class StreamProcessor extends ShellApiWithMongoClass {
 
   @returnsPromise
   async start() {
-    const result = await this._streams._runCommand({
+    return await this._streams._runStreamCommand({
       startStreamProcessor: this.name,
     });
-    return result;
   }
 
   @returnsPromise
   async stop() {
-    const result = await this._streams._runCommand({
+    return await this._streams._runStreamCommand({
       stopStreamProcessor: this.name,
     });
-    return result;
   }
 
   @returnsPromise
   async drop() {
-    const result = await this._streams._runCommand({
+    return await this._streams._runStreamCommand({
       dropStreamProcessor: this.name,
     });
-    return result;
   }
 
   @returnsPromise
   async stats(options: Document = {}) {
-    return this._streams._runCommand({
+    return this._streams._runStreamCommand({
       getStreamProcessorStats: this.name,
       ...options,
     });
@@ -62,7 +55,7 @@ export default class StreamProcessor extends ShellApiWithMongoClass {
 
   @returnsPromise
   async sample(options: Document = {}) {
-    const r = await this._streams._runCommand({
+    const r = await this._streams._runStreamCommand({
       startSampleStreamProcessor: this.name,
       ...options,
     });
@@ -79,7 +72,7 @@ export default class StreamProcessor extends ShellApiWithMongoClass {
     let currentCursorId = cursorId;
     // keep pulling until end of stream
     while (currentCursorId !== 0) {
-      const res = await this._streams._runCommand({
+      const res = await this._streams._runStreamCommand({
         getMoreSampleStreamProcessor: this.name,
         cursorID: currentCursorId,
       });
@@ -100,7 +93,7 @@ export default class StreamProcessor extends ShellApiWithMongoClass {
         const interruptable = this._instanceState.interrupted.asPromise();
         try {
           await Promise.race([
-            sleep(1000), // wait 1 second
+            this._instanceState.shellApi.sleep(1000), // wait 1 second
             interruptable.promise, // unless interruppted
           ]);
         } finally {
--- a/packages/shell-api/src/stream-processor.ts
+++ b/packages/shell-api/src/stream-processor.ts
@@ -74,7 +74,7 @@ export default class StreamProcessor extends ShellApiWithMongoClass {
     while (currentCursorId !== 0) {
       const res = await this._streams._runStreamCommand({
         getMoreSampleStreamProcessor: this.name,
-        cursorID: currentCursorId,
+        cursorId: currentCursorId,
       });
 
       if (res.ok !== 1) {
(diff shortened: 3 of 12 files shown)


packages/service-provider-core/src/connect-info.ts
```diff
@@ -57,9 +58,22 @@ export default function getConnectInfo(
...
     is_data_federation,
+    is_stream: isStream(uri),
     dl_version,
     atlas_version: atlasVersion?.atlasVersion ?? null,
     is_genuine,
     non_genuine_server_name,
   };
 }
+
+function isStream(uri: string): boolean {
```
mcasimir, line 69: NIT: if you have the time for one additional PR: it would be awesome to have this function in https://github.com/mongodb-js/mongodb-build-info where we have all the other similar types of detection, so we will have it also for the other tools.

packages/shell-api/src/shell-instance-state.ts
```diff
@@ -481,6 +487,10 @@ export default class ShellInstanceState {
 
   // eslint-disable-next-line @typescript-eslint/require-await
   async getDefaultPrompt(): Promise<string> {
+    if (this.connectionInfo?.extraInfo?.is_stream) {
```
mcasimir, line 490: Aren't the other parts of the prompt relevant here? At least the dbName .. or in streams `use somedb` should just be disabled?

packages/shell-api/src/stream-processor.ts
```diff
@@ -0,0 +1,121 @@
...
+      if (res.ok !== 1) {
+        return res;
+      }
+
+      currentCursorId = res.cursorId;
+
+      // print fetched documents
+      for (const doc of res.messages) {
+        // eslint-disable-next-line no-console
+        console.log(doc);
```
mcasimir, line 102: Unless this is a WIP, you may want to use something like `this._instanceState.shellApi.print(...)` instead, which is wired with the presentation layer of the REPL.

As far as i know `console.log` wouldn't work when the shell is embedded in Compass, VSCode, Jetbrains etc .. and could probably have unwanted side effects for the CLI too, like not applying user preferences for the output .. although i'm not sure about that.

packages/shell-api/src/stream-processor.ts
```diff
@@ -0,0 +1,121 @@
...
+    });
+
+    if (r.ok !== 1) {
+      return r;
+    }
+
+    return this._sampleFrom(r.cursorId as number);
+  }
+
+  @returnsPromise
```
mcasimir, line 82: I may be misremembering, but if this is a private method that is not meant to be used by users directly you shouldn't need to annotate it.

`@returnsPromise` marks the method such that when a user uses it in the REPL it gets rewritten to be "auto-awaited" and it appears as if it would be synchronous, for example, by marking`findOne` with `@returnsPromise` users are able to do this:

```
db.coll.findOne().x
```

packages/service-provider-core/src/connect-info.ts
shaketbaby: That did cross my mind. Happy to move this to `mongodb-build-info`. Will raise a PR there.

packages/shell-api/src/shell-instance-state.ts
shaketbaby: Database isn't really relevant for streams instance. There isn't any database. We removed the database name from the connection URL for the same reason. Good question about `use samedb`, I am not actually sure. Will need to take it up with the PMs. Maybe the global `db` should not be added even.

packages/shell-api/src/stream-processor.ts
shaketbaby: Good call. Didn't know about `this._instanceState.shellApi.print(...)`, will update to use that.

packages/shell-api/src/stream-processor.ts
shaketbaby: Yes, I added this because it's available and can be called with `sp.someStream._sampleFrom()` even though its name indicates this is private and should not be used directly. I don't mind removing it.

packages/shell-api/src/shell-instance-state.ts
mcasimir: yeah good point about `db`, please loop in also the PMs of the shell

packages/shell-api/src/stream-processor.ts
mcasimir: Ah ok, there should be no harm into keeping it either 🤷‍♂️

packages/service-provider-core/src/connect-info.ts
shaketbaby: Have raised PR https://github.com/mongodb-js/mongodb-build-info/pull/23.

packages/shell-api/src/stream-processor.ts
shaketbaby: Have changed to use `printjson`.
```ts
      for (const doc of res.messages) {
        await this._instanceState.shellApi.printjson(doc);
      }
```

packages/shell-api/src/streams.ts
```diff
@@ -0,0 +1,172 @@
...
+@shellApiClassDefault
+export class Streams extends ShellApiWithMongoClass {
+  public static newInstance(instanceState: ShellInstanceState) {
+    return new Proxy(new Streams(instanceState), {
+      get(target, prop) {
+        const v = (target as any)[prop];
+        if (v !== undefined) {
+          return v;
+        }
+        if (typeof prop === 'string') {
```
addaleax, line 24: Can we add a `prop.startsWith('_')` check here similar to the ones for Database/Collection? That gives us the ability to add underscore-prefixed private properties without a chance of conflicts with user code.

packages/shell-api/src/streams.ts
```diff
@@ -0,0 +1,172 @@
...
+    try {
+      await sp._sampleFrom(cursorId);
+    } catch (err) {
+      // try to stop and drop the temp processor when interrupted
+      if (err instanceof MongoshInterruptedError) {
+        this._instanceState.messageBus.once(
+          'mongosh:interrupt-complete',
+          stopAndDrop
+        );
+      }
```
addaleax, line 90: As the documentation for `interrupt-complete` notes, it’s a testing-only event. Why do we need to wait for it here? Why do we not call `stopAndDrop()` for other errors? Should this just be a 

```js
try { 
  await sp._sampleFrom(cursorId);
} finally { 
 try { 
    await sp.stop();
    await sp.drop();
  } catch { /* ignore */ }
}
```

?

packages/shell-api/src/streams.ts
```diff
@@ -0,0 +1,172 @@
...
+    }
+    const processors = result.streamProcessors;
+    const sps = processors.map((sp: StreamProcessor) =>
+      this.getProcessor(sp.name)
+    );
+    return Object.defineProperty(sps, asPrintable, {
+      value() {
+        return JSON.stringify(processors, null, 2);
+      },
+    });
```
addaleax, line 148: Should this just be

```suggestion
    (sps as any)[asPrintable] = processors;
```

? Why the `JSON.stringify()` here?

packages/service-provider-core/src/connect-info.ts
```diff
@@ -57,6 +58,7 @@ export default function getConnectInfo(
     is_enterprise: getBuildInfo.isEnterprise(buildInfo),
     auth_type,
     is_data_federation,
+    is_stream: getBuildInfo.isAtlasStream(uri),
```
addaleax, line 61: Is there a more reliable check than a URL check? We only do URL checks because there’s nothing we can easily get programmatically out of `buildInfo` etc., but here it seems like we have the opportunity to actually add a proper server-side API for unambiguously indicating a streams processing environment

packages/shell-api/src/streams.ts
shaketbaby: This is guarded by the previous if statement. Things on the `target` itself take precedence.

packages/shell-api/src/streams.ts
shaketbaby: For interrupted error specific, from memory, I think the call will always fail as the underlying connection is closed. So we need to wait until the interruption is handled. 
For other errors, I wasn't sure what other errors could occur. Only error I can think of is network issue. In which case calling the method will also fail. So, I was just leaving that for server to handle. But having said that, we can always call the method on error as you suggested.

packages/shell-api/src/streams.ts
shaketbaby: That was added so that the processor objects are printed in full. By default, deeply nested objects are not printed.

packages/shell-api/src/streams.ts
```diff
@@ -0,0 +1,172 @@
...
+
+  @returnsPromise
+  listConnections(filter: Document) {
+    return this._runCommand({
+      listStreamConnections: 1,
+      filter,
+    });
+  }
+
+  async _runCommand(cmd: Document, options: Document = {}) {
```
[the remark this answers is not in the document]
shaketbaby: Sure, can rename it to `_runStreamsCommand`. Correct me if I am wrong, but from my experience, the normal `runCommand` doesn't seem to handle the interruption, CTRL + C. I remember noticed something like that when working on sample command for the NY.local keynote demo.

packages/shell-api/src/streams.ts
```diff
@@ -0,0 +1,172 @@
...
+          return v;
+        }
+        if (typeof prop === 'string') {
+          return target.getProcessor(prop);
+        }
+      },
+    });
+  }
+
+  constructor(private instanceState: ShellInstanceState) {
```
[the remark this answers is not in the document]
shaketbaby: Happy to change to the explicit form. And no particular reason, this is just the way I found that could get the current `Mongo`. The global `sp` instance is created inside `ShellInstanceState` so I just pass it in directly. Why go through a `Database` when we can pass in the `ShellInstanceState` directly?

packages/shell-api/src/stream-processor.ts
```diff
@@ -0,0 +1,114 @@
...
+      for (const doc of res.messages) {
+        await this._instanceState.shellApi.printjson(doc);
+      }
+
+      // wait before pulling again if no result in this batch
+      if (!res.messages.length) {
+        const interruptable = this._instanceState.interrupted.asPromise();
+        try {
+          await Promise.race([
+            sleep(1000), // wait 1 second
```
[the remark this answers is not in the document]
shaketbaby: Cool, didn't notice `shellApi` has a `sleep` method, will change to that.

packages/shell-api/src/stream-processor.ts
```diff
@@ -0,0 +1,114 @@
...
+  [asPrintable]() {
+    return `Atlas Stream Processor: ${this.name}`;
+  }
+
+  @returnsPromise
+  async start() {
+    const result = await this._streams._runCommand({
+      startStreamProcessor: this.name,
+    });
+    return result;
```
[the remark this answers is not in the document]
shaketbaby: Both do the same thing. But with a local variable, I can easily check the content of `result` when debugging. But can change if that's the standard/preferred way of doing this.

packages/shell-api/src/streams.ts
```diff
@@ -0,0 +1,172 @@
...
+    );
+    return Object.defineProperty(sps, asPrintable, {
+      value() {
+        return JSON.stringify(processors, null, 2);
+      },
+    });
+  }
+
+  @returnsPromise
+  listConnections(filter: Document) {
```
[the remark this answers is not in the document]
shaketbaby: Ok, so `@returnsPromise` can only be used on `async` function, regardless of whether the function returns a promise or not. Noted.

packages/service-provider-core/src/connect-info.ts
shaketbaby: Right now, there isn't. Maybe can potentially change ADF to add something in the response of `buildInfo` to indicate this is a stream instance.

packages/shell-api/src/streams.ts
addaleax: Not quite, though. The reason that Database & Collection have these properties is future-proofing: Right now, `sp._foo` would return the `_foo` stream processor, but if we ever add a `_foo` private property, that would break user code which relies on the previous behavior. If we do not allow accessing underscore-prefixed stream processors this way from the beginning, that won’t happen.

packages/shell-api/src/streams.ts
addaleax: `JSON.stringify()` is not the way to solve that problem, though … you could set a specific result type here (`sps[shellApiType] = '...'`) and then add it to the list here: https://github.com/mongodb-js/mongosh/blob/d5530affb0ccaace873c59f4ba6092df34f96fa8/packages/cli-repl/src/format-output.ts#L128

packages/shell-api/src/streams.ts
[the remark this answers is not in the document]
addaleax: Right, the normal `runCommand` doesn’t interact with interruptions differently than other commands, so I’d suggest doing something to make clear that it’s a different thing. (I think your answer above re: the stopAndDrop() timing answers my second question here)

packages/shell-api/src/streams.ts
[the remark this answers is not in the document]
addaleax: If you want to change all of these classes to take a `Mongo` instance instead of a `Database` instance in this PR, that would be fine with me, it’s semantically the “more correct thing” to do anyway. (Or otherwise, we can also try to do that separately.)

But I’d still try to keep consistent behavior here. Passing `ShellInstanceState` is not quite the same as passing a `Mongo` (or `Database`) object; consider this:

```
db = connect('mongodb://host1');
rs1 = rs;
sp1 = sp;
db = connect('mongodb://host2');
rs2 = rs; // rs2 !== rs1
sp2 = sp; // Would currently be sp2 === sp1, but should not be, one should still be able to access and use sp for host1
```

packages/shell-api/src/stream-processor.ts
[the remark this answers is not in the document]
addaleax: Yeah, no particularly strong feelings, but you can look around for `return await` in the codebase and see that it’s quite the common thing to do here

packages/shell-api/src/streams.ts
addaleax: Thank you, that makes sense! Then one thing that should probably change here is that we’d want to run any code that should wait the connection to be re-established from `onResumeExecution` rather than in response to the `interrupt-complete` event (and in particular keep the control flow local to the shell-api package, instead of relying on the user of the package to always emit this event).

I’ll also try to think whether there is a better way to handle this situation in general – it might be nice to actually allow creating sections of code where the behavior on Ctrl+C is fully overriden, so that there wouldn’t be a connection reset happening in the first place.

packages/shell-api/src/streams.ts
shaketbaby: Right, I see what you mean now. Will add the check.

packages/shell-api/src/streams.ts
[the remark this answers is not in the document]
shaketbaby: One thing I was trying to ensure is for `sp` to always using the current connection. But I can see how that's probably not the right thing in the scenario you gave. Okay, I will change this to accept a database.

packages/service-provider-core/src/connect-info.ts
mcasimir: Mm that's a good point, people do all sort of things to connect, we do detect atlas from the url only, but should be only for telemetry or presentation purposes, if a feature depends on it we may want something returned from the server.

shaketbaby: I definitely agree returning something from server is more reliable and better. But at the moment, this is only used to change the default prompt which user can change. And `sp` is always there regardless of whether this is actually a stream instance. Which makes me wondering how important it is to get it 100% correct.

packages/shell-api/src/streams.ts
```diff
@@ -0,0 +1,174 @@
...
+    const { _mongo, _instanceState } = this;
+    const interruptable = _instanceState.interrupted.asPromise();
+    try {
+      return await Promise.race([
+        _mongo._serviceProvider.runCommand(ADMIN_DB, cmd, options), // run cmd
+        interruptable.promise, // unless interruppted
+      ]);
+    } finally {
+      interruptable.destroy();
+    }
```
addaleax, line 172: @shaketbaby Can you remind me, what was the reason for this logic here, deviating from the core `_runCommand` helper? Is this just to get a specific type of error thrown from this method when execution is interrupted? (Just a question to make sure any changes to the interruption logic to support this here make sense for the streams implementation)

shaketbaby: I'll be honest that it's been a while and I don't remember all the details. I think this was added mainly for `sample` where we don't really need to destroy the underlying connection and just need to return the control to the end user. But looking again now, I am wondering if this is actually really required for all the other commands other than the sample related ones. Because they don't really catch the exception. For the normal core `_runCommand` helper, if user presses `ctrl + c` while waiting for the server to return the command response, the interrupted error will be thrown and the control will be returned to the user, right? If that's the case, then I think this is not needed for non-sample related comamnds.

shaketbaby: Basically, I think if there's a way to take over the handling of interruption on `ctrl + c` completely, then that should be good for all the commands we need to support.

packages/shell-api/src/streams.ts
```diff
@@ -0,0 +1,174 @@
...
+    if (!Array.isArray(pipeline) || !pipeline.length) {
+      throw new MongoshInvalidInputError(
+        'Invalid pipeline',
+        CommonErrors.InvalidArgument,
+        pipeline
+      );
+    }
+    const result = await this._runStreamCommand({
+      processStreamProcessor: pipeline,
+      ...(options ? { options } : {}),
```
sharanlobana, line 61: Has this code been added for future compatibility? As of now, `processStreamProcessor` enforces that no extra fields are specified: https://github.com/10gen/mongohouse/blob/77426dee9b976d20d4a7fc32baffd71bf3acb6e0/internal/mongoserver/command/streams_parser.go#L71C1-L73C3

```
AtlasStreamProcessing> sp.process($pipeline, {limit:5})
MongoServerError: "options" is not a valid argument to processStreamProcessor, correlationID = 177e143da46c62a1befa0aad
AtlasStreamProcessing>
```

shaketbaby: Yes, for now. I believe process command internally call create and then start it immediately. And `createStreamProcessor` needs to support passing in `options` according to the [syntax document](https://docs.google.com/document/d/1dC9ebvAHLjMyqStlhEyA0O2pxIceS0r1xWAfFAIrppo/edit#heading=h.m3fnu0kzh5i6).

packages/shell-api/src/streams.ts
addaleax: This is what my suggestion for this would look like, finally got around to it, please take a look if you have the time: https://github.com/mongodb-js/mongosh/pull/1616

packages/shell-api/src/streams.ts
addaleax: I think it’s a good idea to have some forward compatibility here, even if the server validates that extra options will be rejected for now :+1:

addaleax (approved):

PR #1536, opened 2023-07-25
fix(ci): account for diverging source maps when merging coverage files MONGOSH-1312

addaleax:


scripts/nyc/transform-coverage.ts
```diff
@@ -54,8 +62,24 @@ export function transformCoverageFiles(
...
+      if (unifyFileMetadata === 'keep') {
+        coverage[p] = fileCoverage;
+      } else {
+        // When preparing for the final report, we sometimes end up in a situation
+        // in which the file metadata from different coverage reports are not exact matches
+        // for each other (different source maps, line/column numbers). That leads
+        // nyc to generate lower coverage numbers than it otherwise would.
+        // TODO: Investigate why this mismatch occurs (e.g.: some difference between
+        // running a package's own tests vs. using a package from another package's tests
+        // and then using the transpiled code from lib/ or dist/).
```
himanshusinghs, line 74: If we are closing MONGOSH-1312 with this, maybe we open another for this todo?

himanshusinghs (approved):
Looks good. Would be nice if someone else as well could have a look at this.

scripts/nyc/transform-coverage.ts
addaleax: Sure, added a ticket + reference here! :+1:

scripts/nyc/transform-coverage.ts
```diff
@@ -54,8 +62,24 @@ export function transformCoverageFiles(
...
+        // When preparing for the final report, we sometimes end up in a situation
+        // in which the file metadata from different coverage reports are not exact matches
+        // for each other (different source maps, line/column numbers). That leads
+        // nyc to generate lower coverage numbers than it otherwise would.
+        // TODO(MONGOSH-1551): Investigate why this mismatch occurs (e.g.: some difference
+        // between running a package's own tests vs. using a package from another package's
+        // tests and then using the transpiled code from lib/ or dist/).
+        fileMetadata[p] ??= fileCoverage;
+        const { s, b, f } = fileCoverage;
+        coverage[p] = {
```
lerouxb, line 77: So `inputSourceMap` of the first time we encounter coverage for the path wins out and the assumption is that that happens to be more complete than subsequent ones?

lerouxb: Also: how sure are we that the combined coverage report for lines covered only by those subsequent coverage files won't point to the wrong things?

addaleax: I don’t think “more complete” is the right word, and I don’t know if the source map specifically was causing the issues here (the rest of the file metadata also mismatched and is probably more relevant). The problem is that merging files with different metadata just doesn’t result in meaningful results.

addaleax: They might, but I did look at the actual coverage files before and after this patch, and they definitely seemed more accurate after. We did open https://jira.mongodb.org/browse/MONGOSH-1551 to actually spend time finding the root cause here.

PR #1541, opened 2023-07-31
chore(ci): expand test matrix MONGOSH-1140 MONGOSH-1142 MONGOSH-1146

addaleax:
Add docker-based smoke tests for Debian 12 and e2e tests for RHEL9 and Ubuntu 22.04.

### DIFF
--- a/config/release-package-matrix.js
+++ b/config/release-package-matrix.js
@@ -20,7 +20,7 @@ exports.RELEASE_PACKAGE_MATRIX = [
     compileBuildVariant: 'linux_x64_build',
     packages: [
       { name: 'linux-x64', description: 'Linux Tarball 64-bit', packageOn: 'linux_package', smokeTestKind: 'docker', smokeTestDockerfiles: ['ubuntu20.04-tgz'] },
-      { name: 'deb-x64', description: 'Debian / Ubuntu 64-bit', packageOn: 'linux_package', smokeTestKind: 'docker', smokeTestDockerfiles: ['ubuntu18.04-deb', 'ubuntu20.04-deb', 'ubuntu22.04-deb', 'ubuntu22.04-nohome-deb', 'debian9-deb', 'debian10-deb', 'debian11-deb'] },
+      { name: 'deb-x64', description: 'Debian / Ubuntu 64-bit', packageOn: 'linux_package', smokeTestKind: 'docker', smokeTestDockerfiles: ['ubuntu18.04-deb', 'ubuntu20.04-deb', 'ubuntu22.04-deb', 'ubuntu22.04-nohome-deb', 'debian9-deb', 'debian10-deb', 'debian11-deb', 'debian12-deb'] },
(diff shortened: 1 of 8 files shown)


config/release-package-matrix.js
```diff
@@ -20,7 +20,7 @@ exports.RELEASE_PACKAGE_MATRIX = [
     compileBuildVariant: 'linux_x64_build',
     packages: [
       { name: 'linux-x64', description: 'Linux Tarball 64-bit', packageOn: 'linux_package', smokeTestKind: 'docker', smokeTestDockerfiles: ['ubuntu20.04-tgz'] },
-      { name: 'deb-x64', description: 'Debian / Ubuntu 64-bit', packageOn: 'linux_package', smokeTestKind: 'docker', smokeTestDockerfiles: ['ubuntu18.04-deb', 'ubuntu20.04-deb', 'ubuntu22.04-deb', 'ubuntu22.04-nohome-deb', 'debian9-deb', 'debian10-deb', 'debian11-deb'] },
+      { name: 'deb-x64', description: 'Debian / Ubuntu 64-bit', packageOn: 'linux_package', smokeTestKind: 'docker', smokeTestDockerfiles: ['ubuntu18.04-deb', 'ubuntu20.04-deb', 'ubuntu22.04-deb', 'ubuntu22.04-nohome-deb', 'debian9-deb', 'debian10-deb', 'debian11-deb', 'debian12-deb'] },
```
kmruiz, line 23: Shouldn't we remove support for Debian 9 at some point? It has been EOL for a year:

https://wiki.debian.org/LTS

packages/cli-repl/test/e2e-auth.spec.ts
```diff
@@ -930,7 +936,10 @@ describe('Auth e2e', function () {
     });
     context('with specific auth mechanisms', function () {
       it('can auth with SCRAM-SHA-1', async function () {
-        if (process.env.MONGOSH_TEST_E2E_FORCE_FIPS) {
+        if (
+          process.env.MONGOSH_TEST_E2E_FORCE_FIPS ||
+          process.env.DISTRO_ID === 'rhel92-fips'
```
kmruiz, line 941: Why do we need to check the DISTRO_ID? What would be the impact of, at the CI level, specifying MONGOSH_TEST_E2E_FORCE_FIPS=true on the rhel92-fips variant?

addaleax: `MONGOSH_TEST_E2E_FORCE_FIPS` makes our test runner for the e2e tests always append the `--tlsFIPSMode` flag to the mongosh command line, so mongosh always runs in FIPS mode.

The `rhel92-fips` machines seem to be configured in a similar way; their system OpenSSL configuration always runs everything in FIPS mode (this seems to be different from `rhel83-fips`, by the way; it might be a question of OpenSSL 1.1 vs OpenSSL 3).

If we set `MONGOSH_TEST_E2E_FORCE_FIPS=true` here, we wouldn’t be testing the behavior of mongosh on a FIPS-enforcing machine without the `--tlsFIPSMode` flag anymore. That might be okay, but I think I’d mildly prefer to have the extra test coverage here, even if that means running two e2e test tasks with the same effective behavior.

kmruiz: OK that is a huge detail I wasn't aware of. Thanks a lot for the explanation!

PR #1545, opened 2023-08-01
feat: sh.status() omits deprecated fields from config.version MONGOSH-1331

kmruiz:
Removes deprecated fields on sh.status(), so they are not visible anymore to avoid confusion. Removed fields:

- minCompatibleVersion
- currentVersion
- excluding
- upgradeId
- upgradeState

| Old Version  | New Version |
| ------------- | ------------- |
| Screenshot 2023-08-01 at 17 36 39 | Screenshot 2023-08-01 at 17 37 10  |


packages/shell-api/src/helpers.ts
```diff
@@ -203,6 +203,14 @@ export function processDigestPassword(
   return { digestPassword: true };
 }
 
+const SHARDING_VERSION_FIELD_EXCLUSION = {
+  minCompatibleVersion: 0,
+  currentVersion: 0,
+  excluding: 0,
+  upgradeId: 0,
+  upgradeState: 0,
+};
```
addaleax, line 212: Totally optional, but I think you can just inline this into the place where it’s being used

packages/shell-api/src/helpers.ts
```diff
@@ -346,7 +354,7 @@ export async function getPrintableShardStatus(
     })(),
     (async (): Promise<void> => {
       // Actionlog and version checking only works on 2.7 and greater
-      let versionHasActionlog = false;
+      let versionHasActionlog = !version.currentVersion; // means that we are in a version where it is deprecated (SERVER-68888)
```
addaleax, line 357: We exclude `currentVersion` in the projection above, so `version.currentVersion` should always be undefined, no? What am I mising?

kmruiz: I think the latest commit fixes it.

It keeps the currentVersion if it's on the document, until the end of the `getPrintableShardInfo` function, where we delete it, so it's not printed in the output.

packages/shell-api/src/helpers.spec.ts
```diff
@@ -181,6 +181,21 @@ describe('getPrintableShardStatus', function () {
...
+  describe('hides all internal deprecated fields in shardingVersion', function () {
+    for (const hiddenField of [
+      'minCompatibleVersion',
+      'currentVersion',
+      'excluding',
+      'upgradeId',
+      'upgradeState',
+    ]) {
+      it(`does not show ${hiddenField} in shardingVersion`, async function () {
+        const status = await getPrintableShardStatus(configDatabase, false);
```
addaleax, line 193: Very much an optional suggestion, `getPrintableShardStatus()` does quite a lot of work, it might make sense to run it in a `before()` hook inside this `describe()` block (since it returns the same result for all of the tests)?

addaleax (approved):

packages/shell-api/src/helpers.spec.ts
kmruiz:
```js
describe.only('hides all internal deprecated fields in shardingVersion', async function () {
    const status = await getPrintableShardStatus(configDatabase, false);

    for (const hiddenField of [
      'minCompatibleVersion',
      'currentVersion',
      'excluding',
      'upgradeId',
      'upgradeState',
    ]) {
      it(`does not show ${hiddenField} in shardingVersion`, async function () {
        expect(status.shardingVersion[hiddenField]).to.equal(undefined);
      });
    }
  });
```

Does not work as describe functions can not be async, so I'll need to keep the getPrintableShardStatus inside the `it`. If you are ok with it, I'll keep it as it is, as it seems to be more complicated than it should be 😞 .

addaleax: Yeah, all good then :+1:

addaleax:
Evergreen run (started from the CLI because evergreen is not configured for the v2 base branch): https://spruce.mongodb.com/version/64ca67883066150af0d8ad7f/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

PR #1554, opened 2023-08-04
chore(shell-api): check StaleConfig error codeName in addition to code number

addaleax:
The server team has confirmed that they prefer for downstream teams to check error codeName values rather than (only) code numbers, as the latter may go away when code is being refactored.

(Opened COMPASS-7082 and MONGOSH-1554 to track the other instances of this.)

### DIFF
--- a/packages/shell-api/src/collection.ts
+++ b/packages/shell-api/src/collection.ts
@@ -1853,7 +1853,7 @@ export default class Collection extends ShellApiWithMongoClass {
 
       return await this._aggregateAndScaleCollStats(collStats, scale);
     } catch (e: any) {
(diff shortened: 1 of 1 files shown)


himanshusinghs (approved):

himanshusinghs:
Where are these error codes documented?

addaleax:
@himanshusinghs Not sure what you’re looking for in terms of documentation, but https://github.com/mongodb/mongo/blob/1c3fb19ac8e5690e81617e815f122d8a868804a3/src/mongo/base/error_codes.yml is the source of truth for the _list_ of all of the error codes. I’m not sure if there’s proper documentation for these beyond that file.

PR #1562, opened 2023-08-09
feat: add programmatically accessible download feed MONGOSH-1242

addaleax:


scripts/gather-existing-releases.mjs
```diff
@@ -0,0 +1,121 @@
...
+      const rhel81AndAbove = ['rhel81', 'rhel82', 'rhel83', 'rhel90']
+      const rhel80AndAbove = ['rhel80', ...rhel81AndAbove]
+      const rhel72AndAbove = ['rhel72', ...rhel80AndAbove]
+      const al2AndAbove = ['amazon2', 'amazon2023', ...rhel81AndAbove]
+      const rhel70AndAboveAndRpmBased = ['rhel70', 'rhel71', ...rhel72AndAbove, 'amazon', ...al2AndAbove, 'suse12', 'suse15']
+      const ubuntu1804AndAboveAndDebBased = ['ubuntu1804', 'ubuntu1804', 'ubuntu2004', 'ubuntu2204', 'debian92', 'debian10', 'debian11', 'debian12']
+      const allLinux = [...rhel70AndAboveAndRpmBased, ...ubuntu1804AndAboveAndDebBased]
+
+      for (const [re, targets_] of [
+        [/^mongodb-mongosh(-shared-openssl\d+)?-1.((5|6|7|8|9|1\d+)\.\d+).(aarch64).rpm$/, al2AndAbove],
```
himanshusinghs, line 69: Out of curiosity - Why can't we match the version like `1.(\d+\.\d+)` like we do for the rest?

packages/build/src/download-center/config.ts
```diff
@@ -103,8 +160,117 @@ export function createVersionConfig(
...
+  packageInformation: PackageInformationProvider,
+  publicArtifactBaseUrl: string
+): Promise<JsonFeedVersionEntry> {
+  const { version } = packageInformation('linux-x64').metadata;
+  return {
+    version: version,
+    downloads: await Promise.all(
+      ALL_PACKAGE_VARIANTS.map(async (packageVariant: PackageVariant) => {
+        const arch = getArch(packageVariant);
+        const distro = getDistro(packageVariant);
```
himanshusinghs, line 201: We don't add this distro in the generated json file, just missed out or we can't?

himanshusinghs (approved):
This is amazing! 🤩

packages/build/src/download-center/config.ts
addaleax: There is https://github.com/mongodb-js/mongosh/blob/0b54496fe8024b7982ec1545fedcd02f72e96a3f/packages/build/src/download-center/config.ts#L211 – is that what you mean?

himanshusinghs: Ahh no I mean in here - https://github.com/mongodb-js/mongosh/pull/1562/files#diff-ac13dafadbaacbb5b6de1d253a5ec90b790e2756ca746b2d9bb42177d3cf70b7R106-R115

Where we generate a json feed for existing releases.

addaleax: Oh, yeah. I guess we could try … I’m not sure what the values would be in that case for the pre-1.5.0 releases. I mean, we obviously had some values in the build packages code back then, but I don’t think we could easily apply our _current_ set of `distro` values to those. I’d probably omit it, this can just be a new feature for versions starting now, what do you think?

scripts/gather-existing-releases.mjs
himanshusinghs: Understood. Thanks :)

packages/build/src/download-center/config.ts
himanshusinghs: I feel alright with that. Just wanted to make sure that its not something we missed.

PR #1566, opened 2023-08-10
fix: dropIndexes so it run on the primary regardless the read preference MONGOSH-1478

kmruiz:


packages/shell-api/src/database.ts
```diff
@@ -26,13 +26,14 @@ import {
...
-  WriteConcern,
-  ListCollectionsOptions,
+import {
+  type ChangeStreamOptions,
+  type CommandOperationOptions,
+  type CreateCollectionOptions,
+  type Document,
+  type WriteConcern,
+  type ListCollectionsOptions,
+  ReadPreference,
```
addaleax, line 36: Where is `ReadPreference` being used? Generally speaking, we don’t want to import runtime values from the Node.js driver via `@mongosh/service-provider-core`, only types

packages/shell-api/src/database.ts
```diff
@@ -149,10 +150,25 @@ export default class Database extends ShellApiWithMongoClass {
...
-
   public async _runCommand(
     cmd: Document,
     options: CommandOperationOptions = {}
+  ): Promise<Document> {
+    return this._mongo._serviceProvider.runCommandWithCheck(
+      this._name,
+      adjustRunCommand(cmd, this._instanceState.shellBson),
+      {
+        readPreference: 'primaryPreferred',
```
addaleax, line 161: I would imagine that we don’t want to apply any read preference here at all, to match what drivers and (presumably) the legacy shell do?

packages/shell-api/src/database.ts
kmruiz: Ah yes, you are totally right, this is a leftover. Fixing it!

packages/shell-api/src/database.ts
kmruiz: OK, will fix everything in the same commit as the previous left over.

addaleax (approved):
Sorry, forgot to actually approve this! Feel free to ping me when this happens again :)

addaleax (approved):

PR #1573, opened 2023-08-14
fix(build): fix sharedOpenssl tag for downloads MONGOSH-1242

addaleax:
Should finally be all good then. :)


himanshusinghs (approved):
Good catch! 👀

PR #1579, opened 2023-08-15
chore: use sinon from the sinon package where possible, not ts-sinon MONGOSH-1501

lerouxb:
This change normalises all standard usage of sinon to be from the sinon package, not from the ts-sinon package. The only usage of ts-sinon that's left relates to the usage of the deprecated `stubInterface` function in ts-sinon. Consequently StubbedInstance is still imported from ts-sinon as well. There's a SinonStubbedInstance in sinon itself, but it is not compatible.

I didn't immediately see how to replace stubInterface, but I suppose we can do that in a follow-up. At that point we can remove the ts-sinon package completely.


addaleax (approved):

Anemy (approved):

lerouxb:
Something was up with the evergreen integration here, but this is the green run https://spruce.mongodb.com/version/64dcfb912fbabe75dc5ea469/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

PR #1591, opened 2023-08-17
fix(cli-repl): work around Node.js 20 process.exit() change

addaleax:

PR #1592, opened 2023-08-17
fix(ci): use npm@9.x

addaleax:
Previously, our CI on Node.js 20 was failing silently (e.g. the bug fixed by https://github.com/mongodb-js/mongosh/pull/1591 did make mocha exit with a non-zero exit code, but not our CI).

This is being caused by an incompatibility between npm@8.x and Node.js 20 (which could reasonably be argued to be a Node.js bug, the same one that the PR mentioned above accounts for). Upgrading to npm@9.x should fix this because it includes a fix in npm as well (https://github.com/npm/cli/issues/6399).

### DIFF
--- a/.evergreen/install-node.sh
+++ b/.evergreen/install-node.sh
@@ -5,7 +5,7 @@ export BASEDIR="$PWD/.evergreen"
 if [ "$OS" == "Windows_NT" ]; then
   powershell "$(cygpath -w "$BASEDIR")"/InstallNode.ps1
   . "$BASEDIR/setup-env.sh"
-  mkdir -p "$BASEDIR/npm-8" && (cd "$BASEDIR/npm-8" && echo '{}' > package.json && npm i npm@8.x)
+  mkdir -p "$BASEDIR/npm-8" && (cd "$BASEDIR/npm-8" && echo '{}' > package.json && npm i npm@9.x)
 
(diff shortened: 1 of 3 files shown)


addaleax:
Evergreen: https://spruce.mongodb.com/version/64de0986d1fe071cd92045b5/tasks

.evergreen/install-node.sh
```diff
@@ -5,7 +5,7 @@ export BASEDIR="$PWD/.evergreen"
 if [ "$OS" == "Windows_NT" ]; then
   powershell "$(cygpath -w "$BASEDIR")"/InstallNode.ps1
   . "$BASEDIR/setup-env.sh"
-  mkdir -p "$BASEDIR/npm-8" && (cd "$BASEDIR/npm-8" && echo '{}' > package.json && npm i npm@8.x)
+  mkdir -p "$BASEDIR/npm-8" && (cd "$BASEDIR/npm-8" && echo '{}' > package.json && npm i npm@9.x)
```
lerouxb, line 8: So we're leaving the folder name "npm-8"? Isn't that confusing?

addaleax: Lol yes, I missed this :) Changed!

lerouxb (approved):

addaleax:
Evergreen: https://spruce.mongodb.com/version/64de2a3ce3c3310471c6ec63/tasks

PR #1594, opened 2023-08-17
fix(shell-api): use public BulkWriteResult props MONGOSH-1507 COMPASS-7023

addaleax:
This accounts for removal of these legacy getters in the 6.x driver.

The only catch here is that `.result.upserted` is of a different type than `.upsertedIds`, but since the type of the latter matches what our own `BulkWriteResult` class expects (and also matches `.insertedIds`), this is probably closer to a bug fix than proper breakage.

### DIFF
--- a/packages/shell-api/src/bulk.spec.ts
+++ b/packages/shell-api/src/bulk.spec.ts
@@ -1,5 +1,8 @@
 import { CommonErrors } from '@mongosh/errors';
-import type { ServiceProvider } from '@mongosh/service-provider-core';
+import type {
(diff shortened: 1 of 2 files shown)


addaleax:
Evergreen: https://spruce.mongodb.com/version/64de2e3d3e8e8684e40d116b/tasks

addaleax:
Evergreen: https://spruce.mongodb.com/version/64df4020e3c33171c62b3b78/tasks

mcasimir (approved):

PR #1597, opened 2023-08-17
chore: use lerna@7 MONGOSH-1486

lerouxb:
In theory `lerna bootstrap` is not needed anymore: https://lerna.js.org/docs/legacy-package-management

`npm install` and `npm ci` should do all we need. I considered also adding `&& npm run bootstrap` like we do in compass, but none of the packages in mongosh have a "bootstrap" script, so that's not needed.

I'm still manually testing all of the places we use `lerna`, so consider this in progress for now.


addaleax (approved):

lerouxb:
OK so
NX_NON_NATIVE_HASHER=true was removed https://github.com/nrwl/nx/pull/17338 and that's not an option.

lerouxb:
Closing in favour of https://github.com/mongodb-js/mongosh/pull/1600 and other incremental PRs.

PR #1602, opened 2023-08-21
chore: Remove keyVault single-Document wrapping methods MONGOSH-1556 

kmruiz:
Currently, as in https://jira.mongodb.org/browse/MONGOSH-1266, we are wrapping the KeyVault `getKey` and `getKeyByAltName` return values into a proxy that acts both as a document and a cursor. In mongosh 2.0, we are dropping the proxy and returning either a Document or null if the key doesn't exist, mimicking the driver's behaviour.

This PR implements https://jira.mongodb.org/browse/MONGOSH-1556, that removes the wrapper proxy and simplifies a bit the code base.


kmruiz:
Wrong branch, recreating PR.

PR #1603, opened 2023-08-21
feat(shell-api)!: unwrap and use document | null instead of a cursor MONGOSH-1556

kmruiz:
Currently, as in https://jira.mongodb.org/browse/MONGOSH-1266, we are wrapping the KeyVault getKey and getKeyByAltName return values into a proxy that acts both as a document and a cursor. In mongosh 2.0, we are dropping the proxy and returning either a Document or null if the key doesn't exist, mimicking the driver's behaviour.

This PR implements https://jira.mongodb.org/browse/MONGOSH-1556, that removes the wrapper proxy and simplifies a bit the code base.


addaleax (approved):
:rocket:

Just adjusted the PR title a bit since this is a breaking change (even if it is a very small one) :slightly_smiling_face:

PR #1608, opened 2023-08-23
chore: auto compile locally, remove lerna bootstrap from evergreen setup MONGOSH-1517

lerouxb:
While making sure mongodb runs compile when bootstrapping like compass does I realised that's an issue in CI due to how we install there along some paths and on some operating systems. So in anticipation of [upgrading lerna](https://jira.mongodb.org/browse/MONGOSH-1486), where the latest version won't support some weirder operating systems, I set out to remove lerna as a dependency when installing and compiling.

There are other places we still use lerna in CI, but I'll tackle them in follow-ups.


packages/browser-repl/package.json
```diff
@@ -32,6 +32,7 @@
     "check": "npm run lint && npm run depcheck",
     "depcheck": "depcheck",
     "compile": "tsc -p tsconfig.json",
+    "bootstrap": "npm run compile",
```
lerouxb, line 35: Missed removing this. Already amended the commit. Will push later after this evergreen run finishes.

package.json
```diff
@@ -126,8 +123,31 @@
     "yaml": "^1.10.0"
   },
   "workspaces": [
-    "configs/*",
-    "packages/*",
-    "scripts/docker"
+    "configs/eslint-config-mongosh",
```
lerouxb, line 126: We'll generate this with a hook somewhere. To be decided exactly where. preinstall?

addaleax: Is there any chance we could just use the new monorepo-tools scripts instead of hardcoding a list here?

.evergreen/install-npm-deps.sh
```diff
@@ -8,30 +8,38 @@ set -x
...
   # mongodb-client-encryption cannot be installed everywhere without prerequisites
   # (because it's hard to build from source when there's no prebuilts available
   # because it requires libmongocrypt to be installed globally then), but we still
   # need its types; so if we do not need all packages to be installed,
   # we first try to install mongodb-client-encryption, and if that fails, we fall back
   # to installing with --ignore-scripts (i.e. do not attempt to build addons).
-  (npm run bootstrap-ci -- --scope @mongosh/cli-repl --include-dependencies && test -e packages/service-provider-server/node_modules/mongodb-client-encryption) || \
-    npm run bootstrap-ci -- --scope @mongosh/cli-repl --include-dependencies --ignore-scripts
+  (npm ci -w @mongosh/build -w @mongosh/cli-repl && test -e packages/service-provider-server/node_modules/mongodb-client-encryption) || \
+    npm ci -w @mongosh/cli-repl --ignore-scripts
```
addaleax, line 28: What does this do given that we already ran `npm ci` above in line 9?

lerouxb: We've been trying to figure out for many hours :)

package.json
lerouxb: We're now trying to avoid rewriting lerna since there's quite a bit of scope creep there.

.evergreen/install-npm-deps.sh
lerouxb: `npm run bootstrap-ci -- --scope @mongosh/cli-repl --include-dependencies` ran `lerna bootstrap --ci --scope @mongosh/cli-repl --include-dependencies`.

lerna bootstrap is apparently just npm install nowadays. --scope is -w. npm install always includes dependencies?

package.json
lerouxb: Plan would be adding "workspacePatterns" (which contains what we had before) and it just gets regenerated every so often automatically not too different from a package-lock.json.

This is just a draft / WIP test to see if it would work before we do that.

package.json
```diff
@@ -63,7 +60,8 @@
     "where": "monorepo-where",
     "reformat": "npm run reformat --workspaces --if-present",
     "prepare": "husky install",
-    "precommit": "precommit"
+    "precommit": "precommit",
+    "preinstall": "node scripts/sort-workspaces.js"
```
lerouxb, line 64: This will make sure that workspaces are always sorted in topological order according to what's in lerna.json before we even install. Then `npm nun compile --workspaces` (and any npm run --workspaces command) will run in topological order. ie. We don't need lerna in ci for that command.

Why before install? Suppose you add a new package to mongosh. It would still be missing from workspaces if you don't remember to run sort-workspaces first and that will mess with the next install.

That's also why it is a plain javascript script that only uses builtins and not typescript.

package.json
```diff
@@ -8,11 +8,8 @@
   "homepage": "https://github.com/mongodb-js/mongosh",
   "author": "Compass Team <compass@mongodb.com>",
   "scripts": {
-    "prebootstrap-with-chromium": "npm install",
-    "bootstrap-with-chromium": "lerna bootstrap --concurrency=1",
-    "prebootstrap": "npm install",
-    "bootstrap": "cross-env PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=1 lerna bootstrap --concurrency=1",
-    "bootstrap-ci": "lerna bootstrap --ci",
```
lerouxb, line 15: removed bootstrap-ci because bootstrapping in ci is complicated and I didn't want it split between this package.json, our evergreen yaml and a shell script. Rather just in one place as much as possible.

package.json
```diff
@@ -29,7 +26,7 @@
     "test-connectivity": "npm run test-connectivity --workspaces --if-present",
     "replace-package": "node scripts/replace-package.js",
     "test-nodedriver": "bash .evergreen/test-node-driver.sh",
-    "compile": "lerna run compile",
+    "compile": "npm run compile --workspaces --if-present",
```
lerouxb, line 29: In anticipation of not being able to use the latest lerna in CI anymore (due to them dropping support for exotic platforms) I dealt with this here. We'll have to still tweak a few more commands to make sure we never use lerna in mongosh CI anymore, though. More PRs to follow :)

.github/workflows/cron-tasks.yml
```diff
@@ -37,7 +37,6 @@ jobs:
       - name: Install Dependencies and Compile
         run: |
           npm ci
-          npm run bootstrap-ci
```
lerouxb, line 40: This was just doing npm ci a second time.

lerouxb:
evergreen refresh

addaleax (approved):

PR #1613, opened 2023-08-24
fix: use queryPlanner to keep conventions with other explain methods MONGOSH-1317

kmruiz:
These are the different ways of doing explain on a cursor or a query:

| Case | Code                                            | Provided Execution Stats | Verbosity         |
|------|-------------------------------------------------|--------------------------|-------------------|
| 1    | db.runCommand({…verbosity:’allPlansExecution’}) | yes                      | allPlansExecution |
| 2    | db.foo.explain(‘allPlansExecution’).agg(…)      | yes                      | allPlansExecution |
| 3    | db.foo.explain().agg(…)                         | no                       | queryPlanner      |
| 4    | db.foo.agg(…).explain()                         | no                       | allPlansExecution |

What are we doing in this PR is aligning case 4, so instead of using `allPlansExecution` by default, it uses `queryPlanner`, as the rest of the cases. The table will look like:

| Case | Code                                            | Provided Execution Stats | Verbosity         |
|------|-------------------------------------------------|--------------------------|-------------------|
| 1    | db.runCommand({…verbosity:’allPlansExecution’}) | yes                      | allPlansExecution |
| 2    | db.foo.explain(‘allPlansExecution’).agg(…)      | yes                      | allPlansExecution |
| 3    | db.foo.explain().agg(…)                         | no                       | queryPlanner      |
| 4    | db.foo.agg(…).explain()                         | no                       | queryPlanner |

### DIFF
--- a/packages/shell-api/src/aggregate-or-find-cursor.ts
+++ b/packages/shell-api/src/aggregate-or-find-cursor.ts
@@ -41,9 +41,7 @@ export abstract class AggregateOrFindCursor<
     // TODO: @maurizio we should probably move this in the Explain class?
     // NOTE: the node driver always returns the full explain plan
     // for Cursor and the queryPlanner explain for AggregationCursor.
-    if (verbosity !== undefined) {
-      verbosity = validateExplainableVerbosity(verbosity);
-    }
+    verbosity = validateExplainableVerbosity(verbosity);
     const fullExplain: any = await this._cursor.explain(verbosity);
 
     const explain: any = {
--- a/packages/shell-api/src/helpers.spec.ts
+++ b/packages/shell-api/src/helpers.spec.ts
@@ -5,6 +5,7 @@ import {
   getPrintableShardStatus,
   scaleIndividualShardStatistics,
   tsToSeconds,
+  validateExplainableVerbosity,
 } from './helpers';
 import { Database, Mongo, ShellInstanceState } from './index';
 import constructShellBson from './shell-bson';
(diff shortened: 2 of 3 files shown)


lerouxb (approved):

packages/shell-api/src/aggregate-or-find-cursor.ts
```diff
@@ -44,6 +44,7 @@ export abstract class AggregateOrFindCursor<
     if (verbosity !== undefined) {
       verbosity = validateExplainableVerbosity(verbosity);
     }
```
addaleax, line 46: Doesn’t this conditional explicitly skip applying this logic in the default case?

addaleax (changes requested):
This should target the 2.x branch as it changes a default

packages/shell-api/src/aggregate-or-find-cursor.ts
kmruiz: You are right, but I was testing now the behaviour manually (and have some tests) and it seems to be fixed. I think I'm looking it with the wrong perspective.

Thanks for noticing it!

kmruiz:
Reopened, rebased from 2.0 😄

addaleax (approved):

kmruiz:
Evergreen patch: https://spruce.mongodb.com/version/64e73d4732f417b896319cdb/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

PR #1616, opened 2023-08-24
chore(shell-api): allow more advanced interrupt handling MONGOSH-1555

addaleax:


packages/shell-api/src/interruptor.ts
```diff
@@ -75,12 +86,16 @@ export class InterruptFlag {
...
+   * installed through withOverrideInterruptBehavior() may perform additional
+   * cleanup work before the current connection is ready to be severed.
    */
-  public set(): void {
+  public async set(): Promise<void> {
     this.interrupted = true;
     const err = new MongoshInterruptedError();
     for (const listener of [...this.onInterruptListeners]) {
-      listener(err);
+      await listener(err);
```
shaketbaby, line 98: Do we want to catch any errors that the listener might throw so all listeners will always be notified?
```ts
try {
  await listener(err);
} catch {
  // ignore
}
```

shaketbaby (approved):
Looks good, thanks for doing this 👍

packages/shell-api/src/interruptor.ts
addaleax: Yes, that seems like a good idea :+1:

addaleax:
Evergreen: https://spruce.mongodb.com/version/64ec68b561837d838d2a6308/tasks

PR #1621, opened 2023-08-28
chore: Remove lerna from where it might run on exotic platforms in CI [WIP]

lerouxb:
Includes https://github.com/mongodb-js/mongosh/pull/1608. Merge that first.

Still WIP as there is some usage left in packages/build somewhere. Just testing this so long.


package-lock.json
addaleax, line 1: This seems to be a massive change to the package-lock – let’s hold off on merging this until after 2.0.0 is released?

addaleax (commented):
LGTM!

addaleax:
> This PR just replaces the remaining usage of lerna that could come up in CI with standard npm workspaces commands or custom code and then makes lerna an optional dependency.

I’m still seeing it used here: https://github.com/mongodb-js/mongosh/blob/d26d89b6eedaf61d6cad07c395b45d4cffa82bc3/packages/build/src/npm-packages/bump.ts#L17

packages/build/src/npm-packages/bump.ts
```diff
@@ -13,27 +12,40 @@ export function bumpNpmPackages(
...
-      cwd: PROJECT_ROOT,
-      encoding: 'utf8',
+  const monorepoRootPath = path.resolve(__dirname, '..', '..', '..', '..');
+  const packages = await getPackagesInTopologicalOrder(monorepoRootPath);
+
+  const locations = [monorepoRootPath, ...packages.map((p) => p.location)];
+
+  for (const location of locations) {
+    const packageJsonPath = path.join(location, 'package.json');
+    console.log({ packageJsonPath });
```
lerouxb, line 22: I'll remove this once CI finishes. Forgot it in there and I don't want to start everything over again :)

packages/build/src/npm-packages/bump.ts
```diff
@@ -13,27 +12,40 @@ export function bumpNpmPackages(
...
+      'peerDependencies',
+    ]) {
+      if (!packageJson[grouping]) {
+        continue;
+      }
+
+      for (const [name, currentVersion] of Object.entries(
+        packageJson[grouping]
+      )) {
+        if (currentVersion !== PLACEHOLDER_VERSION) {
```
addaleax, line 39: Wouldn’t this be a more accurate check? I get that in practice we’re not going to include packages in our dependencies that happen to use the placeholder as their actual version, but what we’re really going for here semantically is bumping the packages that are also _from this monorepo_, right?

```suggestion
        if (!packages.map(p => p.name).includes(name)) {
```

(This is optional but fwiw, the test file here needs to be updated as well)

lerouxb: Yeah I have to write tests. Just wanted to see if this fixes the CI failure in the meantime.

PR #1624, opened 2023-08-29
chore: turn some mocha lint rules back into errors, fix the errors MONGOSH-1508

lerouxb:
The are good to reinstate as errors IMHO:
```
'mocha/no-identical-title': 1,
'mocha/no-nested-tests': 1,
```

So I fixed the lines they caught.

I think the following mocha-related overrides make sense to leave as warnings:
```
'mocha/no-exports': 1,
'mocha/max-top-level-suites': 1,
'mocha/no-sibling-hooks': 1,
```

But let me know if others feel differently. I don't feel too strongly about it.

### DIFF
--- a/packages/shell-evaluator/src/shell-evaluator.spec.ts
+++ b/packages/shell-evaluator/src/shell-evaluator.spec.ts
@@ -87,7 +87,7 @@ describe('ShellEvaluator', function () {
       expect(itSpy).to.have.been.calledWith();
     });
 
-    it('forwards the exit/quit command', async function () {
+    it('forwards the exitquit command (exit)', async function () {
       const dontCallEval = () => {
         throw new Error('unreachable');
       };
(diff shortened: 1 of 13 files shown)


packages/shell-evaluator/src/shell-evaluator.spec.ts
```diff
@@ -87,15 +87,15 @@ describe('ShellEvaluator', function () {
       expect(itSpy).to.have.been.calledWith();
     });
 
-    it('forwards the exit/quit command', async function () {
+    it('forwards the exitquit command (exit)', async function () {
```
addaleax, line 90:
```suggestion
    it('forwards the exit/quit command (exit)', async function () {
```

addaleax (approved):

packages/shell-evaluator/src/shell-evaluator.spec.ts
lerouxb: oops. was going to make then forwards the exit command and forwards the quit command but then didn't undo all the typing :)

PR #1629, opened 2023-08-29
feat: add support for local atlas MONGOSH-1574

mabaasit:


packages/shell-api/src/shell-instance-state.ts
```diff
@@ -496,6 +496,8 @@ export default class ShellInstanceState {
     const extraConnectionInfo = this.connectionInfo?.extraInfo;
     if (extraConnectionInfo?.is_data_federation) {
       return 'AtlasDataFederation';
+    } else if (extraConnectionInfo?.is_local_atlas) {
+      return 'Atlas Local Dev';
```
addaleax, line 500: This sounds okay to me, but in case you haven’t, feel free to confirm the exact phrase here with @mmarcon :slightly_smiling_face: 

It might be worth being consistent between ADF and this and either use spaces in both of them or in neither?

packages/service-provider-server/src/cli-service-provider.ts
```diff
@@ -348,14 +348,25 @@ class CliServiceProvider
...
+          'admin',
+          {
+            count: 'atlascli',
+            query: { managedClusterType: 'atlasCliLocalDevCluster' },
+          },
+          this.baseCmdOptions
+        ).catch(() => ({ n: 0 })),
       ]);
 
+    const isLocalAtlasCli = Number(atlascliInfo.n) > 0;
```
addaleax, line 361:
```suggestion
        this.countDocuments(
          'admin',
          'atlascli',
          { managedClusterType: 'atlasCliLocalDevCluster' }
        ).catch(() => 0),
      ]);

    const isLocalAtlasCli = !!atlascliInfo;
```

Given that there has been a lot of back-and-forth around the `count` command’s level of stability and the `count()` driver helper I think it’s best to just stick with a non-deprecated driver helper here

addaleax (commented):
Looks great! :rocket:

packages/shell-api/src/shell-instance-state.ts
mabaasit: add changed it to `LocalDev` and pinged Max on same. If he suggests to change it, I'll do that accordingly.

packages/service-provider-server/src/cli-service-provider.ts
mabaasit: Thanks for suggesting this. I updated it in [4a2ecd0](https://github.com/mongodb-js/mongosh/pull/1629/commits/4a2ecd0db82e12794a1e4f3c9f3377d860875309)

addaleax (approved):

packages/shell-api/src/shell-instance-state.ts
mabaasit: Update from Max - he suggested to use `AtlasLocalDev` and I added that in [57f84af](https://github.com/mongodb-js/mongosh/pull/1629/commits/57f84af45a567f86ea020b1cf1caaf0fd26069e8)

PR #1659, opened 2023-09-05
fix: do not attempt to publish e2e-tests package

addaleax:
The publishing step would currently fail on the `e2e-tests` package because it defined a broken `prepublish` script.

We should not be trying to publish the package at all; mark it `private`.

### DIFF
--- a/packages/build/src/npm-packages/publish.ts
+++ b/packages/build/src/npm-packages/publish.ts
@@ -33,6 +33,7 @@ export function publishNpmPackages(
       [
         'publish',
         'from-package',
+        '--no-private',
         '--no-changelog',
         '--no-push',
         '--exact',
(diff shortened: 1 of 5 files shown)


packages/build/src/npm-packages/publish.ts
```diff
@@ -33,6 +33,7 @@ export function publishNpmPackages(
       [
         'publish',
         'from-package',
+        '--no-private',
```
addaleax, line 36: Not sure if this changes behavior in a meaningful way compared to what’s happening right now, but this is definitely what we _want_ to happen

packages/connectivity-tests/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "connectivity-tests",
+  "name": "@mongosh/connectivity-tests",
```
addaleax, line 2: Shouldn’t have any effect at all, but matches our other packages (including the private ones) and avoids a potential naming conflict

lerouxb (approved):

PR #1678, opened 2023-09-19
feat(cli-repl): do not show MDB version for SPI STREAMS-576

shaketbaby:
Currently, MongoDB version will be printed as part of the greeting message. We'd like to not show it when connecting to a streams processing instance as that info is not very relevant.

### DIFF
--- a/packages/cli-repl/src/mongosh-repl.ts
+++ b/packages/cli-repl/src/mongosh-repl.ts
@@ -173,12 +173,17 @@ class MongoshNodeRepl implements EvaluationListener {
     instanceState.setEvaluationListener(this);
     await instanceState.fetchConnectionInfo();
 
-    let mongodVersion = instanceState.connectionInfo.buildInfo?.version;
-    const apiVersion = serviceProvider.getRawClient()?.serverApi?.version;
-    if (apiVersion) {
-      mongodVersion =
-        (mongodVersion ? mongodVersion + ' ' : '') +
-        `(API Version ${apiVersion})`;
+    const { buildInfo, extraInfo } = instanceState.connectionInfo;
+    let mongodVersion = extraInfo?.is_stream
+      ? 'Atlas Stream Processing'
+      : buildInfo?.version;
+    if (!extraInfo?.is_stream) {
+      const apiVersion = serviceProvider.getRawClient()?.serverApi?.version;
+      if (apiVersion) {
+        mongodVersion =
+          (mongodVersion ? mongodVersion + ' ' : '') +
+          `(API Version ${apiVersion})`;
+      }
     }
     await this.greet(mongodVersion, moreRecentMongoshVersion);
     await this.printBasicConnectivityWarning(instanceState);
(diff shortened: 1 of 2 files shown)


packages/cli-repl/src/mongosh-repl.ts
```diff
@@ -173,12 +173,15 @@ class MongoshNodeRepl implements EvaluationListener {
...
 
-    let mongodVersion = instanceState.connectionInfo.buildInfo?.version;
-    const apiVersion = serviceProvider.getRawClient()?.serverApi?.version;
-    if (apiVersion) {
-      mongodVersion =
-        (mongodVersion ? mongodVersion + ' ' : '') +
-        `(API Version ${apiVersion})`;
+    const { buildInfo, extraInfo } = instanceState.connectionInfo;
+    let mongodVersion = extraInfo?.is_stream ? undefined : buildInfo?.version;
+    if (!extraInfo?.is_stream) {
```
addaleax, line 178: What happens when a user tries to connect to SPI with a stable API version defined?

packages/cli-repl/src/mongosh-repl.ts
```diff
@@ -173,12 +173,15 @@ class MongoshNodeRepl implements EvaluationListener {
...
     await instanceState.fetchConnectionInfo();
 
-    let mongodVersion = instanceState.connectionInfo.buildInfo?.version;
-    const apiVersion = serviceProvider.getRawClient()?.serverApi?.version;
-    if (apiVersion) {
-      mongodVersion =
-        (mongodVersion ? mongodVersion + ' ' : '') +
-        `(API Version ${apiVersion})`;
+    const { buildInfo, extraInfo } = instanceState.connectionInfo;
+    let mongodVersion = extraInfo?.is_stream ? undefined : buildInfo?.version;
```
addaleax, line 177: Are streams endpoint fully unversioned? Why does the `buildInfo` command report a version from the endpoint then? Having an endpoint version available is usually one of the most important things for debugging, so I’d like to make sure we’re omitting it in the right way. It probably makes sense to at least do something like

```suggestion
    let mongodVersion = extraInfo?.is_stream ? 'Atlas Streams Processing' : buildInfo?.version;
```

shaketbaby: Streams currently doesn't have version. Streams reuse the ADF FE and the `version` returned in `buildInfo` is the ADF version.

addaleax: Okay, then I’d still probably leave everything here more or less as it is and use the suggestion I made above?

When users share screenshots or terminal logs with us or TS, it’s going to be much easier to identify that they’re connecting to a streams processing instance if we just explicitly say `Using MongoDB: Atlas Streams Processing` instead of omitting the line entirely.

shaketbaby: Okay, have updated as what you suggested.

PR #1708, opened 2023-10-19
feat: bump to BSON 6.2.0, make string quoting consistent MONGOSH-1452

addaleax:


packages/service-provider-core/src/printable-bson.ts
```diff
@@ -65,14 +57,12 @@ export const bsonStringifiers: Record<
...
-          return `UUID("${asUUID}")`;
+          return `UUID('${asUUID}')`;
         }
       // In case somebody did something weird and used an UUID with a
       // non-standard length, fall through.
       default:
-        return `Binary.createFromBase64("${this.toString('base64')}", ${
-          this.sub_type
-        })`;
+        return binaryInspect.apply(this, args);
```
Anemy, line 65: Do we have a test for this one?

Anemy (approved):

PR #1721, opened 2023-11-03
fix(cli-repl): prevent promisify promise-returning function warning in async repl

baileympearson:
This PR wraps the internals of the async repl `eval()` method in an async function and calls it making `eval` itself an async function.  This prevents warnings from being emitted in Node21+, when the method is promisified.

### DIFF
--- a/packages/cli-repl/src/mongosh-repl.spec.ts
+++ b/packages/cli-repl/src/mongosh-repl.spec.ts
@@ -1286,9 +1286,6 @@ describe('MongoshNodeRepl', function () {
     });
 
     it('does not show a prompt', async function () {
-      process.on('warning', (warning) =>
-        console.error(warning.toString().toUpperCase())
-      );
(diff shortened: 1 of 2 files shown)


packages/cli-repl/src/mongosh-repl.spec.ts
```diff
@@ -1286,6 +1286,9 @@ describe('MongoshNodeRepl', function () {
     });
 
     it('does not show a prompt', async function () {
+      process.on('warning', (warning) =>
+        console.error(warning.toString().toUpperCase())
+      );
```
addaleax, line 1291: Debugging leftover, right? :)

addaleax (approved):

addaleax:
https://github.com/mongodb-js/mongosh/pull/1725 should have taken care of the genuine failures in CI here, I think you’re good to go :+1:

PR #1726, opened 2023-11-07
fix(shell-api): handle db name conflicts with Object.prototype MONGOSH-1630

addaleax:


kmruiz (approved):
Fancy trick using Object.create(null) to not have a prototype, never seen it but it's cool.

PR #1733, opened 2023-11-09
chore: add tests for connecting with a profile from an aws credentials file MONGOSH-1609

lerouxb:


packages/cli-repl/src/run.ts
```diff
@@ -220,6 +221,7 @@ async function main() {
     });
   } catch (e: any) {
     console.error(`${e?.name}: ${e?.message}`);
+    console.error('full error: ' + util.inspect(e, { depth: null }));
```
addaleax, line 224: Leftover? (Fwiw, I do this a ton too…)

packages/e2e-tests/test/e2e-aws.spec.ts
```diff
@@ -295,4 +298,86 @@ describe('e2e AWS AUTH', function () {
       });
     });
   });
+
+  context('with aws profile', function () {
+    const awsConfigPath = path.join(
+      os.tmpdir(),
+      `aws-config-${new Date().getTime()}-${Math.random()}`
+    );
```
addaleax, line 306: `useTmpdir()` from `./repl-helpers.ts`?

packages/cli-repl/src/run.ts
lerouxb: yeah. There's nothing more there anyway ;)

packages/e2e-tests/test/e2e-aws.spec.ts
lerouxb: Didn't know about that. copy/pasted this from another test.

packages/cli-repl/src/run.ts
lerouxb: Would be nice if we could access the most recent error from the test somehow.

lerouxb:
I suppose we can always re-open this if we ever need it.

PR #1751, opened 2023-11-24
fix(deps): build kerberos in RTLD mode for Linux executables MONGOSH-1628

addaleax:
##### fix(deps): build kerberos in RTLD mode for Linux executables MONGOSH-1628

- Update the Kerberos addon to the latest version (2.1.0).
- Build the Kerberos addon using RTLD for Linux mongosh executables
  (i.e. opening the system Kerberos libraries using `dlopen()` to avoid
  the symbol conflict with OpenSSL 3 on RHEL8).
- Update the `.deb` definitions to account for the fact that the
  Kerberos system libraries are no longer a strict dependency
  of mongosh.
- Include the Kerberos version in the `--build-info` output.
- Make sure that if loading the Kerberos addon fails during a
  connection attempt, a meaningful error message gets passed
  to the user, instead of the driver’s default one (which
  swallows all information about the root cause).
- As a drive-by, also do that for mongodb-client-encryption; we
  occasionally receive bug reports about CSFLE/QE not working on
  homebrew installations, and this may help with that.

### DIFF
--- a/config/build.conf.js
+++ b/config/build.conf.js
@@ -165,10 +165,11 @@ module.exports = {
               []
           )
         ],
-        debDepends: 'libc6 (>= 2.17), libgssapi-krb5-2' + (
+        debDepends: 'libc6 (>= 2.17)' + (
           SHARED_OPENSSL_TAG === 'openssl11' ? ', libssl1.1' :
             SHARED_OPENSSL_TAG === 'openssl3' ? ', libssl3' : ''
         ),
+        debRecommends: 'libgssapi-krb5-2',
         fullName: 'MongoDB Shell',
         version: CLI_REPL_PACKAGE_JSON.version,
         description: CLI_REPL_PACKAGE_JSON.description,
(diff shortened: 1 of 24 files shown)


config/build.conf.js
```diff
@@ -165,10 +165,11 @@ module.exports = {
               []
           )
         ],
-        debDepends: 'libc6 (>= 2.17), libgssapi-krb5-2' + (
+        debDepends: 'libc6 (>= 2.17)' + (
           SHARED_OPENSSL_TAG === 'openssl11' ? ', libssl1.1' :
             SHARED_OPENSSL_TAG === 'openssl3' ? ', libssl3' : ''
         ),
+        debRecommends: 'libgssapi-krb5-2',
```
addaleax, line 172: (Kerberos system libraries are no longer strictly a dependency for mongosh)

packages/build/src/compile/signable-compiler.ts
```diff
@@ -168,6 +168,10 @@ export class SignableCompiler {
         // Custom env vars for sccache:
         AWS_ACCESS_KEY_ID: process.env.DEVTOOLS_CI_AWS_KEY,
         AWS_SECRET_ACCESS_KEY: process.env.DEVTOOLS_CI_AWS_SECRET,
+        // https://jira.mongodb.org/browse/MONGOSH-1628
+        ...(process.platform === 'linux' && {
+          GYP_DEFINES: 'kerberos_use_rtld=true',
+        }),
```
addaleax, line 174: ^ This + the version bump are the entire fix, the rest is just testing/cleanup :)

gribnoysup (approved):

PR #1754, opened 2023-11-27
feat: sample sessions sent to telemetry MONGOSH-1651

kmruiz:


packages/cli-repl/src/cli-repl.ts
```diff
@@ -43,6 +43,7 @@ import { promisify } from 'util';
 import { getOsInfo } from './get-os-info';
 import { UpdateNotificationManager } from './update-notification-manager';
 import { markTime } from './startup-timing';
+import { SampledAnalytics } from '@mongosh/logging/lib/analytics-helpers';
```
addaleax, line 46: This is probably not the import you were going for? `SampledAnalytics` should be exported from the `@mongosh/logging` package itself

packages/logging/src/analytics-helpers.spec.ts
```diff
@@ -245,4 +249,51 @@ describe('analytics helpers', function () {
...
+      event: 'hi',
+      properties: { mongosh_version: '1.2.3', session_id: 'abc' },
+    };
+
+    afterEach(function () {
+      delete process.env.MONGOSH_ANALYTICS_SAMPLE;
+    });
+
+    it('should override sampling with the MONGOSH_ANALYTICS_SAMPLE environment variable', function () {
+      process.env.MONGOSH_ANALYTICS_SAMPLE = 'true';
```
addaleax, line 267: This should probably also be set for the existing tests for telemetry in the cli-repl package?

packages/logging/src/analytics-helpers.ts
```diff
@@ -378,3 +378,72 @@ export class ThrottledAnalytics implements MongoshAnalytics {
...
+    configuration.sampling ??= {
+      percentage: 30,
+      samplingFunction: (percentage) => Math.random() * 100 < percentage,
+    };
+
+    const shouldBeSampled = configuration.sampling.samplingFunction(
+      configuration.sampling.percentage
+    );
+
+    this.isEnabled = !!process.env.MONGOSH_ANALYTICS_SAMPLE || shouldBeSampled;
```
addaleax, line 412: I think it might be a bit surprising that this is where we check the environment variable value – generally speaking, we try to keep as much interaction with the process state in `cli-repl.ts` and `run.ts` in the cli-repl package, and in the other parts of the codebase just rely on explicitly passing information

I’d still think that it’s fair to simplify the interface here to taking a `boolean` or maybe `() => boolean`, passing something like `() => !!process.env.MONGOSH_ANALYTICS_SAMPLE || Math.random() < 0.3` from the cli-repl package doesn’t seem too complex or untested to me

packages/cli-repl/src/cli-repl.ts
kmruiz: You are right, fixing it.

packages/logging/src/analytics-helpers.spec.ts
kmruiz: It makes sense, testing it also at the cli-repl package.

addaleax: No, I mean, there are existing tests, right now they wouldn’t work because they would be subject to the sampling and therefore flaky, right?

kmruiz: Ah yes, what I was thinking is on adding a beforeEach on all telemetry tests to enable the environment variable, and adding a test when it's disabled.

packages/logging/src/analytics-helpers.ts
```diff
@@ -378,3 +378,34 @@ export class ThrottledAnalytics implements MongoshAnalytics {
...
+
+  identify(message: AnalyticsIdentifyMessage): void {
+    this.isEnabled && this.target.identify(message);
+  }
+
+  track(message: AnalyticsTrackMessage): void {
+    this.isEnabled && this.target.track(message);
+  }
+
+  flush(callback: (err?: Error | undefined) => void): void {
```
kmruiz, line 408: @addaleax in case the telemetry is disabled, what should we do? ToggleableAnalytics it still runs the flush and the callback even if it's toggled off. I suppose it should be fine to do the same here, as the event queue should be empty.

addaleax: Yes, I think that’s fine – shouldn’t be doing anything in this case :)

packages/cli-repl/src/cli-repl.ts
```diff
@@ -500,12 +501,16 @@ export class CliRepl implements MongoshIOProvider {
...
+      new SampledAnalytics({
+        target: new ThrottledAnalytics({
+          target: this.segmentAnalytics,
+          throttle: {
+            rate: 30,
+            metadataPath: this.shellHomeDirectory.paths.shellLocalDataPath,
+          },
+        }),
+        sampling: () =>
+          !!process.env.MONGOSH_ANALYTICS_SAMPLE || Math.random() < 0.3,
```
addaleax, line 513: Okay, just to check before approving this, is 30% really the value we got from the analytics team? That sounds much higher than what I was expecting (1% or so)

addaleax (commented):
LGTM mod sampling rate question :)

packages/cli-repl/src/cli-repl.ts
kmruiz: It's not decided yet, it's just a placeholder until the analytics team gives me an answer :D. I was waiting for Gaurab to provide me the correct sampling.

If you want we can keep this PR and close it with the proper percentage or merge it and do a small PR changing the number once it's decided.

No strong opinion here.

kmruiz: Let's wait for tomorrow and see if we get a quick answer, and if they mention that they need to think about it more thoroughly I'll merge.

lerouxb (approved):

addaleax (approved):
:rocket:

PR #1756, opened 2023-11-27
feat(shell-api): support `type` field for search index creation MONGOSH-1631

addaleax:
Now that SERVER-83107 is complete, we can add this to mongosh.


packages/shell-api/src/collection.spec.ts
```diff
@@ -2584,6 +2584,61 @@ describe('Collection', function () {
...
+                type: 'vectorSearch',
+                definition: { mappings: { dynamic: true } },
+              },
+            ]
+          );
+        });
+      });
+
+      context('with name, options and type === search', function () {
+        it('calls serviceProvider.createIndexes', async function () {
```
lerouxb, line 2609: serviceProvider.createSearchIndexes?

packages/shell-api/src/collection.spec.ts
```diff
@@ -2584,6 +2584,61 @@ describe('Collection', function () {
           );
         });
       });
+
+      context('with name, options and type !== search', function () {
+        it('calls serviceProvider.createIndexes', async function () {
```
lerouxb, line 2589: serviceProvider.createSearchIndexes?

packages/shell-api/src/collection.spec.ts
```diff
@@ -2584,6 +2584,61 @@ describe('Collection', function () {
...
+          expect(serviceProvider.createSearchIndexes).to.have.been.calledWith(
+            'db1',
+            'coll1',
+            [{ name: 'my-index', definition: { mappings: { dynamic: true } } }]
+          );
+        });
+      });
+
+      context('with options and type but no name', function () {
+        it('calls serviceProvider.createIndexes', async function () {
```
lerouxb, line 2623: and here

lerouxb (approved):

packages/shell-api/src/collection.spec.ts
addaleax:
```suggestion
        it('calls serviceProvider.createSearchIndexes', async function () {
```

PR #1761, opened 2023-11-29
chore: add a new “Startup Time” event for sessions MONGOSH-1653

kmruiz:


packages/logging/src/setup-logger-and-telemetry.ts
```diff
@@ -147,29 +147,33 @@ export function setupLoggerAndTelemetry(
...
-  bus.on('mongosh:start-session', function (args: SessionStartedEvent) {
-    const params = {
-      session_id,
-      userId,
-      telemetryAnonymousId,
-      ...args,
-    };
+  const toSnakeCase = (str: string): string => {
+    const matches = str.match(
+      /[A-Z]{2,}(?=[A-Z][a-z]+[0-9]*|\b)|[A-Z]?[a-z]+[0-9]*|[A-Z]|[0-9]+/g
```
Anemy, line 152: Should we have some tests for this regex? It might also be nice to know what the expected input -> output might be in a comment, a bit hard to grasp at a glance.

packages/cli-repl/src/startup-timing.ts
```diff
@@ -1,30 +1,54 @@
...
+      return data.map(([category, label, time]) => [
+        category,
+        label,
+        Number(time - data[0][2]),
+      ]);
+    },
+  };
+}
+
+export function summariseTimingData(timingData: [string, string, number][]): {
```
baileympearson, line 32: the test in `startup-timing.spec.ts` implies that `timingData` is always of type `[TimingCategory, string, number][]`.  Is that correct?  If so - can we make the type stricter here?

packages/cli-repl/src/cli-repl.ts
```diff
@@ -413,6 +420,11 @@ export class CliRepl implements MongoshIOProvider {
         commandLineLoadFiles,
         evalScripts
       );
+
+      this.bus.emit('mongosh:start-session', {
```
baileympearson, line 424: just a question - why are we emitting `start-session` after evaluating the user's scripts in non-interactive mode?  It seems like when executing scripts, the session has been started after all dependencies have been loaded but before executing any scripts.

packages/logging/src/setup-logger-and-telemetry.ts
kmruiz: Yeah, the thing is that the API is also used in C++ in BoxedNode (with a different implementation, we do dynamic linking), so to ensure that we don't break anything because we are changing how BoxedNode works (for some reason), I still force to snake case the key before sending the event.

The snake casing is tested indirectly in packages/logging/src/setup-logger-and-telemetry.spec.ts, but I agree that it makes sense to have a specific tests for this.

packages/cli-repl/src/startup-timing.ts
kmruiz: We could, but it would be misleading. The thing is that when bundle the application with BoxedNode, timingData can have also strings not included in TimingCategory (it's a C++ library we are linking). Summarise Timing Data should work for both the cases it's using the BoxedNode implementation of getTimingData and the TypeScript version (in the dev environment).

packages/cli-repl/src/cli-repl.ts
kmruiz: Ah, you are right, totally right. start-session should be emitted just before `mongosh:start-loading-cli-scripts`.

Thanks a lot for noticing! I just missed this one.

packages/cli-repl/src/startup-timing.ts
kmruiz: Just to clarify: the only reason TimingCategory exists is to make sure that at least, on the typescript side of things (99% of times) we have a clear convention on what timing categories do. But in BoxedNode we might need sometimes a few more categories, as it does all the bootstrapping.

package-lock.json
```diff
@@ -31575,7 +31575,8 @@
       "dependencies": {
         "@mongosh/async-rewriter2": "0.0.0-dev.0",
         "@mongosh/history": "0.0.0-dev.0",
-        "@mongosh/shell-api": "0.0.0-dev.0"
+        "@mongosh/shell-api": "0.0.0-dev.0",
+        "@mongosh/types": "0.0.0-dev.0"
```
baileympearson, line 31579: I think this is leftover from when `mongosh/types` was a prod dependency of shell-evaluator and should be removed

packages/cli-repl/src/startup-timing.ts
baileympearson: That makes sense but it's sort of misleading that `linkTimingInterface` uses a `TimingInterface` and that when boxednode is linked, it supposedly satisfies that interface.  

Not that important, but this is the kind of "Typescript lie" that I like to avoid because it causes confusion for future devs, and can lead to scenarios where bugs are introduced because developers rely on the types

packages/logging/src/setup-logger-and-telemetry.ts
```diff
@@ -58,6 +59,31 @@ class MultiSet<T extends Record<string, any>> {
   }
 }
 
+/**
+ * It transforms a random string into snake case. Snake case is completely
+ * lowercase and uses '_' to separate words. For example:
+ *
+ * 'Random String' => 'random_string'
```
baileympearson, line 66:
```suggestion
 * It transforms a random string into snake case. Snake case is completely
 * lowercase and uses '_' to separate words. For example:
 *
 * This function defines a "word" as a sequence of characters until the next `.` or capital letter.
 *
 * 'Random String' => 'random_string'
```

just a suggestion.  is my reasoning right here?  It's not actually parsing "words" out of the string.  iiuc, `AaaAaa` would map to `aaa_aaa` and `Aaa` isn't a word in any language

baileympearson (changes requested):
some comments, but one change requested in the package lock

packages/cli-repl/src/startup-timing.ts
kmruiz: We can use TimingCategory | string if you think it's better in summariseTimingData. Semantics are the same, but it's more clear for a developer that this can be either the enum or other unknown strings from other sources.

WDYT?

packages/logging/src/setup-logger-and-telemetry.ts
kmruiz: AaaAaa would be two words in Pascal Case: https://www.freecodecamp.org/news/snake-case-vs-camel-case-vs-pascal-case-vs-kebab-case-whats-the-difference/#pascal-case

I can add the suggestion to make it clearer, I think it's great just to make sure there is no misunderstanding later if someone uses this function.

baileympearson (approved):

packages/types/src/index.ts
```diff
@@ -585,3 +597,25 @@ function isValidUrl(url: string): boolean {
...
+  ResourceFileLoading: 'ResourceFileLoading',
+  AsyncRewrite: 'AsyncRewrite',
+  Eval: 'Eval',
+  EvalFile: 'EvalFile',
+  Telemetry: 'Telemetry',
+  Main: 'Main',
+} as const;
+
+export type TimingCategory =
+  (typeof TimingCategories)[keyof typeof TimingCategories];
```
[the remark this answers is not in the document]
addaleax: Yeah, this would have been a small nice-to-have :)

packages/cli-repl/src/startup-timing.ts
```diff
@@ -1,30 +1,60 @@
...
+export function summariseTimingData(
+  timingData: [string | TimingCategory, string, number][]
+): {
+  [key: string]: number;
+} {
+  const durationByCategory = new Map<string, number>();
+  let lastTimestamp = 0;
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  for (const [category, _, time] of timingData) {
```
addaleax, line 45: Also, just fyi:

```suggestion
  for (const [category,, time] of timingData) {
```

PR #1780, opened 2024-01-08
chore: add a config for running CI against the nightly driver MONGOSH-1473

lerouxb:
For use with https://spruce.mongodb.com/project/mongosh/settings/periodic-builds. Or any other way that you can specify an evergreen config file other than the default one.

PR #1788, opened 2024-01-16
chore: only run the driver nightly on x64 linux, npm i --force MONGOSH-1473

lerouxb:
Follow-up to https://github.com/mongodb-js/mongosh/pull/1780

npm ci doesn't work if your package lock doesn't match what it should be so we have to npm i but then npm i doesn't work on all the exotic platforms due to all the deps we can't install there.. But there's no point running the nightly driver on every platform anyway, so let's only run it on x64 linux while at it.

This pull request now reuses the same templating we have to make .evergreen.yml to also make .evergreen-nightly-driver.yml so we can have one with all the build variants we don't want disabled.

All this to work around the limitations of evergreen's periodic builds. Another option would be to use the evergreen CLI to filter the variants/tasks that will be run, but then we need that set up somewhere we can run it. (github workflows, perhaps?)

### DIFF
--- a/.evergreen-nightly-driver.yml
+++ b/.evergreen-nightly-driver.yml
@@ -1,7 +1,13970 @@
+# Regenerate using `npm run update-evergreen-config`
+# https://jira.mongodb.org/browse/EVG-20276
+unset_function_vars: true
+
 # https://github.com/evergreen-ci/evergreen/blob/main/docs/Project-Configuration/Parameterized-Builds.md#project-config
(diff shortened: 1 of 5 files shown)


lerouxb:
I can't tell if this works because if you schedule a patch manually:

addaleax (approved):

lerouxb:
Closing in favour of https://github.com/mongodb-js/mongosh/pull/1790

PR #1795, opened 2024-01-22
deps(logging): update segment analytics library, use async flush MONGOSH-1690

Anemy:
MONGOSH-1690 also addresses SNYK-JS-AXIOS-6144788


packages/logging/src/analytics-helpers.ts
```diff
@@ -351,31 +349,28 @@ export class ThrottledAnalytics implements MongoshAnalytics {
...
-        );
-        await this.unlock();
-        this.target.flush(callback);
-      } catch (e) {
-        callback(e as Error);
-      }
-    });
+    try {
+      await this.restorePromise;
+    } catch (err) {
```
addaleax, line 364:
```suggestion
    } catch {
```

Doesn't the linter complain about an unused variable otherwise?

addaleax (approved):

Anemy:
Re-requested review since we did a few changes and this pr is from a few weeks ago.

addaleax (approved):

PR #1798, opened 2024-01-23
sign using garasign during the packaging phase

baileympearson:


scripts/docker/rocky8-package.Dockerfile
```diff
@@ -14,5 +14,7 @@ RUN npm i -g npm@9.x
 # set to /root and consequently $HOME/.npm was not writable
 RUN mkdir -p /tmp/home
 ENV HOME=/tmp/home
+RUN mkdir "/tmp/home/.npm"
+RUN sudo chown -R 1000:1000 "/tmp/home/.npm"
```
addaleax, line 18: Why do we need sudo for this? It's a docker build script, it shouldn't need it. A bit more generally speaking, could you add comments that explain why these lines are necessary here but not for other Dockerfiles?

.evergreen.yml
```diff
@@ -134,6 +134,83 @@ functions:
         shell: bash
         script: |
           tar xvzf compiled-ts.tgz
+  spawn-signing-server:
+    # spawn
+    - command: host.create
+      type: setup
+      params:
+        provider: ec2
+        distro: ubuntu2004-large
```
addaleax, line 143: -small should be fine if we're just signing on it, no?

scripts/docker/rocky8-package.Dockerfile
baileympearson: I'm not sure the best path forward here, maybe someone will have thoughts.  Without these changes, I get the following error when I run `package-and-upload-artifact.sh`:

```
npm ERR! Your cache folder contains root-owned files, due to a bug in
npm ERR! previous versions of npm which has since been addressed.
npm ERR! 
npm ERR! To permanently fix this problem, please run:
npm ERR!   sudo chown -R 502:20 "/tmp/home/.npm"
```

The changes to package-and-upload-artifact allow the host to access files written by the container (by running the container as the current user) and are necessary for successful signing.  I'm not sure exactly what is causing this error when I don't `sudo chown` the npm cache.

addaleax: Do you know which npm command writes to `/tmp/home/.npm` with root permissions? If you want to look at that more closely, I think that would be the important question to answer

But also, more specifically, you should not need `sudo` in the Dockerfile, you should be able to just run `chown`

addaleax:
@baileympearson I think this can just be closed now after 2da3c3159545e348f411d77e321f6ada7d170b85?

PR #1802, opened 2024-01-24
[Do Not Merge] PoC for shell-api autocomplete type definitions

addaleax:
```sh
cd packages/shell-api && \
  npm run compile && \
  npx api-extractor run ; \
  npx ts-node bin/api-postprocess.ts ; \
  cat lib/api-processed.d.ts
```

### DIFF
--- a/packages/shell-api/src/mongo.ts
+++ b/packages/shell-api/src/mongo.ts
@@ -49,6 +49,7 @@ import { CommandResult } from './result';
 import { redactURICredentials } from '@mongosh/history';
 import { asPrintable, ServerVersions, Topologies } from './enums';
 import Session from './session';
+import type { GenericServerSideSchema, StringKey } from './helpers';
 import {
   assertArgsDefinedType,
   processFLEOptions,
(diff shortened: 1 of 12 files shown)


packages/shell-api/src/mongo.ts
```diff
@@ -264,14 +268,16 @@ export default class Mongo extends ShellApiClass {
...
     assertArgsDefinedType([db], ['string'], 'Mongo.getDB');
     this._instanceState.messageBus.emit('mongosh:getDB', { db });
     return this._getDb(db);
   }
 
   @returnType('Collection')
-  getCollection(name: string): Collection {
+  getCollection<KD extends StringKey<M>, KC extends StringKey<M[KD]>>(
+    name: `${KD}.${KC}`
+  ): Collection<M, M[KD], M[KD][KC]> {
```
addaleax, line 280: @kmruiz Wanted to highlight this, doubt many people will actually use `.getCollection()` on a Mongo object ... but this is interesting, TS is not able to autocomplete the `name` argument for me (which, like, fair enough), but if I pass a valid `<db>.<collection>` namespace, it will return a collection with the correct schema, kind of interesting that it only partially works

kmruiz: That feels like a bug right? If I recall, it should be able to autocomplete on string templates. But yeah everyone will use db.`collection` as you said, I don't see any reason now for not suggesting the usage of collections using property access instead of getCollection.

addaleax: I think it's struggling a bit with the fact that the second part (collection name) after the `.` depends on the first part (database name)

kmruiz: Makes sense, this seems like a hard problem to solve for a compiler to be fair.

PR #1804, opened 2024-01-29
chore(deps): bump mongodb-constants to 0.8.10

addaleax:


lerna.json
```diff
@@ -1,6 +1,6 @@
 {
   "packages": [
-    "config/*",
+    "configs/*",
```
addaleax, line 3: I don't know why so many things were working before, but this was definitely a typo (see also [Slack thread](https://mongodb.slack.com/archives/GJ2CBBY86/p1705983400964289))

Anemy (approved):

PR #1805, opened 2024-01-29
fix(shell-api): improve error message on reading from secondary via runCommand MONGOSH-1679

paula-stacho:
https://jira.mongodb.org/browse/MONGOSH-1679

Changes:
- `runCommand` in the `shell-api` catches `NotPrimaryNoSecondaryOk` and applies a custom message
- the custom `runCommand` message is ignored in the `mongo-errors` post-processing

Note: I used `codeName` instead of `code` in the new part for readability - I hope it's okay. Let me know if you think it's safe to switch the `mongo-errors` section to also use `codeName`.

### DIFF
--- a/packages/shell-api/src/database.spec.ts
+++ b/packages/shell-api/src/database.spec.ts
@@ -323,6 +323,21 @@ describe('Database', function () {
           }
         );
       });
+
+      it('rephrases the "NotPrimaryNoSecondaryOk" error (version 2)', async function () {
+        const originalError = {
+          message: 'old message',
+          codeName: 'NotPrimaryNoSecondaryOk',
+          code: 13435,
+        }; // Partial<MongoServerError>
+        serviceProvider.runCommandWithCheck.rejects(originalError);
+        const caughtError = await database
+          .runCommand({ someCommand: 'someCollection' })
+          .catch((e) => e);
+        expect(caughtError.message).to.contain(
+          'e.g. db.runCommand({command}, { readPreference: "secondaryPreferred"})'
+        );
+      });
     });
 
     describe('adminCommand', function () {
--- a/packages/shell-api/src/database.spec.ts
+++ b/packages/shell-api/src/database.spec.ts
@@ -31,6 +31,8 @@ import {
   MongoshUnimplementedError,
 } from '@mongosh/errors';
 import type { ClientEncryption } from './field-level-encryption';
+import type { MongoServerError } from 'mongodb';
+
 chai.use(sinonChai);
 
 describe('Database', function () {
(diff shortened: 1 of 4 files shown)


packages/shell-api/src/database.ts
```diff
@@ -365,11 +365,19 @@ export default class Database extends ShellApiWithMongoClass {
...
-      this._emitDatabaseApiCall('runCommand', { cmd, options });
+    try {
+      const hiddenCommands = new RegExp(HIDDEN_COMMANDS);
+      if (!Object.keys(cmd).some((k) => hiddenCommands.test(k))) {
+        this._emitDatabaseApiCall('runCommand', { cmd, options });
+      }
+      return await this._runCommand(cmd, options);
+    } catch (error: any) {
+      if (error.codeName === 'NotPrimaryNoSecondaryOk') {
+        const message = `not primary - consider passing the readPreference option e.g. db.runCommand({command}, { readPreference: "secondaryPreferred"})`;
```
addaleax, line 376:
```suggestion
        const message = `not primary - consider passing the readPreference option e.g. db.runCommand({ command }, { readPreference: 'secondaryPreferred' })`;
```

Optional, but I think it's worth being consistent about the spacing around `{` and `}` (the quotes are even less important, but we've moved towards being consistent about single quotes recently e.g. in https://github.com/mongodb-js/mongosh/commit/cbbd2229d3e0e4cf39ea218c5f99e29bc3344581

packages/shell-api/src/mongo-errors.ts
```diff
@@ -9,8 +9,10 @@ const ERROR_REPHRASES: MongoErrorRephrase[] = [
   {
     // NotPrimaryNoSecondaryOk (also used for old terminology)
     code: 13435,
-    replacement:
-      'not primary and secondaryOk=false - consider using db.getMongo().setReadPref() or readPreference in the connection string',
+    replacement: (message) =>
+      message.includes('db.runCommand')
+        ? message
+        : 'not primary and secondaryOk=false - consider using db.getMongo().setReadPref() or readPreference in the connection string',
```
addaleax, line 15:
```suggestion
        : 'not primary - consider using db.getMongo().setReadPref() or readPreference in the connection string',
```

to match the other adjusted error message?

packages/shell-api/src/database.spec.ts
```diff
@@ -323,6 +323,21 @@ describe('Database', function () {
           }
         );
       });
+
+      it('rephrases the "NotPrimaryNoSecondaryOk" error (version 2)', async function () {
+        const originalError = {
+          message: 'old message',
+          codeName: 'NotPrimaryNoSecondaryOk',
+          code: 13435,
+        }; // Partial<MongoServerError>
```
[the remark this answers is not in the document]
paula-stacho: Can do! Originally I used the constructor and then I took it away as I noticed a note discouraging from outside use, but I read it again and it's specific to the constructor. If the type changes too much we'd want to know anyway

PR #1808, opened 2024-02-01
feat(shell-api): account for unsharded collections in sharding catalog COMPASS-7602

addaleax:


packages/shell-api/src/helpers.ts
```diff
@@ -487,33 +498,45 @@ export async function getPrintableShardStatus(
...
+        .toArray())(),
+    (async () =>
+      await (
+        await configDB.getCollection('collections').find({
+          ...onlyShardedCollectionsInConfigFilter,
+        })
+      )
+        .sort({ _id: 1 })
+        .toArray())(),
+  ]);
```
addaleax, line 517: This isn't directly related to the original change, but it seems that sharding collections doesn't necessarily introduce entries in the `config.databases` collection, which makes sense given that sharding doesn't explicitly need to be enabled for databases anymore since https://jira.mongodb.org/browse/SERVER-60926

packages/shell-api/src/helpers.ts
```diff
@@ -487,33 +498,45 @@ export async function getPrintableShardStatus(
...
-          .find({ _id: new RegExp('^' + escapeRegex(db._id) + '\\.') })
-      )
-        .sort({ _id: 1 })
-        .toArray();
+  result.databases = (
+    await Promise.all(
+      databases.map(async (db) => {
+        const colls = collections.filter((coll) =>
+          coll._id.startsWith(db._id + '.')
+        );
```
addaleax, line 536: If we need to get all collection infos anyway, we might as well avoid doing the extra query that we did _for each database_ here

gribnoysup (approved):

PR #1817, opened 2024-02-09
fix(ci): modify shell.exec as per evergreen instructions to avoid leaking any secrets MONGOSH-1704

himanshusinghs:


.evergreen/run-evergreen-release.sh
```diff
@@ -0,0 +1,8 @@
+#! /usr/bin/env bash
+set -e
+
+echo "//registry.npmjs.org/:_authToken=${devtoolsbot_npm_token}" > .npmrc
```
addaleax, line 4:
```suggestion
echo "//registry.npmjs.org/:_authToken=${devtoolsbot_npm_token}" > .npmrc
set -x
```

addaleax (approved):

himanshusinghs:
Merging this since the only failing test is the vscode one which is also failing on main. Investigating that separately.

PR #1818, opened 2024-02-09
chore: small fixes from perf investigation

addaleax:
No user-facing impact at this time, just two things that I noticed:
- We were consuming Buffers instead of strings in our example script
- We didn't actually run the async rewriter during startup snapshot building, so any lazy loading performed by babel was delayed until runtime; we can save ~75ms startup time by just doing this as part of the snapshot (this will only take effect once snapshots are enabled again, of course).

### DIFF
--- a/packages/shell-evaluator/src/shell-evaluator.ts
+++ b/packages/shell-evaluator/src/shell-evaluator.ts
@@ -25,7 +25,9 @@ try {
 }
 if (v8?.startupSnapshot?.isBuildingSnapshot?.()) {
   v8.startupSnapshot.addSerializeCallback(() => {
+    // Ensure that any lazy loading performed by Babel is part of the snapshot
     eval(new AsyncWriter().runtimeSupportCode());
+    eval(new AsyncWriter().process('1+1'));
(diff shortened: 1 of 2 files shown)


lerouxb (approved):

packages/shell-evaluator/src/shell-evaluator.ts
```diff
@@ -25,7 +25,9 @@ try {
 }
 if (v8?.startupSnapshot?.isBuildingSnapshot?.()) {
   v8.startupSnapshot.addSerializeCallback(() => {
+    // Ensure that any lazy loading performed by Babel is part of the snapshot
     eval(new AsyncWriter().runtimeSupportCode());
+    eval(new AsyncWriter().process('1+1'));
```
kmruiz, line 30: Just to make sure I understand, we are forcing the evaluation of code so we ensure that babel runs and so the async rewriter set up?

kmruiz (approved):

packages/shell-evaluator/src/shell-evaluator.ts
addaleax: Yup, that's exactly it 👍

PR #1845, opened 2024-02-29
fix: upgrade to node-fetch@3.X MONGOSH-1702

paula-stacho:
https://jira.mongodb.org/browse/MONGOSH-1702

## Description

The new version doesn't work with our tsconfig, that's why the "skipLibCheck". I hope it's okay I added it to the common config, I've seen we already have it in [@mongodb-js/tsconfig-devtools](https://github.com/mongodb-js/devtools-shared/blob/f52aa18b703fffd04ef1f69ba991ca033be03049/configs/tsconfig-devtools/tsconfig.common.json#L13)

Code changes are due to:

1. `response.body` can be null - this was probably already true before, just the types were weaker. the upgrade guide only mentions removing a string option. from what I see we've been treating it as a stream. https://github.com/node-fetch/node-fetch/blob/main/docs/v3-UPGRADE-GUIDE.md#reqbody-can-no-longer-be-a-string
2. `response.json()` is `Promise<unknown>`. again, the types were probably just weaker before and it passed as any.
3. node-fetch is now an ES module https://github.com/node-fetch/node-fetch/blob/main/docs/v3-UPGRADE-GUIDE.md#converted-to-es-module -> this one created considerable issues in different builds and per @addaleax's suggestion led to the creation of the `import-node-fetch` wrapper
4. `once` is now recommended when listening for errors https://github.com/node-fetch/node-fetch/blob/main/docs/v3-UPGRADE-GUIDE.md#a-stream-pipeline-is-now-used-to-forward-errors

Note: You'll see the code updates in the build package as well, but not the version change. Originally I thought I might as well upgrade node-fetch everywhere, but the ES Module doesn't work with `ts-node`. Looks like [ESM support in ts-node comes with a warning](https://github.com/TypeStrong/ts-node/issues/1007), and this isn't directly connected to the deprecation warning in homebrew, so I think we can leave this on 2.x for now.


packages/e2e-tests/test/fixtures/curl.js
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env node
 'use strict';
-const fetch = require('node-fetch');
+import fetch from 'node-fetch';
```
addaleax, line 3:
```
$ ./packages/e2e-tests/test/fixtures/curl.js 
/Users/addaleax/src/mongosh/packages/e2e-tests/test/fixtures/curl.js:3
const fetch = require('node-fetch');
              ^

Error [ERR_REQUIRE_ESM]: require() of ES Module /Users/addaleax/src/mongosh/packages/e2e-tests/node_modules/node-fetch/src/index.js from /Users/addaleax/src/mongosh/packages/e2e-tests/test/fixtures/curl.js not supported.
Instead change the require of index.js in /Users/addaleax/src/mongosh/packages/e2e-tests/test/fixtures/curl.js to a dynamic import() which is available in all CommonJS modules.
    at Object.<anonymous> (/Users/addaleax/src/mongosh/packages/e2e-tests/test/fixtures/curl.js:3:15) {
  code: 'ERR_REQUIRE_ESM'
}

Node.js v20.11.1
```

probably needs to be dynamically imported inside the main function here :)

packages/cli-repl/src/update-notification-manager.ts
```diff
@@ -95,9 +95,15 @@ export class UpdateNotificationManager {
...
+    if (!response.body) {
+      throw new Error(
+        `Missing body from ${updateURL}: ${response.status} ${response.statusText}`
+      );
+    }
+
     if (response.status === 304 /* Not Modified, i.e. ETag matched */) {
       response.body
-        .on('error', () => {
+        .once('error', () => {
```
addaleax, line 106: This is probably fine either way – we shouldn't be getting more than one `error` event – but just out of curiosity, did anything in node-fetch change around error handling?

packages/e2e-tests/test/fixtures/curl.js
paula-stacho: I think I already fixed that

packages/cli-repl/src/update-notification-manager.ts
paula-stacho: yes, https://github.com/node-fetch/node-fetch/blob/main/docs/v3-UPGRADE-GUIDE.md#a-stream-pipeline-is-now-used-to-forward-errors

packages/snippet-manager/package.json
```diff
@@ -42,15 +42,15 @@
...
+    "node-fetch": "^3.3.2",
     "tar": "^6.1.15"
   },
   "devDependencies": {
     "@mongodb-js/eslint-config-mongosh": "^1.0.0",
     "@mongodb-js/prettier-config-devtools": "^1.0.1",
     "@mongodb-js/tsconfig-mongosh": "^1.0.0",
     "@types/cross-spawn": "^6.0.2",
-    "@types/node-fetch": "^2.6.4",
+    "@types/node-fetch": "^2.6.11",
```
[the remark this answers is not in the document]
paula-stacho: ah, no we don't, I forgot to remove this

packages/cli-repl/src/update-notification-manager.ts
addaleax: Interesting! Thank you for the link :)

packages/e2e-tests/test/fixtures/curl.js
addaleax: Fwiw, this was with the most recent version of this PR, so it seems like it's still failing here? I don't know how it would work – this is a `.js` file and those don't support top-level `import`. You could try renaming this to `.mjs`, if you want, that might be a "cool" thing to do to fix it :)

paula-stacho: hmm I didn't see this error in the recent version, but there are similar errors from mocha, which I'm currently trying to fix.

packages/cli-repl/src/update-notification-manager.ts
```diff
@@ -95,9 +95,15 @@ export class UpdateNotificationManager {
         : {},
     });
 
+    if (!response.body) {
```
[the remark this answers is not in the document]
paula-stacho: thanks for catching, will fix!

packages/import-node-fetch/src/index.ts
```diff
@@ -0,0 +1,30 @@
...
+  // Use this function to safely import the node-fetch package
+  let module;
+  try {
+    module = await import('node-fetch');
+  } catch (err: unknown) {
+    if (
+      err &&
+      typeof err === 'object' &&
+      'code' in err &&
+      err.code === 'ERR_REQUIRE_ESM'
```
addaleax, line 18: We'll want to have a `typeof __webpack_require__ === 'undefined'` check so that we don't end up trying to run a literal `import()` call if we are in a webpack environment

packages/cli-repl/src/update-notification-manager.ts
```diff
@@ -1,10 +1,10 @@
 import semver from 'semver';
 import { promises as fs } from 'fs';
-import type { RequestInfo, RequestInit, Response } from 'node-fetch';
+import importNodeFetch from '@mongosh/import-node-fetch';
 
 // 'http' is not supported in startup snapshots yet.
-const fetch = async (url: RequestInfo, init?: RequestInit): Promise<Response> =>
-  await (await import('node-fetch')).default(url, init);
+const fetch = async (url: any, init?: any): Promise<any> =>
+  await (await importNodeFetch()).default(url, init);
```
addaleax, line 7: I don't think we need this level of indirection anymore, we should be able to just call `importNodeFetch()` directly where we call `fetch()` below.

If this is necessary, well, let's maybe re-export the necessary types from `@mongosh/import-node-fetch`, it wouldn't be great to have to stick with `any` here

packages/import-node-fetch/package.json
```diff
@@ -0,0 +1,46 @@
...
+    "unitTestsOnly": true
+  },
+  "devDependencies": {
+    "@mongodb-js/eslint-config-mongosh": "^1.0.0",
+    "@mongodb-js/prettier-config-devtools": "^1.0.1",
+    "@mongodb-js/tsconfig-mongosh": "^1.0.0",
+    "depcheck": "^1.4.3",
+    "eslint": "^7.25.0",
+    "prettier": "^2.8.8",
+    "node-fetch": "^3.3.2"
```
addaleax, line 44: That's a regular dependency, not a devDependency, right?

packages/import-node-fetch/src/index.spec.ts
```diff
@@ -0,0 +1,23 @@
+import { expect } from 'chai';
+import importNodeFetch from './';
+
+describe('makeMultilineJSIntoSingleLine', function () {
```
addaleax, line 4:
```suggestion
describe('import-node-fetch', function () {
```

packages/import-node-fetch/package.json
paula-stacho: oh of course 🤦

packages/import-node-fetch/src/index.spec.ts
paula-stacho: thank you!

packages/cli-repl/src/update-notification-manager.ts
addaleax: Yeah, so, since we re-enabled startup snapshot support one of the things we cannot do anymore is to have imports of `http`/`https`/`tls` and a few other built-in Node.js modules at the top level (i.e. loaded when mongosh initially starts). That's hopefully a restriction that's going to go away in #1846, but, well, one step at a time. :)

addaleax (approved):

PR #1848, opened 2024-02-29
chore: make CI a bit greener MONGOSH-1726

addaleax:
- Specifically choose the npm version to be one that matches the Node.js version we install; this is a better version of the fix in a459c995cc6b, which made Windows CI (almost) green but left the test_n16_snippet_manager task failing due to the specific combination of Node.js 16 and npm 10.x on Windows.
- Account for upstream server changes to the PlanCache functionality.

### DIFF
--- a/.evergreen/install-node.sh
+++ b/.evergreen/install-node.sh
@@ -2,10 +2,16 @@ set -e
 set -x
 export BASEDIR="$PWD/.evergreen"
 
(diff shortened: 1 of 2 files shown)


lerouxb (approved):

KidiIT (commented):
All reviewed and approved🫶❤️

PR #1849, opened 2024-02-29
feat: replace Node.js REPL with plain vm context for script usage MONGOSH-1720

addaleax:
[Definitely recommend viewing the diff in whitespace-changes-disabled mode :)]

- Allow running `MongoshNodeRepl` instances either based on:
  - The existing mechanism for running code, which is spinning up
    a Node.js REPL and using it to evaluate code; or
  - A more lightweight mechanism that only creates a `vm` context
    and then run code using that context directly.
- Introduce a new command-line switch, `--jsContext`, that can be
  used to explicitly select the desired behavior, with possible values
  of `repl`, `plain-vm` and `auto`. The default behavior is to switch
  depending on whether the CLI will enter interactive mode or not.

Running in `plain-vm` mode will significantly improve runtime
performance (a 6× reduction of script run time in local testing),
coming from the removal of async context tracking that the REPL
uses to identify async operations which were originally spawned
from the REPL instance in question.

This lack of async context tracking comes with some implications,
in particular asynchronously thrown errors (in the Node.js sense,
e.g. `setImmediate(() => { throw ... })`) will lead to slightly
different behavior (e.g. different error code and error output).
That is probably acceptable breakage in this context.

### DIFF
--- a/packages/cli-repl/src/cli-repl.ts
+++ b/packages/cli-repl/src/cli-repl.ts
@@ -100,7 +100,7 @@ type CliUserConfigOnDisk = Partial<CliUserConfig> &
 export class CliRepl implements MongoshIOProvider {
   mongoshRepl: MongoshNodeRepl;
   bus: MongoshBus;
-  cliOptions: CliOptions;
+  cliOptions: Readonly<CliOptions>;
   getCryptLibraryPaths?: (bus: MongoshBus) => Promise<CryptLibraryPathResult>;
   cachedCryptLibraryPath?: Promise<CryptLibraryPathResult>;
   shellHomeDirectory: ShellHomeDirectory;
@@ -186,8 +186,17 @@ export class CliRepl implements MongoshIOProvider {
       this.bus.emit('mongosh:error', err, 'io');
     });
 
+    let jsContext = this.cliOptions.jsContext;
+    if (jsContext === 'auto' || !jsContext) {
+      jsContext = CliRepl.getFileAndEvalInfo(this.cliOptions)
+        .willEnterInteractiveMode
+        ? 'repl'
+        : 'plain-vm';
+    }
+
     this.mongoshRepl = new MongoshNodeRepl({
       ...options,
+      shellCliOptions: { ...this.cliOptions, jsContext },
       nodeReplOptions: options.nodeReplOptions ?? {
         terminal: process.env.MONGOSH_FORCE_TERMINAL ? true : undefined,
       },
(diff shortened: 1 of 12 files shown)


packages/cli-repl/src/run.ts
```diff
@@ -228,7 +228,10 @@ async function main() {
       ...connectionInfo.driverOptions,
     });
   } catch (e: any) {
-    console.error(`${e?.name}: ${e?.message}`);
+    // for debugging
+    if (process.env.MONGOSH_DISPLAY_STARTUP_STACK_TRACE)
+      console.error(e?.stack);
+    else console.error(`${e?.name}: ${e?.message}`);
```
addaleax, line 234: This is just for us, it's pretty annoying to have the stack trace not showing up for errors occuring during startup

packages/e2e-tests/test/test-shell.ts
```diff
@@ -160,7 +160,7 @@ export class TestShell {
     return this._rawOutput;
   }
 
-  get process(): ChildProcess {
+  get process(): ChildProcessWithoutNullStreams {
```
addaleax, line 163: Just a drive-by to fix type errors showing up in the e2e.spec.ts file

packages/cli-repl/src/mongosh-repl.ts
```diff
@@ -658,8 +730,39 @@ class MongoshNodeRepl implements EvaluationListener {
     input: string,
     filename: string
   ): Promise<ShellResult> {
-    const { repl } = this.runtimeState();
-    return await promisify(repl.eval.bind(repl))(input, repl.context, filename);
+    const { repl, context } = this.runtimeState();
+    if (repl) {
+      return await promisify(repl.eval.bind(repl))(input, context, filename);
+    } else {
```
lerouxb, line 736: Suuuper nitpicky and I'm surprised that our linting didn't complain, but you can probably just remove the else and save one level of indentation because the if is guaranteed to return. But also just ignore me if you prefer this style.

packages/cli-repl/src/run.ts
lerouxb: I should probably already know this answer, but why don't we display the stack trace by default all the time?

packages/cli-repl/src/mongosh-repl.ts
addaleax: Yeah, thanks for pointing this out! Addressed this in 8f7f38f :)

lerouxb (approved):

packages/cli-repl/src/cli-repl.ts
```diff
@@ -186,8 +186,17 @@ export class CliRepl implements MongoshIOProvider {
       this.bus.emit('mongosh:error', err, 'io');
     });
 
+    let jsContext = this.cliOptions.jsContext;
+    if (jsContext === 'auto' || !jsContext) {
+      jsContext = CliRepl.getFileAndEvalInfo(this.cliOptions)
```
kmruiz, line 191: Do we want (didn't see where) to add the jsContext to our telemetry as a dimension to the startup event?

addaleax: @kmruiz Thanks, added it! This is probably not going to be very different from `is_interactive` in practice, but keeping track of the number of people who run it with `js_context === 'repl' && is_interactive === false` might be valuable

kmruiz (approved):

MarcusElevait:
Why is this not versioned as a breaking change? This breaks javascript file executions that uses native things like `require("fs")`

tiagoliveira9:
Yes, this should be versioned as a breaking change, I have several scripts impacted with these changes

tiagoliveira9:
And the helper doesn't show up the new `--jsContext` option

addaleax:
@MarcusElevait @tiagoliveira9 Just for clarity, this was unintended breakage and we released mongosh 2.2.1 soon after to fix it: https://github.com/mongodb-js/mongosh/releases/tag/v2.2.1

PR #1856, opened 2024-03-06
fix(cli-repl): do not run prompt and initial input in parallel MONGOSH-1617

addaleax:
Users were getting incorrect behavior when:

1. A custom asynchronous prompt was set outside of the main REPL input, e.g. in a `.mongoshrc.js` file or a `--eval` script, and
2. The main REPL input was readable before we started evaluating the prompt.

because then, the first lines of input and the prompt may have ended up executing in parallel. Setting the prompt before starting to process input should resolve this issue.


packages/cli-repl/src/mongosh-repl.spec.ts
```diff
@@ -23,6 +23,11 @@ import MongoshNodeRepl from './mongosh-repl';
 import { parseAnyLogEntry } from '../../shell-api/src/log-entry';
 import stripAnsi from 'strip-ansi';
 
+function nonnull<T>(value: T | null | undefined): NonNullable<T> {
+  if (!value) throw new Error();
+  return value;
+}
```
addaleax, line 29: There are some drive-by fixes for type errors in the test files here, mostly things I missed in #1849 because we apply different TS configs for out test and source files

packages/e2e-tests/test/e2e.spec.ts
```diff
@@ -262,6 +262,26 @@ describe('e2e', function () {
...
+        args: [
+          '--nodb',
+          '--shell',
+          '--eval',
+          'prompt = () => {sleep(1);return "x>"}',
+        ],
+      });
+      // The number of newlines here matters
+      shell.writeInput(
+        'sleep(100);print([1,2,3,4,5,6,7,8,9,10].reduce(\n(a,b) => { return a*b; }, 1))\n\n\n\n',
```
paula-stacho, line 277: probably silly question, but why is this sleeping longer if we want to make sure the prompt is evaluated first? also why does the number of newlines matter? 😅

paula-stacho: oh I get it now, thank you for the explanation!

paula-stacho (approved):

PR #1861, opened 2024-03-07
fix(ci): remove unused licenses.json entry

addaleax:
This is no longer necessary since 20df659f9ae0 and made our vuln scan CI job and the GH actions runs for `main` commits fail since then.


addaleax:
I'll go ahead and merge this given the straightforwardness of the change

PR #1869, opened 2024-03-11
feat(cli-repl): make `--quiet` the default for non-interactive usage MONGOSH-1721

addaleax:
This can be disabled through `--no-quiet` or `--verbose` (although `--verbose` may gain other meanings in the future depending on the outcome of MONGOSH-970).


paula-stacho:
it's a bit hard to tell for me even reading the full file 😊  - do we have a test that covers the interactive without explicit quiet=false scenario?

paula-stacho (approved):

PR #1876, opened 2024-03-13
chore(ci): migrate to Garasign from notary MONGOSH-1655

dylrich:


packages/build/src/packaging/run-package.ts
```diff
@@ -39,14 +38,5 @@ export async function runPackage(config: Config): Promise<PackageFile> {
...
 
   const packaged = await runCreatePackage();
-
-  if (packageVariant === 'win32msi-x64') {
-    await notarizeArtifact(packaged.path, {
-      signingKeyName: config.notarySigningKeyName || '',
-      authToken: config.notaryAuthToken || '',
-      signingComment: 'Evergreen Automatic Signing (mongosh)',
-    });
-  }
```
addaleax, line 49: iirc we gave the MSI special handling because it was fairly unusable without the signing – should we keep this? should we apply all signing to the "create package" phase rather than the "draft release" phase?

addaleax (approved):
Looks great, just one question!

packages/build/src/packaging/run-package.ts
dylrich: I dropped this because I noticed that it was being re-signed in `runDraft`. We couldn't find any special reason just the MSI needed to be signed in the packaging step (e.g. for testing), and none of the Windows-specific tests seemed to fail because of this. The reason to drop it is that it'd require the whole host-create + ssh dance as things are written now because packaging happens on particular operating systems instead of just Ubuntu like `runDraft`does. I think a better solution in the future would move signing to something that happens before/after packaging as appropriate for the artifact in question, and ideally this would run on a Linux distro only. But that felt like a bigger refactor than I was prepared to do at this time!

addaleax: Yeah, makes sense – let's leave it like this then :)

PR #1883, opened 2024-03-15
fix(deps): deduplicate devtools-connect by hoisting it to root node_modules

addaleax:


package-lock.json
```diff
@@ -6486,6 +6486,29 @@
         "react": "^17.0.2"
       }
     },
+    "node_modules/@mongodb-js/devtools-connect": {
```
lerouxb, line 6489: You manually edited package-lock.json? Is that advisable? Maybe I'm mistaken but it feels like npm install could just change the package-lock file to be what it thinks it should be in future when someone runs npm install again?

addaleax: > You manually edited package-lock.json? 

No :) I installed `@mongodb-js/devtools-connect` at the top level via `npm install`, then removed it again as a dependency from `package.json`, and ran `npm install` again to get a clean `package-lock.json`.

> Is that advisable? 

I'd avoid that, yeah. But in any case, not what happened here.

> Maybe I'm mistaken but it feels like npm install could just change the package-lock file to be what it thinks it should be in future when someone runs npm install again?

I'm not sure I understand this part. `npm install` by default sticks to what's in the `package-lock.json`, not the other way around, right?

lerouxb: I was referring to the fact that unlike `npm ci`, `npm` i updates the package-lock.json. I have no idea what their intention is vs what the reality is but from experience I know that it rarely removes or downgrades things so we're probably good here.

addaleax: Yeah, that was kind of the problem here, I assume we had conflicting versions or something like that at some point that made npm put things into the individual packages' `node_modules` folders, then that became unnecessary, but it still kept those things in the nested `node_modules` instead of the top-level one because of its reluctancy to move things/update package-lock.json

PR #1897, opened 2024-03-25
Revert "chore(ci): disables test_vscode task (#1888)"

addaleax:
This reverts commit 773eee6fd002aa020be37e43fc24f69159ecf3e2.

These are failures that we should keep being aware of, but the VSCode tests *are* still useful regardless of the failing OIDC tests.

I assume this was disabled for a release, but 97d8c2e966f029699356ed70efc9328b397f6887 already removed this task from the list of tasks required for releases.


himanshusinghs (approved):
I should've looked more closely and thanks :D

PR #1915, opened 2024-03-27
chore: bump Node.js to 20.12.0 MONGOSH-1750 MONGOSH-1691

addaleax:
- Original Node.js update PR in #1914 
- Remove patches that are no longer necessary, specifically https://github.com/nodejs/node/commit/0090c107820 and https://github.com/nodejs/node/commit/e54ddf898f6ca504320b676c984a1cd259987026 which are both now on Node.js 20
- Work around build failures caused by https://github.com/nodejs/node/issues/52229
- Adjust tests to account for slightly different REPL colorization logic after https://github.com/nodejs/node/commit/f94336f95affdc0a72740ed935daaceb7cd21b85 (which we specifically requested as a fix for MONGOSH-1691)


himanshusinghs (approved):

addaleax:
@himanshusinghs Would it be alright to take another look at this? The patches have changed a bit, unfortunately:

- Needed to skip some of the tests around the changed REPL colorization behavior on Node.js 16, that should be fine
- Needed to apply the workaround for https://github.com/nodejs/node/issues/52229 in a patch instead of `CFLAGS`/`CXXFLAGS` because those don't work on Windows

And then, finally, as you know we are building on platforms that predate glibc 2.25, but Node.js broke builds for those older glibc versions by declaring that the `getrandom()` function in glibc would always be available (which they can do, because the official minimum is 2.28), so I:
- Needed to add a patch to only declare the function to be available if it is at build time
- And finally, the one I'm most unsure about, when we install Node.js from source rather than using the official binaries (i.e. exactly those old-glibc systems), I'd pin the Node.js version for that to 20.11.1. That's maybe not a great long-term solution, we _could_ also stick to using the latest 20.x and start applying our build patches to the for-CI-only build as well

himanshusinghs (approved):

PR #1920, opened 2024-03-28
fix(shell-api): improve AbstractCursor iteration performance MONGOSH-1688

addaleax:
- Delegate cursor iteration to the driver where possible
- Do not call the wrapped `.tryNext()` method from inside other shell API methods, and instead only call an internal unwrapped version of it (similar to `db.runCommand()` vs `db._runCommand()`

This results in a 60% improvement of runtime in local testing on the `db_cursor_iteration_plainvm` benchmark.


lerouxb (approved):

lerouxb:
🔥

PR #1922, opened 2024-04-02
fix(shell-evaluator): fix exposed asyncRewrite fn for benchmarks

addaleax:
Painfully obvious bug in 0491491ce96b :) Renamed the benchmark test and fixed it to account for the (now very) different performance measurements after the fix.

### DIFF
--- a/packages/shell-evaluator/src/shell-evaluator.ts
+++ b/packages/shell-evaluator/src/shell-evaluator.ts
@@ -93,8 +93,8 @@ class ShellEvaluator<EvaluationResultType = ShellResult> {
     }
 
     if (this.exposeAsyncRewriter) {
-      (context as any).__asyncRewrite = (input: string) =>
-        this.asyncWriter.process(input);
+      (context as any).__asyncRewrite = (rewriteInput: string) =>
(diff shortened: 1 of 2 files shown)


packages/shell-evaluator/src/shell-evaluator.ts
```diff
@@ -93,7 +93,8 @@ class ShellEvaluator<EvaluationResultType = ShellResult> {
     }
 
     if (this.exposeAsyncRewriter) {
-      (context as any).__asyncRewrite = () => this.asyncWriter.process(input);
+      (context as any).__asyncRewrite = (input: string) =>
+        this.asyncWriter.process(input);
```
Anemy, line 97: [nit, fine as is], how do you feel about renaming `input` to make sure we don't confuse it with the innerEval input?
```suggestion
      (context as any).__asyncRewrite = (rewriteInput: string) =>
        this.asyncWriter.process(rewriteInput);
```

Anemy (approved):

packages/shell-evaluator/src/shell-evaluator.ts
addaleax: sgtm!

PR #1931, opened 2024-04-04
[Do not merge] Prototype: Async rewriter alternative 1 (Rust, WASM, no AST manipulation)

addaleax:


packages/async-rewriter3/src/lib.rs
```diff
@@ -0,0 +1,400 @@
...
+        self.list.pop_back();
+    }
+
+    fn add_variable(&mut self, variable: String) {
+        self.vars.push(variable);
+    }
+}
+
+fn is_block(body: &ExprOrBlock) -> bool {
+    match body {
```
kmruiz, line 68: If you want to make it slimmer (just for fun) you can use matches!

It would be like (with typos for sure):


```rs
matches!(body, ExprOrBlock::Block(_))
```

packages/async-rewriter3/src/lib.rs
```diff
@@ -0,0 +1,400 @@
...
+                if pat.is_some() {
+                    ret.append(&mut add_all_variables_from_declaration([&pat.unwrap()].into_iter()));
+                }
+            },
+            Pattern::AssignPattern(p) => {
+                let key = p.key();
+                if key.is_some() {
+                    ret.append(&mut add_all_variables_from_declaration([&key.unwrap()].into_iter()));
+                }
+            },
```
kmruiz, line 178: You can also use match here to make it simpler to read and avoid that unwrap:

```rs
match p.key()  {
   Some(key) => ret.append(...key)
   _ => 
```
Or even an if with pattern extraction:

```rs
if let Some(key) = p.key {
  ret.append...
}
```

PR #1932, opened 2024-04-05
chore(ci): extract package signing from release-draft MONGOSH-1746

addaleax:


.evergreen/sign-packaged-artifact.sh
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env bash
+set -e
+set -x
+
+export NODE_JS_VERSION=${NODE_JS_VERSION}
+export ARTIFACT_URL_FILE="$PWD/../artifact-url.txt"
+
+# TODO: Remove this once it is confirmed that it's safge to do so
```
mabaasit, line 8: do we need to link the ticket here?

mabaasit: sorry, but do we even need this as we are using garasign? (or is this particular build a special one?)

packages/build/src/index.ts
```diff
@@ -22,6 +22,7 @@ if (require.main === module) {
         'upload',
         'draft',
         'publish',
+        'sign',
         'trigger-release',
       ].includes(command)
     ) {
```
mabaasit: Do we need to add `sign` to the error below as well?

mabaasit (approved):

packages/build/src/index.ts
mabaasit: looked at the next commit and its taken care of :)

.evergreen/sign-packaged-artifact.sh
addaleax: Yeah, thank you for the reminder! I think we can just remove this, that's exactly what I was hoping for, just forgot about this TODO myself 😅

PR #1954, opened 2024-04-19
MONGOSH-1661: Add autocomplete for streams commands

pulkitkalra-mdb:
## Description
[MONGOSH-1661](https://jira.mongodb.org/browse/MONGOSH-1661): Adds autocomplete for `sp.` commands. 
- Works for `sp.` prompts and autocompletes for all standard commands - but not names of stream processors
- Autocompletes when command is in that format: `sp.name.` with standard commands like `start()`, `stop()` etc.

https://github.com/mongodb-js/mongosh/assets/156152710/9c5f699f-9e25-4660-a355-ff27b210d148

## Testing
Unit test

### DIFF
--- a/packages/autocomplete/src/index.spec.ts
+++ b/packages/autocomplete/src/index.spec.ts
@@ -836,4 +836,22 @@ describe('completer.completer', function () {
       expect(await completer(apiStrictParams, i)).to.deep.equal([[], i]);
     });
   });
+
+  context('with stream processing sp', function () {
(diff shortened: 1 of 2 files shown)


andym03 (approved):
LGTM! Can't see what's causing those tests to timeout, perhaps just flakey. 

Should wait for the team to approve as well.

addaleax (approved):
LGTM!

PR #1955, opened 2024-04-19
fix(cli-repl): throw error when changing telemetry setting while `forceDisableTelemetry: true` MONGOSH-1762

addaleax:


gribnoysup (approved):

Ghostbird:
Ah, so this is why our docker containers were failing. This was the startup command of our docker container.

```sh
bash -mc "mongod --replSet rs0 & sleep 3; mongosh --eval 'disableTelemetry(); rs.initiate()'; fg
```

This used to work fine, but now crashed, and the replica set was never initialised. The weird thing is that apparently `forceDisableTelemetry` is set to `true`, but I have no clue _why_ it's set to `true`.

It doesn't seem to be done by https://github.com/search?q=repo%3Adocker-library%2Fmongo%20forceDisableTelemetry&type=code, nor by this repo.

I think this should have only thrown when trying to _change_ the telemetry setting. It shouldn't throw when setting it to the current value.

addaleax:
@Ghostbird And sorry for the delay, but just to present you with an update, we just released mongosh 2.2.6 with the fix – thanks again for bringing this up!

Ghostbird:
Thanks for the fix!

PR #1964, opened 2024-04-25
feat(shell-api): warn users when using background option in aggregate MONGOSH-1766

mabaasit:


lerouxb (approved):

packages/shell-api/src/collection.ts
```diff
@@ -177,6 +177,11 @@ export default class Collection extends ShellApiWithMongoClass {
       options = {};
       pipeline = args || [];
     }
+    if ('background' in options) {
+      this._instanceState.printWarning(
+        'the background option is not currently supported and will be ignored.'
```
tzembo, line 182: if we're already adding this warning just for Atlas Data Federation, could we improve it to suggest what users should do instead (instead of leading to confusion or a HELP ticket)?

something like:

"the background option is not supported by the aggregate method and will be ignored, use runCommand to use {background: true} with Atlas Data Federation"

mabaasit: Hi @tzembo, i made the suggested change in [49eebaf](https://github.com/mongodb-js/mongosh/pull/1964/commits/49eebafa1ea3ec783b42bf0970c8ec6b9f48b6a0).

PR #1989, opened 2024-05-10
chore(deps): bump 1st party deps to versions with available CodeQL scans

addaleax:


packages/cli-repl/package.json
```diff
@@ -78,13 +78,13 @@
     "@mongosh/types": "0.0.0-dev.0",
     "@segment/analytics-node": "^1.3.0",
     "ansi-escape-sequences": "^5.1.2",
-    "askcharacter": "^1.0.0",
-    "askpassword": "^1.2.4",
+    "askcharacter": "^2.0.4",
+    "askpassword": "^2.0.2",
```
addaleax, line 82: Major version bumps because these included changes to the tested (= supported) Node.js versions. I stopped doing these after a bit because it turned out lots of packages needed updates 🙂

packages/cli-repl/package.json
```diff
@@ -78,13 +78,13 @@
...
     "escape-string-regexp": "^4.0.0",
     "is-recoverable-error": "^1.0.3",
     "js-yaml": "^4.1.0",
-    "mongodb-connection-string-url": "^3.0.0",
-    "mongodb-log-writer": "^1.4.0",
+    "mongodb-connection-string-url": "^3.0.1",
+    "mongodb-log-writer": "^1.4.2",
```
addaleax, line 87: Some of these versions got bumped twice because it turns out that CodeQL jobs with cron workflow triggers get disabled after 60 days of no activity in a repo and I didn't know that when doing the first set of releases 😕

packages/cli-repl/package.json
```diff
@@ -115,8 +115,8 @@
   "optionalDependencies": {
     "get-console-process-list": "^1.0.4",
     "glibc-version": "^1.0.0",
-    "macos-export-certificate-and-key": "^1.1.1",
-    "mongodb-crypt-library-version": "^1.0.3",
-    "win-export-certificate-and-key": "^1.1.1"
+    "macos-export-certificate-and-key": "^1.1.2",
```
addaleax, line 118: Was almost driven insane by the "fact" that this bump of the library made CI fail even though no code changed (it surely can't be a coincidence, right, this PR bumps the package that provides the error message that's leading to the CI errors??), then finally thought to check other runs on `main`, turns out the CI redness happens completely independently of this PR without any code changes at all: https://github.com/mongodb-js/mongosh/pull/1991 🫠

mcasimir (approved):

packages/cli-repl/package.json
mabaasit: activity in the repo means commits to the main branch?

mabaasit (approved):

packages/cli-repl/package.json
addaleax: I don't know, this is basically all the information available to me: 

Screenshot 2024-05-10 at 21 12 42

PR #2012, opened 2024-06-03
fix(deps): bump pack-resolver version MONGOSH-1791

paula-stacho:
https://jira.mongodb.org/browse/MONGOSH-1791

From the vulnerability description:
"An attacker can manipulate a system that uses `isLoopback()`, `isPrivate()` and `isPublic` functions to guard outgoing network requests to treat certain IP addresses as globally routable by supplying specially crafted IP addresses."

There is no fix for `ip`, but `pac-resolver` has a patch that removes the dependency on `ip` (implements it's own `isLoopback` method). Unfortunately this fix hasn't bubbled up our dependency tree yet, which is why I'm proposing an override for the sub dependency. I'm not too happy about overrides, because they tend to be forgotten later - so I'm curious if we have a process for dealing with such situations

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18156,11 +18156,6 @@
         "loose-envify": "^1.0.0"
       }
     },
-    "node_modules/ip": {
-      "version": "1.1.9",
-      "resolved": "https://registry.npmjs.org/ip/-/ip-1.1.9.tgz",
-      "integrity": "sha512-cyRxvOEpNHNtchU3Ln9KC/auJgup87llfQpQ+t5ghoC/UhL16SWzbueiCsdTnWmqAWl7LadfuwhlqmtOaqMHdQ=="
-    },
     "node_modules/ip-address": {
       "version": "9.0.5",
       "resolved": "https://registry.npmjs.org/ip-address/-/ip-address-9.0.5.tgz",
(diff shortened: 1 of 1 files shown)


addaleax:
> Unfortunately this fix hasn't bubbled up our dependency tree yet, which is why I'm proposing an override for the sub dependency. I'm not too happy about overrides, because they tend to be forgotten later - so I'm curious if we have a process for dealing with such situations

Is there anything speaking against updating `pac-resolver` directly instead of using overrides? e.g.

```
npm i pac-resolver@7.0.1 && git checkout -- package.json && npm i
```

on a clean checkout of `main` seems to do the trick for me.

paula-stacho:
> > Unfortunately this fix hasn't bubbled up our dependency tree yet, which is why I'm proposing an override for the sub dependency. I'm not too happy about overrides, because they tend to be forgotten later - so I'm curious if we have a process for dealing with such situations
> 
> Is there anything speaking against updating `pac-resolver` directly instead of using overrides? e.g.
> 
> ```
> npm i pac-resolver@7.0.1 && git checkout -- package.json && npm i
> ```
> 
> on a clean checkout of `main` seems to do the trick for me.

ah, yes 🤦 I don't know why it didn't occur to me to try that :D

addaleax (approved):
:rocket:

PR #2022, opened 2024-06-11
fix(build): fix tests for new platforms MONGOSH-1806

addaleax:
Our `build` package tests have been failing since 904162f5e89f8. This commit fixes:

- Ubuntu 24.04 is not known as part of the server distro list yet, but we can reliably infer its name tag based on the format used for existing Ubuntu distros
- Oracle Linux 9 is not known as a server distro and will likely never be.


Anemy (approved):
TIL oracle linux

PR #2023, opened 2024-06-11
chore(MONGOSH-1808): migrate mongosh to new mongodb-client-encryption repository

nbbeeken:
Changes prep-fle-addon script to use new repo and depend on new repo's script for downloading/building bindings. The version of libmongocrypt is now determined by the binding's package.json file.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22876,18 +22876,46 @@
       }
     },
     "node_modules/mongodb-client-encryption": {
-      "version": "6.0.0",
-      "resolved": "https://registry.npmjs.org/mongodb-client-encryption/-/mongodb-client-encryption-6.0.0.tgz",
-      "integrity": "sha512-GtqkqlSq19acX006/U1odA3l+gwhvABeoTUlvvgtvSs6qcN3qSHPnur3Z5N4oKOv6fZ7EtT8rIsWP2riI0+Eyg==",
+      "version": "6.1.0-alpha.0",
+      "resolved": "https://registry.npmjs.org/mongodb-client-encryption/-/mongodb-client-encryption-6.1.0-alpha.0.tgz",
+      "integrity": "sha512-pcypn4m2EdAYPkfw0V4cfu64E5spJDHJTVSiIU8ErNbiIk+zJe1HcnfMRbyiXsulI4B46aWoOzNNd7NUGUBRwg==",
       "hasInstallScript": true,
+      "license": "Apache-2.0",
       "optional": true,
       "dependencies": {
         "bindings": "^1.5.0",
         "node-addon-api": "^4.3.0",
-        "prebuild-install": "^7.1.1"
+        "prebuild-install": "^7.1.2"
       },
       "engines": {
         "node": ">=16.20.1"
+      }
+    },
+    "node_modules/mongodb-client-encryption/node_modules/prebuild-install": {
+      "version": "7.1.2",
+      "resolved": "https://registry.npmjs.org/prebuild-install/-/prebuild-install-7.1.2.tgz",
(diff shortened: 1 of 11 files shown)


nbbeeken:
v6.0.0 and v6.0.1 tags in the other repo do not have a working version of the download/build script. This PR should be picked up once we have tagged and released from the new repo.

nbbeeken:
Currently blocked on:
- [error: /Applications/Xcode12.4.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/libtool: unknown option character `r' in: -arch](https://parsley.mongodb.com/evergreen/mongosh_darwin_arm64_compile_artifact_patch_f2b3c27a4de31f39df484ace59b665077aa30193_6671ed04ff0ae60007ee1a21_24_06_18_20_25_15/0/task?bookmarks=0,13952&shareLine=13862)
    - Not sure what's going on there, it is during the boxednode build. [FLE completes its build here](https://parsley.mongodb.com/evergreen/mongosh_darwin_compile_artifact_patch_f2b3c27a4de31f39df484ace59b665077aa30193_6671ed04ff0ae60007ee1a21_24_06_18_20_25_15/0/task?bookmarks=0,13792&shareLine=11366) Where would the arch flag be modified after that? 
- [CMake 3.12 or higher is required.  You are running version 3.11.0](https://parsley.mongodb.com/evergreen/mongosh_win32_build_compile_artifact_patch_f2b3c27a4de31f39df484ace59b665077aa30193_6671ed04ff0ae60007ee1a21_24_06_18_20_25_15/0/task?bookmarks=0,11635&shareLine=11534)
    - I switched to building on Windows because the libmongocrypt prebuilds were failing for me. Maybe I missed something and could go back to that.

addaleax:
> * Not sure what's going on there, it is during the boxednode build. [FLE completes its build here](https://parsley.mongodb.com/evergreen/mongosh_darwin_compile_artifact_patch_f2b3c27a4de31f39df484ace59b665077aa30193_6671ed04ff0ae60007ee1a21_24_06_18_20_25_15/0/task?bookmarks=0,13792&shareLine=11366) Where would the arch flag be modified after that?

I think it's not complaining about the value of `-arch`, it doesn't seem to understand the `-arch` flag as all and rather interpret it as `-a -r -c -h` (and doesn't know `-r` specifically).

These flags are coming from here: https://github.com/mongodb-js/mongodb-client-encryption/blob/95849ddca9a452b1f697189a29a483d354a394bd/binding.gyp#L27-L33 – they should probably be in the `build_type=="dynamic"` sub-section?

> CMake 3.12 or higher is required. You are running version 3.11.0

Fwiw, we happily adjust our build `PATH` as needed in `.evergreen/setup-env.sh` 🙂

nbbeeken:
Good call on the -arch flag being split up. I'll look into that, but I don't think that it should be in the "dynamic" section, IIUC that is unused, we build static prebuilds and we're trying to build statically for mongosh here. 

It would be nice if I could somehow feature detect, but maybe I can add a new variable we pipe in for our prebuilds to set -arch and otherwise it is omitted. I will investigate further.

nbbeeken:
An update, so I've enabled the ability to only build for the specific macos platform, however, even when only building for the local platform node-gyp is still putting `-arch` into the build flags which is making libtool unhappy on BigSur, maybe we've been using a newer version of node-gyp or some other tool that makes an assumption about support for that flag. 

Window, still need to find a newer cmake version, haven't started on that. I don't want to have to download it but may have to.

nbbeeken:
A lot of red but none of them are compile_artifact 🎉

mabaasit:
Closing in favour of #2076

PR #2032, opened 2024-06-14
fix(build): disable publishing to 8.0 Linux repos for now

addaleax:
Our build scripts try to publish to all available combinations of Linux repositories. It seems that, for now, barque has started forbidding the implicit creation of repositories that way; for example, the publishing steps for 2.2.8 failed because we cannot publish to the Debian 10 repo for the 8.0 server release line (which does not support Debian 10).

For now, let's disable these repos to unblock the release, and follow up with a ticket to re-enable the exact subset of repos that we can and should publish to.


addaleax:
(Received LGTM over Slack)

PR #2048, opened 2024-07-01
fix(shell-api): sh.status() should show full list of tag ranges in verbose mode only MONGOSH-1327

alenakhineika:
Show a full tag/chunk list when there are 20 or fewer items and cut a tag/chunk list when there are more than 20 items.

### DIFF
--- a/packages/shell-api/src/helpers.ts
+++ b/packages/shell-api/src/helpers.ts
@@ -556,6 +556,7 @@ export async function getPrintableShardStatus(
                 { noBalance: coll.noBalance },
               ];
             }
+
             const chunksRes = [];
             const chunksCollMatch = buildConfigChunksCollectionMatch(coll);
             const chunks = await (
@@ -566,21 +567,20 @@ export async function getPrintableShardStatus(
                 { $sort: { shard: 1 } },
               ])
             ).toArray();
-            let totalChunks = 0;
+
             collRes.chunkMetadata = [];
+
             chunks.forEach((z: any) => {
-              totalChunks += z.nChunks;
               collRes.chunkMetadata.push({
                 shard: z.shard,
                 nChunks: z.nChunks,
               });
             });
 
-            // NOTE: this will return the chunk info as a string, and will print ugly BSON
-            if (totalChunks < 20 || verbose) {
-              for await (const chunk of (
-                await chunksColl.find(chunksCollMatch)
-              ).sort({ min: 1 })) {
+            for await (const chunk of (
+              await chunksColl.find(chunksCollMatch)
+            ).sort({ min: 1 })) {
+              if (chunksRes.length < 20 || verbose) {
                 const c = {
                   min: chunk.min,
                   max: chunk.max,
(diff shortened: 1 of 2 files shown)


packages/shell-api/src/helpers.ts
```diff
@@ -614,18 +614,30 @@ export async function getPrintableShardStatus(
...
-                tag: tag.tag,
-                min: tag.min,
-                max: tag.max,
-              });
+            )
+              .sort({ min: 1 })
+              .toArray();
+            const tagsRes: any[] = [];
+
+            // NOTE: this will return tags as a string, and will print ugly BSON
```
addaleax, line 626: I'm not quite sure I understand this comment – the tags are still regular tags, right?

packages/shell-api/src/helpers.ts
```diff
@@ -614,18 +614,30 @@ export async function getPrintableShardStatus(
               );
             }
 
-            const tagsRes: any[] = [];
-            for await (const tag of (
+            const tags = await (
```
addaleax, line 617: It seems like, based on the ticket, the legacy shell printed at least some of the tags even if there were more than the given threshold, and then added the "too many tags" message after that. Should we do the same? That would also allow us to keep the `for await` instead of having to load the entire tag list into memory with `.toArray()`

alenakhineika: The comment and this behavior reflect the behavior we already have for chunks, that happen above in the same method. I assumed this was the intended behavior for the new shell. Is it really helpful to print some tags id we hide the rest anyway? How big is the chance that they want to see the first 5 out of 100..

addaleax: Hm yeah ... I'd still think it results in a better UX (for both cases, tbh) if we print a few and then tell the user that there's more – that gives them an impression of what these tag ranges look like, if nothing else.

Is there any reason not to do that?

alenakhineika: No reason. Should we also update chunks then?

addaleax: I think that would be an UX improvement, yes 🙂

packages/shell-api/src/shard.spec.ts
```diff
@@ -2629,4 +2677,90 @@ describe('Shard', function () {
...
+        (d) => d.database._id === 'test'
+      ).collections[ns].chunks;
+      expect(chunks.length).to.equal(21);
+      expect(
+        !!chunks.find(
+          (tag) =>
+            tag ===
+            'too many chunks to print, use verbose if you want to force print'
+        )
+      ).to.equal(true);
```
addaleax, line 2763: Totally optional, but I think you can simplify this to

```suggestion
      expect(chunks).to.include(
        'too many chunks to print, use verbose if you want to force print'
      );
```

or maybe use a stronger assertion like

```suggestion
      expect(
        chunks.indexOf(
          'too many chunks to print, use verbose if you want to force print'
        )
      ).to.equal(20);
```

addaleax (approved):

PR #2049, opened 2024-07-01
wip: chore(worker-runtime): remove worker proxy MONGOSH-1416

mabaasit:


packages/node-runtime-worker-thread/src/index.ts
```diff
@@ -222,7 +231,7 @@ class WorkerRuntime implements Runtime {
 
   async interrupt() {
     await this.initWorkerPromise;
-    return this.childProcessRuntime.interrupt();
+    return this.childProcess.kill('SIGINT');
```
addaleax, line 234: This means that we're now losing all state (e.g. variables, global object, etc.) in the shell when using Ctrl+C, correct?

That was one of the reasons why we wanted to use a Worker thread, and why we wanted to remove the child process indirection rather than the Worker thread one here ... I remember we talked about making the integration with Electron's web workers possible, I guess that's out of the question here?

addaleax: Cool, yeah – I hadn't tested this, so if everything's working fine, then feel free to disregard me 🙂 Basit can probably confirm once he's back

Anemy: Screenshot 2024-07-23 at 5 53 35 PM
[this pr on the left | current Compass on the right]
Basit mentioned this might not work with ctrl + c on windows, and we may want to use a new worker directly, and use worker.interupt, instead of spawning a process. Looking into that now.

Anemy:
Removed the child process usage. We now use the worker thread directly. Trying it out in Compass now. Let me know if this is what you had in mind @mabaasit @addaleax

package.json
```diff
@@ -136,14 +136,14 @@
   "workspaces": [
     "configs/eslint-config-mongosh",
     "configs/tsconfig-mongosh",
+    "packages/java-shell",
+    "packages/js-multiline-to-singleline",
```
Anemy, line 140: 🤔 My bootstrap keeps doing this. I'll remove this change from this pr, maybe it's a version issue or I have some old workspaces leftover in my project?

packages/node-runtime-worker-thread/src/index.ts
```diff
@@ -11,184 +9,162 @@ import type {
...
-import { ChildProcessEvaluationListener } from './child-process-evaluation-listener';
+import { createCaller, cancel, exposeAll } from './rpc';
 import type { WorkerRuntime as WorkerThreadWorkerRuntime } from './worker-runtime';
 import {
   deserializeEvaluationResult,
   serializeConnectOptions,
 } from './serializer';
-import { ChildProcessMongoshBus } from './child-process-mongosh-bus';
 import type { CompassServiceProvider } from '@mongosh/service-provider-server';
+import { SHARE_ENV, Worker } from 'worker_threads';
```
addaleax, line 20: @Anemy iirc what we tried to use in Compass were Web Workers with `nodeIntegrationInWorker: true` ... does this work?

Anemy: Hmmm adding that option here: https://github.com/mongodb-js/compass/blob/66e4d196949cb48787fde3c582b2e3d052ac103a/packages/compass/src/main/window-manager.ts#L124 
On a quick test pulling this code into Compass is getting me:
Screenshot 2024-07-26 at 10 15 51 AM

```
The V8 platform used by this instance of Node does not support creating Workers
Call Stack
 new Worker
  node:internal/worker:213:21
 WorkerRuntime.initWorker
  MongoDBCompass/../compass-shell/src/node-runtime-worker-thread/index.ts:62:27
```

Branch: test-poc-worker
https://github.com/mongodb-js/compass/compare/main...test-poc-worker

addaleax: Yeah – Electron doesn't support Node.js `worker_threads` (hence the idea to use Web Workers instead)

gribnoysup: Yep, that was the reason we had a child_process added in the initial implementation: to work around Electron runtimes not supporting those

Anemy: TIL, thanks, I'll update this to web workers and see if things work smoothly.

Anemy: Updated https://github.com/mongodb-js/compass/compare/main...test-poc-worker to use a web worker, appears to run smoothly, woo! Updating this pr next and will see how we can test it.

packages/shell-api/src/mongo.ts
```diff
@@ -1,3 +1,4 @@
+import { setTimeout } from 'node:timers';
```
addaleax, line 1: This doesn't change all imports of `setTimeout` in shell-api, and there are users of shell-api that do not have Node.js libraries available

So ... what does this do? :)

packages/node-runtime-worker-thread/src/worker-runtime.ts
```diff
@@ -1,7 +1,6 @@
 /* istanbul ignore file */
-/* ^^^ we test the dist directly, so isntanbul can't calculate the coverage correctly */
+/* ^^^ we test the dist directly, so istanbul can't calculate the coverage correctly */
```
addaleax, line 2: 😂

packages/node-runtime-worker-thread/package.json
```diff
@@ -52,6 +52,7 @@
   },
   "dependencies": {
     "interruptor": "^1.0.1",
-    "system-ca": "^2.0.1"
+    "system-ca": "^2.0.1",
+    "web-worker": "^1.3.0"
```
addaleax, line 56: Hm ... maybe we actually need to create a separate SBOM for this package: https://jira.mongodb.org/browse/MONGOSH-1856

Not something for this PR, but a good reminder

packages/shell-api/src/mongo.ts
mabaasit: i should have added a comment why: so when trying this locally with compass, we have two autocompletions ([1](https://github.com/mongodb-js/mongosh/blob/MONGOSH-1416-remove-process-indirection/packages/shell-api/src/mongo.ts#L349), [2](https://github.com/mongodb-js/mongosh/blob/MONGOSH-1416-remove-process-indirection/packages/shell-api/src/database.ts#L284)) where we do `setTimeout(...).unref()`. With webworkers, this throws and autocomplete is not working well. (even with `nodeIntegrationInWorker` set to true in Compass, this throws. I am not sure why, but this change will help the integration on the Compass side). 

As for other uses, (as Sergey suggested to me), we can do what driver [did](https://github.com/mongodb/node-mongodb-native/blob/5565d5008812f03e9ff86a7b6326447ac5916244/.eslintrc.json#L30-L56). I can create a follow for that if that sounds good to you.

addaleax: Ah, I don't think we want to do either of these – we'll just want to call `.unref()` conditionally, i.e. replace `setTimeout(...).unref()` with `setTimeout(...)?.unref?.()`?

Because a) these `.unref()` calls are still relevant and b) we really can't just assume that the `timers` module is available as a built-in module in this context

addaleax (approved):
LGTM once CI is green 👍

PR #2076, opened 2024-07-12
chore: update libmongocrypt and fix tests 

mabaasit:
In this PR, we are using new mongodb-client-encryption repository for libmongocrypt (MONGOSH-1808) and updating driver to add support for multiple KMS (MONGOSH-1550).


packages/build/src/packaging/download-crypt-library.ts
```diff
@@ -34,7 +34,7 @@ export async function downloadCryptLibrary(
   );
   // Download mongodb for latest server version, including rapid releases
   // (for the platforms that they exist for, i.e. for ppc64le/s390x only pick stable releases).
-  let versionSpec = 'continuous';
+  let versionSpec = process.env.MONGOSH_SERVER_TEST_VERSION || 'continuous';
```
mabaasit, line 37: @addaleax i used this approach to correctly pull the crypt-shared library. Let me know if this looks good to you.

mabaasit: okay, this change broke many tests and i added [d1086d8](https://github.com/mongodb-js/mongosh/pull/2076/commits/d1086d81636787d577be5add724643ca59206b6e) to explicitly specify the version.

addaleax: Mhhh no, I don't think we want to artificially use a different `crypt_shared` library here – that would mean that we'd test mongosh differently than what our users actually see, which kind of makes the test pointless, no?

Can we just use a 8.x rc version here for the time being?

PR #2084, opened 2024-07-17
fix(shell-api): apply readPreference to admin read commands MONGOSH-1837

addaleax:
This mirrors 949c2b5ee87, apart from applying corresponding changes to `_runAdminCommand()` instead of `_runCommand()`.

### DIFF
--- a/packages/shell-api/src/database.ts
+++ b/packages/shell-api/src/database.ts
@@ -185,10 +185,14 @@ export default class Database extends ShellApiWithMongoClass {
     cmd: Document,
     options: CommandOperationOptions = {}
   ): Promise<Document> {
-    return this.getSiblingDB('admin')._runCommand(cmd, {
-      ...(await this._baseOptions()),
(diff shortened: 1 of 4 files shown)


packages/shell-api/src/database.ts
```diff
@@ -185,10 +185,14 @@ export default class Database extends ShellApiWithMongoClass {
     cmd: Document,
     options: CommandOperationOptions = {}
   ): Promise<Document> {
-    return this.getSiblingDB('admin')._runCommand(cmd, {
-      ...(await this._baseOptions()),
```
kmruiz, line 189: Why are we removing the baseOptions here? Not sure the reason.

kmruiz (approved):

packages/shell-api/src/database.ts
addaleax: `_runCommand()` already applies it, right?

PR #2092, opened 2024-07-24
feat(shell-api): add shard moveCollection and unshardCollection helpers

mabaasit:


packages/i18n/src/locales/en_US.ts
```diff
@@ -2045,6 +2045,30 @@ const translations: Catalog = {
                 'Returns a cursor with information about metadata inconsistencies',
               example: 'sh.checkMetadataConsistency(<options>)',
             },
+            moveCollection: {
+              link: 'https://docs.mongodb.com/manual/reference/method/sh.moveCollection',
```
mabaasit, line 2049: The current link to this helper (moveCollection) is https://www.mongodb.com/docs/upcoming/reference/method/sh.moveCollection/#mongodb-method-sh.moveCollection and i've added is to the future one, follows the same pattern as other existing commands here. 
For the other 3 helpers, there exists no link in upcoming docs and i anyway put it here. 

If this does not feel ok, I'd drop it.

packages/shell-api/src/shard.ts
```diff
@@ -693,4 +693,58 @@ export default class Shard extends ShellApiWithMongoClass {
...
+    });
+  }
+
+  @serverVersions(['8.0.0', ServerVersions.latest])
+  @apiVersions([])
+  @returnsPromise
+  async abortUnshardCollection(ns: string): Promise<Document> {
+    assertArgsDefinedType([ns], ['string'], 'Shard.abortUnshardCollection');
+    this._emitShardApiCall('abortUnshardCollection', {
+      abortMoveCollection: ns,
```
addaleax, line 744:
```suggestion
      abortUnshardCollection: ns,
```

addaleax (approved):

PR #2097, opened 2024-07-26
chore(shell-api): show warning when connection is not mongos MONGOSH-1328

mabaasit:


packages/shell-api/src/collection.ts
```diff
@@ -2065,6 +2066,12 @@ export default class Collection extends ShellApiWithMongoClass {
   async getShardDistribution(): Promise<CommandResult> {
     this._emitCollectionApiCall('getShardDistribution', {});
 
+    if (!isMongosConnection(this._mongo._serviceProvider)) {
+      throw new MongoshInvalidInputError(
+        'Please connect to a mongos before calling this function.'
+      );
```
addaleax, line 2072: The ticket asks for a warning rather than an error – what would currently happen if this is run against a single shard? It only displays the information for a that shard, I guess?

mabaasit: with warning, we continue the execution or do an early return? Early return means that user running it again, will not see any output.
if its run against a single shard, it throws with `Collection reviews is not sharded`.

addaleax: > if its run against a single shard, it throws with `Collection reviews is not sharded`.

Oh. I think this might just be about updating the error message for that case then?

Generally, I'd have expected a behavior that's closer to what our `getConfigDB()` helper does, i.e. check whether the server self-reports to be a mongos and just print a warning in that case, without fully taking away the option to run the command from the user (i.e. just make them _aware_ that the command may not be giving them the information they seek).

mabaasit: just updated the PR with that. so with that we logged a warning and also did not take away this option to the run command from the user.

addaleax (approved):

PR #2098, opened 2024-07-26
chore(shell-api): unified shard distribution helper MONGOSH-1694

mabaasit:


packages/shell-api/src/shard.ts
```diff
@@ -694,6 +694,29 @@ export default class Shard extends ShellApiWithMongoClass {
...
+  ): Promise<Document> {
+    this._emitShardApiCall('shardAndDistributeCollection', {
+      ns,
+      key,
+      unique,
+      options,
+    });
+    await this.shardCollection(ns, key, unique, options);
+    // SERVER-92762: Prevent unequal data distribution by setting
+    // numInitialChunks to 1000.
```
addaleax, line 713: Should users be able to override this through `options`?

addaleax (approved):

packages/shell-api/src/shard.ts
mabaasit: so that would make options to this helper include both - options for shardCollection + this option. 
yeah, this sounds good to me and i'll make the change.

mabaasit: just realized that `numInitialChunks` is also used in shardCollection command. added [7bc01c9](https://github.com/mongodb-js/mongosh/pull/2098/commits/7bc01c9965785351fbb886ca5b205b31693b6f8d)

addaleax (approved):

PR #2105, opened 2024-07-29
chore(shell-api): use strict TS config for testing

addaleax:


packages/shell-api/src/shell-api.spec.ts
```diff
@@ -845,49 +849,49 @@ describe('ShellApi', function () {
...
         });
       });
 
       context('with a no-config evaluation listener', function () {
-        let config;
+        let config: ShellConfig;
 
         beforeEach(function () {
-          config = instanceState.context.config;
+          config = instanceState.shellApi.config;
```
mabaasit, line 894: if all the previous assertions were running on instanceState.context, i expect either them to be wrong/fail or something i am not able to follow.

addaleax: Ah, these are the same objects, except that using `.shellApi` instead of `.context` makes this typesafe

`.context` is the global object in which shell commands are being executed, it more or less has to be typed as `any`, and properties from `.shellApi` are just copied over to it:

https://github.com/mongodb-js/mongosh/blob/1570f12d8dcdbf4668e2323edc124fd9c467e9f5/packages/shell-api/src/shell-instance-state.ts#L310-L311

mabaasit (approved):

Anemy (approved):
Nice!!

PR #2112, opened 2024-07-31
 feat: add debug flag for dumping OIDC tokens to output MONGOSH-1845 

addaleax:
Depends on #2109, not logically but since there are a number of adjacent source locations being modified, there are quite a few conflicts otherwise, plus on https://github.com/mongodb-js/oidc-plugin/pull/194 for including the IdP server info

### DIFF
--- a/packages/cli-repl/src/cli-repl.ts
+++ b/packages/cli-repl/src/cli-repl.ts
@@ -47,7 +47,7 @@ import path from 'path';
 import { getOsInfo } from './get-os-info';
 import { UpdateNotificationManager } from './update-notification-manager';
 import { getTimingData, markTime, summariseTimingData } from './startup-timing';
-import { IdPInfo, OIDCCallbackParams } from 'mongodb';
+import type { IdPInfo } from 'mongodb';
 
 /**
  * Connecting text key.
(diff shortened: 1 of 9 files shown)


packages/cli-repl/src/cli-repl.ts
```diff
@@ -1254,4 +1257,96 @@ export class CliRepl implements MongoshIOProvider {
...
+    ): unknown {
+      if (!token) return token;
+      const jwtParts = token.split('.');
+      if (
+        jwtParts.length === 3 &&
+        jwtParts.every(
+          (part) =>
+            Buffer.from(part, 'base64url')
+              .toString('base64url')
+              .replace(/=+$/, '') === part
```
mabaasit, line 1274: Trying to understand what does this condition ensure? 🤔

addaleax: Yeah, added a comment here 🙂 This bit is trying to validate that the parts are base64url-encoded (required for JWTs) by decoding it from base64url and re-encoding it back, then checking whether that's equal to the original string

mabaasit (approved):

PR #2113, opened 2024-07-31
feat: report hostname for atlas COMPASS-8093

alenakhineika:
For every detected Atlas cluster track the resolved hostname without the port when connecting.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6749,10 +6749,11 @@
       "integrity": "sha512-YcRlJ/HkKMahWvaPFTN/al5MGGX6CKQTsBHFTpNduxmeHd40jO8Cj5W7dfrqJAoit5E/G+GOwMaEfWdNrdlihg=="
     },
     "node_modules/@mongodb-js/oidc-mock-provider": {
-      "version": "0.9.0",
-      "resolved": "https://registry.npmjs.org/@mongodb-js/oidc-mock-provider/-/oidc-mock-provider-0.9.0.tgz",
-      "integrity": "sha512-VuUkdUz/EZ17lE/tK9DCVy6e8C+lHdJG2pEXWO+TmhWlenn/EmqVLZ5qZGuJIhAmgoHjlsvKZxUIht8GJYg9Xw==",
+      "version": "0.10.0",
+      "resolved": "https://registry.npmjs.org/@mongodb-js/oidc-mock-provider/-/oidc-mock-provider-0.10.0.tgz",
+      "integrity": "sha512-NTJLomPT6Lew5Wx8f9qqDF9P2Xj/itPPxqPQ+BDbTSD8oRZET345eG+y/mlYhpqA+SN2dYSGV2zuv/VlaqnugQ==",
       "dev": true,
+      "license": "Apache-2.0",
       "dependencies": {
         "yargs": "17.7.2"
       },
@@ -6802,9 +6803,10 @@
       }
     },
     "node_modules/@mongodb-js/oidc-plugin": {
-      "version": "1.0.2",
-      "resolved": "https://registry.npmjs.org/@mongodb-js/oidc-plugin/-/oidc-plugin-1.0.2.tgz",
-      "integrity": "sha512-hwTbkmJ31RPB5ksA6pLepnaQOBz6iurE+uH89B1IIJdxVuiO0Qz+OqpTN8vk8LZzcVDb/WbNoxqxogCWwMqFKw==",
+      "version": "1.1.1",
+      "resolved": "https://registry.npmjs.org/@mongodb-js/oidc-plugin/-/oidc-plugin-1.1.1.tgz",
+      "integrity": "sha512-u2t3dvUpQJeTmMvXyZu730yJzqJ3aKraQ7ELlNwpKpl1AGxL6Dd9Z2AEu9ycExZjXhyjBW/lbaWuEhdNZHEgeg==",
+      "license": "Apache-2.0",
       "dependencies": {
         "express": "^4.18.2",
         "open": "^9.1.0",
@@ -31655,7 +31657,7 @@
       },
       "devDependencies": {
         "@mongodb-js/eslint-config-mongosh": "^1.0.0",
-        "@mongodb-js/oidc-mock-provider": "^0.9.0",
+        "@mongodb-js/oidc-mock-provider": "^0.10.0",
         "@mongodb-js/prettier-config-devtools": "^1.0.1",
         "@mongodb-js/tsconfig-mongosh": "^1.0.0",
         "@types/chai-as-promised": "^7.1.3",
@@ -32063,7 +32065,7 @@
       "license": "Apache-2.0",
       "dependencies": {
         "@mongodb-js/devtools-connect": "^3.0.5",
-        "@mongodb-js/oidc-plugin": "^1.0.2",
+        "@mongodb-js/oidc-plugin": "^1.1.1",
         "@mongosh/errors": "0.0.0-dev.0",
         "@mongosh/service-provider-core": "0.0.0-dev.0",
         "@mongosh/types": "0.0.0-dev.0",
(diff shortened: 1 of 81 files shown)


packages/service-provider-core/src/connect-info.ts
```diff
@@ -43,11 +120,12 @@ export default function getConnectExtraInfo(
     : null;
   const { serverOs: server_os, serverArch: server_arch } =
     getBuildInfo.getBuildEnv(buildInfo);
+  const isAtlas = !!atlasVersion?.atlasVersion || getBuildInfo.isAtlas(uri);
 
   return {
-    is_atlas: !!atlasVersion?.atlasVersion || getBuildInfo.isAtlas(uri),
-    is_localhost: getBuildInfo.isLocalhost(uri),
-    is_do: getBuildInfo.isDigitalOcean(uri),
+    ...(await getHostInformation(resolvedHostname)),
```
addaleax, line 126: This can potentially block for quite some time (it does DNS lookups and HTTP calls), and except for telemetry, we do not currently need this information. 

Can we make this part of the `mongosh:connect` listener that actually emits telemetry instead?

package-lock.json
```diff
@@ -22950,6 +23262,24 @@
         "node": ">=16.20.1"
       }
     },
+    "node_modules/mongodb-cloud-info": {
+      "version": "2.1.3",
+      "resolved": "https://registry.npmjs.org/mongodb-cloud-info/-/mongodb-cloud-info-2.1.3.tgz",
+      "integrity": "sha512-SxvjPZ+4CsPhe0Now9aLpiSHmQGwm1jYu5hshBMkq3hAKnLrBuOwK/3TovY4SQl6h6Utqkx2rBHqLbO9incEZA==",
+      "dependencies": {
+        "cross-fetch": "^3.1.6",
+        "gce-ips": "^1.0.2",
```
addaleax, line 23271: This isn't being used as a production dependency of `mongodb-cloud-info` and adds a decent number of extra dependencies, let's make `gce-ips` a dev dependency in `mongodb-cloud-info` first before we start pulling them in?

packages/types/src/index.ts
```diff
@@ -44,22 +44,28 @@ export interface ShowEvent {
 
 export interface ConnectEvent {
   is_atlas?: boolean;
+  resolved_hostname?: string;
   is_localhost?: boolean;
-  is_do?: boolean;
+  is_do_url?: boolean;
```
addaleax, line 49: Just checking – this is an intentional change to the data format we use for telemetry?

packages/types/src/index.ts
```diff
@@ -44,22 +44,28 @@ export interface ShowEvent {
...
   server_version?: string;
+  mongosh_version: string;
```
addaleax, line 51: Similarly here – this is intentional? We always add this in `setup-logger-and-telemetry.ts` anyway

packages/shell-api/src/shell-instance-state.ts
```diff
@@ -217,6 +220,8 @@ export default class ShellInstanceState {
   }
 
   async fetchConnectionInfo(): Promise<ConnectionInfo | undefined> {
+    const { version } = require('../package.json');
```
addaleax, line 223: This will currently work fine because we're releasing packages using the same version, but there's no guarantee that that would stay that way indefinitely – are you sure this is what we need?

packages/service-provider-core/src/connect-info.ts
```diff
@@ -21,11 +18,59 @@ export interface ConnectionExtraInfo {
...
+
+export type HostInformation = {
+  is_localhost?: boolean;
+  is_atlas_url?: boolean;
+  is_do_url?: boolean; // Is digital ocean url.
+};
+
+export function getHostnameForConnection(topology: any): string | undefined {
+  const resolvedHost =
+    topology?.s?.servers?.values().next().value.description.address ?? '';
```
addaleax, line 31: I know this is already part of the code here, but it's still worth asking – can we avoid this? `service-provider-core` isn't a place for code that depends on Node.js driver logic specifically (that's kinda the point of the package), and it would be great if we could avoid using internal properties like `.s` here

packages/logging/src/setup-logger-and-telemetry.ts
```diff
@@ -141,32 +141,38 @@ export function setupLoggerAndTelemetry(
...
-      'connect',
-      'Connecting to server',
-      params
-    );
 
-    analytics.track({
-      ...getTelemetryUserIdentity(),
-      event: 'New Connection',
-      properties: {
+    (async () => {
```
addaleax, line 150: There's no `await` here

packages/types/src/index.ts
alenakhineika: We align this prom name with the rest of the dev tools.

packages/types/src/index.ts
alenakhineika: Removed together with requiring package.json.

packages/service-provider-core/src/connect-info.ts
alenakhineika: Refactored to get topology from the topologyDescriptionChanged event.

packages/service-provider-server/src/cli-service-provider.ts
```diff
@@ -247,12 +263,18 @@ class CliServiceProvider
         connectionString.toString(),
         clientOptions,
         bus,
-        MongoClientCtor
+        MongoshMongoClient
```
addaleax, line 266: nit: Do you also want to change L257 to use this ctor? It shouldn't make a difference in behavior right now, but it seems like it would be a bit more consistent

packages/service-provider-server/src/cli-service-provider.ts
```diff
@@ -264,6 +286,12 @@ class CliServiceProvider
   public baseCmdOptions: OperationOptions; // public for testing
   private bus: MongoshBus;
 
+  /**
+   * Stores the most recent topology description from the server's SDAM events:
```
addaleax, line 290: This doc comment is the same as the data-service one and seems to imply the same behavior, but it's not; this one only returns the last topology at the time when `.connect()` finishes.

I think it's worth either adjusting the doc comment, or, preferably, actually keeping this value up to date beyond the `CliServiceProvider` instantiation

packages/service-provider-core/src/connect-info.ts
```diff
@@ -21,36 +19,84 @@ export interface ConnectionExtraInfo {
...
+  connectionString,
+  buildInfo,
+  atlasVersion,
+  topology,
+  isLocalAtlas,
+}: {
+  connectionString?: ConnectionString;
+  buildInfo: any;
+  atlasVersion: any;
+  topology: any;
```
addaleax, line 76:
```suggestion
  topology: TopologyDescription;
```

?

packages/service-provider-core/src/connect-info.ts
```diff
@@ -21,36 +19,84 @@ export interface ConnectionExtraInfo {
...
   is_local_atlas?: boolean;
+} & HostInformation;
+
+export type HostInformation = {
+  is_localhost?: boolean;
+  is_atlas_url?: boolean;
+  is_do_url?: boolean; // Is digital ocean url.
+};
+
+export function getHostnameForConnection(topology: any): string | undefined {
```
addaleax, line 30:
```suggestion
export function getHostnameForConnection(topology: TopologyDescription): string | undefined {
```

and maybe even

```suggestion
// TODO: Move this into service-provider-server
export function getHostnameForConnection(topology: TopologyDescription): string | undefined {
```

packages/logging/src/setup-logger-and-telemetry.spec.ts
```diff
@@ -78,14 +81,114 @@ describe('setupLoggerAndTelemetry', function () {
...
+      is_localhost: false,
+      is_atlas: true,
+      resolved_hostname: 'test-data-sets-00-02-a011bb.mongodb.net',
       node_version: 'v12.19.0',
-    } as any);
+    });
+
+    // Make sure cloud info is resolved.
+    await wait(500);
+
```
addaleax, line 145:
```suggestion
```

packages/logging/src/setup-logger-and-telemetry.spec.ts
```diff
@@ -78,14 +81,114 @@ describe('setupLoggerAndTelemetry', function () {
...
       uri: 'mongodb://localhost/',
       is_localhost: true,
       is_atlas: false,
+      resolved_hostname: 'localhost',
+      node_version: 'v12.19.0',
+    });
+
+    // Make sure cloud info is resolved.
+    await wait(500);
+
```
addaleax, line 94:
```suggestion
```

packages/logging/src/setup-logger-and-telemetry.spec.ts
```diff
@@ -3,10 +3,13 @@ import { expect } from 'chai';
 import { MongoLogWriter } from 'mongodb-log-writer';
 import { setupLoggerAndTelemetry } from './';
 import { EventEmitter } from 'events';
+import { promisify } from 'util';
 import { MongoshInvalidInputError } from '@mongosh/errors';
 import type { MongoshBus } from '@mongosh/types';
 import { toSnakeCase } from './setup-logger-and-telemetry';
 
+const wait = promisify(setTimeout);
+
```
addaleax, line 12:
```suggestion
```

packages/logging/src/setup-logger-and-telemetry.spec.ts
```diff
@@ -3,10 +3,13 @@ import { expect } from 'chai';
 import { MongoLogWriter } from 'mongodb-log-writer';
 import { setupLoggerAndTelemetry } from './';
 import { EventEmitter } from 'events';
+import { promisify } from 'util';
```
addaleax, line 6:
```suggestion
```

addaleax (commented):
Nice! This looks pretty good already 🙂

packages/logging/src/setup-logger-and-telemetry.spec.ts
alenakhineika: Thx for noticing!

addaleax (approved):

PR #2116, opened 2024-08-01
fix(cli-repl): prompt for password when redirecting output MONGOSH-1136

mabaasit:


addaleax (approved):
LGTM, feel free to add a test but this is probably just fine as it is as well

packages/cli-repl/src/run.spec.ts
```diff
@@ -97,4 +97,52 @@ describe('CLI entry point', function () {
...
+      const [code] = await once(proc, 'exit');
+      expect(code).to.equal(1);
+      expect(promptedForPassword).to.be.true;
+    });
+
+    it('requests password when user is redirecting output', async function () {
+      const args = [
+        ...pathToRun,
+        'mongodb://amy@localhost:27017',
+        '> /dev/null',
```
addaleax, line 127: This doesn't work on Windows, does it? I think you'd need to end up doing something like `stdio: ['pipe', 'ignore', pipe']` or maybe actually opening a file and passing that fd as the stdout entry?

addaleax (approved):

packages/cli-repl/src/run.spec.ts
mabaasit: yep, it does not! will check it!

mabaasit: cleaned up in [83511f0](https://github.com/mongodb-js/mongosh/pull/2116/commits/83511f0fc5e6faf0a8a76c8d2c86f00e5e271b7b)

addaleax (approved):

PR #2119, opened 2024-08-05
feat: enable tlsUseSystemCA by default MONGOSH-1852

alenakhineika:
Enable tlsUseSystemCA by default.

### DIFF
--- a/config/webpack.base.config.js
+++ b/config/webpack.base.config.js
@@ -31,7 +31,6 @@ module.exports = {
   externals: {
     "node:crypto": "commonjs2 crypto",
     electron: "commonjs2 electron", // optional dep of the OIDC plugin
-    ssh2: "commonjs2 ssh2",
   },
 
   optimization: {
--- a/config/webpack.base.config.js
+++ b/config/webpack.base.config.js
@@ -22,7 +22,9 @@ module.exports = {
       // only provide features that Node.js also provides out of the box.
       browserslist: path.resolve(__dirname, '..', 'scripts', 'dummy-browserslist.js'),
       tr46: path.resolve(__dirname, '..', 'scripts', 'tr46-stub.js'),
+      // Optional native-addon dependencies of ssh2
       cpufeatures: false,
+      './crypto/build/Release/sshcrypto.node': false
     }
   },
 
--- a/config/webpack.base.config.js
+++ b/config/webpack.base.config.js
@@ -23,14 +23,14 @@ module.exports = {
       browserslist: path.resolve(__dirname, '..', 'scripts', 'dummy-browserslist.js'),
       tr46: path.resolve(__dirname, '..', 'scripts', 'tr46-stub.js'),
       // Optional native-addon dependencies of ssh2
-      cpufeatures: false,
-      './crypto/build/Release/sshcrypto.node': false
+      'cpu-features': false,
+      './crypto/build/Release/sshcrypto.node': false,
     }
   },
 
   externals: {
-    "node:crypto": "commonjs2 crypto",
-    electron: "commonjs2 electron", // optional dep of the OIDC plugin
+    'node:crypto': 'commonjs2 crypto',
+    electron: 'commonjs2 electron', // Optional dep of the OIDC plugin
   },
 
   optimization: {
--- a/config/webpack.base.config.js
+++ b/config/webpack.base.config.js
@@ -22,12 +22,14 @@ module.exports = {
       // only provide features that Node.js also provides out of the box.
       browserslist: path.resolve(__dirname, '..', 'scripts', 'dummy-browserslist.js'),
       tr46: path.resolve(__dirname, '..', 'scripts', 'tr46-stub.js'),
+      cpufeatures: false,
     }
   },
 
   externals: {
     "node:crypto": "commonjs2 crypto",
-    electron: "commonjs2 electron" // optional dep of the OIDC plugin
+    electron: "commonjs2 electron", // optional dep of the OIDC plugin
+    ssh2: "commonjs2 ssh2",
   },
 
   optimization: {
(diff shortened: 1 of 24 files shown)


packages/arg-parser/src/arg-mapper.ts
```diff
@@ -222,7 +222,7 @@ const MAPPINGS: {
   tlsCertificateKeyFile: (i, v) => setUrlParam(i, 'tlsCertificateKeyFile', v),
   tlsCertificateKeyFilePassword: (i, v) =>
     setUrlParam(i, 'tlsCertificateKeyFilePassword', v),
-  tlsUseSystemCA: (i, v) => setDriver(i, 'useSystemCA', v),
+  tlsUseSystemCA: (i) => i,
```
addaleax, line 225:
```suggestion
```

This is the default already, we can just remove it

packages/e2e-tests/test/e2e-oidc.spec.ts
```diff
@@ -413,8 +413,15 @@ describe('OIDC auth e2e', function () {
...
     // so the best we can verify here is that auth failed *on the server*
     shell.assertContainsOutput(/MongoServerError: Authentication failed/);
+  });
+
+  it('does not fail without --tlsUseSystemCA', async function () {
+    await fs.mkdir(path.join(tmpdir.path, 'certs'), { recursive: true });
+    await fs.copyFile(
+      getCertPath('ca.crt'),
+      path.join(tmpdir.path, 'certs', 'somefilename.crt')
+    );
```
addaleax, line 423: Where is this file used?

packages/e2e-tests/test/e2e-oidc.spec.ts
```diff
@@ -413,8 +413,15 @@ describe('OIDC auth e2e', function () {
     // We cannot make the mongod server accept the mock IdP's certificate,
     // so the best we can verify here is that auth failed *on the server*
     shell.assertContainsOutput(/MongoServerError: Authentication failed/);
+  });
```
addaleax, line 416: The test here should probably be renamed, and `--tlsUseSystemCA` dropped from its command line, right?

packages/arg-parser/src/arg-mapper.ts
addaleax, line 1: The option is also mentioned in `cli-options.ts` in this directory

addaleax (commented):
🚀

packages/service-provider-server/src/cli-service-provider.ts
```diff
@@ -236,7 +236,6 @@ class CliServiceProvider
       delete clientOptionsCopy.oidc;
       delete clientOptionsCopy.parentHandle;
       delete clientOptionsCopy.parentState;
```
addaleax, line 238: Completely optional, but feel free to add these (since it goes hand in hand with the rest of the devtools-connect update), otherwise I'll do it in a follow-up PR:

```suggestion
      delete clientOptionsCopy.proxy;
      delete clientOptionsCopy.applyProxyToOIDC;
```

packages/e2e-tests/test/e2e-oidc.spec.ts
alenakhineika: Right, we don't need it for this test.

packages/e2e-tests/test/e2e-oidc.spec.ts
alenakhineika: Ah actually, the test should be definately renamed, but I wanted to keep `--tlsUseSystemCA` so we verify that setting this cli option does not error. The test right under this does not set `--tlsUseSystemCA`.

packages/arg-parser/src/arg-mapper.ts
alenakhineika: But do we want to allow users to keep using this option?

addaleax: The `CliOptions` type in `packages/arg-parser/src/cli-options.ts` defines the options that are available to mongosh _internally_. We don't want to use this option anymore, including by accident, so we should drop it there.

The options that users can pass are defined in `packages/cli-repl/src/arg-parser.ts`. We want users to be able to specify it, so it can keep being listed there.

config/webpack.base.config.js
```diff
@@ -22,12 +22,14 @@ module.exports = {
       // only provide features that Node.js also provides out of the box.
       browserslist: path.resolve(__dirname, '..', 'scripts', 'dummy-browserslist.js'),
       tr46: path.resolve(__dirname, '..', 'scripts', 'tr46-stub.js'),
+      cpufeatures: false,
```
addaleax, line 25:
```suggestion
      // Optional native-addon dependencies of ssh2
      cpufeatures: false,
      './crypto/build/Release/sshcrypto.node': false
```

config/webpack.base.config.js
```diff
@@ -22,12 +22,14 @@ module.exports = {
...
     }
   },
 
   externals: {
     "node:crypto": "commonjs2 crypto",
-    electron: "commonjs2 electron" // optional dep of the OIDC plugin
+    electron: "commonjs2 electron", // optional dep of the OIDC plugin
+    ssh2: "commonjs2 ssh2",
```
addaleax, line 32:
```suggestion
```

packages/cli-repl/src/run.ts
```diff
@@ -12,6 +12,7 @@ function enableFipsIfRequested() {
   }
 }
 enableFipsIfRequested();
+(process as any).noDeprecation = true;
```
addaleax, line 15: I don't think we need this anymore, since we replaced `new Buffer()` in `socksv5`

packages/e2e-tests/test/e2e-oidc.spec.ts
addaleax: @alenakhineika I think the test here still needs to be renamed, right?

addaleax (approved):
LGTM once we get CI green

PR #2129, opened 2024-08-14
fix(shell-api): expand fallback condition for $collstats MONGOSH-1425

paula-stacho:


packages/shell-api/src/collection.spec.ts
```diff
@@ -1467,62 +1467,83 @@ describe('Collection', function () {
...
+                ...new Error('Stale Config'),
+                codeName: 'StaleConfig',
+              },
+              {
+                ...new Error('Failed to Parse'),
+                codeName: 'FailedToParse',
+              },
+            ];
+
+            fallbackErrors.forEach((mockError) => {
```
addaleax, line 1485: for-of?

addaleax (approved):

packages/shell-api/src/collection.spec.ts
paula-stacho: sure! convention, or any other reason?

addaleax: Well, yeah, convention, and that convention's mostly in place because for-of expands nicely to other use cases like iterables and for-await-of

addaleax (approved):

PR #2154, opened 2024-09-04
chore(deps): bump node-gyp in top-level package.json

addaleax:
node-gyp 10.x is required for compatibility with recent Python versions (which has been an issue at least for me locally).


lerouxb (approved):

addaleax:
Okay, this doesn't seem to work because we also have to support Python 3.6, which is the oldest one that's available out-of-the-box on the rhel70 hosts.

PR #2165, opened 2024-09-11
chore(cli-repl): skip startup warnings for local atlas environments MONGOSH-1776

kraenhansen:
Adding a check on the connection info extra info to determine if the server is "local atlas" to conditionally skip calling the "startupWarnings" command when initializing the REPL.


addaleax (approved):
Just re-affirming my LGTM

kraenhansen:
Merging as tests pass locally and it seems we'd need a more stable CI to rely on it as a blocker.

PR #2187, opened 2024-10-01
MONGOSH-1703: FIx mongosh error when converting invalid regular expression in db.currentOp()

gagik:


packages/e2e-tests/test/e2e-current-op.spec.ts
```diff
@@ -0,0 +1,84 @@
+import { expect } from 'chai';
+import {
+  skipIfApiStrict,
+  startSharedTestServer,
+} from '../../../testing/integration-testing-hooks';
+import { TestShell } from './test-shell';
+
+describe('e2e currentOp', function () {
+  skipIfApiStrict();
```
gagik, line 9: I kept it as most e2e had it but this can be removed right? not too familiar with full implications of the apiStrict option

packages/e2e-tests/test/e2e-current-op.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+      // Insert a dummy object so find commands will actually run with the delay.
+      await helperShell.executeLine('db.coll.insertOne({})');
+    });
+
+    it('should return the correct operation', async function () {
+      const regexOperation = helperShell.executeLine(
+        `db.coll.find({$where: function() { sleep(${OPERATION_TIME}) }}).projection({test: 1})`
+      );
+      helperShell.assertNoErrors();
+      await currentOpShell.executeLine(`sleep(${CURRENT_OP_WAIT_TIME})`);
```
gagik, line 40: I could use an actual JS timeout-based sleep here.

packages/e2e-tests/test/e2e-current-op.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+        `db.coll.find({$where: function() { sleep(${OPERATION_TIME}) }}).projection({re: BSONRegExp('${regExpString}')})`
+      );
+      helperShell.assertNoErrors();
+
+      await currentOpShell.executeLine(`sleep(${CURRENT_OP_WAIT_TIME})`);
+
+      const currentOpCall = await currentOpShell.executeLine(`db.currentOp()`);
+      currentOpShell.assertNoErrors();
+
+      const regExpStringWithoutEscapes = regExpString.replace(/\\/g, '');
```
gagik, line 77: Okay so I find this to be a bit weird. I am not sure why the output would end up being without the escapes.
So running the command
```
(...).projection({re: BSONRegExp('^(?i)\Qchho0842\E')}
```
leads to currentOp
```
projection: { re: BSONRegExp('^(?i)Qchho0842E', '') }
```
This seems odd. Aren't the escapes important in this case?

packages/e2e-tests/test/e2e-current-op.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+      expect(currentOpCall).not.to.include("find: 'coll'");
+      expect(currentOpCall).not.to.include(
+        `filter: { '$where': Code('function() { sleep(${OPERATION_TIME}) }') }`
+      );
+      expect(currentOpCall).not.to.include('projection: { test: 1 }');
+    });
+
+    it('should work when the operation contains regex', async function () {
+      const regExpString = '^(?i)\\\\Qchho0842\\E';
+      // The values from currentOp removes redundant escapes such as \E
```
gagik, line 66: Seems like expected behavior right?

gagik:
I realize CI isn't passing but I have trouble tracking down anything related to my change (and I know some tests have always been flakey), let me know if there's some way I can access better logs.

gagik:
Remaining failing CI tests seem unrelated to the changes 🤞

packages/e2e-tests/test/e2e-current-op.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+      await regexOperation;
+
+      currentOpCall = await currentOpShell.executeLine(`db.currentOp()`);
+
+      currentOpShell.assertNoErrors();
+      expect(currentOpCall).not.to.include("find: 'coll'");
+      expect(currentOpCall).not.to.include(
+        `filter: { '$where': Code('function() { sleep(${OPERATION_TIME}) }') }`
+      );
+      expect(currentOpCall).not.to.include('projection: { test: 1 }');
```
kraenhansen, line 59: I wonder if you could make the test faster and still test what you want, by simply killing the test shell instead of awaiting the `regexOperation` and asserting the current op gets reset?

packages/e2e-tests/test/e2e-current-op.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+describe('e2e currentOp', function () {
+  skipIfApiStrict();
+
+  const testServer = startSharedTestServer();
+
+  context('with 2 shells', function () {
+    let helperShell: TestShell;
+    let currentOpShell: TestShell;
+
+    const OPERATION_TIME = 1_000;
```
kraenhansen, line 17: Could this be smaller without causing issues? 
```suggestion
    const OPERATION_TIME = CURRENT_OP_WAIT_TIME * 2;
```

packages/shell-api/src/database.spec.ts
```diff
@@ -1796,10 +1796,11 @@ describe('Database', function () {
         },
       });
 
-      const READ_PREFERENCE = {
+      const AGGREGATE_OPTIONS = {
```
kraenhansen, line 1799: What's the context for these changes?

packages/shell-api/src/database.ts
```diff
@@ -1063,7 +1063,10 @@ export default class Database extends ShellApiWithMongoClass {
     }
 
     const adminDb = this.getSiblingDB('admin');
-    const aggregateOptions = { $readPreference: { mode: 'primaryPreferred' } };
+    const aggregateOptions = {
+      $readPreference: { mode: 'primaryPreferred' },
+      bsonRegExp: true,
```
kraenhansen, line 1068: That was it? Does this deserve a comment in code?

kraenhansen (approved):
👍 Nothing blocking.

packages/e2e-tests/test/e2e-current-op.spec.ts
gagik: I agree it'd still be the test I want for the purposes of changes in this PR; I thought this is nice in terms of it provides a full e2e usecase in terms of currentOp also working after the operation ended but that's not responsibility of this change

packages/shell-api/src/database.ts
gagik: Yes, from my understanding it was just a matter of passing the correct set options in this case. I'll add a comment.

packages/shell-api/src/database.spec.ts
gagik: Since the change just adds a new field to the default constant aggregate options, I just updated the test to match (and the naming since what is being checked is the options as a whole)

packages/e2e-tests/test/e2e-current-op.spec.ts
addaleax: We just add this for tests that cover features which aren't part of the stable API – and $currentOp, in particular, is not part of that.

Fwiw I think this is also something that could be added to the existing tests in `e2e.spec.ts`, unless we expect to add more tests for currentOp-specific behavior in the long run

packages/e2e-tests/test/e2e-current-op.spec.ts
addaleax: Yeah, that might just be a bit easier 🙂

packages/e2e-tests/test/e2e-current-op.spec.ts
addaleax: Okay, so, the way I read this, it's not even $currentOp that removes the E, it's just normal JS escaping behavior that applies before we run any code at all; the code here translates to 

```js
executeLine(`db.coll.find(...).projection({re: BSONRegExp('${regExpString}')})`)
```

which expands to 


```js
db.coll.find(...).projection({re: BSONRegExp('^(?i)\\Qchho0842\E')})
```

which is evaluated as JS code, where `'^(?i)\\Qchho0842\E' === '^(?i)\\Qchho0842E'`.

If you want to actually test that $currentOp drops the `\` in front of an `E`, you'd need to escape it like the `Q` here.

I feel like this would be a bit easier to read if we used `String.raw` for defining these literals and `JSON.stringify()` for inserting them into JS code without having to worry about escaping logic; i.e.

```js
const regExpString = String.raw `^(?i)\Qchho0842\E`;
...
helperShell.executeLine(`db.coll.find(...).projection({re: BSONRegExp(${JSON.stringify(regExpString)})})`);
```

packages/e2e-tests/test/e2e-current-op.spec.ts
gagik: @kraenhansen  Ah right, looking the exact code again. I actually made 2 tests, since this is also introducing e2e test to currentOp in general:
1. **one "generic"  e2e (unrelated to the changes):** which just makes sure the currentOp works as expected; I think here it's nice to await the command (which confused me because I still call it `regexOperation`, it actually has no `regex` so I'll rename this) and make sure that running currentOp again updates accordingly. I think it's nice to have it just for general currentOp testing.
2. **the test related to these changes** in which case I no longer care about the operation and it doesn't get awaited as you describe.

packages/e2e-tests/test/e2e-current-op.spec.ts
gagik: Okay today was the day I learnt 😄 
```tsx
"\\Q"  === "\\\Q" === String.raw`\Q`
```
Surprised I hadn't used `String.raw` much before, good to know, thank you!

packages/e2e-tests/test/util-helpers.ts
```diff
@@ -0,0 +1,3 @@
+export function sleep(ms: number) {
```
gagik, line 1: I could also inline it in the test's scope it but I figure there's a good chance this kind of file will be needed? At least I was looking for that.

packages/e2e-tests/test/e2e.spec.ts
```diff
@@ -1938,4 +1939,80 @@ describe('e2e', function () {
...
+        expect(currentOpCall).not.to.include(
+          `filter: { '$where': Code('function() { sleep(${OPERATION_TIME}) }') }`
+        );
+        expect(currentOpCall).not.to.include('projection: { test: 1 }');
+      });
+
+      it('should work when the operation contains regex', async function () {
+        const regExpString = String.raw`^(?i)\Qchho0842\E`;
+
+        // Stringify the reg exp and drop the quotation marks.
```
gagik, line 1995: Stringify wraps it in `"` and `currentOp` output on my machine uses `'`

packages/e2e-tests/test/util-helpers.ts
addaleax: I wouldn't overthink it tbh – currently, we use inline code like `await new Promise((resolve) => setTimeout(resolve, 400));` in a bunch of tests, but this obviously also works.

We do tend to prefer adding TS return types wherever possible, by the way 🙂 Only reason it's not a linter rule is that the codebase currently still has a bunch of exceptions to that rule

addaleax (approved):

packages/e2e-tests/test/util-helpers.ts
gagik: Sounds good, I just copy-pasted this from an old `node-runtime-worker-thread` spec and didn't change it. I like the readability of `sleep` though don't mind a promise either; maybe it's worth keeping? We use sleep in the worker thread tests anyhow.

Re: TS return types, great to hear! I love strong typing, just didn't think much about it here (and similarly influenced by the lack of it right now).

gagik: Yeah, I can make a small chore PR later for that.

gagik:
Rest of the failing tests are not related to the changes, going ahead.

PR #2189, opened 2024-10-03
fix(shell-api): Fix support for sharded time series collections with getShardDistribution MONGOSH-1447

gagik:
When the getShardDistribution command is invoked on a sharded time series collection, the output says that the collection is not actually sharded. If the user invokes the same command on the related bucket collection, it produces an incorrect output.

This PR fixes both.

### DIFF
--- a/packages/shell-api/src/collection.ts
+++ b/packages/shell-api/src/collection.ts
@@ -2078,7 +2078,8 @@ export default class Collection extends ShellApiWithMongoClass {
 
   /**
    * Helper for getting collection info of sharded timeseries collections
-   * @returns the bucket count and collection info based on given collStats
+   * @returns collection info based on given collStats, null if the information is not found or
+   * if the collection is not timeseries.
    */
   async _getShardedTimeseriesCollectionInfo(
     config: Database,
--- a/packages/shell-api/src/collection.ts
+++ b/packages/shell-api/src/collection.ts
@@ -2076,6 +2076,47 @@ export default class Collection extends ShellApiWithMongoClass {
     });
   }
 
+  /**
+   * Helper for getting collection info of sharded timeseries collections
+   * @returns the bucket count and collection info based on given collStats
+   */
+  async _getShardedTimeseriesCollectionInfo(
+    config: Database,
+    collStats: Document[]
+  ): Promise<{
+    timeseriesBucketCount: number | null;
+    timeseriesCollectionInfo: Document | null;
+  }> {
+    const timeseriesShardStats = collStats.find(
+      (extractedShardStats) =>
+        typeof extractedShardStats.storageStats.timeseries !== 'undefined'
+    );
+
+    if (!timeseriesShardStats) {
+      return {
+        timeseriesCollectionInfo: null,
+        timeseriesBucketCount: null,
+      };
+    }
+
+    const { storageStats } = timeseriesShardStats;
+
+    const timeseries: Document = storageStats['timeseries'];
+
+    const timeseriesBucketCount: number | null =
+      timeseries['bucketCount'] ?? null;
+    const timeseriesBucketNs: string | undefined = timeseries['bucketsNs'];
+
+    const timeseriesCollectionInfo = await config
+      .getCollection('collections')
+      .findOne({
+        _id: timeseriesBucketNs,
+        ...onlyShardedCollectionsInConfigFilter,
+      });
+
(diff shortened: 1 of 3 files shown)


packages/shell-api/src/collection.ts
```diff
@@ -2131,39 +2177,42 @@ export default class Collection extends ShellApiWithMongoClass {
...
             numChunks: numChunks,
-            avgObjSize: extShardStats.storageStats.avgObjSize,
+            avgObjSize: extractedShardStats.storageStats.avgObjSize,
           };
 
           const key = `Shard ${shardStats.shardId} at ${shardStats.host}`;
 
-          const estChunkData =
+          // In sharded timeseries collections, count is 0.
+          const shardStatsCount = shardStats.count ?? 0;
```
gagik, line 2195: A weird side-effect is that count (and consequently `docs` and `estimated docs per chunk`) will always be 0. This seems consistent with `collStats`

https://github.com/mongodb-js/mongosh/blob/a2fa6b458187ed32547e002d11d1492c7d7c016a/packages/shell-api/src/collection.ts#L1676-L1680

Is this what we want?

packages/shell-api/src/collection.ts
```diff
@@ -2115,12 +2138,35 @@ export default class Collection extends ShellApiWithMongoClass {
...
+        _id: ns,
+        ...onlyShardedCollectionsInConfigFilter,
+      });
+
+    if (!existingConfigCollectionsInfo) {
+      const timeseriesCollectionInfo =
+        await this._getShardedTimeseriesCollectionInfo(config, collStats);
+
+      if (!timeseriesCollectionInfo) {
+        throw new MongoshInvalidInputError(
```
gagik, line 2156: there could be more things here that have gone wrong, so one could differentiate more errors potentially

packages/shell-api/src/collection.ts
```diff
@@ -2076,6 +2076,44 @@ export default class Collection extends ShellApiWithMongoClass {
...
+
+    if (!timeseriesShardStats) {
+      return null;
+    }
+
+    const { storageStats } = timeseriesShardStats;
+
+    const timeseries: Document = storageStats['timeseries'];
+
+    const timeseriesBucketNs: string | undefined = timeseries['bucketsNs'];
```
addaleax, line 2101: I'm a bit surprised the linter doesn't say anything, but either way, we can just use regular JS property accesses here, right?

```suggestion
    const timeseries: Document = storageStats.timeseries;

    const timeseriesBucketNs: string | undefined = timeseries.bucketsNs;
```

packages/shell-api/src/collection.ts
```diff
@@ -2115,12 +2139,35 @@ export default class Collection extends ShellApiWithMongoClass {
...
+        throw new MongoshInvalidInputError(
+          `Collection ${this._name} is not sharded`,
+          ShellApiErrors.NotConnectedToShardedCluster
+        );
+      }
+
+      configCollectionsInfo = timeseriesCollectionInfo;
+    } else {
+      configCollectionsInfo = existingConfigCollectionsInfo;
+    }
```
addaleax, line 2165: Just a feeling, but my gut reaction here is that it would make sense to move this conditional into a helper like `_getShardedTimeseriesCollectionInfo()` as well – if we end up re-using `_getShardedTimeseriesCollectionInfo()` at some point, there's a good chance that we'll want the rest of this block as well too

packages/shell-api/src/collection.ts
addaleax: Feels odd to say this, but this honestly sounds like a perfect use case for `NaN` – we don't know the number, and actually calculating it would be prohibitively expensive, so we might as well just forward this "we don't know" information to the caller?

addaleax (approved):
Awesome work!

packages/shell-api/src/collection.ts
gagik: If I understand correctly, that'd mean actually returning intentionally returning `NaN` for `docs` and `estimated docs per chunk` in the result? Could this break something for potential users if they would otherwise expect some number? (i.e. maybe even Compass?)

addaleax: Compass doesn't use shell methods under the hood (currently)

I guess it could break scripts, yeah, but it still seems better than providing incorrect results, doesn't it?

gagik:
All the remaining failing CI tasks seem unrelated, going ahead 🤞

PR #2191, opened 2024-10-04
fix(cli-repl): multiline dots and prompts

kraenhansen:
For now this PR adds a failing test demonstrating how writing a multiline command prints an extraneous prompt before multiline dots (the "..." prefixed after every line-break in the input).


packages/e2e-tests/test/e2e.spec.ts
```diff
@@ -571,6 +571,30 @@ describe('e2e', function () {
...
+      // Wait for some multiline dots
+      await eventually(() => {
+        shell.assertContainsOutput('... ...');
+      });
+      const outputOffsetAfterDots = shell.output.length;
+      const output = shell.output
+        .substring(outputOffsetBefore, outputOffsetAfterDots)
+        .trim();
+      // Expect just the multiline dots
+      expect(output.startsWith('... ...')).equals(
```
addaleax, line 592: This test does fail for me, locally, but unfortunately I think that's actually expected behavior? The first prompt printed by mongosh in response to the `.writeInputLine()` call is a regular prompt, not a multiline prompt, but that's happening because the first line of the `writeInputLine()` argument is empty/whitespace-only.

That being said ... yeah, there's a good chance that that causes the flakiness in the FLE test at least. I feel like it's probably a good idea to just always `.trim()` the input to `.writeInputLine()`? For raw input, there's always `.writeInput()` itself

PR #2196, opened 2024-10-07
chore(deps): pin cookie to ^0.7.2

addaleax:
This is being reported as an issue by our vuln checker, although we are not affected by the vulnerability itself as mongosh does not use HTTP cookies in production code.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13824,10 +13824,10 @@
       "dev": true
     },
     "node_modules/cookie": {
-      "version": "0.4.2",
-      "resolved": "https://registry.npmjs.org/cookie/-/cookie-0.4.2.tgz",
-      "integrity": "sha512-aSWTXFzaKWkvHO1Ny/s+ePFpvKsPnjc551iI41v3ny/ow6tBG5Vd+FuqGNhh1LxOmVzOlGUriIlOaokOvhaStA==",
-      "dev": true,
+      "version": "0.7.2",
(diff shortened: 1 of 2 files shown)


kmruiz (approved):

addaleax:
@kmruiz Fair question! There's two parts to this:

- Based on `npm ls`, we only use `cookie` in production as a dependency of `express`, i.e. in the local server for OIDC Auth Code Flow. We don't rely on cookies in this express server, though.
- In our update notification code, there _is_ a piece of statefulness for server/client communication, but mostly we're using HTTP ETag here to avoid duplicate downloads of the mongosh downloads JSON file: https://github.com/mongodb-js/mongosh/blob/7803ee15e2d6d239bb6b57713216310cda2d0749/packages/cli-repl/src/update-notification-manager.ts#L105 So, here as well, we just don't use cookies currently, and I don't think we'd be using this package if we did (but who knows, I haven't looked into that hypothetical too deeply).

kraenhansen:
Did you investigate upgrading the dependencies bringing in `cookie` instead? It looks to me that `@mongosh/browser-repl` is pulling it in transitively via `karma`, `@pmmmwh/react-refresh-webpack-plugin` and `webpack-dev-server` which all have new versions available.

addaleax:
@kraenhansen Yes! Unfortunately, the path you're mentioning isn't a path which brings this package into our production dependencies. The relevant one is via `express` (transitively through `@mongodb-js/oidc-plugin`), and unfortunately `express` pins its dependency versions to exact versions and there's no new `express` release out yet.

PR #2203, opened 2024-10-08
feat(shell-api): Account for orphan documents count in getShardDistribution() helper MONGOSH-1838

gagik:
`getShardDistribution()` should correctly account orphan documents counts when calculating statistics. This count is present under the "numOrphanDocs" field in $collStats output in version (>= 6.0)


packages/shell-api/src/result.ts
```diff
@@ -127,3 +127,21 @@ export class CursorIterationResult extends ShellApiValueClass {
     this.documents = [];
   }
 }
+
+export type GetShardDistributionResult = {
```
gagik, line 131: This is just my attempt at introducing some more typing into this area of the code, let me know what you think.
I see it more as documentation more than anything; not sure the location of it (inside `result.ts`) is best but probably simplest; could also be in a `command-result-types.ts`.

packages/shell-api/src/collection.spec.ts
```diff
@@ -2276,6 +2276,70 @@ describe('Collection', function () {
...
+        });
+
+        it('should account for numOrphanDocs when calculating size', async function () {
+          const shardDistribution = await collection.getShardDistribution();
+
+          const { storageStats } = mockedShardStats;
+          expect(shardDistribution.type).equals('StatsResult');
+          const adjustedSize =
+            storageStats.size -
+            storageStats.numOrphanDocs * storageStats.avgObjSize;
```
gagik, line 2329: I tried to look deeper into the collections code to understand whether the size of oprhan docs would be `storageStats.numOrphanDocs * storageStats.avgObjSize`; I haven't found too much certainty but can't think of any other way I'd get this value.
https://github.com/mongodb/mongo/blob/d59d165c43c59f7a0a9b30d912bd7daa5da91a92/src/mongo/db/catalog/collection_impl.h#L305

packages/shell-api/src/collection.ts
```diff
@@ -2179,10 +2182,21 @@ export default class Collection extends ShellApiWithMongoClass {
               .findOne({ _id: extractedShardStats.shard }),
             config.getCollection('chunks').countDocuments(countChunksQuery),
           ]);
+
+          // Since 6.0, there can be orphan documents indicated by numOrphanDocs.
+          // These orphan documents need to be accounted for in the size calculation.
+          const orphanDocumentsSize = !extractedShardStats.storageStats
+            .numOrphanDocs
+            ? 0
+            : extractedShardStats.storageStats.numOrphanDocs *
```
gagik, line 2191: Based on:
https://github.com/mongodb/mongo/blob/d3313627b309a13a0a8050e2cf16b784b4c1ded3/jstests/sharding/sharded_data_distribution.js#L60

packages/shell-api/src/result.ts
addaleax: I'd move it to either `collection.ts`, i.e. where it's used or `helpers.ts`. This file here is mostly for directly defining the types of results that a consumer of shell-api will see, but for `getShardDistribution()`, we still use the `CommandResult` wrapper (and that's fine).

> This is just my attempt at introducing some more typing into this area of the code, let me know what you think.

Definitely an improvement! You may want to know for context that we don't currently have any consumers of shell-api that typically make use of our typings, since most usage of this package is through regular, untyped JS written by shell users. Sometimes we do re-use methods internally, and in those cases, the typings are helpful of course.

However, we have an upcoming project for improving autocomplete functionality for shell code across our products based on a PoC we did around the start of the year (with code in #1802 and https://github.com/mongodb-js/poc-mongodb-autocomplete). This will make use of TS definitions for our API, so it's definitely worth adding TS definitions for all cases which a user may reasonably want to autocomplete 🙂

packages/shell-api/src/result.ts
```diff
@@ -127,3 +127,21 @@ export class CursorIterationResult extends ShellApiValueClass {
...
+  Totals: {
+    data: string;
+    docs: number;
+    chunks: number;
+  } & {
+    [individualShardDistribution: string]: string[];
+  };
+} & {
+  [individualShardResult: string]: {
```
addaleax, line 140: Optional, feel free to take it or leave it:

```suggestion
    [individualShardDistribution: `Shard ${string} at ${string}`]: string[];
  };
  [individualShardResult: `Shard ${string}`]: {
```

Using template literals we can both be a bit more precise here and avoid the potential conflicts between `Totals`, `data`, `docs`, `chunks` and the generated keys

addaleax (approved):

packages/shell-api/src/result.ts
gagik: That's awesome! I was thinking of check in this direction but didn't think template literals were that advanced in TS, super cool!

packages/shell-api/src/collection.ts
```diff
@@ -2179,10 +2182,22 @@ export default class Collection extends ShellApiWithMongoClass {
               .findOne({ _id: extractedShardStats.shard }),
             config.getCollection('chunks').countDocuments(countChunksQuery),
           ]);
+
+          // Since 6.0, there can be orphan documents indicated by numOrphanDocs.
+          // These orphan documents need to be accounted for in the size calculation.
+          const orphanDocumentsCount =
+            typeof extractedShardStats.storageStats.numOrphanDocs === 'number'
```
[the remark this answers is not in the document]
gagik: Agreed, I had it as `??` originally but then ended up changing to this to check some failing tests (which turned out to be related to `avgObjSize`)

packages/shell-api/src/result.ts
gagik: That's great context, thank you! autocompletitions would be super cool

PR #2207, opened 2024-10-09
feat(cli-repl): Ask for tlsCertificateKeyFilePassword when it is needed MONGOSH-1730

gagik:
When using TLS enabled with a Certificate that requires a password, the mongosh does not prompt for a passphrase automatically, failing with an error. This adds a password prompt for that case.


addaleax (approved):
Nice!

PR #2214, opened 2024-10-11
feat(shell-api): add shardedDataDistribution to sh.status() MONGOSH-1326

gagik:
Considering the balancing of a sharded namespace would be sensitive to orphans and data volume, and in order to count with a mechanism that easily allows checking the current state of the cluster, it would be good if the sh.status() provides the following information per sharded namespaces:

    Estimated number of orphan documents per shard
    Data volume per shard

This is supported in `6.0.3+`


packages/shell-api/src/helpers.ts
```diff
@@ -644,10 +668,83 @@ export async function getPrintableShardStatus(
...
+          [count: number]:
+            | 'Success'
+            | `Failed with error '${string}', from ${string} to ${string}`;
+        };
+    'Balancer active window is set between'?: `${string} and ${string} server local time`;
+    'Last reported error'?: string;
+    'Time of Reported error'?: string;
+    'Collections with active migrations'?: `${string} started at ${string}`[];
+  };
+  shardedDataDistribution?: ShardedDataDistribution;
```
gagik, line 733: I am adding it as a separate field at the moment since that's just a simple solution, but we could make it cleaner/nicer for the user. Open to suggestions

packages/shell-api/src/helpers.ts
```diff
@@ -295,25 +299,27 @@ export async function getPrintableShardStatus(
...
-        ).toArray()) as any[]
+        ).toArray()) as { _id: string; num: number }[]
       ).map((z: { _id: string; num: number }) => {
         return { [z._id]: z.num };
       });
     }
   }
 
-  const balancerRes: Record<string, any> = {};
+  const balancerRes = {} as ShardingStatusResult['balancer'];
```
lerouxb, line 322: Why the preference for it this way around as opposed to:

```
const balancerRes: ShardingStatusResult['balancer'] = {};
```

gagik: I'm not a fan of this myself but it'd actually be invalid; we're essentially doing a cast here to an invalid object, kind of breaking type safety; still better than not doing a record. Here's a [playground example of what I mean](https://www.typescriptlang.org/play/?#code/PTAEHUFMBsGMHsC2lQBd5oBYoCoE8AHSAZVgCcBLA1UABWgEM8BzM+AVwDsATAGiwoBnUENANQAd0gAjQRVSQAUCEmYKsTKGYUAbpGF4OY0BoadYKdJMoL+gzAzIoz3UNEiPOofEVKVqAHSKymAAmkYI7NCuqGqcANag8ABmIjQUXrFOKBJMggBcISGgoAC0oACCbvCwDKgU8JkY7p7ehCTkVDQS2E6gnPCxGcwmZqDSTgzxxWWVoASMFmgYkAAeRJTInN3ymj4d-jSCeNsMq-wuoPaOltigAKoASgAywhK7SbGQZIIz5VWCFzSeCrZagNYbChbHaxUDcCjJZLfSDbExIAgUdxkUBIursJzCFJtXydajBVDtOYAXlAAG9FCUStx8ldUJROMwANyKAC+wQQnEENAYADVHKAabSeWJhBVuaLHAFXDSAEQAZhV3IFQrEYrIACYWVVJTytY0dQqyGqjRK6czQOqVabgiocM0PGQvIh4H0GMD2DQvm4zMx2AxmJB+LBoOpEn74HoRF4VQBRVYMRALfQqpLYlXgByoADkwgAcpAJCqgioAPJfMjvQSRrSQI6oRwKVzSPCgJzenTDLD6FAIRDQ4SXIMSH3REQGDjYhZMVgcHhBEpAA)

lerouxb (approved):
Looks good. I'm no expert in this department, though.

packages/shell-api/src/helpers.ts
lerouxb: Wondering if it is better to only assert once it is correct rather than cast. Just a style nit, really.

packages/shell-api/src/helpers.ts
```diff
@@ -644,10 +668,83 @@ export async function getPrintableShardStatus(
...
+    'Balancer active window is set between'?: `${string} and ${string} server local time`;
+    'Last reported error'?: string;
+    'Time of Reported error'?: string;
+    'Collections with active migrations'?: `${string} started at ${string}`[];
+  };
+  shardedDataDistribution?: ShardedDataDistribution;
+  databases: { database: Document; collections: Document }[];
+};
+
+export type ShardedDataDistribution = {
```
lerouxb, line 737: Don't know where to leave this comment, but I noticed on the ticket the comment:  "if it's possible I think the order should be ownedSizeBytes, numOwnedDocuments, orphanedSizeBytes, numOrphanedDocs".

Wondering how much effort we should put in for that request. Probably not worth it and kinda weird precedent to be setting. Wouldn't want sorting/ordering to break this when they add fields, for example.

gagik: Yeah, wanted to comment on that too; I think the idea of controlling ordering is weird, I imagine it'd be consistent with whatever order the DB returns and I doubt that'd change much if at all. I will try to run it on an actual and see if that's the case.

packages/shell-api/src/helpers.ts
gagik: I think the benefit of the cast is that it gives us IDE hints so at least it adds some checks while the fields are being set; though I agree it also can give misleading idea of what fields are already set depending on context.

Seems fine for initialization though; do you mean like a runtime assertion function?

lerouxb: I just mean that at line 322 it isn't really a ShardingStatusResult['balancer']. But I don't have better ideas.

packages/shell-api/src/helpers.ts
```diff
@@ -644,10 +668,83 @@ export async function getPrintableShardStatus(
...
 
+  let shardedDataDistribution: ShardedDataDistribution | undefined;
+
+  try {
+    // Available since >= 6.0.3
+    const adminDb = db.getSiblingDB('admin');
+    shardedDataDistribution = (await (
+      await adminDb.aggregate([{ $shardedDataDistribution: {} }])
+    ).toArray()) as ShardedDataDistribution;
+  } catch (e) {
```
addaleax, line 679:
```suggestion
  } catch {
```

packages/shell-api/src/helpers.ts
```diff
@@ -644,10 +668,83 @@ export async function getPrintableShardStatus(
     )
   ).filter((dbEntry) => !!dbEntry);
 
+  let shardedDataDistribution: ShardedDataDistribution | undefined;
+
+  try {
+    // Available since >= 6.0.3
+    const adminDb = db.getSiblingDB('admin');
+    shardedDataDistribution = (await (
+      await adminDb.aggregate([{ $shardedDataDistribution: {} }])
```
addaleax, line 677: Can we do this in parallel to other operations, like we do around ~L527?

packages/shell-api/src/helpers.ts
```diff
@@ -331,13 +337,13 @@ export async function getPrintableShardStatus(
...
       // Is the balancer currently active
-      let balancerRunning = 'unknown';
+      let balancerRunning: 'yes' | 'no' | 'unknown' = 'unknown';
       try {
         const balancerStatus = await configDB.adminCommand({
           balancerStatus: 1,
         });
         balancerRunning = balancerStatus.inBalancerRound ? 'yes' : 'no';
-      } catch (err: any) {
+      } catch (err) {
```
addaleax, line 346:
```suggestion
      } catch {
```

packages/shell-api/src/helpers.ts
```diff
@@ -224,10 +224,12 @@ export function processDigestPassword(
...
 export async function getPrintableShardStatus(
-  configDB: Database,
+  db: Database,
   verbose: boolean
-): Promise<Document> {
-  const result = {} as any;
+): Promise<ShardingStatusResult> {
+  const result = {} as ShardingStatusResult;
+
+  const configDB = await getConfigDB(db);
```
addaleax, line 232: Uh... this is an unintentional breaking change, no? You can't pass a `configDB` object to `sh.status()` explicitly anymore after this

packages/shell-api/src/helpers.ts
```diff
@@ -364,7 +370,9 @@ export async function getPrintableShardStatus(
       if (activeLocks?.length > 0) {
         balancerRes['Collections with active migrations'] = activeLocks.map(
           (lock) => {
-            return `${lock._id} started at ${lock.when}`;
+            // This type assertion is necessary for the string literal type check
+            // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
+            return `${lock._id} started at ${lock.when}` as `${string} started at ${string}`;
```
addaleax, line 375:
```suggestion
            return `${lock._id} started at ${lock.when}` as const;
```

addaleax (changes requested):

packages/shell-api/src/helpers.ts
gagik: Oh right, my bad, somehow I thought the `getConfigDB` was always happening already.

I suppose if `configDB` is passed explicitly then we won't end up doing shardedDataDistribution? or does `getSiblingDb` imply one can get `admin` db from a given `config` db?

packages/shell-api/src/change-stream-cursor.spec.ts
```diff
@@ -342,7 +342,6 @@ describe('ChangeStreamCursor', function () {
     it('isExhausted fails', function () {
       try {
         cursor.isExhausted();
-        expect.fail('missed exception');
```
addaleax, line 345: Unrelated change slipped in here?

packages/shell-api/src/helpers.ts
```diff
@@ -472,11 +491,12 @@ export async function getPrintableShardStatus(
...
             ])
-          ).toArray()
+          ).toArray()) as MigrationResult[]
         );
 
-        const migrationsRes: Record<number, string> = {};
-        migrations.forEach((x: any) => {
+        const migrationsRes: ShardingStatusResult['balancer']['Migration Results for the last 24 hours'] =
+          {};
+        migrations.forEach((x) => {
```
addaleax, line 499: Just a note for the future, but if you run into existing `.forEach()` into the future, don't be shy to refactor them to `for ... of` 🙂

addaleax (approved):

PR #2218, opened 2024-10-14
fix(shell-api): Add support for running aggregate database with a single stage MONGOSH-1868

gagik:
Running aggregate() on a database with a non-array pipeline will now run a single stage aggregation. This will make it consistent with `db.{collection}aggregate`  behavior so
```
db.aggregate({$listLocalSessions: {}}
```
is same as 
```
db.aggregate([{$listLocalSessions: {}}])
```

### DIFF
--- a/packages/shell-api/src/collection.ts
+++ b/packages/shell-api/src/collection.ts
@@ -164,7 +164,7 @@ export default class Collection extends ShellApiWithMongoClass {
   async aggregate(
     pipeline: Document[],
     options: Document & { explain: ExplainVerbosityLike }
-  ): Promise<Document>;
+  ): Promise<AggregationCursor>;
   async aggregate(...stages: Document[]): Promise<AggregationCursor>;
   @returnsPromise
   @returnType('AggregationCursor')
@@ -191,7 +191,10 @@ export default class Collection extends ShellApiWithMongoClass {
       this._database._name,
       this._name,
       pipeline,
-      { ...(await this._database._baseOptions()), ...aggOptions },
+      {
+        ...(await this._database._baseOptions()),
+        ...aggOptions,
+      },
       dbOptions
     );
     const cursor = new AggregationCursor(this._mongo, providerCursor);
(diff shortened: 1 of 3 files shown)


packages/shell-api/src/database.ts
```diff
@@ -1429,6 +1437,7 @@ export default class Database extends ShellApiWithMongoClass {
         CommonErrors.CommandFailed
       );
     }
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
```
gagik, line 1440: autoformat kept removing the `as Document[]` for me so I had to add this comment

Anemy (approved):

packages/shell-api/src/collection.ts
```diff
@@ -164,7 +164,7 @@ export default class Collection extends ShellApiWithMongoClass {
   async aggregate(
     pipeline: Document[],
     options: Document & { explain: ExplainVerbosityLike }
-  ): Promise<Document>;
+  ): Promise<AggregationCursor>;
```
addaleax, line 167: But the existing typings here are correct? The return type in this overload is an explain document, not a cursor

packages/shell-api/src/database.ts
```diff
@@ -412,6 +412,14 @@ export default class Database extends ShellApiWithMongoClass {
...
   }
 
+  async aggregate(
+    singleStage: Document,
+    options?: Document
+  ): Promise<AggregationCursor>;
+  async aggregate(
+    pipeline: Document[],
+    options?: Document
+  ): Promise<AggregationCursor>;
```
addaleax, line 422: Totally optional, but as far as I can tell these overloads don't differ in return type or allowed parameter combinations, so you don't need to include them as explicit overloads:

```suggestion
```

addaleax (approved):

gagik:
Now that I'm looking into documentation more, perhaps it'd make sense to completely sync up with `.aggregate()` with how `db.coll.aggregate` works so it's less confusing; i.e. you can also specify aggregation stages as`aggregate({ stage1: ... }, {stage2: ...}` (in which case `options` will not be possible). It seems more reasonable to have a 1:1 mapping between these two aggregate functions' params.

PR #2219, opened 2024-10-15
fix(cli-repl): Fix depcheck with a more permissive workaround for the searchParams type

gagik:


lerouxb (approved):

packages/cli-repl/src/cli-repl.ts
```diff
@@ -1017,12 +1017,11 @@ export class CliRepl implements MongoshIOProvider {
...
-    searchParams: ReturnType<
-      typeof ConnectionString.prototype.typedSearchParams<DevtoolsConnectOptions>
-    >
+    searchParams: unknown
   ): Promise<boolean> {
-    const tlsCertificateKeyFile = searchParams.get('tlsCertificateKeyFile');
-    const tlsCertificateKeyFilePassword = searchParams.get(
+    const searchMap = searchParams as Map<keyof DevtoolsConnectOptions, string>;
+    const tlsCertificateKeyFile = searchMap.get('tlsCertificateKeyFile');
+    const tlsCertificateKeyFilePassword = searchMap.get(
```
addaleax, line 1024: This could have used a bit more description of what it is fixing, plus maybe a TODO comment to revert it at some point (if that is possible) since we're losing type safety here?

PR #2227, opened 2024-10-18
feat(tests): Add individual evergreen test results with XUnit

gagik:
Allows for individual testing results to be viewed and makes it easier to detect specific flaky tests with Foliage in the future.


.evergreen.yml
```diff
@@ -12448,6 +12516,7 @@ tasks:
           executable_os_id: win32
       - func: run_e2e_tests
         vars:
+          task_name_var: "${task_name}"
```
addaleax, line 12519: Keep in mind that this file is automatically generated from `.evergreen/evergreen.yml.in` 🙂

gagik: Oh! so are my changes not doing anything at all? I have suspicion that `${task_name}` is always empty right now for the `run_e2e_tests` and have been battling it for so long

gagik:
```
npm run update-evergreen-config
```
oh cool!

addaleax: Yeah, although any changes you make to `.evergreen.yml` should already have an effect (they'd just get overwritten after merging).

I don't know what exactly's going on, but you shouldn't need the `task_name_var` indirection here, using `${task_name}` directly in the `env:` list should already work

gagik: Yeah, I am suspecting this expansion is either not working or somehow getting lost in the process. I looked through evergreen docs and nothing suspicious there.

gagik: Okay seems like the env variable just somehow doesn't reach the bash script

gagik:
evergreen validate -p mongosh

.logs/empty.xml
```diff
@@ -0,0 +1,3 @@
+<!-- An empty test suite which is needed as not all tasks produce XUnit results and the current Evergreen setup always expects some file to be uploaded -->
+<testsuite name="empty" tests="0" failures="0" errors="0" skipped="0" timestamp="Fri, 18 Oct 2024 09:21:06 GMT" time="8.907">
```
gagik, line 2: I can't think of a better way to go about this... evergreen seems to fail the job completely when no files are found in the attach results step

addaleax (approved):

PR #2229, opened 2024-10-21
fix(shell-api): Align database and collection aggregate functions MONGOSH-1868

gagik:
Looking into the documentation for both database and collection functions made me realize it'd be a lot clearer to align them completely. Instead of https://github.com/mongodb-js/mongosh/pull/2218 which would allow
```ts
db.aggregate({ $stage: 1 }, { options: true }) 
// => db.aggregate([{ $stage: 1 }], { options: true })
```
It'd be better to align with how `db.{collection}.aggregate` works i.e.
```ts
db.aggregate({ $stage: 1 }, { $stage2: true }) 
// => db.aggregate([{ $stage: 1 }, { $stage2: true })]);
```

This does mean if not using array, one cannot pass options but this is already the behavior of collection `aggregate` so it'd be best (and most intuitive) to keep it consistent.

### DIFF
--- a/packages/shell-api/src/collection.ts
+++ b/packages/shell-api/src/collection.ts
@@ -66,6 +66,7 @@ import type {
   UpdateOptions,
   DropCollectionOptions,
   CheckMetadataConsistencyOptions,
+  AggregateOptions,
 } from '@mongosh/service-provider-core';
 import type { RunCommandCursor, Database } from './index';
 import {
@@ -159,26 +160,27 @@ export default class Collection extends ShellApiWithMongoClass {
    */
   async aggregate(
     pipeline: Document[],
-    options: Document & { explain?: never }
+    options: AggregateOptions & { explain?: never }
   ): Promise<AggregationCursor>;
   async aggregate(
     pipeline: Document[],
-    options: Document & { explain: ExplainVerbosityLike }
+    options: AggregateOptions & { explain: ExplainVerbosityLike }
   ): Promise<Document>;
   async aggregate(...stages: Document[]): Promise<AggregationCursor>;
   @returnsPromise
   @returnType('AggregationCursor')
   @apiVersions([1])
-  async aggregate(...args: any[]): Promise<any> {
-    let options;
-    let pipeline;
+  async aggregate(...args: unknown[]): Promise<unknown> {
+    let options: AggregateOptions;
+    let pipeline: Document[];
     if (args.length === 0 || Array.isArray(args[0])) {
       options = args[1] || {};
-      pipeline = args[0] || [];
+      pipeline = (args[0] as Document[]) || [];
     } else {
       options = {};
-      pipeline = args || [];
+      pipeline = (args as Document[]) || [];
     }
+
     if ('background' in options) {
       await this._instanceState.printWarning(
         aggregateBackgroundOptionNotSupportedHelp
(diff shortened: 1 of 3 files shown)


packages/shell-api/src/collection.ts
```diff
@@ -159,26 +160,27 @@ export default class Collection extends ShellApiWithMongoClass {
    */
   async aggregate(
     pipeline: Document[],
-    options: Document & { explain?: never }
-  ): Promise<AggregationCursor>;
+    options: AggregateOptions & { explain: ExplainVerbosityLike }
```
gagik, line 163: I realize removing `& {explain?: never}` seems not intuitive but this actually was causing issues in cases where.

```ts
const options: AggregateOptions = {};
// This would complain that neither of the overrides match the type of options
db.coll.aggregate({}, options);
```
With this setup, what we're looking for still happens as the first overload takes priority so when explain is defined, it'd pick this and return `Document` as expected

packages/shell-api/src/collection.ts
```diff
@@ -159,26 +160,27 @@ export default class Collection extends ShellApiWithMongoClass {
...
+    options: AggregateOptions
+  ): Promise<AggregationCursor>;
   async aggregate(...stages: Document[]): Promise<AggregationCursor>;
   @returnsPromise
   @returnType('AggregationCursor')
   @apiVersions([1])
-  async aggregate(...args: any[]): Promise<any> {
-    let options;
-    let pipeline;
+  async aggregate(...args: unknown[]): Promise<unknown> {
```
gagik, line 173: I prefer switching `any`s to `unknown`s to make the type cast moments more visible (it kind of seems to be a lint rule considering `any` gets underlined?) but I am open to using `any` if it messes the code too much

packages/shell-api/src/collection.ts
gagik: Hm, I tried again with the never and now I am not seeing it; a bit weird

addaleax: This exact example seems to work for me – can you share a reproduction/the actual error for this case?

gagik: @addaleax actually I only see it in the database.ts file; adding `& {explain: never}` leads to the sql function complaining about it, but maybe there's something specific there

gagik:
```
  Overload 1 of 3, '(pipeline: Document[], options: AggregateOptions & { explain: ExplainVerbosityLike; }): Promise<Document>', gave the following error.
    Argument of type 'AggregateOptions | undefined' is not assignable to parameter of type 'AggregateOptions & { explain: ExplainVerbosityLike; }'.
      Type 'undefined' is not assignable to type 'AggregateOptions & { explain: ExplainVerbosityLike; }'.
        Type 'undefined' is not assignable to type 'AggregateOptions'.
  Overload 2 of 3, '(pipeline: Document[], options: (AggregateOptions & { explain: never; }) | undefined): Promise<AggregationCursor>', gave the following error.
    Argument of type 'AggregateOptions | undefined' is not assignable to parameter of type '(AggregateOptions & { explain: never; }) | undefined
```
I think it's specifically the case where `AggregateOptions | undefined`?

gagik: Yeah what I was describing is specifically a problem with
```
    await this.aggregate([], {} as AggregateOptions | undefined);
```

addaleax: Fwiw, if you "just" want to fix this here, you could probably make that work by using `this.aggregate([...], options ?? {})` in the .sql() function, but I don't know if I have a strong preference for that really

addaleax (approved):

PR #2230, opened 2024-10-21
chore(deps): Add node-gyp version control across different distros MONGOSH-1891

gagik:
Recent CI update likely led to Python being bumped to 3.12+ which means our current node-gyp library can no longer support it without some more manual setup. We need to bump node-gyp and also handle some older dependencies (such as RHEL which should work with 3.6).


package.json
```diff
@@ -112,7 +112,7 @@
     "mocha": "^10.2.0",
     "mongodb": "^6.9.0",
     "mongodb-runner": "^5.7.0",
-    "node-gyp": "^9.0.0",
+    "node-gyp": "^10.2.0",
```
addaleax, line 115: Idk if this will work out in CI, but if you use

```suggestion
    "node-gyp": "^9.0.0 || ^10.2.0",
```

you could probably replace the `npm run replace-package` bit with just `npm install node-gyp@9` without breaking CI

gagik:
Need to add some of the older Linux distros to the filter as well and should be good!

gagik:
I lost and re-found the issue. e2e_tests_linux_x64_60x Setup Failed still happens so probably needs manual handling

addaleax (approved):

paula-stacho (approved):

gagik:
Rest of the failing tests seem historically flaky and don't seem to be related to node-gyp so going ahead.

PR #2232, opened 2024-10-22
chore: rename service-provider-server and `CliServiceProvider`

addaleax:
`ServiceProvider` is our abstraction for how mongosh interacts with the database.
The two notable implementations of that that are actually being used are based on
the Node.js driver and the Java driver, respectively.

Neither `CliServiceProvider` (has no inherent connection with CLI usage) nor
`service-provider-server` (connects to the same servers as any other service
provider would) are great names and only reflect the historical architecture for
mongosh at this point.

So, let's rename this to something that accurately reflects what this service
provider implementation is and what distinguishes it from others!


gagik (approved):
Definitely clearer naming with this!

PR #2236, opened 2024-10-24
chore: Migrate rhel81-power8 to rhel8-power9 MONGOSH-1835

nirinchev:
This is the second half of MONGOSH-1835 - it moves to building for ppc to the new rhel8-power9 hosts. We had to bump the mongodb toolchain to v4 for these hosts due to an issue compiling `stat.h` with the v3 toolchain. This is temporary and likely to go away once MONGOSH-864 is completed.


.evergreen/setup-env.sh
```diff
@@ -1,8 +1,17 @@
 set -e
 set -x
 
+OS_ARCH="$(uname "-m")"
+if [ "$OS_ARCH" = "aarch64" ] || [ "$OS_ARCH" = "arm64" ] ; then
+    echo "[INFO] Choosing v4 because OS_ARCH is $OS_ARCH"
+    export TOOLCHAIN_PATH='/opt/mongodbtoolchain/v4/bin'
```
addaleax, line 7: The idea here is that we'd pick v4 on ppc64le and s390x – not sure what the `uname -m` values are for that, but we should probably be using those?

nirinchev: Yeah, I was absentmindedly reverting that commit and got distracted before I changed the values 🤦‍♂️ One thing though - do we need to pick v4 for s390x builds - those are already passing successfully with the v3 toolchain.

addaleax: Yeah, if v3 works for s390x, then it probably doesn't matter for now 🤷‍♀️

.evergreen/install-node-source.sh
```diff
@@ -7,6 +7,16 @@ EVGDIR="$ROOT_DIR/.evergreen"
 NVM_DIR="$EVGDIR/.nvm"
 ORIGINAL_PATH="${PATH}"
 
+OS_ARCH="$(uname "-m")"
+if [ "$OS_ARCH" = "ppc64le" ] || [ "$OS_ARCH" = "ppc64" ] ; then
+    echo "[INFO] Choosing v4 because OS_ARCH is $OS_ARCH"
```
addaleax, line 12: Just a small note, in scripts that already have `set -x` set, echoing back debug information can be nice for formatting but doesn't usually provide other benefits, since bash will already echo back the values for OS_ARCH, PATH, etc. when they are being set

addaleax (approved):

.evergreen/install-node-source.sh
nirinchev: Yeah, I guess there's a small benefit of making this easier to grep if you notice it in the parsley logs + that was the original code before it got removed in https://github.com/mongodb-js/mongosh/commit/315fefa444e36c982a8e6905976199aaf2f5793d, so figured might keep it like that. I don't have strong opinions though and happy to remove it to keep the script more concise.

PR #2239, opened 2024-10-25
fix(ci): allocate more memory to fix test_vscode MONGOSH-1892

gagik:
Fixes [test_vscode](https://spruce.mongodb.com/version/674075f3fd09fd00073611ec/tasks?page=0&sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC&taskName=test_vscode) which had started to run out of memory by bumping the max allocation size.


addaleax (approved):
Just in case you happen to know, have you figured out what part of the test takes this much memory? I think this is perfectly fine as a workaround but it's slighly concerning that the extension tests take this much time, and I feel like this might point in the direction of a memory leak that's bad enough to cause real-world trouble

gagik:
@addaleax Hm yeah, I just tried to create and use a mocha reporter which kept track of `process.memoryUsage().heapUsed` before and after a test and did not see anything >10mb. Probably kind of a flawed method but I am wondering if the memory usage issue is just coming from the electron being spun up? I could create a ticket in VSCODE to potentially investigate this in the future.

lerouxb:
My hunch is that it happens when something fails. In my experience you often get something trying to pretty-print gigabytes of something on error. Or similar :P

PR #2245, opened 2024-10-29
feat(shell-api): add stream processor modify MONGOSH-1864

mongodb-matthew-normyle:
This PR adds a stream processor `modify` method. It can be used to modify the definition of a stream processor.

### DIFF
--- a/packages/shell-api/src/stream-processor.ts
+++ b/packages/shell-api/src/stream-processor.ts
@@ -75,7 +75,7 @@ export default class StreamProcessor extends ShellApiWithMongoClass {
     if (Array.isArray(pipelineOrOptions)) {
       options['pipeline'] = pipelineOrOptions;
     } else if (typeof pipelineOrOptions === 'object') {
-      if (Object.keys(options).length != 0) {
+      if (Object.keys(options).length !== 0) {
         throw new MongoshInvalidInputError(
           'If the first argument to modify is an object, the second argument should not be specified.',
           CommonErrors.InvalidArgument
--- a/packages/shell-api/src/stream-processor.ts
+++ b/packages/shell-api/src/stream-processor.ts
@@ -62,6 +62,7 @@ export default class StreamProcessor extends ShellApiWithMongoClass {
    * modify can be used like so:
    *  Change the pipeline:
    *    sp.name.modify(pipeline)
+   *  Change the pipeline with additional options:
    *    sp.name.modify(pipeline, {resumeFromCheckpoint: false})
    *  For modify requests that don't change the pipeline:
    *    sp.name.modify({resumeFromCheckpoint: false})
--- a/packages/shell-api/src/stream-processor.ts
+++ b/packages/shell-api/src/stream-processor.ts
@@ -64,24 +64,21 @@ export default class StreamProcessor extends ShellApiWithMongoClass {
    *    sp.name.modify(newPipeline)
    *  Keep the same pipeline, change other options:
    *   sp.name.modify({resumeFromCheckpoint: false})
+   *  Change the pipeline and set additional options:
+   *    sp.name.modify(newPipeline, {resumeFromCheckpoint: false})
    */
   async modify(options: Document): Promise<Document>;
   async modify(pipeline: Document[], options?: Document): Promise<Document>;
 
-  /**
-   * modify is used to modify a stream processor definition, like below:
-   *  Change the pipeline and set additional options:
-   *    sp.name.modify(newPipeline, {resumeFromCheckpoint: false})
-   */
   @returnsPromise
   async modify(
     pipelineOrOptions: Document[] | Document,
-    options: Document = {}
+    options?: Document
   ): Promise<Document> {
     if (Array.isArray(pipelineOrOptions)) {
       options = { ...options, pipeline: pipelineOrOptions };
     } else if (typeof pipelineOrOptions === 'object') {
-      if (Object.keys(options).length !== 0) {
+      if (options) {
         throw new MongoshInvalidInputError(
           'If the first argument to modify is an object, the second argument should not be specified.',
           CommonErrors.InvalidArgument
--- a/packages/shell-api/src/stream-processor.ts
+++ b/packages/shell-api/src/stream-processor.ts
@@ -1,4 +1,5 @@
 import type { Document } from '@mongosh/service-provider-core';
+import { CommonErrors, MongoshInvalidInputError } from '@mongosh/errors';
 
 import type Mongo from './mongo';
 import { asPrintable } from './enums';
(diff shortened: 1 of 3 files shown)


mongodb-matthew-normyle:
I also manually tested this change against the ASP service with the below:

```
AtlasStreamProcessing> sp.createStreamProcessor("testedit2", [{$source: {connectionName: "DemoKafka", topic: "t1"}}, {$project: {device_id: 1, foo: "bar"}}, {$project: {_stream_meta: 0}}, {$emit: {connectionName: "DemoKafka", topic: "testedit2"}}])
Atlas Stream Processor: testedit2
AtlasStreamProcessing> sp.testedit2.start()
{ ok: 1 }
AtlasStreamProcessing> sp.testedit2.stop()
{ ok: 1 }
AtlasStreamProcessing> sp.testedit2.modify([{$source: {connectionName: "DemoKafka", topic: "t1"}}, {$project: {device_id: 1, foo: "BAZ"}}, {$project: {_stream_meta: 0}}, {$emit: {connectionName: "DemoKafka", topic: "testedit2"}}])
{ ok: 1 }
AtlasStreamProcessing> sp.testedit2.start()
{ ok: 1 }

```

packages/shell-api/src/stream-processor.ts
```diff
@@ -57,6 +58,37 @@ export default class StreamProcessor extends ShellApiWithMongoClass {
...
+   *    sp.name.modify(pipeline)
+   *  Change the pipeline with additional options:
+   *    sp.name.modify(pipeline, {resumeFromCheckpoint: false})
+   *  For modify requests that don't change the pipeline:
+   *    sp.name.modify({resumeFromCheckpoint: false})
+   */
+  @returnsPromise
+  async modify(
+    pipelineOrOptions: Document[] | Document,
+    options: Document = {}
```
mongodb-matthew-normyle, line 73: The method parameters were defined in the [Scope](https://docs.google.com/document/d/1wQ_e2TrJKegyei0nDFajVejE6nqahmftieyZ3UEMnXE/edit?tab=t.0) and [PD](https://docs.google.com/document/d/1Qvw4fI6DrSxqJbhODnHaY1p3aJIz0632KsJ-qK401d8/edit?tab=t.0#heading=h.wdpdgdpcthgg). I think [watch](https://www.mongodb.com/docs/manual/reference/method/db.collection.watch/) has a similar structure that support arguments of different types. But if this is against the best mongosh practices we can modify the scope.

mongodb-matthew-normyle:
I'm waiting on an evergreen build to see if there are any other linting errors to fix.

packages/shell-api/src/stream-processor.ts
addaleax: This is fine, but:
- Please add TS overloads that describe the different signatures, i.e. `modify(options: Document): ...` and `modify(pipeline: Document[], options: Document): ...`
- Please add return types to the definitions

packages/shell-api/src/stream-processor.ts
```diff
@@ -57,6 +58,43 @@ export default class StreamProcessor extends ShellApiWithMongoClass {
...
+  ) {
+    if (Array.isArray(pipelineOrOptions)) {
+      options['pipeline'] = pipelineOrOptions;
```
addaleax, line 76:
```suggestion
      options.pipeline = pipelineOrOptions;
```

or, even better:

```suggestion
      options = { ...options, pipeline: pipelineOrOptions };
```

since best practice is not to modify arguments passed to functions

packages/shell-api/src/stream-processor.ts
```diff
@@ -57,6 +58,43 @@ export default class StreamProcessor extends ShellApiWithMongoClass {
...
+    } else if (typeof pipelineOrOptions === 'object') {
+      if (Object.keys(options).length !== 0) {
```
addaleax, line 78:
```suggestion
      if (options) {
```

packages/shell-api/src/stream-processor.ts
mongodb-matthew-normyle: Done, thanks!

packages/shell-api/src/stream-processor.ts
mongodb-matthew-normyle: I think `if (options) {}` evaluates to true when options is `{}` (the default argument). So, if I make this change,  I'll need to change `options: Document = {}` to `options: Document = null`. Do you have a preference? I think keeping `options: Document = {}` and `if (Object.keys(options).length !== 0) ` is slightly better.

Apologies if I'm missing something-- this is the first typescript PR I've written :)

packages/shell-api/src/stream-processor.ts
mongodb-matthew-normyle: Thanks, done.

packages/shell-api/src/stream-processor.ts
```diff
@@ -57,6 +58,48 @@ export default class StreamProcessor extends ShellApiWithMongoClass {
...
   }
 
+  /**
+   * modify is used to modify a stream processor definition, like below:
+   *  Change the pipeline:
+   *    sp.name.modify(newPipeline)
+   *  Keep the same pipeline, change other options:
+   *   sp.name.modify({resumeFromCheckpoint: false})
+   */
+  async modify(pipelineOrOptions: Document[] | Document): Promise<Document>;
```
addaleax, line 68:
```suggestion
  async modify(pipelineOrOptions: Document[] | Document): Promise<Document>;
  async modify(pipeline: Document[], options: Document): Promise<Document>;
```

packages/shell-api/src/stream-processor.ts
addaleax: Yeah, the `options: Document = {}` above would become `options?: Document` in this case 



> I think keeping `options: Document = {}` and `if (Object.keys(options).length !== 0) ` is slightly better.

I'd say checking the # of keys of an argument to check whether it was passed or not is definitely a bit unidiomatic. For example, it will just throw an unexpected exception if somebody passes `null` as an option objects.

mongodb-matthew-normyle: Ah yes that is much better, thanks. I didn't realize that options?: Document would work with the spread operator in `{...options, pipeline: pipelineOrOptions }`.

addaleax (approved):

addaleax:
@mongodb-matthew-normyle The outstanding CI failures are known flaky/failing tests, so this would be good to merge if you're read 👍

mongodb-matthew-normyle:
> @mongodb-matthew-normyle The outstanding CI failures are known flaky/failing tests, so this would be good to merge if you're read 👍

Hi @addaleax, I'm ready to merge :) It appears I don't have write access to merge my PR. Are you able to hit merge for me? Thanks in advance.

PR #2247, opened 2024-11-01
chore(deps): bump devtools-shared packages, add regression test for MONGOSH-1878

addaleax:


packages/e2e-tests/test/e2e-tls.spec.ts
```diff
@@ -304,6 +314,73 @@ describe('e2e TLS', function () {
...
+          ],
+        });
+        const result = await shell.waitForPromptOrExit();
+        expect(result.state).to.equal('prompt');
+      });
+
+      it('works with matching CA (system certs)', async function () {
+        if (process.platform !== 'linux') {
+          return this.skip();
+        }
```
nirinchev, line 357: Are we skipping this test on other platforms because it's more annoying to orchestrate or some other reason?

addaleax: It's because Linux is the only platform where system certificates are loaded from the disk, not through an OS API, so yeah, because it's much harder to "mock" another system CA store on those

nirinchev (approved):

gagik (approved):

PR #2254, opened 2024-11-06
feat: add support for fetching a greeting CTA as part of the update flow MONGOSH-1897

nirinchev:
This adds support for fetching a greeting CTA via the versions json. The proposed design adds support for reading a `cta` field in the version objects:

```diff
 {
   "version": [
     {
       "version": "1.2.3",
       "downloads": [ ... ],
+      "cta": {
+        "chunks": [
+          { "text": "Fill out our survey ", "style": "bold" }
+          { "text": "https://mongodb.com/survey", "style": "mongosh:uri" }
+        ]
+      }
     }
   ]
 }
```

These are optional and nothing will be displayed if they're not present. If they are present and the version matches the current version, we'll format the chunks with their corresponding styles and display the message as part of the greeting.

Additionally, we add the following to the `update-metadata.json`:

```diff
 {
     "updateURL": "https://downloads.mongodb.com/compass/mongosh.json",
     "lastChecked": 1730881696340,
     "etag": "\"7f84f631518c4872e3e469c765d1eb03\"",
     "latestKnownMongoshVersion": "1.2.3",
+    "cta": {
+        "1.2.3-alpha.1": {
+           "chunks": [
+                { "text": "This is an alpha version! Use with caution!", "style": "bold" }
+            ]
+        },
+        "1.2.3": {
+           "chunks": [
+                { "text": "This is a stable version and very reliable!", "style": "bold" }
+            ]
+        },
+    }
 }
```

The greeting CTA for the current version based on package.json will be picked, but locally we're saving both the current and the latest version's CTAs to make it readily available once the user upgrades.


packages/build/src/index.ts
```diff
@@ -38,29 +40,46 @@ if (require.main === module) {
...
-        .filter(Boolean)[0];
-      if (cliBuildVariant) {
-        config.packageVariant = cliBuildVariant[1] as PackageVariant;
-        validatePackageVariant(config.packageVariant);
-      }
+        ctaConfig.isDryRun ||= process.argv.includes('--dry-run');
 
-      config.isDryRun ||= process.argv.includes('--dry-run');
+        await updateJsonFeedCTA(ctaConfig);
+        break;
```
addaleax, line 60: I think this is the only step in this approach that I don't fully follow – is the idea that we'd locally run the `update-cta` command? How would we get the credentials to access the relevant S3 bucket?

We're definitely moving towards a world in which less and less access will be given to individual developers' machines, so I don't think this is something we could easily do, and we may just need to fall back to a variant in which we update the CTA entries along with the version feed during releases only – I think that would be okay though, we're generally free to do mongosh releases whenever necessary.

(In the future, we'll need to make sure that even this kind of flow where we update an existing file in S3 as part of the releases is supported – see e.g. https://jira.mongodb.org/browse/WRITING-16730)

nirinchev: I was hoping we could decouple releases from CTA updates. I had the following in mind - let me know how it sounds:
1. Updates are published when `cta.conf.js` is updated. This will have to go through the normal PR flow of accepting changes.
2. When the change is merged into `main`, a github action triggers that has access to the S3 credentials through environment secrets and runs the `update-cta` command.
3. The secrets are stored in an environment - let's say "production" - that will have protection rules configured, which means the deployment will be paused until the protection rules are satisfied (e.g. approval required by specified people/wait specified number of hours).

addaleax: > a github action triggers that has access to the S3 credentials through environment secrets and runs the update-cta command.

We can do that but I think we'd be going for Evergreen rather than GHA here? That's the only platform I'd really expect to be available to access release infrastructure in the long run. I think we'd encounter some resistance if we wanted to put downloads.mongodb.com secrets into another CI environment

nirinchev: We could sort it out - my personal preference is towards GHA and we already have a handful of products being signed and released with secrets stored in GHA, as well as all our code being hosted by github. That being said, if there's a strong desire to only do that on evergreen, we could work with those limitations - the important part here is that the flow of publishing CTA updates happens on CI and in a somewhat automated and reviewable fashion.

nirinchev:
This is ready for review - I tested it with a custom S3 bucket and feel fairly confident that it works. The only remaining task is provisioning credentials for the downloads S3 bucket, adding them as a secret to GHA, and testing the workflow for updating the cta. Waiting on devprod there and I'll make sure to test it out before merging.

configs/eslint-config-mongosh/index.js
```diff
@@ -93,6 +93,7 @@ module.exports = {
         ...common.testRules,
         ...extraJSRules,
         ...extraTypescriptRules,
+        '@typescript-eslint/no-non-null-assertion': 'off',
```
addaleax, line 96: Just going to mention that I think that this is on almost everywhere in our repos and that hasn't been a major issue for the most part

packages/cli-repl/src/update-notification-manager.ts
```diff
@@ -123,17 +164,36 @@ export class UpdateNotificationManager {
...
       ?.sort(semver.rcompare)?.[0];
 
+    this.currentVersionGreetingCTA =
+      jsonContents?.versions?.filter((v) => v.version === currentVersion)?.[0]
+        ?.cta ?? jsonContents?.cta;
+
+    const latestKnownVersionCTA =
+      jsonContents?.versions?.filter(
+        (v) => v.version === this.latestKnownMongoshVersion
+      )?.[0]?.cta ?? jsonContents?.cta;
```
addaleax, line 180:
```suggestion
      jsonContents?.versions?.find(
        (v) => v.version === this.latestKnownMongoshVersion
      )?.cta ?? jsonContents?.cta;
```

packages/cli-repl/src/update-notification-manager.ts
```diff
@@ -123,17 +164,36 @@ export class UpdateNotificationManager {
...
-    const jsonContents = (await response.json()) as { versions?: any[] };
+    const jsonContents = (await response.json()) as MongoshVersionsContents;
     this.latestKnownMongoshVersion = jsonContents?.versions
-      ?.map((v: any) => v.version as string)
+      ?.map((v) => v.version)
       ?.filter((v) => !semver.prerelease(v))
       ?.sort(semver.rcompare)?.[0];
 
+    this.currentVersionGreetingCTA =
+      jsonContents?.versions?.filter((v) => v.version === currentVersion)?.[0]
```
addaleax, line 174:
```suggestion
      jsonContents?.versions?.find((v) => v.version === currentVersion)
```

addaleax (approved):

packages/build/package.json
```diff
@@ -69,6 +70,7 @@
     "@mongodb-js/monorepo-tools": "^1.1.16",
     "@mongodb-js/signing-utils": "^0.3.7",
     "@octokit/rest": "^17.9.0",
+    "ajv": "^8.17.1",
```
addaleax, line 73: This reminded me to finally do this: https://github.com/mongodb-js/mongosh/pull/2360

Honestly I'd consider it for this one too, zod really just has a very neat integration with TS, and there are converters from/to JSON schemas available

packages/build/src/config/config.ts
```diff
@@ -7,6 +8,18 @@ interface ManPageConfig {
   fileName: string;
 }
 
+// TODO: this is duplicated in update-notification-manager.ts
+export interface GreetingCTADetails {
+  chunks: {
+    text: string;
+    style?: string; // TODO: this is actually clr.ts/StyleDefinition
```
addaleax, line 15: Is addressing the open TODOs planned to be part of this PR?

.github/workflows/update-cta.yml
```diff
@@ -0,0 +1,46 @@
...
+          cache: "npm"
+
+      - name: Install Dependencies and Compile
+        run: |
+          npm ci
+          npm run compile
+
+      - name: Update greeting CTA
+        run: |
+          npm run update-cta ${{ github.event.inputs.dry-run && '-- --dry-run' || '' }}
```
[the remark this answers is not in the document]
addaleax: ^ should be addressed, right?

packages/build/src/config/config.ts
```diff
@@ -7,6 +8,18 @@ interface ManPageConfig {
...
+  }[];
+}
+
+export type CTAConfig = {
+  [version: string | '*']: GreetingCTADetails;
```
gagik, line 20:
```suggestion
  [version: `${number}.${number}.${number}` | '*']: GreetingCTADetails;
```
(if that makes sense)

gagik (approved):

packages/build/src/config/config.ts
nirinchev: Good point - I'm not sure honestly. I can change the TODO to be merely a comment as I feel it's going to be significant undertaking to extract the common type and reference it in the cli-repl and the build projects. The actual permitted values are already enforced in json schema, so treating them as strings here is probably not the end of the world.

packages/build/src/config/config.ts
nirinchev: The version can be any semver string, which includes prereleases. The permitted format is enforced by a regex in the json schema definition, so leaving it as a string is probably easiest here.

packages/build/package.json
nirinchev: Is your suggestion to port the current json schema to zod schema and use that for validation or to retain the json schema and use json-schema-to-zod to convert it into a zod schema and then use that for validations?

.github/workflows/update-cta.yml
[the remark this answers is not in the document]
nirinchev: Yep, will do.

packages/build/package.json
addaleax: The former – I'd at least consider it as an option here, just for a bit more consistency. Totally understand if you want to go with "this is already in place now and there's no strong reason to change it", too

nirinchev: I do like having a json schema since vscode supports that nicely and gives you autocomplete and warnings for violations. I'm not fundamentally opposed to incorporating zod somewhere in there - I'll try and think about how to seamlessly do it.

nirinchev: After some playing with this, I'm inclined to keep things as they are. I explored using `json-schema-to-zod` to conver the existing json schema to a zod schema and validate it, but that package is intended to be used to generate static schemas and not at runtime (see [note here](https://github.com/StefanTerdell/json-schema-to-zod?tab=readme-ov-file#use-at-runtime)). Alternatively, we could define the schema using the zod syntax and generate a json schema from it, but I couldn't find an elegant way to do it to make sure both are always in sync other than adding additional tooling like a pre-commit hook or a step to the cron-tasks.yml. Considering we want to keep the json schema for the vscode intellisense and validation integration, I believe both approaches fell short of what we currently have.

addaleax (approved):

nirinchev:
Evergreen failures point to issues that exist on main, so going to go ahead and merge this.

PR #2255, opened 2024-11-08
chore: drop node 16 from CI matrix MONGOSH-1898

nirinchev:
Drops Node 16 from the test matrix.


addaleax (approved):

addaleax:
> cli-repl/package.json uses node >= 16.15.0 - should we bump that as well?

Yeah, if we don't test it, it's probably best to bump it 👍

packages/e2e-tests/test/e2e.spec.ts
```diff
@@ -1059,6 +1059,13 @@ describe('e2e', function () {
     });
 
     it('reads and runs the vscode extension example playground', async function () {
+      if (
+        process.platform === 'win32' &&
+        process.versions.node.startsWith('22')
+      ) {
+        // This test fails on Windows with node 22 likely due to https://github.com/nodejs/node/issues/51766
+        return this.skip();
+      }
```
addaleax, line 1068: Hm, this might be something we want to track in our own JIRA? Not sure if we should consider this a blocker to adopting Node.js 22 for our official builds, but we should probably at least discuss it

nirinchev: Agreed - was hoping to get this to a stable state and then I'll go back and compile a list of the hacks/questionable changes that we should double check on prior to merging.

nirinchev:
@addaleax I've split this PR and only left the changes related to removing Node 16 from the testing matrix. There were still some issues with the Node 22 tests, so I'll investigate those separately.

addaleax (approved):

PR #2258, opened 2024-11-12
fix: suppress experimental warning for node 23 MONGOSH-1895

nirinchev:
- [x] Suppresses experimental warnings on node 23
- [x] Updates smoke test to catch those from stderr
- [x] Adds smoke tests on gha with node 20/22/23
- [x] Adds a smoke test on gha to run `brew install mongosh` and then smoke test it
- [ ] Posts a slack message when the homebrew test fails (needs webhook url - requested here: https://help-it.mongodb.com/hc/en-us/requests/492654).


.github/workflows/homebrew.yml
```diff
@@ -0,0 +1,39 @@
+name: "Smoke Test Homebrew install"
+
+on:
+  schedule:
+    - cron: "0 0 * * *"
+  pull_request: # TODO: remove pull_request trigger
```
nirinchev, line 6: Waiting on IT to create the webhook, after which I'll clean up the TODOs in this file.

packages/cli-repl/src/smoke-tests.ts
```diff
@@ -397,7 +397,7 @@ async function runSmokeTest({
   // eslint-disable-next-line
   const { spawn } = require('child_process') as typeof import('child_process');
   const proc = spawn(executable, [...args], {
-    stdio: ['pipe', 'pipe', includeStderr ? 'pipe' : 'inherit'],
+    stdio: 'pipe',
```
nirinchev, line 400: Changing to `pipe` indiscriminately so that we can then validate there's no warnings in stderr.

packages/cli-repl/src/smoke-tests.ts
```diff
@@ -420,13 +420,14 @@ async function runSmokeTest({
...
     stdout,
-    stderr,
+    stderr: includeStderr ? stderr : '',
     executable,
     actualExitCode,
     args: args.map((arg) => redactURICredentials(arg)),
   };
   try {
     assert.match(includeStderr ? `${stdout}\n${stderr}` : stdout, output);
+    assert.doesNotMatch(stderr, /ExperimentalWarning/);
```
nirinchev, line 430: I'm tempted to change this to match more broadly just `Warning`, but I'm not sure if that might result in false positives.

.github/workflows/homebrew.yml
```diff
@@ -0,0 +1,39 @@
...
+    strategy:
+      matrix:
+        runner: [macos-13, macos-14, macos-15]
+      fail-fast: false
+    steps:
+      - run: brew install mongosh
+      - run: mongosh --smokeTests
+      - name: Report failure
+        # TODO: replace with failure()
+        if: ${{ success() }}
```
addaleax, line 21: Open TODO?

package.json
```diff
@@ -167,4 +169,4 @@
   "overrides": {
     "cookie": "^0.7.2"
   }
-}
+}
```
addaleax, line 172: Missing newline at end of file?

.github/workflows/smoke-tests.yml
```diff
@@ -0,0 +1,36 @@
...
+          cache: "npm"
+
+      - name: Install npm@10
+        run: npm install -g npm@10
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Run smoke tests
+        run: npm run test-smoke
```
addaleax, line 36: Let's do something like

```suggestion
        run: npx mongodb-runner exec -- sh -c 'env MONGOSH_SMOKE_TEST_SERVER=$MONGODB_URI npm run test-smoke'
```

so that we run the smoke tests that require a MongoDB server as well

addaleax (approved):
Just a couple small comments, nothing big!

.github/workflows/homebrew.yml
nirinchev: Yeah, I'll fix that before we merge - just wanted to test it first once I get the slack webhook from IT.

package.json
nirinchev: Ugh, will need to take a look at my vscode settings - I thought it was prettier that removes it 😬

nirinchev:
I've removed the homebrew smoke test workflow and will re-add it in a separate PR. Addressed comments as well, so I'm going to go ahead and merge this so we can reap some of the benefits, particularly around reduced CI times due to the chromium downloads (in my tests, shaves 3-4 minutes from each `npm ci`).

PR #2274, opened 2024-11-22
feat(shell-api): add options in stream processor start, stop, and drop MONGOSH-1920

mongodb-matthew-normyle:
* Add an options parameter to the stream processor start, stop, and drop methods. We already do this for sample and stats methods.
* This allows Atlas Stream Processing to support features like STREAMS-872.

### DIFF
--- a/packages/shell-api/src/stream-processor.ts
+++ b/packages/shell-api/src/stream-processor.ts
@@ -26,27 +26,30 @@ export default class StreamProcessor extends ShellApiWithMongoClass {
   }
 
   @returnsPromise
-  async start() {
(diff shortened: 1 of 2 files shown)


mongodb-matthew-normyle:
> some versions of mongosh, options will silently be ignored and, in newer versions, will be applied is probably going to trip up a couple users but I also don't really

Noted. We will clarify in streams documentation that various new features (i.e. support for `.start(resumeFromCheckpoint=false)`) require the latest version of mongosh. This should be fine given our customer base is small-ish (but growing 😄 )

PR #2275, opened 2024-11-22
chore: push auto-generated files directly instead of using PRs

nirinchev:
Changes the workflow to update auto-generated files to directly push to main. This should eliminate the extra PR we open after every merge. The devtoolsbot user has been exempted from the PR requirement, so should be able to push directly to main. While this is not ideal, it's not less secure than what we previously had where the github actions bot would open a PR and then immediately merge it with no checks/approvals required.

If this works well, I'll copy it over to the other repos we manage.

### DIFF
--- a/.github/workflows/update-node-js.yaml
+++ b/.github/workflows/update-node-js.yaml
@@ -19,7 +19,6 @@ jobs:
           # this is important so git log can pick up on
           # the whole history to generate the list of AUTHORS
           fetch-depth: "0"
-          token: ${{ secrets.SVC_DEVTOOLSBOT_TOKEN }}
 
       - name: Setup git
         run: |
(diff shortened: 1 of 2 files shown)


addaleax (approved):
No strong feelings here, but I admit that I kind of like the idea of getting a notification for every change, even if it's automatically generated. But I can obviously also see how it's a lot of noise 🙂

.github/workflows/update-node-js.yaml
```diff
@@ -38,16 +38,14 @@ jobs:
...
-          git commit --no-allow-empty -m "chore: update node.js version" || true
 
       - name: Create Pull Request
         id: cpr
-        uses: peter-evans/create-pull-request@v6
+        uses: peter-evans/create-pull-request@5e914681df9dc83aa4e4905692ca88beb2f9e91f # 7.0.5
         with:
-          token: ${{ secrets.GITHUB_TOKEN }}
-          commit-message: 'chore: update node.js'
+          token: ${{ secrets.SVC_DEVTOOLSBOT_TOKEN }}
```
nirinchev, line 46: This is a drive-by - I noticed the update-node-js workflow didn't trigger GHA workflow since it created a PR with the workflow token.

nirinchev:
Let's see how it goes - if you do find you're missing out on things, we could setup a slack/email integration and send out notifications for these autogenerated commits.

PR #2277, opened 2024-11-22
chore(ci): login to docker

lerouxb:
This should help with rate limiting.

Hopefully I got all the places that need the env vars. If not then `docker-credential-from-env` will hopefully tell us.

### DIFF
--- a/scripts/docker/build.sh
+++ b/scripts/docker/build.sh
@@ -7,7 +7,7 @@ cd "$(dirname "$0")"
 [ -x node_modules/mongodb-crypt-library-version ] || ((cd ../.. && npm ci) && cp -r ../../node_modules node_modules)
 
 # we don't have credentials for registry.suse.com and docker now requires them due to our config
-if [[ "$1" == suse* ]]
+if [[ "$1" == suse* ]]; then
   unset DOCKER_CONFIG
 fi
 


addaleax (approved):

.evergreen/docker-config/config.json
```diff
@@ -0,0 +1,6 @@
+{
+	"auths": {
+		"https://index.docker.io/v1/": {}
+	},
+	"credsStore": "from-env"
```
lerouxb, line 5: A problem is that by specifying [credsStore](https://docs.docker.com/reference/cli/docker/login/#credential-stores), the docker-credential-from-env script gets used for all stores. And at that point I can't tell if there's a way for the script to signal that a specific registry should be used without authentication.

What then happens is that we don't have credentials for registry.suse.com and then our script errors.

There is an alternative to credsStore which is [credHelpers](https://docs.docker.com/reference/cli/docker/login/#credential-helpers) where you can specify a credential helper per registry domain. I tried it, but no matter what I used as the registry domain for docker hub I couldn't get it to use our script, so everything just worked unauthenticated. I looked online and can't find any examples where people specify the default docker hub registry domain in there so I'm beginning to think that it doesn't work for that case.

Unless I misread the docs or had some silly typo throughout all my tests? Not sure.

So what I'm doing now is to just specify credsStore which at least works in that it executes our script and then I just remove the DOCKER_CONFIG var for the suse cases. Which is an ugly hack and will probably stop working if we mix suse and regular docker images together in one script.

PR #2280, opened 2024-11-26
chore(ci): use a github app for the generated files credentials MONGOSH-1927

nirinchev:


.github/workflows/cron-tasks.yml
```diff
@@ -7,6 +7,7 @@ on:
   push:
     branches:
       - main
+  workflow_dispatch: # This should be removed once the workflow is tested and working
```
addaleax, line 10: Leaving a comment here just as a reminder :)

addaleax (approved):

PR #2283, opened 2024-11-27
chore(e2e-tests): add workload OIDC test MONGOSH-1832

addaleax:
Now that the driver has support for the Kubernetes OIDC workload integration, it's quite easy to add a test for OIDC workload usage with mongosh.


gagik (approved):
Not very familiar with this side of things but LGTM as far as I can see 🙂

PR #2284, opened 2024-11-28
fix(browser-repl): keep operation in progress COMPASS-8576

lerouxb:
Required for https://github.com/mongodb-js/compass/pull/6538

Changes:

* output, history and isOperationInProgress are now all controlled props. So initialOutput and initialHistory was removed.
* I took the chance to rewrite Shell to be a function component. So I also converted the tests to use testing-library.


packages/browser-repl/package.json
```diff
@@ -35,7 +35,8 @@
     "depcheck": "depcheck",
     "compile": "tsc -p tsconfig.json",
     "prettier": "prettier",
-    "reformat": "npm run prettier -- --write . && npm run eslint --fix"
+    "reformat": "npm run prettier -- --write . && npm run eslint --fix",
+    "sync-to-compass": "node scripts/sync-to-compass.js"
```
lerouxb, line 39: Useful in combination with [this](https://github.com/mongodb-js/compass/pull/6538/files#r1862161800). Based on https://github.com/mongodb-js/compass/blob/main/packages/compass-web/scripts/sync-dist-to-mms.js

This way you can save a file in browser-repl in mongosh and it hot-reloads in compass.

packages/browser-repl/src/components/shell-input.tsx
```diff
@@ -110,10 +110,12 @@ export class ShellInput extends Component<ShellInputProps, ShellInputState> {
 
   private onEnter = async (): Promise<void> => {
     if (this.props.onInput) {
-      await this.props.onInput(this.state.currentValue);
+      const value = this.state.currentValue;
+      // clear the value before evaluating the input because it could take a
+      // long time
+      this.setState({ currentValue: '' });
```
lerouxb, line 116: drive-by fix

packages/browser-repl/src/components/shell.tsx
```diff
@@ -127,13 +127,17 @@ interface ShellProps {
    */
   initialHistory: readonly string[];
 
+  /**
+   * A boolean so we can keep the isOperationInProgress state between sessions.
```
addaleax, line 131: If the expectation is that `onOperationStarted` and `onOperationEnd` are used to keep track of this state outside of the shell component itself, then I'd mention that in the doc comment here.

I'm wondering, though, if it doesn't make sense to just pass this prop as the initial value of `operationInProgress` and keep tracking this in the state of the shell itself, semantically that definitely seems like the right thing to do.

If you decided against that, you'll also want to update `src/sandbox.tsx` to adopt the new way of keeping track of this state.

lerouxb: > and keep tracking this in the state of the shell itself

I'm not following? Where do you mean in the shell? The bug here is that we lose that variable when the shell re-renders, so I'm making it a controlled prop.

lerouxb: (I should definitely add some comments to document this)

addaleax: Right, I'm saying, if we expect the parent of this component (= `Shell`) to keep track of the state, then that should be documented here and we should adjust the sandbox example as well

But the real underlying issue seems to be that all shell state gets lost in this case, that's why we have the `initial...` properties on the component, and I'm wondering why we treat this particular property differently (i.e. don't also provide an `initialOperationInProgress` and use that to set the default value of the flag in the Shell component state)

lerouxb: My thinking is that if we keep track of initialOperationInProgress to exactly match the internal operationInProgress state at all times because at any point the component could be replaced, then why have both? Same actually for the other props and state this component has that acts like that. Tempting to just change them all and turn this into a function component while at it.

lerouxb: I changed inputText, output, history and isOperationInProgress into controlled props and updated both the README and sandbox accordingly. While at it I just changed the whole thing into a function component.

I have to basically rewrite the tests now because they were still using enzyme so I just removed them for the moment. Will replace in a separate commit as I think it is easier that way around at this stage.

packages/browser-repl/scripts/sync-to-compass.js
```diff
@@ -0,0 +1,73 @@
+/* eslint-disable no-console */
+'use strict';
+const fs = require('fs');
+const path = require('path');
+const child_process = require('child_process');
+const { debounce } = require('lodash');
```
kraenhansen, line 6: Any particular reason this isn't ESM?

kraenhansen: Ahhh - it was copied from the `sync-dist-to-mms.js` file.

lerouxb:
@kraenhansen I don't have a better explanation than https://github.com/mongodb-js/compass/blob/117c32f8ddd5d2bf4f8fceec5b324bf9866de66e/packages/compass-web/scripts/sync-dist-to-mms.js#L47-L52

I don't think any of us have managed to get that work. This is by far the closest to "it just works, instantly and reliably" that I've managed.

packages/browser-repl/src/components/shell.tsx
```diff
@@ -153,94 +163,123 @@ const normalizeInitialEvaluate = (initialEvaluate: string | string[]) => {
...
+  const editorChanged = useCallback(
+    (editor: EditorRef | null) => {
+      setEditor(editor);
+      onEditorChanged?.(editor);
+    },
+    [onEditorChanged]
+  );
+
+  const listener = useMemo(() => {
+    const evaluationListener: RuntimeEvaluationListener = {
```
addaleax, line 224: nit: could probably just do this if we're going for type safety

```suggestion
  const listener = useMemo<RuntimeEvaluationListener>(() => {
    return {
```

packages/browser-repl/src/components/shell.tsx
```diff
@@ -153,94 +163,123 @@ const normalizeInitialEvaluate = (initialEvaluate: string | string[]) => {
...
+
+        setPasswordPrompt(question);
+
+        return ret;
+      },
+      onClearCommand: (): void => {
+        onOutputChanged?.([]);
+      },
+    };
+    return evaluationListener;
```
addaleax, line 274: (would go with the suggestion above)

```suggestion
```

packages/browser-repl/src/components/shell.tsx
```diff
@@ -253,235 +292,181 @@ export class _Shell extends Component<ShellProps, ShellState> {
...
+      return outputLine;
+    },
+    [listener, onOperationEnd, onOperationStarted, runtime, updateShellPrompt]
+  );
 
-    const outputLine = await this.evaluate(code);
+  const onInput = useCallback(
+    async (code: string) => {
+      const newOutput = (output ?? []).slice();
+      const newHistory = (history ?? []).slice();
```
addaleax, line 343: nit: usually copying is a bit easier to read with object/array spread imo:

```suggestion
      const newOutput = [...output ?? []];
      const newHistory = [...history ?? []];
```

packages/browser-repl/src/components/shell-input.tsx
lerouxb: I suppose. A shell-input without onInput isn't going to be useful, but we might as well keep it sane.

packages/browser-repl/scripts/sync-to-compass.js
Anemy: Having this typescript would make me happy. Could do it another time though.

lerouxb: I copy/pasted it and it is just a developer productivity script. We can improve it whenever.

packages/browser-repl/README.md
```diff
@@ -31,14 +31,20 @@ const runtime = new IframeRuntime(serviceProvider);
...
+- `onEditorChanged?: (editor: EditorRef | null) => void`: A function called each time the editor ref changes. Can be used to call editor methods.
+- `onOperationStarted?: () => void`: A function called when an operation has begun.
+- `onOperationEnd?: () => void`: A function called when an operation has completed (both error and success).
 - `redactInfo?: boolean`: If set, the shell will omit or redact entries containing sensitive info from history. Defaults to `false`.
 - `maxOutputLength?: number`: The maxiumum number of lines to keep in the output. Defaults to `1000`.
 - `maxHistoryLength?: number`: The maxiumum number of lines to keep in the history. Defaults to `1000`.
-- `initialOutput?: readonly ShellOutputEntry[]`: An array of entries to be displayed in the output area. Can be used to restore the output between sessions, or to setup a greeting message. **Note**: new entries will not be appended to the array.
-- `initialHistory?: readonly string[]`: An array of history entries to prepopulate the history.
+- `initialEvaluate?: string|string[]`: A set of input strings to evaluate right after shell is mounted.
+- `inputText?: string`: Initial text for the input field.
```
Anemy, line 43: How do you feel about prefixing this with initial, `initialInputText`?

packages/browser-repl/README.md
```diff
@@ -31,14 +31,20 @@ const runtime = new IframeRuntime(serviceProvider);
...
+- `output?: ShellOutputEntry[]`: An array of entries to be displayed in the output area. Can be used to restore the output between sessions, or to setup a greeting message. **Note**: new entries will not be appended to the array.
```
Anemy, line 44: This sounds like another `initial` value. Maybe we follow that `initial` prefix naming convention? Same for `history` and `isOperationInProgress`?

packages/browser-repl/src/components/shell.tsx
```diff
@@ -1,4 +1,6 @@
-import React, { Component } from 'react';
+/* eslint-disable jsx-a11y/no-static-element-interactions */
+/* eslint-disable jsx-a11y/click-events-have-key-events */
```
Anemy, line 2: nit: can we turn these off for the specific lines involved? That way we can avoid adding more instances of this by accident.

Anemy (commented):
Some small things I noticed initially, trying it out now, and taking a look at the usage in Compass.
Nice!

packages/browser-repl/README.md
lerouxb: The way I see it it isn't an initial value, it is the controlled value. Same as a leafygreen text input's `value` or a modal's `open`. But I also don't feel too strongly about it.

I should update the readme line, though.

packages/browser-repl/README.md
lerouxb: Same comment as above - these are controlled values, not initial.

lerouxb: Maybe I don't understand react conventions, though?

packages/browser-repl/README.md
lerouxb: OK I'm going with initialText since that's what we pass down to the input.

packages/browser-repl/src/components/shell.tsx
```diff
@@ -253,235 +287,185 @@ export class _Shell extends Component<ShellProps, ShellState> {
...
+      capLengthEnd(newOutput, maxOutputLength);
+      onOutputChanged?.(newOutput);
+
+      // update history
+      newHistory.unshift(code);
+      changeHistory(
+        newHistory,
+        redactInfo ? 'redact-sensitive-data' : 'keep-sensitive-data'
+      );
+      capLengthStart(newHistory, maxHistoryLength);
```
gribnoysup, line 372: Not sure if this makes any difference, but I wonder if we'd want to cap before we call the change above? Feels off that this method is getting a different history compared to what will be in the ui state

packages/browser-repl/src/components/shell.tsx
```diff
@@ -253,235 +287,185 @@ export class _Shell extends Component<ShellProps, ShellState> {
...
+
+    void updateShellPrompt().then(async () => {
+      if (initialEvaluate) {
+        const evalLines = normalizeInitialEvaluate(initialEvaluate);
+        for (const input of evalLines) {
+          await onInput(input);
+        }
+      }
+    });
+  }, [initialEvaluate, onInput, scrollToBottom, updateShellPrompt, output]);
```
gribnoysup, line 406: As you're switching to an effect and as `output` is controlled, this effect will be running every time the output is changed (so every user evaluation). To get a "true" on mount effect, this should have an empty dependency list, usually a way to do this is to ref everything that's being used here

packages/browser-repl/src/components/shell.tsx
```diff
@@ -153,94 +159,122 @@ const normalizeInitialEvaluate = (initialEvaluate: string | string[]) => {
...
-  componentDidUpdate(): void {
-    this.scrollToBottom();
-  }
+const capLengthStart = (elements: unknown[], maxLength: number) => {
+  elements.splice(maxLength);
+};
 
-  private evaluate = async (code: string): Promise<ShellOutputEntry> => {
-    let outputLine: ShellOutputEntry;
+// eslint-disable-next-line react/display-name
```
gribnoysup, line 174: I think this shouldn't be needed anymore

packages/browser-repl/src/components/shell.tsx
lerouxb: `output` shouldn't even used here here. Don't know what I was doing there. I put initialEvaluate in a ref and use that. The others are all callbacks that shouldn't be changing I think - manual testing shows that it isn't. Is this good enough?

packages/browser-repl/src/components/shell.tsx
```diff
@@ -153,94 +159,122 @@ const normalizeInitialEvaluate = (initialEvaluate: string | string[]) => {
...
-    } catch (error) {
-      outputLine = {
-        format: 'error',
-        value: error,
-      };
-    } finally {
-      await this.updateShellPrompt();
-      this.props.onOperationEnd();
-    }
+  const [editor, setEditor] = useState<EditorRef | null>(null);
```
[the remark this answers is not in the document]
lerouxb: I changed this now. Does have the problem where it isn't set in time for the useEffect in compass, though.

packages/browser-repl/src/components/shell.tsx
lerouxb: ok so there were more props being depended on indirectly.. 😢

packages/browser-repl/src/components/shell.tsx
```diff
@@ -153,94 +159,122 @@ const normalizeInitialEvaluate = (initialEvaluate: string | string[]) => {
...
+  >(() => noop);
+
+  const focusEditor = useCallback(() => {
+    editor?.focus();
+  }, [editor]);
+
+  const editorChanged = useCallback(
+    (editor: EditorRef | null) => {
+      setEditor(editor);
+      onEditorChanged?.(editor);
```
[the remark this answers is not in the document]
lerouxb: Should be good now?

packages/browser-repl/src/sandbox.tsx
```diff
@@ -156,18 +156,20 @@ const IframeRuntimeExample: React.FunctionComponent = () => {
   const [redactInfo, setRedactInfo] = useState(false);
   const [maxOutputLength, setMaxOutputLength] = useState(1000);
   const [maxHistoryLength, setMaxHistoryLength] = useState(1000);
-  const [initialInput, setInitialInput] = useState('');
   const [initialEvaluate, setInitialEvaluate] = useState<string[]>([]);
-  const [initialOutput] = useState<ShellOutputEntry[]>([
+
+  const [initialText, setInitialText] = useState('');
+  const [output, setOutput] = useState<ShellOutputEntry[]>([
```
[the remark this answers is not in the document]
Anemy: https://github.com/user-attachments/assets/9e885223-02ac-4fa0-b4e0-00678a73784a

Apologies if this is expected in the recent changes and is WIP.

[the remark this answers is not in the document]
lerouxb: OK so what happened there was I was trying to get the useEffect that does the initial work of updating the prompt and running the initial commands to not fire every time the output changes. But it depends on onInput and onInput has to use and update the output so IT returns on output. So output changes, there's a new onInput and then the useEffect runs again.

So I was trying to be clever and use a ref for the output inside onInput, but then every time onInput changes the output it was reusing the initial value.

So now I just have a boolean as state. Which seems like a hack. But isn't every solution to that a hack?

gribnoysup (approved):

packages/browser-repl/src/components/shell.tsx
```diff
@@ -253,235 +323,203 @@ export class _Shell extends Component<ShellProps, ShellState> {
...
+      return (runtime as WorkerRuntime).interrupt();
+    }
 
-    const output = this.addEntriesToOutput([inputLine]);
+    return Promise.resolve(false);
+  }, [isOperationInProgress, runtime]);
 
-    this.setState({ output });
-  }
+  const [isFirstRun, setIsFirstRun] = useState(true);
```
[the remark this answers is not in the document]
lerouxb: Like this? https://github.com/mongodb-js/mongosh/pull/2284/commits/4e66b190405a7c17fea7b28484ad1dc5bcbc968f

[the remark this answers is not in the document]
lerouxb: Don't know why I didn't consider just updating the ref. This stuff still isn't muscle memory for me - somehow I still don't do enough react often enough.

gribnoysup (approved):

PR #2286, opened 2024-12-03
fix(e2e-tests): increase timeouts for AWS tests MONGOSH-1924

addaleax:
This should hopefully address AWS auth test flakiness on s390x, where DNS queries can take multiple seconds each due to the host CI setup.

Example strace that indicates a 5 second delay for a DNS query: https://docs.google.com/document/d/1jWxKTZWma5nN5I3a9iTANnulzm_9GlLanEUbf9q9vh0/edit?tab=t.0

### DIFF
--- a/packages/e2e-tests/test/e2e-aws.spec.ts
+++ b/packages/e2e-tests/test/e2e-aws.spec.ts
@@ -83,7 +83,11 @@ function getConnectionString(username?: string, password?: string): string {
 }
 
 describe('e2e AWS AUTH', function () {
-  this.timeout(60_000); // AWS auth tests can take longer than the default timeout in CI
+  // AWS auth tests can take longer than the default timeout in CI
+  // DNS resolution for many hosts in particular can be time-intensive in some
+  // CI environments
+  this.timeout(80_000);
+  const initialWaitForPromptTimeoutOptions = { timeout: 60_000 };
   let expectedAssumedRole: string;
 
   before(function () {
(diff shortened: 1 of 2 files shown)


nirinchev (approved):
Is there something we can work on together with devprod to reduce DNS lookup times on these hosts? If you've already talked to them, might be good to add a comment in the code that summarizes the motivation/intricacies of this setup, as well as any steps we might want to take in the future. To be clear, this seems like a reasonable bandaid, but ideally, we'd want to improve the latency to avoid wasting CI resources.

addaleax:
@nirinchev Oh yeah, I was chatting with Thomas from the DevProd team who also helped with providing debug access to a host. I don't think there's a known root cause here – but they're eager to investigate this on their side.

nirinchev:
Cool, if they have a ticket tracking this, mind if you link to it in a code comment as it'll provide valuable context for future readers.

addaleax:
@nirinchev I don't think there's a ticket right now. We do have:

> We recently got honeycomb metrics functional on this platform, so hopefully stuff like that will start showing up in traces

so that's likely going to help the DevProd team with keeping an eye on things, and if we feel that this is really an issue that is significant enough to require special attention, we can always open a new ticket with them. But I also have to be honest and say that this problem doesn't seem bad enough that it _requires_ extra attention at this point – if the tests start timing out with the new 1 minute limit I'd be a lot more concerned.

nirinchev:
@addaleax fair enough - I agree it's sufficiently unimportant that creating a ticket is not something we need to do at this point.

PR #2300, opened 2024-12-12
fix(cli-repl): only ever run MongoshRepl.onExit once MONGOSH-1943

addaleax:
(Depends on #2299 as the first commit here and on #2298 logically – without #2298, this just fails consistently now instead of flakily)

Previously, a task in our CI was flaky because it was running `exit\n` as literal input to the shell, expecting it to reliably make the shell exit.

`exit` as a command leads to the `MongoshRepl.onExit` function called, which then calls a) `.close()` and, indirectly, `repl.close()` on the Node.js REPL instance as well b) the `CliRepl.exit` function (and through that, indirectly, `process.exit()`).

However, closing the Node.js REPL instance makes it flush the history file to disk as a typically fast but asynchronous operation, and, after that, emit `exit` on the REPL; that in turn called `MongoshRepl.onExit` again (i.e. calling `MongoshRepl.onExit` leads to another call to the same function, just asynchronously with a short delay).

The flaky test in CI failed because of a combination of issues, including the fact that `process.exit()` did not necessarily actually exit the process because of coverage tooling interfering with it (which in turn happened because it was running with an altered working directory pointing at a temporary directory where `nyc` had not created a subdirectory for storing coverage output).

In typical operations, the REPL would flush history quickly, and calling `.onExit()` again would allow the process to exit normally (by ways of calling `CliRepl.exit` and `process.exit()` a second time). However, in the flaky failure case, the history flushing operation would take long enough that the original `process.exit()` call (which failed due to nyc) would set the internal `process._exiting` flag before throwing, which in turn prevents `process.nextTick()` from scheduling callbacks, and which then ultimately prevented the REPL from cleaning itself up, leading to the process timing out instead of exiting.

This commit introduces safeguards against calling `.onExit()` twice, always returning the same (never-resolving) `Promise` instance, and explicitly keeps track of whether the REPL was closed or not.

### DIFF
--- a/packages/cli-repl/src/mongosh-repl.ts
+++ b/packages/cli-repl/src/mongosh-repl.ts
@@ -127,6 +127,7 @@ type Mutable<T> = {
 class MongoshNodeRepl implements EvaluationListener {
   _runtimeState: MongoshRuntimeState | null;
   closeTrace?: string;
+  exitPromise?: Promise<never>;
   input: Readable;
   lineByLineInput: LineByLineInput;
   output: Writable;
@@ -545,6 +546,7 @@ class MongoshNodeRepl implements EvaluationListener {
     });
 
     repl.on('exit', () => {
+      if (this._runtimeState) this._runtimeState.repl = null;
       this.onExit().catch(() => {
         /* ... */
       });
@@ -1051,7 +1053,10 @@ class MongoshNodeRepl implements EvaluationListener {
     if (rs) {
       this._runtimeState = null;
       this.closeTrace = new Error().stack;
-      rs.repl?.close();
+      if (rs.repl) {
+        rs.repl.close();
+        await once(rs.repl, 'exit');
+      }
       await rs.instanceState.close(true);
       await new Promise((resolve) => this.output.write('', resolve));
     }
@@ -1063,8 +1068,10 @@ class MongoshNodeRepl implements EvaluationListener {
    * @param exitCode The user-specified exit code, if any.
    */
   async onExit(exitCode?: number): Promise<never> {
-    await this.close();
-    return this.ioProvider.exit(exitCode);
+    return (this.exitPromise ??= (async () => {
+      await this.close();
+      return this.ioProvider.exit(exitCode);
+    })());
   }
 
   /**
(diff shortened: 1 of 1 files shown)


packages/cli-repl/src/mongosh-repl.ts
```diff
@@ -1028,7 +1031,13 @@ class MongoshNodeRepl implements EvaluationListener {
    */
   runtimeState(): MongoshRuntimeState {
     if (this._runtimeState === null) {
-      throw new MongoshInternalError('Mongosh not initialized yet');
+      // This error can be really hard to debug, so we always attach stack traces
+      // from both when .close() was called and when
+      throw new MongoshInternalError(
+        `mongosh not initialized yet\nCurrentTrace: ${
+          new Error().stack
+        }\nClose trace: ${this.closeTrace}\n`
```
gagik, line 1039: is there a reason we're ending with a newline?

packages/cli-repl/src/mongosh-repl.ts
```diff
@@ -544,6 +546,7 @@ class MongoshNodeRepl implements EvaluationListener {
     });
 
     repl.on('exit', () => {
+      if (this._runtimeState) this._runtimeState.repl = null;
```
gagik, line 549: what are the implications of this being set to null? is this to force a quicker garbage collection / flushing? might be worth leaving a comment, though perhaps I could just be lacking context

gagik (approved):
Nice, seems tricky to catch this!

packages/cli-repl/src/mongosh-repl.ts
addaleax: Oh yeah, comment is a good idea – also added in 3bcdc5295025302154e06fa5f746f150aa5ca17e. (The shell can also exit when stdin ends, which would result in us calling this listener here first, before we call `.onExit()` through some other means)

packages/cli-repl/src/mongosh-repl.ts
addaleax: Yeah no, good point. I think I have a tendency to always terminate string that contain multiple newlines with one (just like most editors do for files), but Errors are a bit special in that their messages are usually displayed with an extra newline anyway. Removed in 3bcdc5295025302154e06fa5f746f150aa5ca17e!

PR #2301, opened 2024-12-12
chore(e2e-tests): consistently wait for successful exits of shell MONGOSH-1943

addaleax:
Do not use unqualified `shell.waitForExit()` in cases where we do not validate the exit code or check that no errors are displayed in the output, just assuming that the shell exited successfully, but instead intentionally always validate that these conditions are true in the cases in which we expect them to be.


nirinchev (approved):
Nice restructuring - I like the more explicit helper names. The test failures seem legit though and we should fix them before merging.

addaleax:
@nirinchev Yeah – https://github.com/mongodb-js/mongosh/pull/2298 and https://github.com/mongodb-js/mongosh/pull/2300 should address these. Part of the motivation for this change here is that, had we always used this "explicitly await success or failure", that issue would have been much more visible since it would have shown up as a consistent failure, not a flaky one 🙂

gagik (approved):

PR #2313, opened 2025-01-10
chore(docs): add architecture diagrams to repository

addaleax:


gagik (approved):

gagik:
Could it make sense to include the image inside some README.md?

addaleax:
@gagik Thought about that, but I don't know what that would look like right now – unless it would end up being a full description for the architecture, but I don't think that's something we have the time to produce right now 🙂

PR #2316, opened 2025-01-14
chore(ci): fix issues with bumping and publishing workflows

gagik:
Fixes some uncaught issues from before and some quick quality of life improvements.

### DIFF
--- a/.github/workflows/bump-auxiliary-packages.yml
+++ b/.github/workflows/bump-auxiliary-packages.yml
@@ -1,9 +1,6 @@
 name: Bump Auxiliary Packages
 on:
   workflow_dispatch:
-  push:
-    branches:
-      - main
 
 jobs:
   update_generated_files:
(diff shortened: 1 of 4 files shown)


.github/workflows/bump-auxiliary-packages.yml
```diff
@@ -36,11 +33,12 @@ jobs:
 
       - name: Install Dependencies
         run: |
-          npm ci
+          npm ci --workspace @mongosh/build
```
gagik, line 36: this is just to help speed up the CI a little

addaleax: I'm a bit worried that this would lead to inconsistent behavior between local an CI usage (e.g. because some dependency might have optional dependencies that are present locally but absent in this task) but it's no big deal either way

addaleax (approved):

.github/workflows/bump-auxiliary-packages.yml
gagik: I feel like that'd actually be a good thing because it'd help force us to be more intentional with defining our dependencies but yeah if it causes some unexpected behavior, that could be annoying

PR #2317, opened 2025-01-14
chore(cli-repl): clean up autocomplete implementation and fix flaky test MONGOSH-1908 MONGOSH-1960

addaleax:
##### chore(cli-repl): clean autocomplete implementation up a bit

Best viewed as a whitespace-diff-disabled change, doesn't include
any functional work.

##### chore(cli-repl): fix flaky autocompletion test MONGOSH-1960

Our current autocomplete logic does not always wait for the collection
or database name caches to be filled, in case it takes longer than 200ms
to do so. By ensuring that the caches are filled before we run the test,
we get consistent results.

##### chore(cli-repl,shell-api): filter autocompletions for CCs MONGOSH-1908

Collection and database names, as well as other completion items, can
mostly contain arbitrary characters with only a few exceptions.
Since some characters may mess with our REPL input, we should filter
those autocompletions out.

### DIFF
--- a/packages/cli-repl/src/cli-repl.spec.ts
+++ b/packages/cli-repl/src/cli-repl.spec.ts
@@ -2449,6 +2449,24 @@ describe('CliRepl', function () {
         await waitCompletion(cliRepl.bus);
         expect(output).to.include('res.acknowledged');
       });
+
+      it('completes only collection names that do not include control characters', async function () {
+        if (!hasCollectionNames) return;
+
+        input.write(
+          'db["actestcoll1"].insertOne({}); db["actestcoll2\\x1bfooobar"].insertOne({})\n'
+        );
+        await waitEval(cliRepl.bus);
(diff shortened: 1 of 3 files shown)


packages/cli-repl/src/cli-repl.spec.ts
```diff
@@ -2449,6 +2449,24 @@ describe('CliRepl', function () {
         await waitCompletion(cliRepl.bus);
         expect(output).to.include('res.acknowledged');
       });
+
+      it('completes only collection names that do not include control characters', async function () {
+        if (!hasCollectionNames) return;
```
gagik, line 2454: very nit: we could `this.skip()` instead

gagik (approved):

packages/cli-repl/src/cli-repl.spec.ts
addaleax: Sure, done!

PR #2319, opened 2025-01-15
chore(release): use a different release process when publishing auxiliary packages

gagik:
Some refactoring to make it different scenarios clearer and use a different function when just publishing auxiliary packages.
We should just be publishing to npm in this case (and let lerna push tags).


addaleax (approved):
Only small bit of feedback I'd have is that the commit message here could be used to explain what didn't work with the previous approach/how exactly this approach differs from it 🙂

PR #2322, opened 2025-01-17
feat(logging): custom log entries MONGOSH-1989

alenakhineika:
Allow users to add custom entries to mongosh logs in their [scripts](https://www.mongodb.com/docs/mongodb-shell/write-scripts/).

```js
log.info('This is an info message');
log.warn('Warning!');
```

The custom log message will sequentially appear alongside internal mongosh messages, and will look similar to this:

```log
{"t":{"$date":"2025-01-06T16:14:23.023Z"},"s":"I","c":"MONGOSH","id":1000000012,"ctx":"shell-api","msg":"Loading file via load()","attr":{"nested":false,"filename":"/Users/alena.khineika/custom-logging.js"}}
{"t":{"$date":"2025-01-06T16:14:23.041Z"},"s":"I","c":"MONGOSH-SCRIPTS","id":1000000054,"ctx":"custom-log","msg":"This is an info message"}
{"t":{"$date":"2025-01-06T16:14:23.041Z"},"s":"W","c":"MONGOSH-SCRIPTS","id":1000000054,"ctx":"custom-log","msg":"Warning!"}
```


packages/shell-api/src/shell-instance-state.ts
```diff
@@ -362,7 +362,48 @@ export default class ShellInstanceState {
       });
     }
 
-    this.messageBus.emit('mongosh:setCtx', { method: 'setCtx', arguments: {} });
+    const bus = this.messageBus;
+    if (contextObject.log === undefined) {
```
addaleax, line 366: This can just be `??=`

packages/shell-api/src/shell-instance-state.ts
```diff
@@ -362,7 +362,48 @@ export default class ShellInstanceState {
...
+        },
+        debug(message: string, attr?: unknown, level?: 1 | 2 | 3 | 4 | 5) {
+          bus.emit('mongosh:write-custom-log', {
+            method: 'debug',
+            message,
+            attr,
+            level,
+          });
+        },
+      };
```
addaleax, line 404: I'd split this out info a separate object, similar to how we've already split out the `ShellApi` class separately.

packages/e2e-tests/test/e2e.spec.ts
```diff
@@ -1506,6 +1511,45 @@ describe('e2e', function () {
...
+            `connect(${JSON.stringify(connectionString)})`
+          );
+          const filename = path.resolve(
+            __dirname,
+            '..',
+            '..',
+            'cli-repl',
+            'test',
+            'fixtures',
+            'custom-log-info.js'
```
addaleax, line 1543: Why put the fixture in `cli-repl` if it's used here? :)

packages/e2e-tests/test/e2e.spec.ts
```diff
@@ -1506,6 +1511,45 @@ describe('e2e', function () {
...
+          );
+          await shell.executeLine(`load('${filename}')`);
```
addaleax, line 1545:
```suggestion
          await shell.executeLine(`load(${JSON.stringify(filename)})`);
```

packages/e2e-tests/test/e2e.spec.ts
```diff
@@ -1506,6 +1511,45 @@ describe('e2e', function () {
             ).to.have.lengthOf(1);
           });
         });
+
+        it('writes custom log directly', async function () {
+          const connectionString = await testServer.connectionString();
+          await shell.executeLine(
+            `connect(${JSON.stringify(connectionString)})`
+          );
```
addaleax, line 1519: What does this do?

alenakhineika: It was going to check that both commands write to the same log. It is coming :)

packages/shell-api/src/shell-log.ts
```diff
@@ -0,0 +1,54 @@
+import type { MongoshBus } from '@mongosh/types';
+
+export interface ShellLog {
```
nirinchev, line 3: Should this extend `ShellApi`?

nirinchev, line 3: Should this extend `ShellApi`?

packages/logging/src/setup-logger-and-telemetry.spec.ts
```diff
@@ -822,4 +822,105 @@ describe('setupLoggerAndTelemetry', function () {
...
+
+    bus.emit('mongosh:write-custom-log', {
+      method: 'fatal',
+      message: 'Fatal!',
+    });
+
+    bus.emit('mongosh:write-custom-log', {
+      method: 'debug',
+      message: 'Debug',
+      level: 1,
```
nirinchev, line 872: Can we also add a test where we log at debug severity without supplying a level?

packages/logging/src/setup-logger-and-telemetry.spec.ts
```diff
@@ -822,4 +822,105 @@ describe('setupLoggerAndTelemetry', function () {
...
+    expect(logOutput).to.have.lengthOf(0);
+    expect(analyticsOutput).to.be.empty;
+
+    bus.emit('mongosh:connect', {
+      uri: 'mongodb://localhost/',
+      is_localhost: true,
+      is_atlas: false,
+      resolved_hostname: 'localhost',
+      node_version: 'v12.19.0',
+    });
```
nirinchev, line 846: I'm sorry if this is silly, but as someone who's not super proficient with the mongosh testing harness, it's not clear to me why we need this as part of this test. Is it used to test that non-custom logs are still emitted or for some other reason?

packages/e2e-tests/test/e2e.spec.ts
```diff
@@ -1506,6 +1511,46 @@ describe('e2e', function () {
...
+          const filename = path.resolve(
+            __dirname,
+            '..',
+            '..',
+            'e2e-tests',
+            'test',
```
addaleax, line 1537:
```suggestion
```

packages/e2e-tests/test/e2e.spec.ts
```diff
@@ -578,6 +578,11 @@ describe('e2e', function () {
       shell.assertNoErrors();
     });
 
+    it('runs a custom log command', async function () {
+      await shell.executeLine("log.info('Try me')");
+      shell.assertNoErrors();
+    });
```
addaleax, line 584: I was about to comment that this should assert that the message is actually logged, but then it seems a test further below already checks this. Does this test do anything that the other doesn't? The `shell.assertNoErrors()` call can and should be added to the other tests as well.

packages/e2e-tests/test/e2e.spec.ts
```diff
@@ -1506,6 +1511,46 @@ describe('e2e', function () {
...
           });
         });
+
+        it('writes custom log directly', async function () {
+          await shell.executeLine("log.info('This is a custom entry')");
+          await eventually(async () => {
+            const log = await readLogfile();
+            expect(
+              log.filter((logEntry) =>
+                logEntry.msg.includes('This is a custom entry')
```
addaleax, line 1521: Let's try to verify the other attributes of `logEntry` as well?

packages/shell-api/src/shell-log.ts
addaleax: You mean `ShellApiClass`? Yes, it probably should for consistency (and to ensure that we add e.g. `help` text for the new methods).

nirinchev: Yeah, the same - must have gotten distracted halfway through typing it 😵‍💫

alenakhineika: This was my initial thought as well, but then I looked at the ShellBson and now I am wondering how it is different. Or should ShellBson also extend ShellApi?

packages/logging/src/setup-logger-and-telemetry.spec.ts
alenakhineika: Yes, that is exactly the reason to check that all of those records are located in the same log file and the custom log is not being saved separately to keep the ordered log entries in one place to provide a clear picture for debugging.

packages/shell-api/src/shell-log.ts
addaleax: `ShellBson` groups a few classes/constructors together, but this one is a class that the user would interact with directly, so yes, they are a bit different.

packages/shell-api/src/shell-log.ts
```diff
@@ -0,0 +1,68 @@
...
+  // Use symbols to make sure these are *not* among the things copied over into
+  // the global scope.
+  [instanceStateSymbol]: ShellInstanceState;
+  log: {
+    info: (message: string, attr?: unknown) => void;
+    warn: (message: string, attr?: unknown) => void;
+    error: (message: string, attr?: unknown) => void;
+    fatal: (message: string, attr?: unknown) => void;
+    debug: (message: string, attr?: unknown, level?: 1 | 2 | 3 | 4 | 5) => void;
+  };
```
addaleax, line 20: This is surprising – I'd have expected `ShellLog` to be a class with `info`, `warn`, etc. methods. Right now, this class doesn't really do anything on its own, right?

packages/shell-api/src/shell-instance-state.ts
```diff
@@ -362,6 +365,7 @@ export default class ShellInstanceState {
       });
     }
 
+    Object.assign(contextObject, { log: this.shellLog });
```
addaleax, line 368: This is an involved way of writing

```suggestion
    contextObject.log = this.shellLog;
```

🙂

packages/shell-api/src/shell-log.ts
```diff
@@ -0,0 +1,60 @@
+import type ShellInstanceState from './shell-instance-state';
+import { shellApiClassDefault, ShellApiClass } from './decorators';
+
+const instanceStateSymbol = Symbol.for('@@mongosh.instanceState');
+
+/**
+ * This class contains the *global log* property that is considered part of the immediate shell API.
+ */
+@shellApiClassDefault
+export default class ShellLog extends ShellApiClass {
```
addaleax, line 10: tiny suggestion: let's avoid default exports for new code? :)

addaleax (approved):

gagik (approved):

PR #2325, opened 2025-01-22
feat(logging): add configuration to disable logging MONGOSH-1988

gagik:
Adds ability to disable logging completely both through global file and local configurations.


packages/logging/src/logging-and-telemetry.ts
```diff
@@ -0,0 +1,785 @@
...
+import type {
+  AnalyticsIdentifyMessage,
+  MongoshAnalytics,
+  MongoshAnalyticsIdentity,
+} from './analytics-helpers';
+import { hookLogger } from '@mongodb-js/devtools-connect';
+import { MultiSet, toSnakeCase } from './helpers';
+import { Writable } from 'stream';
+
+export class MongoshLoggingAndTelemetry {
```
gagik, line 49: I think structuring it as a class provided much better readability and an easier interface for us to test and debug the functionality of this compared to a function. Open to using the function structure though if there's a reason I may be missing here, I started doing it that way and just realized the ability to stub and access the scope of the class may be worth some greater changes. 

Naming of it I'm less sure of.

packages/logging/src/logging-and-telemetry.ts
```diff
@@ -0,0 +1,785 @@
...
+
+  /** Start processing pending events after mongosh:logger-initialized */
+  private onLoggerInitialized() {
+    this.hasMongoLogWriterInitialized = true;
+    for (const pendingEvent of this.pendingLogEvents) {
+      pendingEvent();
+    }
+    this.pendingLogEvents = [];
+
+    hookLogger(this.bus, this.log, 'mongosh', (uri) =>
```
gagik, line 119: Worth noting that the `devtools-connect` hooks right now only get initialized after the actual logger has been initialized. the reason for that is that, at least with our current setup, we end up passing the dummy log into this function, causing a bunch of event listeners to listen to even with the old log even after we swap it (since the reference of `this.log` doesn't change in this external case). 
If we want to be able to do the pending events in these cases, we may need to consider i.e. updating `devtools-connect` to pass a function to retrieve to get the logger instead.

packages/logging/src/logging-and-telemetry.ts
```diff
@@ -0,0 +1,785 @@
...
+
+  telemetryAnonymousId: string | undefined;
+  userId: string | undefined;
+  trackingProperties: {
+    mongosh_version: string;
+    session_id: string;
+  };
+  userTraits: AnalyticsIdentifyMessage['traits'] & {
+    [key: string]: unknown;
+  };
```
addaleax, line 61: Let's mark the things that are readonly here with `readonly`/`Readonly<>` to make it a bit easier to follow what's state of the class and what's just parameters.

packages/logging/src/logging-and-telemetry.ts
addaleax: A couple of thoughts:

- Turning this into a class seems fine as a way of acknowleding that this code has a bit of state that it manages on its own.
- "he ability to stub and access the scope of the class" is usually a _bad_ reason to turn something into a class – exposing internal state as a way to test something is often a code smell that means that the API of what you're testing doesn't directly allow you to observe behavior that should be externally visible
- This kind of change should probably be mentioned in the tech design and ideally addressed as a separate item, so it's more obvious what the refactor and the functional changes are, each (probably not necessary at this point since disentangling this is probably more work than not) – and so the review process doesn't get too bloated 🙂

packages/logging/src/logging-and-telemetry.ts
```diff
@@ -0,0 +1,785 @@
...
+      listener: (...args: T) => void
+    ) => {
+      this.bus.on(event, (...args: T) => {
+        if (this.hasMongoLogWriterInitialized) {
+          listener(...args);
+        } else {
+          this.pendingLogEvents.push(() => listener(...args));
+        }
+      });
+    };
```
addaleax, line 142: The template makes this looks like it would be type-safe, but right now it doesn't actually do anything towards that end (i.e. the `T` could as well just be `any[]` here) – here's a typesafe version of this:

```suggestion
    const onBus = <K extends keyof MongoshBusEventsMap>(
      event: K,
      listener: (...args: Parameters<MongoshBusEventsMap[K]>) => void
    ) => {
      this.bus.on(event, ((...args: Parameters<MongoshBusEventsMap[K]>) => {
        if (this.hasMongoLogWriterInitialized) {
          listener(...args);
        } else {
          this.pendingLogEvents.push(() => listener(...args));
        }
      }) as MongoshBusEventsMap[K]);
    };
```

Otherwise there's nothing that correlates the listener parameters to the actual event name (which is what's supposed to be defining the listener parameters).

packages/logging/src/logging-and-telemetry.ts
```diff
@@ -0,0 +1,785 @@
...
+
+    // We emit different analytics events for loading files and evaluating scripts
+    // depending on whether we're already in the REPL or not yet. We store the
+    // state here so that the places where the events are emitted don't have to
+    // be aware of this distinction.
+    let hasStartedMongoshRepl = false;
```
addaleax, line 148: Should this be part of the class state?

packages/logging/src/logging-and-telemetry.ts
```diff
@@ -0,0 +1,785 @@
...
+    this.trackingProperties.session_id = this.log.logId;
+  }
+
+  public getTelemetryUserIdentity = (): MongoshAnalyticsIdentity => ({
+    anonymousId: this.telemetryAnonymousId ?? (this.userId as string),
+  });
+
+  /** Start processing pending events after mongosh:logger-initialized */
+  private onLoggerInitialized() {
+    this.hasMongoLogWriterInitialized = true;
```
addaleax, line 113: Should there be a guard against calling this twice (same comment for `setupLogger()` if we don't merge the two anyway)?

packages/logging/src/logging-and-telemetry.ts
```diff
@@ -0,0 +1,785 @@
...
+    this.userTraits.session_id = this.log.logId;
+    this.trackingProperties.session_id = this.log.logId;
+  }
+
+  public getTelemetryUserIdentity = (): MongoshAnalyticsIdentity => ({
+    anonymousId: this.telemetryAnonymousId ?? (this.userId as string),
+  });
+
+  /** Start processing pending events after mongosh:logger-initialized */
+  private onLoggerInitialized() {
```
addaleax, line 112: It seems redundant and fragile that we are attaching a listener on `mongosh:logger-initialized` if we already get informed about the logger being initialized via a direct call to `.setupLogger()` – can we just wrap this method into the latter?

packages/logging/src/logging-and-telemetry.ts
```diff
@@ -0,0 +1,785 @@
...
+      });
+    };
+
+    // We emit different analytics events for loading files and evaluating scripts
+    // depending on whether we're already in the REPL or not yet. We store the
+    // state here so that the places where the events are emitted don't have to
+    // be aware of this distinction.
+    let hasStartedMongoshRepl = false;
+
+    /* eslint prefer-arrow-callback: "error" */
```
addaleax, line 150:
```suggestion
```

This duplicates a line from the top of the file

packages/logging/src/logging-and-telemetry.ts
```diff
@@ -0,0 +1,785 @@
...
+      session_id: this.log.logId,
+    };
+
+    this.trackingProperties = {
+      mongosh_version: mongoshVersion,
+      session_id: this.log.logId,
+    };
+  }
+
+  public setup() {
```
addaleax, line 92: Is there any case where this isn't called immediately after the constructor?

packages/cli-repl/src/cli-repl.ts
```diff
@@ -346,23 +353,31 @@ export class CliRepl implements MongoshIOProvider {
...
+      this.warnAboutInaccessibleFile(err as Error);
     }
 
     this.globalConfig = await this.loadGlobalConfigFile();
     markTime(TimingCategories.UserConfigLoading, 'read global config files');
 
+    const disableLogging = this.getConfig('disableLogging');
+    if (disableLogging !== true) {
+      await this.startLogging(loggingAndTelemetry);
+    }
```
addaleax, line 379: You'll want to add an e2e test for this as well, maybe even including one that tests that the global config file can disable logging via this option

addaleax (changes requested):

packages/logging/src/logging-and-telemetry.ts
addaleax: > at least with our current setup, we end up passing the dummy log into this function, causing a bunch of event listeners to listen to even with the old log even after we swap it (since the reference of this.log doesn't change in this external case).

This is a hypothetical case, right? As if, this is what would happen if we were to call `hookLogger` from the beginning, correct?

I think that's fine then. If we do need to do something to address this, we could also pass an object to `hookLogger` that buffers up log events until the real log destination is available, then forward events to that destination, similar to how we handle the events here.

packages/logging/src/logging-and-telemetry.ts
gagik: That makes sense, I did think it may be too big of a change. I do not mind writing this into a function, I was thinking of doing that first and then introducing an extra commit I could reverse but seemed easier for me to reason to have this state first.

packages/cli-repl/src/cli-repl.ts
```diff
@@ -856,9 +872,7 @@ export class CliRepl implements MongoshIOProvider {
    * Implements getConfig from the {@link ConfigProvider} interface.
    */
   // eslint-disable-next-line @typescript-eslint/require-await
-  async getConfig<K extends keyof CliUserConfig>(
-    key: K
-  ): Promise<CliUserConfig[K]> {
+  getConfig<K extends keyof CliUserConfig>(key: K): CliUserConfig[K] {
```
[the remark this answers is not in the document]
gagik: ah that makes sense, I didn't notice the eslint comment and didn't realize this was an intentional decision for the sake of being consistent with an interface (I guess I would have expected some type error)

will revert 🙂

packages/logging/src/logging-and-telemetry.ts
gagik: As in, it could be left as a class? 

Also, with this:
> exposing internal state as a way to test something is often a code smell that means that the API of what you're testing doesn't directly allow you to observe behavior that should be externally visible

do I understand correctly that the flaw here is that tests should focus on testing expected behavior and not some arbitrary internal state one exposes?

packages/logging/src/logging-and-telemetry.ts
gagik: Ah yeah I meant 

> with our current setup ***if we were to initialize them like we do all the other onBus listeners setup()**..

packages/logging/src/logging-and-telemetry.ts
addaleax: Yeah, leaving it as a class seems fine. I'd maybe consider marking as many of the properties `private` as possible, but otherwise it's not like "function with a bunch of local variables" and "class with a bunch of properties" are terribly different concepts.

> > exposing internal state as a way to test something is often a code smell that means that the API of what you're testing doesn't directly allow you to observe behavior that should be externally visible
> 
> do I understand correctly that the flaw here is that tests should focus on testing expected behavior and not some arbitrary internal state one exposes?

Yeah, exactly. If I get to a point where I'd want to test internal state, that's usually a good point to wonder if maybe the class can be broken down into smaller pieces that can be unit tested individually, rather than starting to rely on the ability to inspect and/or mock internals.

packages/logging/src/logging-and-telemetry.ts
gagik: Yeah, I have had people prefer a constructor not to have "side effects" so wasn't sure about combining it. Though I myself don't see a problem, at least in this case.

addaleax: Yeah, fair enough, I didn't really think of setting up listeners as side effects here 🙂

packages/logging/src/logging-and-telemetry.ts
gagik: > otherwise it's not like "function with a bunch of local variables" and "class with a bunch of properties" are terribly different concepts

good throwback to pre-ES6 JavaScript I guess.

sounds good, thank you for the explanation 😀

packages/logging/src/logging-and-telemetry.ts
gagik: to put it in context of the tech design, would `setupLogger` also emit `mongosh:logger-initialized` but we'd not necessarily attach any listeners to it? 

or would we get rid of `mongosh:logger-initialized` altogether?

addaleax: > would `setupLogger` also emit `mongosh:logger-initialized` but we'd not necessarily attach any listeners to it?

This is what I'd lean towards, since "logger initialized" is generally a meaningful event that another part of mongosh could be interested in listening for it. But that's just theoretical, so, no strong feelings about keeping or removing it altogether.

packages/logging/src/logging-and-telemetry.ts
```diff
@@ -119,35 +115,43 @@ export class MongoshLoggingAndTelemetry {
...
+    };
+  }
+
+  /** We emit different analytics events for loading files and evaluating scripts
+   * depending on whether we're already in the REPL or not yet. We store the
+   * state here so that the places where the events are emitted don't have to
+   * be aware of this distinction. */
+  private hasStartedMongoshRepl = false;
+
+  private apiCallTracking = {
```
gagik, line 134: Seemed best to move all variables into the class state and keep the setup listeners function scope clean of state but if it's making the class too bloated, could reconsider (or maybe scope them under `eventListenersState` or something)

packages/logging/src/logging-and-telemetry.ts
```diff
@@ -0,0 +1,791 @@
...
+
+  public setup(): void {
+    if (this.isSetup) {
+      throw new Error('Setup can only be called once.');
+    }
+    this._setupBusEventListeners();
+    this.isSetup = true;
+  }
+
+  public setupLogger(logger: MongoLogWriter): void {
```
gagik, line 100: This might be too destructive, I could limit it to only event defined here but then we don't end up handling `hookLogger` things

packages/cli-repl/src/cli-repl.ts
```diff
@@ -605,6 +626,19 @@ export class CliRepl implements MongoshIOProvider {
...
+  async setLoggingEnabled(enabled: boolean): Promise<void> {
+    if (enabled) {
+      if (!this.logWriter) {
+        await this.startLogging();
+      }
+      // Do nothing if there is already an active log writer.
+    } else {
+      this.loggingAndTelemetry?.detachLogger();
+      this.logWriter?.destroy();
+      this.logWriter = undefined;
```
gagik, line 638: So we create a new logger (and new logId) each time logging gets reenabled, which also means if someone disable and then enables logging, a new file is created.
I think that is reasonable unless we'd rather make sure it writes to the same file again (thinking of it more of a "pause")

packages/cli-repl/src/cli-repl.ts
```diff
@@ -605,6 +626,19 @@ export class CliRepl implements MongoshIOProvider {
     );
   }
 
+  async setLoggingEnabled(enabled: boolean): Promise<void> {
```
gagik, line 629: do we want to consider `disableLogging()` and `enableLogging()` shorthands similar to telemetry?

addaleax: We only have those because they predate the `config` object, otherwise we probably wouldn't

packages/logging/src/logging-and-telemetry.ts
```diff
@@ -114,10 +155,24 @@ export class MongoshLoggingAndTelemetry {
     this.trackingProperties.session_id = this.log.logId;
 
     this.isDummyLog = false;
+    if (this.hookedExternalListenersLogId !== this.log.logId) {
+      hookLogger(
+        {
+          ...this.bus,
```
gagik, line 161: this is quite hacky... we probably should rather update the `hookLogger` instead?

i.e. instead of `this.log` we could update it to take `() => this.log` which can also support nullability

gagik: alternatively we call `this.bus.removeAllEventListeners()` when stopping logging and set up the event listeners again for the rest

packages/logging/src/logging-and-telemetry.ts
```diff
@@ -0,0 +1,884 @@
...
+    [key: string]: unknown;
+  };
+
+  private pendingLogEvents: CallableFunction[] = [];
+
+  private telemetryAnonymousId: string | undefined;
+  private userId: string | undefined;
+  private isDummyLog = true;
+  private isSetup = false;
+  private hookedExternalListenersLogId: string | undefined = undefined;
```
addaleax, line 67: I feel like this class's state space exploded a bit now, which makes this pretty hard to review and probably also hard to work with in the future

e.g. it's hard to tell whether `isDummyLog = false, isSetup = true, log = null, pendingLogEvents = [...], hookedExternalListenersLogId = 'foo'` is a valid state for this class or not and what that means specifically

I'll try to get more specific at individual points here but I think we might want to rethink this and look for opportunities to make this simpler

packages/logging/src/logging-and-telemetry.ts
```diff
@@ -0,0 +1,884 @@
...
+    };
+  }
+
+  public setup(): void {
+    if (this.isSetup) {
+      throw new Error('Setup can only be called once.');
+    }
+    this._setupBusEventListeners();
+    this.isSetup = true;
+  }
```
addaleax, line 101: As one step here, how about we:

- Define the interface of this class as a separate TS `interface`
- Make sure the class implements this interface
- Stop exporting the class from this file and instead export a function that creates the instance + calls `.setup()` (as well as export the interface itself)
- Use the interface as that function's return type 

That feels like a better way to get out of the ".setup() must always be called after the ctor because ctors should not have side effects" dilemma – ultimately the "ctors should not have side effects" rule is about communication and managing expectations of readers/maintainers, not necessarily about code structure

packages/logging/src/logging-and-telemetry.ts
gagik: oh yeah, didn't think about having the writer be the adapter, sounds good

packages/logging/src/logging-and-telemetry.ts
```diff
@@ -0,0 +1,884 @@
...
+      this.hookedExternalListenersLogId = this.log.logId;
+    }
+
+    this.runAndCleanPendingEvents();
+
+    this.bus.emit('mongosh:log-initialized');
+  }
+
+  public detachLogger() {
+    this.log = null;
```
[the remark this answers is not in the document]
gagik: The idea was to differentiate initial buffering of events from intentional discarding of the writer but yeah this did not end up worth all the extra state thinking that came along with it. Will change

packages/logging/src/logging-and-telemetry.ts
```diff
@@ -0,0 +1,875 @@
...
+            code: mongoshError.code,
+            scope: mongoshError.scope,
+            metadata: mongoshError.metadata,
+          },
+        });
+      }
+    });
+
+    onBus('mongosh:write-custom-log', (event: WriteCustomLogEvent) => {
+      if (this.log) {
```
addaleax, line 354: Is this still necessary? If not, we can remove it, if it is, we can still simply this with `?.`

packages/logging/src/logging-and-telemetry.ts
```diff
@@ -880,5 +851,25 @@ export class MongoshLoggingAndTelemetry {
         );
       }
     );
+
+    hookLogger(
+      this.bus,
+      {
+        info: (...args: Parameters<typeof this.log.info>) => {
```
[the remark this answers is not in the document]
addaleax: > Do we want to specify them here or potentially update LogWriter i.e. use an interface?

I think that would be a very nice follow-up, but not required for this PR since `hookLogger` already defines the interface it needs

addaleax (commented):
Looks great! Just a few (final 🤞) comments

gagik:
@addaleax @alenakhineika 
do we usually merge stacked PRs i.e. https://github.com/mongodb-js/mongosh/pull/2355 like this or do we want to revert that, get this one into main and then merge the #2355 into main as well?

packages/e2e-tests/test/e2e.spec.ts
```diff
@@ -22,6 +22,7 @@ import { once } from 'events';
 import type { AddressInfo } from 'net';
 const { EJSON } = bson;
 import { sleep } from './util-helpers';
+import type { LogEntry } from '@mongosh/shell-api/src/log-entry';
```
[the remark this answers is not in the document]
gagik: good catch, replaced with `MongoLogEntry`
https://github.com/mongodb-js/devtools-shared/blob/01e5c1c9c3563b47b1eefcfb6f0ffd6b9a2ef0f7/packages/mongodb-log-writer/src/mongo-log-writer.ts#L25

alenakhineika:
From my side, I would merge them all into one and then merge them all into the main to be able to test the whole thing first. Let's wait for @addaleax thoughts about it.

addaleax:
> From my side, I would merge them all into one and then merge them all into the main to be able to test the whole thing first. Let's wait for @addaleax thoughts about it.

I don't see why we'd be more or less able to test this depending on how we merge things?

If this was just a "feature branch" PR, I think that would be okay, but right now it's a) a "feature branch" PR plus b) a refactor PR plus c) an actual feature PR, and that's something I'd like to get merged as quickly as possible because otherwise there's no way to properly the review the feature and refactor in isolation

packages/e2e-tests/test/e2e-tls.spec.ts
```diff
@@ -241,13 +241,13 @@ describe('e2e TLS', function () {
         expect(prompt.state).to.equal('prompt');
 
         const logPath = path.join(logBasePath, `${shell.logId}_log`);
-        const logContents = await readReplLogfile(logPath);
+        const logContents = await readReplLogFile(logPath);
         expect(
-          logContents.find((line) => line.id === 1_000_000_049).attr
+          logContents.find((line) => line.id.__value === 1_000_000_049)?.attr
```
addaleax, line 246: Looks like CI doesn't like this modification, and I also can't see where this change is coming from

addaleax (approved):

PR #2326, opened 2025-01-22
feat(cli-repl): add configuration to set log location MONGOSH-1983

gagik:
Adds ability to configure log location.

### DIFF
--- a/packages/types/src/index.ts
+++ b/packages/types/src/index.ts
@@ -586,6 +586,13 @@ export class CliUserConfigValidator extends SnippetShellUserConfigValidator {
           value !== undefined &&
           (typeof value !== 'string' || !path.isAbsolute(value))
         ) {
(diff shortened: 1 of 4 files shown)


packages/types/src/index.ts
```diff
@@ -574,6 +575,11 @@ export class CliUserConfigValidator extends SnippetShellUserConfigValidator {
           return `${key} must be a valid URL or empty`;
         }
         return null;
+      case 'logLocation':
+        if (value !== undefined && typeof value !== 'string') {
```
gagik, line 579: I am wondering if there's a good way to check for valid path. I was thinking regex but worried it'd be end up disallowing valid paths in weirder OSes. Running `fs.access` could be a way, just seems to be a bit more proactive validator than others.

addaleax (approved):

alenakhineika (approved):

PR #2328, opened 2025-01-23
feat(cli-repl): add support for bracketed paste in REPL MONGOSH-1909

addaleax:
Bracketed paste allows us to receive a copy-pasted piece of mongosh as a single block, rather than interpreting it line-by-line. For now, this requires some monkey-patching of Node.js internals, so a follow-up ticket will include work to upstream support for this into Node.js core.

### DIFF
--- a/packages/cli-repl/package.json
+++ b/packages/cli-repl/package.json
@@ -43,7 +43,7 @@
     "access": "public"
   },
   "engines": {
-    "node": ">=16.15.0"
+    "node": ">=18.19.0"
   },
   "mongosh": {
     "ciRequiredOptionalDependencies": {
--- a/packages/e2e-tests/test/e2e.spec.ts
+++ b/packages/e2e-tests/test/e2e.spec.ts
@@ -298,6 +298,19 @@ describe('e2e', function () {
       await shell.waitForSuccessfulExit();
       shell.assertContainsOutput('3628800');
     });
+    it('ignores control characters in TTY input', async function () {
+      shell = this.startTestShell({
+        args: ['--nodb'],
+        forceTerminal: true,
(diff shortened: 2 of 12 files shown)


packages/cli-repl/package.json
```diff
@@ -43,7 +43,7 @@
     "access": "public"
   },
   "engines": {
-    "node": ">=16.15.0"
+    "node": ">=18.19.0"
```
addaleax, line 46: Minimum version with https://github.com/nodejs/node/pull/47150

packages/e2e-tests/test/e2e.spec.ts
```diff
@@ -1510,9 +1540,6 @@ describe('e2e', function () {
 
       describe('history file', function () {
         it('persists between sessions', async function () {
-          if (process.arch === 's390x') {
-            return this.skip(); // https://jira.mongodb.org/browse/MONGOSH-746
-          }
```
addaleax, line 1515: Drive-by because this is also fixed by the env var fixup in .evergreen/setup-env.sh

gagik (approved):

lerouxb (approved):

gagik:
Since this touches `@mongosh/e2e-tests` might be worth holding on until https://github.com/mongodb-js/mongosh/pull/2332 is merged and the temporary push-tags workflow is run just so we can have an accurate tag representing the currently published version.

gagik:
> Since this touches `@mongosh/e2e-tests` might be worth holding on until #2332 is merged and the temporary push-tags workflow is run just so we can have an accurate tag representing the currently published version.

Tags have been pushed 😄

PR #2332, opened 2025-01-27
fix(ci): add publishing tags step when publishing auxiliary packages MONGOSH-1871

gagik:
lerna publish [doesn't automatically push tags](https://github.com/lerna/lerna/issues/2085) when using from-package so we have to create and push the tags manually.

This is different from releasing mongosh as it does its own tag push already (without `mongosh@` prefix): https://github.com/mongodb-js/mongosh/blob/e0a051c30f3198fccd6257c77d4bf243a0cd173f/packages/build/src/npm-packages/publish.ts#L59-L73

### DIFF
--- a/.github/workflows/publish-auxiliary-packages.yml
+++ b/.github/workflows/publish-auxiliary-packages.yml
@@ -58,10 +58,3 @@ jobs:
           npm config list
           echo "Publishing packages as $(npm whoami)"
           npm run publish-auxiliary
-
-      - name: "Publish tags"
-        run: |
-          npx lerna list -a --json | \
-            jq -r '.[] |  select(.name != "mongosh") | .name + "@" + .version' | \
-            xargs -i sh -c "git tag -a {} -m {} || true"
-          git push --follow-tags
--- a/.github/workflows/publish-auxiliary-packages.yml
+++ b/.github/workflows/publish-auxiliary-packages.yml
@@ -58,3 +58,10 @@ jobs:
           npm config list
           echo "Publishing packages as $(npm whoami)"
           npm run publish-auxiliary
+
+      - name: "Publish tags"
+        run: |
+          npx lerna list -a --json | \
+            jq -r '.[] |  select(.name != "mongosh") | .name + "@" + .version' | \
+            xargs -i sh -c "git tag -a {} -m {} || true"
+          git push --follow-tags
--- a/.github/workflows/temp-push-auxiliary-tags.yml
+++ b/.github/workflows/temp-push-auxiliary-tags.yml
@@ -12,7 +12,7 @@ jobs:
     environment: Production
 
     steps:
-      - uses: mongodb-js/devtools-shared/actions/setup-bot-token@main
+      - uses: mongodb-js/devtools-shared/actions/setup-bot-token@d638357cabd624ae626d9abb62cbac2760aea932
         id: app-token
         with:
           app-id: ${{ vars.DEVTOOLS_BOT_APP_ID }}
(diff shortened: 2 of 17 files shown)


.github/workflows/publish-auxiliary-packages.yml
```diff
@@ -58,3 +58,10 @@ jobs:
           npm config list
           echo "Publishing packages as $(npm whoami)"
           npm run publish-auxiliary
+
+      - name: "Publish tags"
```
gagik, line 62: essentially same as https://github.com/mongodb-js/compass/blob/main/.github/workflows/publish-packages.yaml#L60 with one adjustment to not push tags for `mongosh` in this case

.github/workflows/publish-auxiliary-packages.yml
```diff
@@ -58,3 +58,10 @@ jobs:
...
+        run: |
+          npx lerna list -a --json | \
+            jq -r '.[] |  select(.name != "mongosh") | .name + "@" + .version' | \
```
addaleax, line 65: Is `mongosh` excluded here because it's not scoped with `@mongosh/`? Maybe add a comment or even a test somewhere that reminds people to update this line if that ever changes

gagik: I am excluding it here as it is the publishing auxiliary packages which would mean it'd try to publish `mongosh@x.x.x` multiple times and make the tag meaningless.

Come to think of it, it'd be trivial to extract this to a function as well and make this distinction easier to reason about (and test/enforce) so I'll do that

packages/build/src/npm-packages/push-tags.ts
```diff
@@ -0,0 +1,88 @@
...
+    } else {
+      console.warn(`${newVersionTag} tag already exists. Skipping...`);
+    }
+  }
+
+  spawnSync('git', ['push', '--follow-tags'], commandOptions);
+}
+
+/** Returns true if the tag exists in the remote repository. */
+export function existsTag(tag: string): boolean {
```
gagik, line 75: I added this as another sanity check though maybe it's a bit of an overkill / slow when done for every auxiliary package.

packages/build/src/publish-mongosh.ts
```diff
@@ -22,6 +23,7 @@ export async function publishMongosh(
   barque: Barque,
   createAndPublishDownloadCenterConfig: typeof createAndPublishDownloadCenterConfigFn,
   publishToNpm: typeof publishToNpmType,
+  pushTags: typeof pushTagsType,
```
gagik, line 26: these helper functions are getting out of hand and only seem to exist for the purpose of spying/stubbing with sinon. I am keeping same structure here but think it'd be good to wrap them in a namespace or something in the future.

.github/workflows/temp-push-auxiliary-tags.yml
```diff
@@ -0,0 +1,49 @@
...
+    steps:
+      - uses: mongodb-js/devtools-shared/actions/setup-bot-token@d638357cabd624ae626d9abb62cbac2760aea932
+        id: app-token
+        with:
+          app-id: ${{ vars.DEVTOOLS_BOT_APP_ID }}
+          private-key: ${{ secrets.DEVTOOLS_BOT_PRIVATE_KEY }}
+
+      - uses: actions/checkout@v4
+        with:
+          # don't checkout a detatched HEAD
```
addaleax, line 23:
```suggestion
          # don't checkout a detached HEAD
```

packages/build/src/npm-packages/push-tags.ts
addaleax: I think this is fine 👍

addaleax (approved):

PR #2334, opened 2025-01-28
fix(evergreen): bump v8 to 11.8.50

ELDiablO59152:
Mongosh's build is broken on Freebsd 14 and 15 due to [a change in llvm18+](https://github.com/llvm/llvm-project/commit/e30a148b098d462d0267c479cd9e4783363c2761) (Aug 3 2023).
```bash
In file included from ../deps/v8/src/inspector/v8-string-conversions.cc:5:
In file included from ../deps/v8/src/inspector/v8-string-conversions.h:9:
/usr/include/c++/v1/string:820:42: error: implicit instantiation of undefined template 'std::char_traits<unsigned short>'
  820 |   static_assert(is_same<_CharT, typename traits_type::char_type>::value,
      |                                          ^
../deps/v8/src/inspector/v8-string-conversions.cc:387:26: note: in instantiation of template class 'std::basic_string<unsigned short>' requested here
  387 | std::basic_string<UChar> UTF8ToUTF16(const char* stringStart, size_t length) {
      |                          ^
/usr/include/c++/v1/__fwd/string.h:23:29: note: template is declared here
   23 | struct _LIBCPP_TEMPLATE_VIS char_traits;
      |                             ^
In file included from ../deps/v8/src/inspector/v8-string-conversions.cc:5:
In file included from ../deps/v8/src/inspector/v8-string-conversions.h:9:
In file included from /usr/include/c++/v1/string:647:
/usr/include/c++/v1/string_view:300:42: error: implicit instantiation of undefined template 'std::char_traits<unsigned short>'
  300 |   static_assert(is_same<_CharT, typename traits_type::char_type>::value,
```

source: https://reviews.llvm.org/D157058
> The base template for ``std::char_traits`` has been removed. If you are using
  ``std::char_traits`` with types other than ``char``, ``wchar_t``, ``char8_t``,
  ``char16_t``, ``char32_t`` or a custom character type for which you
  specialized ``std::char_traits``, your code will no longer work. The Standard
  does not mandate that a base template is provided, and such a base template is

source: https://github.com/llvm/llvm-project/blob/e30a148b098d462d0267c479cd9e4783363c2761/libcxx/docs/ReleaseNotes/18.rst?plain=1#L50-56

The fix appeared in V8 11.8.50 (Aug 21 2023) so I bumped the version to this ref but feel free to upgrade to a more recent version.

The following line:
``` c
using UChar = uint16_t;
```
was change to:
``` c
using UChar = char16_t;
```
source : https://chromium.googlesource.com/v8/v8/+/refs/heads/11.8.50/src/inspector/v8-string-conversions.cc

### DIFF
--- a/packages/build/src/npm-packages/bump.ts
+++ b/packages/build/src/npm-packages/bump.ts
@@ -95,6 +95,7 @@ export function bumpAuxiliaryPackages() {
     encoding: 'utf8',
     env: {
       ...process.env,
+      LAST_BUMP_COMMIT_MESSAGE: 'chore(release): bump auxiliary packages',
       SKIP_BUMP_PACKAGES: [
         ...EXCLUDE_RELEASE_PACKAGES,
         ...MONGOSH_RELEASE_PACKAGES,


ELDiablO59152:
Hi @addaleax, thank you very much for your fast reply !
Ok, I see, I will upgrade from node 20 to 22. I didn't know it was the new LTS.

Btw you can take a look at the build process at https://github.com/freebsd/freebsd-ports/tree/main/databases/mongosh

I had to make some patches to link all dynamic libs.
If you find some time to support FreeBSD natively in the source code, it would be great for the BSD community.

This is how I managed to build mongosh on BSD:
(Node sources has to be downloaded before building process when you use Poudriere)
```bash
SEGMENT_API_KEY=dummy \
GYP_DEFINES="libmongocrypt_link_type=dynamic" \
NODE_JS_VERSION=file://${_DISTDIR}/${NODEJS_TARBALL} \ 
BOXEDNODE_CONFIGURE_ARGS="--shared-brotli,--shared-cares,--shared-libuv,--shared-nghttp2,--shared-zlib,--shared-openssl,--openssl-use-def-ca-store,--without-npm" \
/usr/local/bin/npm run compile-exec
```

I have another issue also on x86 and arm architecture, nx-freebsd-arm64 does not exist and Mongosh build fails at build stp.
More infos here: https://github.com/nrwl/nx/issues/29257
Do you know how to build it ?

addaleax:
> This is how I managed to build mongosh on BSD:

This looks pretty fine to me! We don't currently have any plans to change our build processes in a way that would break this.

> I have another issue also on x86 and arm architecture, nx-freebsd-arm64 does not exist and Mongosh build fails at build stp.
More infos here: https://github.com/nrwl/nx/issues/29257
Do you know how to build it ?

Does adding a `echo > scripts/sort-workspaces.js` line to the build script fix this? It's a bit hacky but I've also had to do that to get mongosh to build on more exotic architectures locally.

ELDiablO59152:
> Does adding a `echo > scripts/sort-workspaces.js` line to the build script fix this? It's a bit hacky but I've also had to do that to get mongosh to build on more exotic architectures locally.

I see what you mean, I have this patch to fix a bug on BSD
```diff
--- scripts/sort-workspaces.js.orig	2024-05-03 16:42:16 UTC
+++ scripts/sort-workspaces.js
@@ -22,7 +22,7 @@ async function main() {
   ));
 
   // should use the scopes in lerna.json
-  const { stdout } = await exec('npx -y lerna ls --all --no-since --toposort --json');
+  const { stdout } = await exec('npx -y lerna ls --all --toposort --json');
   packageJSON.workspaces = JSON.parse(stdout).map(({ location }) => path.relative(monorepoRoot, location));
 
   await fs.writeFile(
```

PR #2338, opened 2025-01-29
fix(cli-repl): support bracketed paste on Windows MONGOSH-1999

addaleax:
Split out into a separate piece because this adds a patch to libuv that has non-trivial downsides, as explained in the referenced PR. I'd assume that mitigating this through a `process.on('exit')` handler is sufficient and that we don't need to plan specifically for e.g. hard crashes of mongosh here, but I didn't want to merge this together with the original work in 4232a9847cd5d2486 but rather have it stand on its own (including for the simple reason that it's easier to revert if we decide that we have to).

Refs: https://github.com/libuv/libuv/pull/4688

### DIFF
--- a/scripts/nodejs-patches/006-windows-virtual-terminal-input-libuv-4688.patch
+++ b/scripts/nodejs-patches/006-windows-virtual-terminal-input-libuv-4688.patch
@@ -1,0 +1,46 @@
+diff --git a/deps/uv/src/win/tty.c b/deps/uv/src/win/tty.c
+index 7e1f15544b17..37bc2c4ace57 100644
+--- a/deps/uv/src/win/tty.c
++++ b/deps/uv/src/win/tty.c
+@@ -58,6 +58,9 @@
+ #ifndef ENABLE_VIRTUAL_TERMINAL_PROCESSING
+ #define ENABLE_VIRTUAL_TERMINAL_PROCESSING 0x0004
+ #endif
++#ifndef ENABLE_VIRTUAL_TERMINAL_INPUT
++#define ENABLE_VIRTUAL_TERMINAL_INPUT 0x0200
++#endif
+ 
+ #define CURSOR_SIZE_SMALL     25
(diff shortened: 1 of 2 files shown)


scripts/nodejs-patches/006-windows-virtual-terminal-input-libuv-4688.patch
```diff
@@ -0,0 +1,46 @@
+diff --git a/deps/uv/src/win/tty.c b/deps/uv/src/win/tty.c
+index 7e1f15544b17..37bc2c4ace57 100644
+--- a/deps/uv/src/win/tty.c
++++ b/deps/uv/src/win/tty.c
+@@ -58,6 +58,9 @@
+ #ifndef ENABLE_VIRTUAL_TERMINAL_PROCESSING
+ #define ENABLE_VIRTUAL_TERMINAL_PROCESSING 0x0004
+ #endif
++#ifndef ENABLE_VIRTUAL_TERMINAL_INPUT
++#define ENABLE_VIRTUAL_TERMINAL_INPUT 0x0200
```
gagik, line 10: what is the significance of the 0x0200 value? is it just an available value?

addaleax: You can look up the values here: https://learn.microsoft.com/en-us/windows/console/setconsolemode#parameters

This `#define`, like the one above, is added in case the compilation happens on a Windows machine that doesn't have the most recent system headers and doesn't know about the flag yet, so you can still compile a program that can make use of it when run on more recent Windows machines.

That's not a practical concern for mongosh, but for the libuv PR it's relevant.

gagik (approved):
Very unfamiliar territory for me but seems to make sense.

addaleax:
Alright, I'll go ahead with this, even though we know it's quite likely that we'll need to adjust the libuv patch here a bit more.

PR #2343, opened 2025-01-31
fix(e2e-tests): make version test independent of package.json files

addaleax:
`--version`, `version()` and `--build-info` all source their version information differently. It should be sufficient to include a test to verify that they all match, without having to rely on the specific dependency from the e2e-test package.json file (which is in line with the general idea that e2e tests should not actually depend on the specific shell implementation and only test externally visible behavior of that implementation).


gagik (approved):
Sounds good to me!
(Sorry for taking over the branch, figured it'd be good to have this quick fix pass CI and ready to merge)

PR #2344, opened 2025-02-02
fix(tests): update VSCode runner to use 24.04

gagik:


addaleax (approved):
Looks good if it works 👍

.evergreen.yml
```diff
@@ -3597,9 +3597,13 @@ functions:
           set -e
           {
           source .evergreen/setup-env.sh
-          (cd scripts/docker && docker build -t ubuntu22.04-xvfb -f ubuntu22.04-xvfb.Dockerfile .)
+          (cd scripts/docker && docker build -t ubuntu24.04-xvfb -f ubuntu24.04-xvfb.Dockerfile .)
+          apt list --installed
+          sudo apt-get update
+          sudo apt-get install -y curl=7.68.0-1ubuntu2.22 libcurl4=7.68.0-1ubuntu2.22 libcurl4-openssl-dev=7.68.0-1ubuntu2.22
```
addaleax, line 3603: Were these lines supposed to go into the dockerfile config? Because otherwise this here would be affecting the global scope of the evergreen host, which we'd almost never want to do

gagik: yeah they should though none of these things work at the moment these are more of just different attempts at trying to fix it. Good reminder that evergreen maintains state though, that's one thing I find weird coming from GitHub Actions.

PR #2347, opened 2025-02-03
Revert "fix(ci): add Node.js patch for OpenSSL Windows build fix MONGOSH-1632 (#2285)"

orgads:
Already applied upstream.

This reverts commit 0130549a3218736a84f663458a6381c853cd9bd5.


addaleax:
@orgads This patch doesn't seem to be applied to Node.js 20.x, which is what mongosh currently builds against. If you're building mongosh locally, you'll want to set the `NODE_JS_VERSION` environment variable to a 20.x version.

orgads:
I see, thanks.

PR #2353, opened 2025-02-04
fix(ci): fix build tests for pushing tags

gagik:
--follow-tags was replaced with --tags but the corresponding tests were not fully updated.


MohsenPrg (approved):
c6d8afe60cc8084348411a62b64325c6b45cc75a

MohsenPrg (commented):
I'm sorry  , I don't know how to use this app

PR #2356, opened 2025-02-04
fix(ci): make bumping mongosh apply only to mongosh packages

gagik:
mongosh bump would actually end up overriding most auxiliary packages. This was likely not noticable before as the versions were already aligned and the ordering of the auxiliary bump was after the mongosh bump in tests

### DIFF
--- a/packages/build/src/npm-packages/bump.spec.ts
+++ b/packages/build/src/npm-packages/bump.spec.ts
@@ -10,12 +10,16 @@ import path from 'path';
 import { PROJECT_ROOT } from './constants';
 
 describe('npm-packages bump', function () {
-  let fsReadFile: SinonStub;
   let fsWriteFile: SinonStub;
+  const shellApiSrc = path.join(
+    PROJECT_ROOT,
+    'packages',
+    'shell-api',
+    'src',
+    'mongosh-version.ts'
+  );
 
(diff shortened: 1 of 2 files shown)


packages/build/src/npm-packages/bump.spec.ts
```diff
@@ -35,9 +46,112 @@ describe('npm-packages bump', function () {
...
+      });
+
+      expect(
+        JSON.parse(
+          fsWriteFile.withArgs('packages/autocomplete/package.json').firstCall
+            .args[1] as string
+        )
+      ).deep.equals({
+        name: '@mongosh/autocomplete',
+        version: '1.2.3',
```
addaleax, line 140: We do want to bump these packages, just not with the same version as mongosh, right?

Otherwise the published `@mongosh/cli-repl` package.json would specify a version that's not actually the same code as the one in the compiled executable?

gagik: yeah, this happens at the previous step with `bumpAuxiliaryPackages`

gagik: it'd be good to have a test mixing the two though yes

addaleax (approved):

PR #2359, opened 2025-02-05
chore(build): update build README.md for releases for new tag merge process

addaleax:


gagik (approved):
We can get this in and if there's a follow-up to automate it, that can just adjust the docs accordingly.

PR #2360, opened 2025-02-06
chore(snippet-manager): migrate from joi to zod MONGOSH-2010

addaleax:
This aligns us with the validators we've standardized on in Compass, improves TypeScript integration, and reduces executable size by 564 kB and startup time by 1.5% (locally on an M3).


gagik (approved):
Nice! Actually used zod before, it's great.

PR #2361, opened 2025-02-07
feat(cli-repl): add configuration to set max log files size MONGOSH-1985

gagik:
Requires https://github.com/mongodb-js/devtools-shared/pull/504 to be released.

Builds on top of https://github.com/mongodb-js/mongosh/pull/2354

### DIFF
--- a/packages/types/src/index.ts
+++ b/packages/types/src/index.ts
@@ -515,6 +515,7 @@ export class CliUserConfig extends SnippetShellUserConfig {
   logRetentionDays = 30;
   logMaxFileCount = 100;
   logCompressionEnabled = false;
+  logRetentionGB: number | undefined = undefined;
 }
 
 export class CliUserConfigValidator extends SnippetShellUserConfigValidator {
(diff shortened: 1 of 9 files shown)


packages/types/src/index.ts
```diff
@@ -539,6 +540,11 @@ export class CliUserConfigValidator extends SnippetShellUserConfigValidator {
           return `${key} must be a positive integer`;
         }
         return null;
+      case 'logRetentionGB':
+        if (value !== undefined && (typeof value !== 'number' || value < 0)) {
+          return `${key} must be a positive number or empty`;
```
addaleax, line 545:
```suggestion
          return `${key} must be a positive number or undefined`;
```

we also mention the specific value in other validators, and for users less familiar with JS, this might be a hint that `config.set('...', undefined)` is what they'd want to do if they want to unset this

gagik: should change this as well? https://github.com/mongodb-js/mongosh/blob/0aec4d594b518ed6325d506974e1c4d91458e4a6/packages/types/src/index.ts#L589-L592

gagik:
On the topic of unexpected green CI, https://github.com/mongodb-js/mongosh/actions/runs/13199134044/job/36846954054?pr=2361 seems not to actually run tests?

packages/types/src/index.ts
addaleax: That's actually talking about an empty string though, right?

addaleax: A bit, yes.

But we do use `undefined` to signify "default behavior", and for the `updateURL` example, the default behavior _is_ a non-empty string, so it does make sense to me that a user would explicitly have to unset it by specifying an empty string instead of setting it to `undefined`.

addaleax (approved):

PR #2363, opened 2025-02-10
feat(logging): print help for the global log property MONGOSH-1995

alenakhineika:
When running the ShellApi `help` command, print the description for the ShellLog methods.

### DIFF
--- a/packages/i18n/src/locales/en_US.ts
+++ b/packages/i18n/src/locales/en_US.ts
@@ -137,7 +137,6 @@ const translations: Catalog = {
           description: 'Shell log methods',
           link: '#',
           attributes: {
-            // TODO(MONGOSH-1995): Print help for the global log property.
             info: {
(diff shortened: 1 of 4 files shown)


packages/i18n/src/locales/en_US.ts
```diff
@@ -166,6 +165,10 @@ const translations: Catalog = {
           description: 'Shell Help',
           link: 'https://docs.mongodb.com/manual/reference/method',
           attributes: {
+            log: {
+              description:
+                "'log.info(<msg>)': Writes a custom info/warn/error/fatal/debug message to the log file",
```
addaleax, line 170:
```suggestion
                "'log.info(<msg>)': Write a custom info/warn/error/fatal/debug message to the log file",
```

The `show` entries also all use imperative

addaleax (approved):

gagik (approved):

PR #2365, opened 2025-02-11
feat(cli-repl): use mongosh_ prefix for logs in custom locations MONGOSH-2012

gagik:
This helps differentiate mongosh logs from others in case those files get shared.

### DIFF
--- a/packages/cli-repl/src/cli-repl.ts
+++ b/packages/cli-repl/src/cli-repl.ts
@@ -255,12 +255,11 @@ export class CliRepl implements MongoshIOProvider {
       throw new Error('Logging and telemetry not setup');
     }
 
-    const defaultDirectory = this.shellHomeDirectory.localPath('.');
-    const directory = (await this.getConfig('logLocation')) || defaultDirectory;
+    const customLogLocation = await this.getConfig('logLocation');
 
     this.logManager ??= new MongoLogManager({
-      directory,
-      prefix: directory !== defaultDirectory ? 'mongosh_' : undefined,
+      directory: customLogLocation || this.shellHomeDirectory.localPath('.'),
+      prefix: customLogLocation ? 'mongosh_' : undefined,
       retentionDays: await this.getConfig('logRetentionDays'),
       gzip: await this.getConfig('logCompressionEnabled'),
       maxLogFileCount: await this.getConfig('logMaxFileCount'),
--- a/packages/cli-repl/src/cli-repl.ts
+++ b/packages/cli-repl/src/cli-repl.ts
@@ -255,10 +255,12 @@ export class CliRepl implements MongoshIOProvider {
       throw new Error('Logging and telemetry not setup');
     }
 
+    const defaultDirectory = this.shellHomeDirectory.localPath('.');
+    const directory = (await this.getConfig('logLocation')) || defaultDirectory;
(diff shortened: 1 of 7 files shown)


packages/cli-repl/src/cli-repl.ts
```diff
@@ -255,10 +255,12 @@ export class CliRepl implements MongoshIOProvider {
...
 
+    const defaultDirectory = this.shellHomeDirectory.localPath('.');
+    const directory = (await this.getConfig('logLocation')) || defaultDirectory;
+
     this.logManager ??= new MongoLogManager({
-      directory:
-        (await this.getConfig('logLocation')) ||
-        this.shellHomeDirectory.localPath('.'),
+      directory,
+      prefix: directory !== defaultDirectory ? 'mongosh_' : undefined,
```
addaleax, line 263: Telling whether two directories are the same is not trivial, unfortunately. Comparing the strings can easily fail to account for path normalization (e.g. `/usr//bin` and `/usr/bin` being the same directory on UNIX, or `C:\foo` and `\\?\C:\foo` being the same on Windows). You could use `path.normalize()` on those paths, and that is probably the minimum we'd want to do here before comparing paths.

What it wouldn't account for is actual file system semantics (`path.[...]` only does in-memory computation on the string, possibly accounting for the current working directory). You _can_ do that via comparing `stat` results, e.g.

```js
async function sameFile(a, b) {
  if (a === b) return true;
  a = path.normalize(a), b = path.normalize(b);
  if (a === b) return true;
  const opt = { bigint: true };
  const [sa, sb] = await Promise.all([fs.stat(a, opt), fs.stat(b, opt)]);
  return sa.dev === sb.dev && sa.ino === sb.ino;
}
```

This comes at the cost of 2 file system calls, but otherwise is the most reliable way to identify whether two files or directories are really the same, accounting for symlinks, mount aliases, etc.

I know we're trying to be sensitive about startup time, but I think this would be acceptable here (in particular, it would only trigger when paths are actually different).

gagik: Hm, I see. I was already wondering do we really want to be handling the case where a user has specified a custom log directory which is the default log directory?
How often would that happen and even if so, would it really be an issue to have `mongosh_` prefix for logs there then? It's a bit weird but doesn't feel "breaking" in the sense that it'd require a user to use this newly introduced configuration.
I don't mind this otherwise, just not sure if it's worth directory calls for all actual custom location cases because of kind of weird user behavior.

addaleax: Yeah, I think this would be happening a lot, I just wanted to mention it for completeness.

gagik: by this, do you mean the file system calls or the case where someone would set the custom location to the default location manually?

should we not handling this case that way and just default to `mongosh_` if any not-undefined logLocation is set?

addaleax: I just meant that people would use `undefined` if they intentionally want to set the directory to the default one 🙂 

> should we not handling this case that way and just default to mongosh_ if any not-undefined logLocation is set?

Yeah, that might be an easier solution here 👍

addaleax (approved):

PR #2366, opened 2025-02-12
 chore(tests): use `logMaxFileCount` instead of a environment variable in tests MONGOSH-2009

gagik:
Previously we relied on `MONGOSH_TEST_ONLY_MAX_LOG_FILE_COUNT` for setting max log files. Now we can do so by defining a global config for tests.


testing/tests-globalconfig.conf
```diff
@@ -0,0 +1,3 @@
+# Default global configuration used by tests
+mongosh:
```
gagik, line 2: Not sure if this is the best way to go about it (also will see how CI reacts...)

alenakhineika: Maybe we can create a file during testing, write to it, and then delete it as a test suit cleanup. Similarly how we do here: https://github.com/mongodb-js/mongosh/blob/main/packages/e2e-tests/test/e2e.spec.ts#L1463-L1467

addaleax: Yeah, or you could just do `config.set(...)` in one shell and then start another one, we also do that in a couple of places (has the advantage of not requiring any explicit disk I/O)

gagik: my thinking was to make it consistent with everything that would otherwise be using the environment variable so at least in both `cli-repl` and `e2e-tests` which is why I thought to keep the actual settings being set in `setup-env`

PR #2369, opened 2025-02-13
feat(cli-repl): add history() command MONGOSH-2013

gagik:
To Do:
- [x] Potentially do the setup in a different place/way?
- [x] Tests

### DIFF
--- a/packages/cli-repl/src/cli-repl.ts
+++ b/packages/cli-repl/src/cli-repl.ts
@@ -54,10 +54,6 @@ import type {
   DevtoolsProxyOptions,
 } from '@mongodb-js/devtools-proxy-support';
 import { useOrCreateAgent } from '@mongodb-js/devtools-proxy-support';
-import type { TypeSignature } from '@mongosh/shell-api';
-import { signatures } from '@mongosh/shell-api';
-import util from 'util';
-import { formatOutput } from './format-output';
 
 /**
  * Connecting text key.
@@ -500,8 +496,6 @@ export class CliRepl implements MongoshIOProvider {
       });
     }
 
-    this.setupHistoryCommand();
-
     Editor.create({
       input: this.input,
       vscodeDir: this.shellHomeDirectory.rcPath('.vscode'),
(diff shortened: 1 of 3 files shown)


packages/cli-repl/src/cli-repl.ts
```diff
@@ -488,6 +492,8 @@ export class CliRepl implements MongoshIOProvider {
       });
     }
 
+    this.setupHistoryCommand();
```
addaleax, line 495: This should go into mongosh-repl, right? It's not concerned with external I/O

packages/cli-repl/src/cli-repl.ts
```diff
@@ -902,6 +908,47 @@ export class CliRepl implements MongoshIOProvider {
...
+              // eslint-disable-next-line @typescript-eslint/no-explicit-any
+              value: formattedHistory,
+            },
+            { colors: true, maxArrayLength: Infinity }
+          );
+        },
+      };
+    };
+
+    getHistory.isDirectShellCommand = true;
```
addaleax, line 935: Do we want this to be callable as just `history`? I don't see a major downside in making it a regular function like `history()`, and then you wouldn't need all this extra setup here either

gagik: an argument for `history` could be that it'd make it more intuitive coming from a Unix shell so maybe it's worth it?

packages/cli-repl/src/mongosh-repl.spec.ts
```diff
@@ -526,18 +527,69 @@ describe('MongoshNodeRepl', function () {
...
+                formatOutput(
+                  {
+                    value: history.slice(history.length - 10).reverse(),
+                  },
+                  { colors: false, maxArrayLength: Infinity }
+                )
+              );
+            });
+
+            it('works without ()', async function () {
```
gagik, line 576: one con of `history` is that one can't do `history().length` etc operations on it so maybe it's confusing to have both. I think keeping it JS-like and not implementing i.e. `history -10` might be the way to go

packages/cli-repl/src/mongosh-repl.ts
```diff
@@ -391,6 +394,31 @@ class MongoshNodeRepl implements EvaluationListener {
...
+      const formattedHistory =
+        // Remove the history call from the formatted history
+        replHistory.slice(1).reverse();
+
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      formattedHistory[util.inspect.custom as any] = (() => {
+        return formatOutput(
+          {
+            // Providing a copy of the history avoids a circular reference.
+            value: formattedHistory.concat(),
```
addaleax, line 410: Doesn't `.slice().reverse()` above already copy?

packages/cli-repl/src/mongosh-repl.ts
```diff
@@ -391,6 +394,31 @@ class MongoshNodeRepl implements EvaluationListener {
...
+          },
+          { colors: true, maxArrayLength: Infinity }
```
addaleax, line 412: Don't unconditionally set `colors: true` if you don't know if the output supports colors (or wants colors – disabling them can be an accessibility feature). We have `this.formatOutput()` on this class specifically to account for that.

All that being said, you'll likely want to follow best practices for Node.js custom inspect functions here and not use our `formatOutput()` at all and instead use the `inspectOptions` and `inspect` arguments that already get passed to the custom inspect function automatically.

packages/cli-repl/src/mongosh-repl.ts
gagik: it copies it for the sake of not mutating the original `history` but needs to be copied again

addaleax: Ah wait, this is because the `formattedHistory` value here has the custom inspect function?

I think I was misled by the comment here a bit, we're not avoiding a circular reference but rather infinite recursion, right?

packages/cli-repl/src/cli-repl.ts
GaurabAryal: For now the ability to see history and C+P historical commands sounds like a good enough value drive for customers, without having to support something complicated like `!n` in bash. For `history` vs `history()`, I don't have a strong feeling either way but `history` being intuitive from bash is a good argument for putting a little extra effort to supporting it.

gagik: I think `history` will still be fine as far as discoverability goes as they'd see that it's a function in response and then do `history()` and use it to its fullest potential that way

packages/cli-repl/src/mongosh-repl.ts
gagik: yeah, I thought of it as a circular reference since the inspect ends up (I'm assuming) looking into the given object's [util.inspect.custom] as well and then finding the original function again. But yeah it's meant to avoid infinite recursion either way, I'll reword that.

addaleax (approved):

PR #2370, opened 2025-02-14
feat(shell-api): throw for EJSON.serialize(cursor) MONGOSH-2002

alenakhineika:
Throw a custom exception when a user attempts to serialize a cursor, pointing them to the fact that they likely want to call `.toArray()` first.

### DIFF
--- a/packages/e2e-tests/test/e2e.spec.ts
+++ b/packages/e2e-tests/test/e2e.spec.ts
@@ -737,13 +737,10 @@ describe('e2e', function () {
 
     it('throws the .toArray() suggestion when a user attempts to serialize a cursor', async function () {
       await shell.executeLine('const b = db.test.find()');
-      await shell.executeLine('b.b = b');
       await shell.executeLine('console.log(JSON.stringify(b));');
-      await eventually(() => {
-        shell.assertContainsOutput(
-          'Error serializing a cursor. Did you mean to call .toArray() first?'
-        );
-      });
+      shell.assertContainsOutput(
+        'Cannot serialize a cursor to JSON. Did you mean to call .toArray() first?'
+      );
     });
 
     it('expands explain output from aggregation indefinitely', async function () {
--- a/packages/e2e-tests/test/e2e.spec.ts
+++ b/packages/e2e-tests/test/e2e.spec.ts
@@ -735,6 +735,17 @@ describe('e2e', function () {
       );
     });
 
+    it('throws the .toArray() suggestion when a user attempts to serialize a cursor', async function () {
+      await shell.executeLine('const b = db.test.find()');
+      await shell.executeLine('b.b = b');
+      await shell.executeLine('console.log(JSON.stringify(b));');
+      await eventually(() => {
+        shell.assertContainsOutput(
+          'Error serializing a cursor. Did you mean to call .toArray() first?'
+        );
+      });
+    });
+
     it('expands explain output from aggregation indefinitely', async function () {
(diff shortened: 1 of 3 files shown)


packages/shell-api/src/cursor.ts
```diff
@@ -34,6 +34,21 @@ export default class Cursor extends AggregateOrFindCursor<ServiceProviderFindCur
...
+   * Throw a custom exception when a user attempts to serialize a cursor,
+   * pointing to the fact that .toArray() needs to be called first.
+   *
+   * @param {CursorFlag} flag - The cursor flag.
+   *
+   * @returns {void}
+   */
+  toJSON(): void {
+    throw new MongoshInvalidInputError(
+      'Error serializing a cursor. Did you mean to call .toArray() first?',
```
addaleax, line 47: This sounds like there was a previous error, which isn't the case

```suggestion
      'Cannot serialize a cursor to JSON. Did you mean to call .toArray() first?',
```

packages/shell-api/src/cursor.spec.ts
```diff
@@ -793,6 +793,29 @@ describe('Cursor', function () {
...
+      let shellApiCursor: Cursor;
+
+      beforeEach(function () {
+        spCursor = stubInterface<ServiceProviderCursor>();
+        shellApiCursor = new Cursor(mongo, spCursor);
+      });
+
+      it('throws with the .toArray() suggestion', function () {
+        try {
+          shellApiCursor.toJSON();
```
addaleax, line 807: Should this use `JSON.stringify()`?

packages/e2e-tests/test/e2e.spec.ts
```diff
@@ -735,6 +735,17 @@ describe('e2e', function () {
       );
     });
 
+    it('throws the .toArray() suggestion when a user attempts to serialize a cursor', async function () {
+      await shell.executeLine('const b = db.test.find()');
+      await shell.executeLine('b.b = b');
```
addaleax, line 740: I don't quite understand this line here and what it does – the test should work just fine without it, right?

packages/e2e-tests/test/e2e.spec.ts
```diff
@@ -735,6 +735,17 @@ describe('e2e', function () {
...
+      await shell.executeLine('console.log(JSON.stringify(b));');
+      await eventually(() => {
```
addaleax, line 742: In this case, the return value of `executeLine()` should already contain this output, so there's no need to use a less exact method like `eventually()`

packages/shell-api/src/cursor.spec.ts
alenakhineika: Isn't it just the direct calling of the method in the unit test, the one that will be called using JSON.stringify? I have also e2e tests with JSON.stringify().

packages/e2e-tests/test/e2e.spec.ts
alenakhineika: Yeah, this is not needed here. I will remove it. Tried previously without a MongoDB document and with the plain object instead, like from the example in [NODE-6708](https://jira.mongodb.org/browse/NODE-6708).

packages/shell-api/src/cursor.spec.ts
addaleax: Well, it's generally preferable to have tests that verify externally visible behavior when there's no significant difference in complexity, right?

addaleax (approved):

PR #2374, opened 2025-02-19
refactor(build): restructure build helpers to use classes MONGOSH-2007

gagik:
A lot of the helper do dependency injection by using some of its arguments to optionally pass other (stubbed) helpers. This is getting out of hand and now we have things like:
 
```ts
export async function publishMongosh(
  config: Config,
  mongoshGithubRepo: GithubRepo,
  mongodbHomebrewForkGithubRepo: GithubRepo,
  homebrewCoreGithubRepo: GithubRepo,
  barque: Barque,
  createAndPublishDownloadCenterConfig: typeof createAndPublishDownloadCenterConfigFn,
  publishToNpm: typeof publishToNpmType,
  pushTags: typeof pushTagsType,
  writeBuildInfo: typeof writeBuildInfoType,
  publishToHomebrew: typeof publishToHomebrewType,
  shouldDoPublicRelease: typeof shouldDoPublicReleaseFn = shouldDoPublicReleaseFn,
  getEvergreenArtifactUrl: typeof getArtifactUrlFn = getArtifactUrlFn,
  bumpMongoshReleasePackages: typeof bumpMongoshReleasePackagesFn = bumpMongoshReleasePackagesFn,
  bumpAuxiliaryPackages: typeof bumpAuxiliaryPackagesFn = bumpAuxiliaryPackagesFn,
  spawnSync: typeof spawnSyncFn = spawnSyncFn 
...
```

To minimize this, this PR restructures these functions to be inside classes based on responsibility. 

There are no real functionality changes beyond combining some helpers into a single class. 
Some noteworthy changes:
1. `push-tags` => combined with `publish` in `PackagePublisher`
2. `markBumpedFilesAsAssumeUnchanged` was completely removed as it was unused. 

For the sake of keeping this not larger than it already is, some helper functions were left the same and only more core ones were changed.

I am not changing the file names here so diffs can be more readable but it'd be worth doing that in a follow-up PR.


packages/build/src/npm-packages/bump.spec.ts
```diff
@@ -19,9 +16,14 @@ describe('npm-packages bump', function () {
...
   );
 
+  let testBumper: PackageBumper;
+
   beforeEach(function () {
     fsWriteFile = sinon.stub(fs, 'writeFile');
     fsWriteFile.resolves();
+
+    testBumper = new PackageBumper();
+    sinon.stub(testBumper, 'spawnSync').resolves();
```
addaleax, line 26: Can we turn these 'stored as properties with the expectation to mock them later' methods into regular properties that get passed through the constructor?

The previous dependency injection code was primarily annoying imo because functions would take more and more arguments, but the general principle of explicitly passing in a dependency is still a good one

addaleax (approved):

PR #2377, opened 2025-02-20
fix(shell-api): create vector search index with description options MONGOSH-1851

alenakhineika:
Add support for calling `createSearchIndex` with a description to match the driver behavior: https://www.mongodb.com/docs/drivers/node/current/fundamentals/indexes/#create-a-search-index

```
createSearchIndex(definition)
createSearchIndex(name, definition)
createSearchIndex(name, type, definition)
createSearchIndex(description) // New.
```


lerouxb (approved):

packages/shell-api/src/collection.ts
```diff
@@ -2388,17 +2388,30 @@ export default class Collection extends ShellApiWithMongoClass {
   @apiVersions([])
   // TODO(MONGOSH-1471): use SearchIndexDescription once available
   async createSearchIndex(
-    indexName?: string | Document,
+    indexNameOrOptions?: string | Document,
```
gagik, line 2391: could we define a stricter type here instead of `Document`?

alenakhineika: Well, we can make it `string | SearchIndexDescription  | Document`, where:

```
interface SearchIndexDescription {
  definition: Document;
  name?: string;
  type?: string;
}
```

But `definition` will still have the `Document` type. Not sure if this brings a lot of value here.

gagik: I think it's still good to have even as purely documentation

alenakhineika: I can add it if this helps to read the code 👍

addaleax: Yes, I think that's a great idea – and maybe it's worth adding [TS function overloads](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads) to this method just to describe what the actual valid call signatures are 🙂

packages/shell-api/src/collection.ts
```diff
@@ -2383,34 +2385,69 @@ export default class Collection extends ShellApiWithMongoClass {
...
+    type?: 'search' | 'vectorSearch',
+    definition?: SearchIndexDefinition
+  ): Promise<string>;
+  async createSearchIndex(
+    definition?: SearchIndexDefinition,
+    type?: 'search' | 'vectorSearch'
+  ): Promise<string>;
+  async createSearchIndex(
+    description?: SearchIndexDescription
+  ): Promise<string>;
```
addaleax, line 2403: I don't think any of the arguments should be optional here, except type in the third one?

packages/shell-api/src/collection.ts
```diff
@@ -2383,34 +2385,69 @@ export default class Collection extends ShellApiWithMongoClass {
...
-      indexName = undefined;
+
+    if (
+      typeof nameOrOptions === 'object' &&
+      nameOrOptions !== null &&
+      nameOrOptions.definition
+    ) {
+      indexName = nameOrOptions.name;
+      indexType = nameOrOptions.type;
+      definition = nameOrOptions.definition;
```
addaleax, line 2428: If the user provides a SearchIndexDescription, we should just pass that value through instead of decomposing it and then re-composing it into a SearchIndexDescription

packages/shell-api/src/collection.ts
alenakhineika: Then we would need two more overloads:
```
async createSearchIndex(): Promise<string>;
async createSearchIndex(name: string): Promise<string>;
```
Does it make sense to add those? I feel like with the current state it is pretty descriptive.

addaleax: Yes, exactly, those aren't valid overloads so they shouldn't be allowed 👍 As mentioned on Slack, there are four variants we want to/can support:

```
createSearchIndex(definition)
createSearchIndex(name, definition)
createSearchIndex(name, type, definition)
createSearchIndex(description)
```

A definition or description is always necessary.

alenakhineika: Hmm we have a test tho that claims that just a name is allowed: https://github.com/mongodb-js/mongosh/blob/3456adced8738f135befb3fbc122e1f018bb52e2/packages/shell-api/src/collection.spec.ts#L2584-L2596

addaleax: That test doesn't have anything to do with this function, though

alenakhineika: Wrong lines. Here it is: https://github.com/mongodb-js/mongosh/blob/3456adced8738f135befb3fbc122e1f018bb52e2/packages/shell-api/src/collection.spec.ts#L2648-L2658

addaleax: Yeah, `definition: {}` is not a valid definition and will be rejected by the server, this isn't a supported use case

addaleax (approved):

gagik (approved):

PR #2379, opened 2025-02-20
chore(ci): add automatic merging of the release tag MONGOSH-2018

gagik:
The new system requires us to manually merge into main. This automates it through a workflow.

It can also be triggered manually with a workflow dispatch.

### DIFF
--- a/.github/workflows/merge-release-tag.yml
+++ b/.github/workflows/merge-release-tag.yml
@@ -22,26 +22,11 @@ jobs:
 
       - uses: actions/checkout@v4
         with:
-          ref: ${{ github.ref }}
+          ref: main
           fetch-depth: "0"
           token: ${{ steps.app-token.outputs.token }}
 
-      - name: Extract version from tag
-        uses: actions-ecosystem/action-regex-match@d50fd2e7a37d0e617aea3d7ada663bd56862b9cc
-        id: version-match
-        with:
-          text: ${{ github.ref }}
-          regex: 'refs/tags/mongosh@([0-9]+\.[0-9]+\.[0-9]+.*)'
-
-      - uses: actions/github-script@v7
-        if: steps.version-match.outputs.group1 == ''
-        with:
-          script: |
-            core.setFailed('Could not extract version from tag: ${{ github.ref }}');
-
       - name: Merge release tag into main
         run: |
-          git checkout -b package-release/v${{ steps.version-match.outputs.group1 }} ${{ github.ref }}
-          git checkout main
-          git merge package-release/v${{ steps.version-match.outputs.group1 }}
+          git merge ${{ github.ref }} -m "chore(release): merge changes from ${{ github.ref_name }}"
           git push origin main
--- a/.github/workflows/merge-release-tag.yml
+++ b/.github/workflows/merge-release-tag.yml
@@ -3,7 +3,8 @@ name: Merge Release Tag
 on:
   push:
     tags:
-      - 'mongosh@[0-9]+.[0-9]+.[0-9]+*'
+      - 'mongosh@[0-9]+.[0-9]+.[0-9]+'
+      - 'mongosh@[0-9]+.[0-9]+.[0-9]-rc.[0-9]+'
   workflow_dispatch:
 
 permissions:
--- a/.github/workflows/merge-release-tag.yml
+++ b/.github/workflows/merge-release-tag.yml
@@ -1,0 +1,54 @@
+name: Merge Release Tag
+
+on:
+  push:
+    tags:
+      - 'mongosh@[0-9]+.[0-9]+.[0-9]+'
+  workflow_dispatch:
+
+jobs:
+  merge-release-tag:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: mongodb-js/devtools-shared/actions/setup-bot-token@main
+        id: app-token
+        with:
+          app-id: ${{ vars.DEVTOOLS_BOT_APP_ID }}
+          private-key: ${{ secrets.DEVTOOLS_BOT_PRIVATE_KEY }}
+
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.ref }}
+          fetch-depth: "0"
+          token: ${{ steps.app-token.outputs.token }}
+
+      - name: Extract version from tag
+        id: version
+        run: echo "RELEASE_VERSION=$(echo ${GITHUB_REF} | sed -n 's/refs\/tags\/mongosh@\([0-9]\+\.[0-9]\+\.[0-9]\+\)/\1/p')" >> $GITHUB_ENV
+
+      - name: Validate release version
+        shell: bash
+        run: |
+          if [ -z "${RELEASE_VERSION}" ]; then
+            echo "RELEASE_TAG is not set or is empty"
+            exit 1
+          fi
+
+          if git rev-parse "release/v$RELEASE_VERSION" >/dev/null 2>&1; then
+            echo "Error: Branch release/v$RELEASE_TAG already exists"
+            echo "If this version has already been released consider using a different one."
+            exit 1
+          fi
+
+      - name: Create and push branch from tag
+        run: |
+          git checkout -b release/v$RELEASE_VERSION ${{ github.ref }}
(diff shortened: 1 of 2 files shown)


.github/workflows/merge-release-tag.yml
```diff
@@ -0,0 +1,54 @@
+name: Merge Release Tag
+
+on:
+  push:
+    tags:
+      - 'mongosh@[0-9]+.[0-9]+.[0-9]+'
```
nirinchev, line 6: This assumes we don't do prereleases - is this a safe assumption to make?

.github/workflows/merge-release-tag.yml
```diff
@@ -0,0 +1,54 @@
...
+
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.ref }}
+          fetch-depth: "0"
+          token: ${{ steps.app-token.outputs.token }}
+
+      - name: Extract version from tag
+        id: version
+        run: echo "RELEASE_VERSION=$(echo ${GITHUB_REF} | sed -n 's/refs\/tags\/mongosh@\([0-9]\+\.[0-9]\+\.[0-9]\+\)/\1/p')" >> $GITHUB_ENV
```
nirinchev, line 28: Consider using a regex matching action here instead of sed. Alternatively, we probably want to store the release version as output rather than add it as an environment variable.

.github/workflows/merge-release-tag.yml
```diff
@@ -0,0 +1,54 @@
...
+          if git rev-parse "release/v$RELEASE_VERSION" >/dev/null 2>&1; then
+            echo "Error: Branch release/v$RELEASE_TAG already exists"
+            echo "If this version has already been released consider using a different one."
+            exit 1
+          fi
+
+      - name: Create and push branch from tag
+        run: |
+          git checkout -b release/v$RELEASE_VERSION ${{ github.ref }}
+          git push origin release/v$RELEASE_VERSION
```
nirinchev, line 47: Why are we pushing this branch only to delete it later?

.github/workflows/merge-release-tag.yml
gagik: I'll make it a bit more permissive just in case

.github/workflows/merge-release-tag.yml
```diff
@@ -0,0 +1,57 @@
+name: Merge Release Tag
+
+on:
+  push:
+    tags:
+      - 'mongosh@[0-9]+.[0-9]+.*'
```
nirinchev, line 6: We can still enforce the 3 components:

```suggestion
      - 'mongosh@[0-9]+\.[0-9]+\.[0-9]+.*'
```

Though I'm not sure how strict we have to be here - I don't really expect we'll be pushing malformed `mongosh@...` tags

.github/workflows/merge-release-tag.yml
```diff
@@ -0,0 +1,57 @@
...
+          if [[ -z "${{ steps.version-match.outputs.group1 }}" ]]; then
+            echo "Error: Could not extract version from tag"
+            exit 1
+          fi
+
+          if git rev-parse "release/v${{ steps.version-match.outputs.group1 }}" >/dev/null 2>&1; then
+            echo "Error: Branch release/v$RELEASE_TAG already exists"
+            echo "If this version has already been released consider using a different one."
+            exit 1
+          fi
```
nirinchev, line 49: This seems redundant now, since we're not pushing that branch, there's no way for it to exist - is there?

.github/workflows/merge-release-tag.yml
```diff
@@ -0,0 +1,57 @@
...
+          text: ${{ github.ref }}
+          regex: 'refs/tags/mongosh@([0-9]+\.[0-9]+\.*)'
+
+      - name: Validate release version
+        shell: bash
+        id: validate
+        run: |
+          if [[ -z "${{ steps.version-match.outputs.group1 }}" ]]; then
+            echo "Error: Could not extract version from tag"
+            exit 1
```
nirinchev, line 42: I don't think this will show up in the check failure description, which is not a huge problem, but we can make it nicer - e.g. using the [`github-script`](https://github.com/actions/github-script) action:

```yaml
- uses: actions/github-script@v7
  if: steps.version-match.outputs.group1 == ''
  with:
    script: |
      core.setFailed('Could not extract version from tag: ${{ github.ref }}');
```

.github/workflows/merge-release-tag.yml
gagik: not sure escaping is right, this isn't actually regex: https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

.github/workflows/merge-release-tag.yml
```diff
@@ -0,0 +1,57 @@
...
+          fi
+
+          if git rev-parse "release/v${{ steps.version-match.outputs.group1 }}" >/dev/null 2>&1; then
+            echo "Error: Branch release/v$RELEASE_TAG already exists"
```
addaleax, line 46: `release/` is actually reserved as a prefix for releases we kick off from other release lines (e.g. if we wanted to do a 1.x.x release now, we'd do it from a `release/v1.x.x` branch) – maybe call this `package-release/` or something similar?

.github/workflows/merge-release-tag.yml
```diff
@@ -0,0 +1,47 @@
...
+    steps:
+      - uses: mongodb-js/devtools-shared/actions/setup-bot-token@main
+        id: app-token
+        with:
+          app-id: ${{ vars.DEVTOOLS_BOT_APP_ID }}
+          private-key: ${{ secrets.DEVTOOLS_BOT_PRIVATE_KEY }}
+
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.ref }}
```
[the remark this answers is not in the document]
gagik: I was under impression that tag merging wasn't possible for some reason but yeah makes sense.

.github/workflows/merge-release-tag.yml
```diff
@@ -0,0 +1,32 @@
...
+
+      - uses: actions/checkout@v4
+        with:
+          ref: main
+          fetch-depth: "0"
+          token: ${{ steps.app-token.outputs.token }}
+
+      - name: Merge release tag into main
+        run: |
+          git merge ${{ github.ref }} -m "chore(release): merge changes from ${{ github.ref_name }}"
```
gagik, line 31: also added a message seems otherwise it may either prompt us to do it or use the default which maybe isn't best?

.github/workflows/merge-release-tag.yml
```diff
@@ -0,0 +1,32 @@
+name: Merge Release Tag
+
+on:
+  push:
+    tags:
+      - 'mongosh@[0-9]+.[0-9]+.[0-9]+*'
```
addaleax, line 6:
```suggestion
      - 'mongosh@[0-9]+.[0-9]+.[0-9]+'
```

I assume this is a regexp, in which case `+*` would be equivalent to just `*`

addaleax (approved):

.github/workflows/merge-release-tag.yml
addaleax: Oh, and, as Nikola suggested above, shouldn't it be

```suggestion
      - 'mongosh@[0-9]+\.[0-9]+\.[0-9]+'
```

then?

gagik: It's GitHub regex-esque pattern: https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet
I think the idea is to allow `+rc0` though if it's just this we can also do
```yaml
      - 'mongosh@[0-9]+.[0-9]+.[0-9]+'
      - 'mongosh@[0-9]+.[0-9]+.[0-9]-rc[0-9]+'
```

gagik: I am not sure how escapes would work but GitHub Docs themselves don't use them and use i.e `v[12].[0-9]+.[0-9]+`

gagik:
Going to go ahead with merge and can follow up if other changes are needed.

PR #2381, opened 2025-02-21
fix: default to server-defined numInitialChunks MONGOSH-1949

nirinchev:
This removes our default of 1000 for numInitialChunks in `shardAndDistributeCollection`. Additionally, it fixes what I believe was a bug where the `collation` option was ignored when calling `reshardCollection`.

### DIFF
--- a/packages/shell-api/src/shard.ts
+++ b/packages/shell-api/src/shard.ts
@@ -710,16 +710,24 @@ export default class Shard extends ShellApiWithMongoClass {
       options,
     });
     await this.shardCollection(ns, key, unique, options);
-    // SERVER-92762: Prevent unequal data distribution by setting
-    // numInitialChunks to 1000.
-    const numInitialChunks =
-      typeof unique === 'object'
-        ? unique.numInitialChunks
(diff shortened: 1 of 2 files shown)


packages/shell-api/src/shard.ts
```diff
@@ -710,16 +710,24 @@ export default class Shard extends ShellApiWithMongoClass {
...
+    const reshardOptions: Document = {
       forceRedistribution: true,
-    });
+    };
+
+    if (options?.numInitialChunks !== undefined) {
+      reshardOptions.numInitialChunks = options.numInitialChunks;
+    }
+
+    if (options?.collation !== undefined) {
```
nirinchev, line 726: I'm not 100% for this one, but reading the [docs for reshardCollection](https://www.mongodb.com/docs/manual/reference/command/reshardCollection/), it appears that there are cases where `collation` must be included. We would already be passing it correctly to `shardCollection`, but were previously ignoring it for `reshardCollection`, which would have resulted in the latter failing.

addaleax (approved):

PR #2382, opened 2025-02-21
fix: set a consistent key for output elements rather than using the array index MONGOSH-2021

lerouxb:
By default we have maxOutputLines == 1000. Once you get to 1000 elements it starts to splice the output to only keep the last 1000. But then the same array index maps to a different output entry component. Which means all of them get re-rendered at once.

By adding a persistent key to each shell output entry react won't lose track of or think that the components completely changed.

As a reminder, sync-to-compass is your friend for testing this locally:

```
~/mongo/mongosh/packages/browser-repl/scripts % export COMPASS_HOME=/Users/leroux.bodenstein/mongo/compass
~/mongo/mongosh/packages/browser-repl/scripts % node sync-to-compass.js
```

### DIFF
--- a/packages/browser-repl/src/components/shell.tsx
+++ b/packages/browser-repl/src/components/shell.tsx
@@ -357,12 +357,14 @@ const _Shell: ForwardRefRenderFunction<EditorRef | null, ShellProps> = (
         runtime.setEvaluationListener(listener);
         const result = await runtime.evaluate(code);
         outputLine = {
+          key: lastKey++,
           format: 'output',
           type: result.type,
           value: result.printable,
         };
       } catch (error) {
         outputLine = {
+          key: lastKey++,
           format: 'error',
           value: error,
         };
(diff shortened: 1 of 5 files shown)


addaleax (approved):

packages/browser-repl/src/components/shell.tsx
```diff
@@ -262,6 +264,7 @@ const _Shell: ForwardRefRenderFunction<EditorRef | null, ShellProps> = (
           ...(outputRef.current ?? []),
           ...result.map(
             (entry): ShellOutputEntry => ({
+              key: lastKey++,
```
addaleax, line 267: Also needs to be accounted for on L362/L396, right? Maybe these are easier to catch if we don't make this an optional property?

lerouxb: yeah I plan on making it required - that was just temporary so I didn't have to change who knows how many places to get things to build and test it out.

lerouxb: Actually I think I overreacted. I saw a bunch of tabs in vscode go red and thought it would touch way more places, but it was just a few.

addaleax (approved):
:rocket:

PR #2386, opened 2025-02-25
chore(deps): update patch for bracketed paste on Windows MONGOSH-1999

addaleax:
Update the patch to the commit merged into libuv and add a small Node.js source modification that will need to go into an upcoming Node.js release in order for the patch to take effect.

Refs: https://github.com/libuv/libuv/commit/843b64faf5fbbd19c04475bafa38c0c91514efb8

### DIFF
--- a/scripts/nodejs-patches/006-windows-virtual-terminal-input-libuv-4688.patch
+++ b/scripts/nodejs-patches/006-windows-virtual-terminal-input-libuv-4688.patch
@@ -1,5 +1,116 @@
+diff --git a/deps/uv/docs/src/tty.rst b/deps/uv/docs/src/tty.rst
+index f1acfdc13729..f3c94390ef6d 100644
+--- a/deps/uv/docs/src/tty.rst
++++ b/deps/uv/docs/src/tty.rst
+@@ -27,10 +27,15 @@ Data types
+       typedef enum {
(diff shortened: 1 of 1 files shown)


scripts/nodejs-patches/006-windows-virtual-terminal-input-libuv-4688.patch
```diff
@@ -12,35 +123,307 @@ index 7e1f15544b17..37bc2c4ace57 100644
...
+   TTYWrap* wrap;
+   ASSIGN_OR_RETURN_UNWRAP(
+       &wrap, args.This(), args.GetReturnValue().Set(UV_EBADF));
+-  int err = uv_tty_set_mode(&wrap->handle_, args[0]->IsTrue());
++  int err = uv_tty_set_mode(
++      &wrap->handle_,
++      args[0]->IsTrue() ? UV_TTY_MODE_RAW_VT : UV_TTY_MODE_NORMAL);
+   args.GetReturnValue().Set(err);
+ }
+
```
addaleax, line 429: We'll likely need to do a separate PR to Node.js after the libuv changes are included there for this part here

gagik (approved):
Not an area I have a great grasp of but nothing stands out

PR #2387, opened 2025-02-25
fix(build): github api calls for homebrew release fail MONGOSH-1928

alenakhineika:
When sending requests via Octokit, a situation can arise where the server closes the connection. However, the client still believes it’s open and attempts to write to it, which leads to receiving an `EPIPE` error from the OS, indicating the connection has already been closed. In such cases, retrying the request can help establish a new, functional connection.


packages/build/src/homebrew/generate-formula.ts
```diff
@@ -1,15 +1,33 @@
...
+ */
+async function getFileWithRetry(
+  homebrewCore: GithubRepo,
+  remainingRetries = 3
+) {
+  try {
+    return await homebrewCore.getFileContent('Formula/m/mongosh.rb', 'master');
+  } catch (error: any) {
+    if (error.message.includes('EPIPE') && remainingRetries > 0) {
+      return await getFileWithRetry(homebrewCore, remainingRetries - 1);
```
addaleax, line 18: Just a small suggestion, but I'd add `console.error(error)` in this branch so that it's a bit clearer whether there were any retries

packages/build/src/homebrew/generate-formula.ts
```diff
@@ -1,15 +1,33 @@
 import * as semver from 'semver';
 import type { GithubRepo } from '@mongodb-js/devtools-github-repo';
 
+/**
+ * When sending requests via Octokit, a situation can arise where the server closes the connection,
+ * but the client still believes it’s open and attempts to write to it,
+ * what leads to receiving an EPIPE error from the OS, indicating the connection has already been closed.
+ * In such cases, retrying the request can help establish a new, functional connection.
+ */
+async function getFileWithRetry(
```
addaleax, line 10: Since this function is currently specific to fetching the formula file from the homebrew core repo, let's make the name also more specific to describe that use case

packages/build/src/homebrew/generate-formula.ts
```diff
@@ -1,15 +1,37 @@
...
+async function getFormulaFromRepositoryWithRetry(
+  homebrewCore: GithubRepo,
+  remainingRetries = 3
+) {
+  try {
+    return await homebrewCore.getFileContent('Formula/m/mongosh.rb', 'master');
+  } catch (error: any) {
+    if (error.message.includes('EPIPE') && remainingRetries > 0) {
+      console.error(error);
+      return await getFormulaFromRepositoryWithRetry(
```
gagik, line 19: would it be beneficial to sleep here for i.e. 100 milliseconds or whatever before retrying?

addaleax (approved):

gagik (approved):

PR #2390, opened 2025-02-27
chore(ci): don't use gcc on macOS MONGOSH-2029

nirinchev:
gcc is not officially supported on macOS (per [Build Platform Support Tiers]( https://docs.google.com/document/d/1G7JioblpjLdi8WiJP2A7OaZfNBF7yZ6-yXSCEarU2Sg/edit?tab=t.0#heading=h.4itnffmo0xv8)) and would typically not be installed, which is why the scripts have worked previously. It appears that there was a rogue task that installed the linux toolchain on those macs and is currently being removed. Still, I think the changes in this PR make sense as a way to protect us against future incidents.

### DIFF
--- a/.evergreen/install-node-source.sh
+++ b/.evergreen/install-node-source.sh
@@ -18,7 +18,13 @@ fi
 export PATH="$TOOLCHAIN_PATH:${ORIGINAL_PATH}"
 
 
-if [ `uname` != Darwin ]; then
+if [ `uname` = Darwin ]; then
+    echo "Using clang version:"
+    (which clang && clang --version)
+
+    echo "Using clang++ version:"
+    (which clang++ && clang++ --version)
+else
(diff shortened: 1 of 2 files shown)


.evergreen/install-node-source.sh
```diff
@@ -15,11 +15,19 @@ else
...
+
+if [ `uname` != Darwin ]; then
+    export CC=gcc
+    export CXX=g++
+
+    echo "Using gcc version:"
+    (which gcc && gcc --version)
+
+    echo "Using g++ version:"
+    (which g++ && g++ --version)
```
addaleax, line 29: Can we still print the compiler versions in the darwin case?

addaleax (approved):

addaleax (approved):

PR #2391, opened 2025-03-03
chore(ci): don't download chrome in bump-auxiliary-packages

nirinchev:
This is a minor change that sets the `PUPPETEER_SKIP_DOWNLOAD` environment variable in `bump-auxiliary-packages.yml` as we don't really need chrome there. Additionally, it adds a description for the workflow and removes some redundant committing.

### DIFF
--- a/.github/workflows/bump-auxiliary-packages.yml
+++ b/.github/workflows/bump-auxiliary-packages.yml
@@ -5,6 +5,14 @@ on:
 permissions:
   contents: none # We use the github app to checkout and create PR
 
+env:
+  PUPPETEER_SKIP_DOWNLOAD: "true"
+
+description: |
+  This workflow increases the versions of the auxiliary (i.e. non-mongosh) packages and creates a PR
+  as a first step for a release. Use this workflow when you don't want to do a full mongosh release,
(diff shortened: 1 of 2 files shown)


.github/workflows/bump-auxiliary-packages.yml
```diff
@@ -41,8 +49,6 @@ jobs:
       - name: Bump packages
         run: |
           npm run bump-auxiliary
-          git add .
-          git commit --no-allow-empty -m "chore(release): bump packages for auxiliary release" || true
```
nirinchev, line 45: peter-evans/create-pull-request already commits all changes, so there's no reason for us to manually do it.

.github/workflows/bump-auxiliary-packages.yml
```diff
@@ -5,6 +5,14 @@ on:
 permissions:
   contents: none # We use the github app to checkout and create PR
 
+env:
+  PUPPETEER_SKIP_DOWNLOAD: "true"
```
gagik, line 9: oh nice, same can be applied to `publish-auxiliary-packages.yml`

gagik (approved):
LGTM, can probably be extended to the publish workflow also

gagik (approved):

PR #2393, opened 2025-03-04
feat(deps): bump driver to 6.14.2, libmongocrypt to 1.13.0 MONGOSH-2037 MONGOSH-2038

addaleax:
Includes support for $lookup (e2e test TBD until COMPASS-8816, will likely need to wait for 8.1 release) and an OIDC connection fix


package-lock.json
```diff
@@ -23897,6 +23898,12 @@
         "node": ">=10"
       }
     },
+    "node_modules/prebuild-install/node_modules/napi-build-utils": {
```
nirinchev, line 23901: Not a huge deal, but it looks like we now have duplication of `prebuild-install` and `napi-build-utils` because of the different versions being requested by kerberos and mongodb-client-encryption. Might be a good idea to dedupe them if possible.

nirinchev (approved):

package-lock.json
nirinchev: As long as it's a pinky promise, we can let this one slide 😆

gagik (approved):

PR #2394, opened 2025-03-04
feat: update sh.xxxBalancing to use configureCollectionBalancing MONGOSH-1932

nirinchev:
This updates `sh.(enable|disable)Balancing` to use the `configureCollectionBalancing` admin command on 8.1+ servers.

### DIFF
--- a/packages/shell-api/src/shard.ts
+++ b/packages/shell-api/src/shard.ts
@@ -492,34 +492,52 @@ export default class Shard extends ShellApiWithMongoClass {
     });
   }
 
-  @returnsPromise
-  @apiVersions([])
-  async enableBalancing(ns: string): Promise<UpdateResult> {
-    assertArgsDefinedType([ns], ['string'], 'Shard.enableBalancing');
-    this._emitShardApiCall('enableBalancing', { ns });
+  private async updateBalancing(
+    ns: string,
+    enabled: boolean
+  ): Promise<Document> {
+    const apiCall = enabled ? 'enableBalancing' : 'disableBalancing';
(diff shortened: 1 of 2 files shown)


packages/shell-api/src/shard.ts
```diff
@@ -492,34 +492,52 @@ export default class Shard extends ShellApiWithMongoClass {
...
+    const apiCall = enabled ? 'enableBalancing' : 'disableBalancing';
+    assertArgsDefinedType([ns], ['string'], `Shard.${apiCall}`);
+    this._emitShardApiCall(apiCall, { ns });
+
+    const serverVersion = (
+      this._instanceState.cachedConnectionInfo() ??
+      (await this._instanceState.fetchConnectionInfo())
+    )?.extraInfo?.server_version;
+
+    if (serverVersion && semver.gte(serverVersion, '8.1.0-alpha3')) {
```
addaleax, line 508: Did we get an answer from the server team regarding how they'd prefer for us to do feature detection?

I think if we were to rely on a server version check, it would still be better to use the wire version – the only downside is that it's not granular enough to distinguish between "8.0 without a backport of this feature" and "8.0 with a backport of this feature"

nirinchev: Hey, yeah, I wasn't sure if I understood you correctly about the wire version - originally I thought you meant the max supported wire protocol version, but then reached the same conclusion that we wouldn't know if we're connecting to 8.0.x that has the backport or not. Let's see what we come up with tomorrow and will update this check.

PR #2398, opened 2025-03-05
chore(releasing): use github app token when cloning the repo for releasing MONGOSH-2049

nirinchev:


addaleax (approved):

nirinchev:
This was a draft because I wanted to check if there's any other places where we need write access and couldn't find any. @gagik since you recently overhauled a bunch of the release process, let me know if I may have missed something

gagik (approved):
Seems reasonable

PR #2400, opened 2025-03-06
feat(shell-api): add sh.isConfigShardEnabled and listShards helpers MONGOSH-1919

nirinchev:
This adds the `sh.isConfigShardEnabled` helper for server versions 8.0 and above. It uses `listShards` and then looks for a shard with `_id: 'config'`.

Additionally, it adds the `sh.listShards` helper that calls the `listShards` admin command.

### DIFF
--- a/packages/shell-api/src/shard.spec.ts
+++ b/packages/shell-api/src/shard.spec.ts
@@ -2108,6 +2108,114 @@ describe('Shard', function () {
         expect(caughtError).to.equal(expectedError);
       });
     });
+
+    describe('isConfigShardEnabled', function () {
+      this.beforeEach(function () {
+        serviceProvider.runCommandWithCheck.resolves({
+          ok: 1,
+          msg: 'isdbgrid',
+        });
+      });
+
+      it('calls serviceProvider.runCommandWithCheck', async function () {
+        await shard.isConfigShardEnabled();
+
+        expect(serviceProvider.runCommandWithCheck).to.have.been.calledWith(
+          ADMIN_DB,
+          {
+            listShards: 1,
+          }
+        );
+      });
+
+      it("returns false when listShards doesn't contain config shard", async function () {
+        serviceProvider.runCommandWithCheck.resolves({
+          ok: 1,
+          msg: 'isdbgrid',
+          shards: [
+            {
+              _id: 'shard1',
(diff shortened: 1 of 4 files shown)


packages/shell-api/src/shard.ts
```diff
@@ -811,4 +811,26 @@ export default class Shard extends ShellApiWithMongoClass {
...
+  @serverVersions(['8.0.0', ServerVersions.latest])
+  @apiVersions([])
+  @returnsPromise
+  async isConfigShardEnabled(): Promise<Document> {
+    this._emitShardApiCall('isConfigShardEnabled', {});
+    await getConfigDB(this._database);
+
+    const shards = (await this._database.adminCommand({ listShards: 1 }))[
+      'shards'
+    ] as Document[] | undefined;
```
addaleax, line 824: Surprised the linter doesn't complain, we'd probably want roughly this, right?

```suggestion
    const shards = (await this._database.adminCommand({ listShards: 1 })).shards as Document[] | undefined;
```

addaleax (approved):
Looks good! 

Makes me wonder though, do we also want to expose a generic `sh.listShards()` helper that forwards to `adminCommand({ listShards: 1 })`? That feels like something that should be there in some form, no?

nirinchev:
ping @addaleax - I know you approved the PR, but I did end up adding a `listShards` command, so figured it's a "big enough" commit, that you might want to take another look 👀

packages/shell-api/src/shard.spec.ts
```diff
@@ -2408,11 +2583,11 @@ describe('Shard', function () {
         expect(
           (await sh.removeShardFromZone(`${shardId}-1`, 'zone1')).ok
         ).to.equal(1);
-        expect((await sh.status()).value.shards[1].tags).to.deep.equal([]);
+        expect((await sh.listShards())[1].tags).to.deep.equal([]);
         expect((await sh.removeShardTag(`${shardId}-0`, 'zone0')).ok).to.equal(
           1
         );
-        expect((await sh.status()).value.shards[0].tags).to.deep.equal([]);
+        expect((await sh.listShards())[0].tags).to.deep.equal([]);
```
addaleax, line 2590: 💙

addaleax (approved):

PR #2401, opened 2025-03-06
chore(ci): adopt own toolchain MONGOSH-864

addaleax:


.evergreen/setup-env.sh
```diff
@@ -26,38 +19,47 @@ fi
...
 
     echo "Using clang++ version:"
     (which clang++ && clang++ --version)
+
+    export NVM_DIR="$BASEDIR/.nvm"
+    echo "Setting NVM environment home: $NVM_DIR"
+    [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
+    set +x # nvm is very verbose
+    echo nvm use $NODE_JS_VERSION || nvm use 20.11.1
+    nvm use $NODE_JS_VERSION || nvm use 20.11.1 # see install-node.sh
```
gagik, line 34:
```suggestion
    # Node.js 20.11.1 is the last 20.x that builds out of the box on RHEL7
    # https://github.com/nodejs/node/issues/52223
    nvm use $NODE_JS_VERSION || nvm use 20.11.1
```
(since we remove this comment otherwise; though not sure how relevant it still is)

gagik (approved):
Nice, will be great to have the quicker CI 🚀

.evergreen/setup-env.sh
addaleax: Yeah, great catch, we can actually remove the fallback here altogether since we only use nvm on macOS now: 745915d65a028ad5dc7a2e6c4714af88e400639a

PR #2403, opened 2025-03-10
chore(ci): correctly pass args to mongodb-runner MONGOSH-2011

nirinchev:
On windows with node 22/23, when running mongodb-runner with npx, the `--` will be applied to npx command rather than the exec, so we need another set of dashes to make yargs happy.


nirinchev:
Yeah, it seems like this is only happening on Windows with node 22/23. I didn't test with bash (only pwsh) since I figured this is a good enough solution.

PR #2407, opened 2025-03-13
chore(releasing): validate tarball sha1 before publishing to homebrew MONGOSH-2059

nirinchev:
This is related to MONGOSH-2059 where we published a homebrew formula with the incorrect sha. There doesn't seem to be anything wrong with the way we're doing these sha updates, so my best guess is that it's a case of incomplete/corrupted download from npm that caused us to compute the wrong sha256. This PR changes it so we use the `shasum` field from the npm API to validate the tarball integrity before publishing the formula to homebrew, as well as adds a retry mechanism to attempt to recover in case we download the wrong thing.

### DIFF
--- a/packages/build/src/homebrew/utils.ts
+++ b/packages/build/src/homebrew/utils.ts
@@ -1,13 +1,68 @@
 import crypto from 'crypto';
 import https from 'https';
 
-export function httpsSha256(url: string): Promise<string> {
-  return new Promise((resolve, reject) => {
+export async function npmPackageSha256(
+  packageUrl: string,
+  httpGetFn: typeof httpGet = httpGet
+): Promise<string> {
+  const json = await httpGetFn(packageUrl, 'json');
+  const tarballUrl = json.dist.tarball;
+  const shasum = json.dist.shasum;
+
+  const tarball = await getTarballWithRetries(tarballUrl, shasum, httpGetFn);
+  const hash = crypto.createHash('sha256');
+  hash.update(tarball);
+  return hash.digest('hex');
+}
+
+async function getTarballWithRetries(
+  url: string,
+  shasum: string,
+  httpGetFn: typeof httpGet,
+  attempts = 3
+): Promise<Buffer> {
+  try {
+    const tarball = await httpGetFn(url, 'binary');
+    const hash = crypto.createHash('sha1').update(tarball).digest('hex');
+    if (hash !== shasum) {
+      throw new Error(`shasum mismatch: expected '${shasum}', got '${hash}'`);
+    }
+
+    return tarball;
+  } catch (err) {
+    if (attempts === 0) {
+      throw err;
+    }
+
+    return getTarballWithRetries(url, shasum, httpGetFn, attempts - 1);
+  }
+}
+
+export function httpGet(url: string, response: 'json'): Promise<any>;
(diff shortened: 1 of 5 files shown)


lerouxb (approved):

packages/build/src/homebrew/utils.ts
```diff
@@ -1,13 +1,68 @@
...
+  }
+}
+
+export function httpGet(url: string, response: 'json'): Promise<any>;
+export function httpGet(url: string, response: 'binary'): Promise<Buffer>;
+
+export async function httpGet(
+  url: string,
+  responseType: 'json' | 'binary'
+): Promise<any | Buffer> {
```
addaleax, line 47: You'd need to define overloads if you actually want this return type to be interpreted as anything other than `Promise<any>` by TS.

packages/build/src/homebrew/utils.ts
```diff
@@ -1,13 +1,68 @@
...
+  const response = await new Promise<string | Buffer[]>((resolve, reject) => {
     https.get(url, (stream) => {
-      const hash = crypto.createHash('sha256');
+      let data: string | Buffer[] = responseType === 'json' ? '' : [];
       stream.on('error', (err) => reject(err));
-      stream.on('data', (chunk) => hash.update(chunk));
-      stream.on('end', () => resolve(hash.digest('hex')));
+      stream.on('data', (chunk) => {
+        if (typeof data === 'string') {
+          data += chunk;
```
addaleax, line 54: You'll want to call `stream.setEncoding('utf8');` in this case in order to properly handle multibyte characters.

packages/build/src/homebrew/utils.ts
nirinchev: Hm... that one has overloads on line 41 and 42 - or am I misunderstanding what you mean here?

packages/build/src/homebrew/utils.ts
```diff
@@ -1,13 +1,72 @@
...
+  }
+}
+
+export function httpGet(url: string, response: 'json'): Promise<any>;
+export function httpGet(url: string, response: 'binary'): Promise<Buffer>;
+
+export async function httpGet(
+  url: string,
+  responseType: 'json' | 'binary'
+): Promise<any | Buffer> {
```
gagik, line 47:
```suggestion
export function httpGet<T>(url: string, response: 'json'): Promise<T>;
export function httpGet(url: string, response: 'binary'): Promise<Buffer>;
export async function httpGet<T>(
  url: string,
  responseType: 'json' | 'binary'
): Promise<T | Buffer> {
```

packages/build/src/homebrew/utils.ts
```diff
@@ -1,13 +1,72 @@
 import crypto from 'crypto';
 import https from 'https';
 
-export function httpsSha256(url: string): Promise<string> {
-  return new Promise((resolve, reject) => {
+export async function npmPackageSha256(
+  packageUrl: string,
+  httpGetFn: typeof httpGet = httpGet
+): Promise<string> {
+  const json = await httpGetFn(packageUrl, 'json');
```
gagik, line 8:
```suggestion
  const json = await httpGetFn<{
  dist: {
	tarball: string;
	shasum: string;
	}
  }>(packageUrl, 'json');
```

gagik (approved):

PR #2411, opened 2025-03-19
feat(telemetry): add device ID logging COMPASS-8443

gagik:
To standardize identification across DevTools and Atlas CLI, this introduces a `native-machine-id` dependency which uses the same base system calls for determining the device Id as https://github.com/denisbrodbeck/machineid that is used for the device ID by the Atlas CLI.

Both libraries use hashing to protect the machine-specific information.

We could using `anonymousId` as `deviceId` as done in the Atlas CLI but the concerns are: a) this transition may break many existing user associations we have and b) if for whatever reason device ID cannot be determined on a given OS, we'd lose `anonymousId` altogether. Therefore this instead adds a new identity field for Segment.

We're using our own `native-machine-id` library for this as opposed to `node-machine-id` as:
- The node library isn't actively being maintained.
- Depends on spawning child processes that call OS-specific functions which can be troublesome.
- In case of nicher OS environments which may not provide the system calls it relies on, its behavior may be unpredictable.

**To Do:**
- [x] Verify the hashed device ID is the same as the Atlas Device ID.

### DIFF
--- a/packages/connectivity-tests/test/all.sh
+++ b/packages/connectivity-tests/test/all.sh
@@ -20,7 +20,7 @@ fi
 git clone https://github.com/mongodb-js/devtools-docker-test-envs.git test-envs
 cd test-envs
 
-git checkout ac8e6675fcd769f0bfc868c3ac73cdd8bcdcc792
+git checkout cbec3a65efe9def50638994ae6b36ffa3b4708ee
 
 "$CONNECTIVITY_TEST_SOURCE_DIR/ldap.sh"
 "$CONNECTIVITY_TEST_SOURCE_DIR/localhost.sh"


packages/logging/src/logging-and-telemetry.ts
```diff
@@ -52,10 +52,23 @@ import type {
...
 export function setupLoggingAndTelemetry(
   props: MongoshLoggingAndTelemetryArguments
 ): MongoshLoggingAndTelemetry {
+  if (!props.deviceId) {
+    try {
+      props.deviceId = machineIdSync();
+    } catch (error) {
+      props.bus.emit(
+        'mongosh:error',
+        new Error('Failed to get device ID'),
```
gagik, line 66: is it safe for us to pass the error to our telemetry? In Atlas CLI they do https://github.com/mongodb/mongodb-atlas-cli/blob/4c36ccfeecc95f208b7d276126c70cf16c5cde58/internal/telemetry/event.go#L156 but seems risky to me considering this potentially has traces with OS-level call output or whatnot.

packages/logging/src/logging-and-telemetry.ts
```diff
@@ -52,10 +52,23 @@ import type {
   MongoshLoggingAndTelemetryArguments,
   MongoshTrackingProperties,
 } from './types';
+import { machineIdSync } from 'node-machine-id';
 
 export function setupLoggingAndTelemetry(
   props: MongoshLoggingAndTelemetryArguments
 ): MongoshLoggingAndTelemetry {
+  if (!props.deviceId) {
```
gagik, line 60: Seemed more logical for me for the device Id determination to exist here rather than in the class itself as then its state is immutable

packages/logging/src/logging-and-telemetry.ts
```diff
@@ -52,10 +52,23 @@ import type {
...
+    try {
+      props.deviceId = machineIdSync();
```
addaleax, line 62: Let's do our very best to avoid synchronous I/O calls in the mongosh startup path, if we can at all? We already have a mechanism for deferring telemetry in place here, maybe we could use that with the async version of this function?

packages/logging/src/logging-and-telemetry.ts
addaleax: You mean, the original `error`? Yeah I'd include that at least for logging ... maybe we could make it the `cause` of this new error object, and then add logging for the cause in the `mongosh:error` listener (but not analytics)?

packages/logging/package.json
```diff
@@ -22,7 +22,8 @@
     "@mongosh/history": "2.4.6",
     "@mongosh/types": "3.6.0",
     "mongodb-log-writer": "^2.3.1",
-    "mongodb-redact": "^1.1.5"
+    "mongodb-redact": "^1.1.5",
+    "node-machine-id": "^1.1.12"
```
addaleax, line 26: This is an unmaintained package which always runs multiple child processes to get its result, I do wonder if there are better alternatives out there, even if it's just hashing `os.hostname()` and maybe the mac addresses reported in `os.networkInterfaces()`?

gagik: yeah I agree, I wasn't sure about the package though it is used by some pretty big projects and with no good alternatives. If we go in a different direction we'd likely want the Atlas CLI to adopt the same as well but considering the library talks about lack of reliability of Mac addresses etc., seem like it won't be that reliable.

I'm wondering if we could instead do a shared mongodb devtools like `~/.mongodb-tools` directory or something where we write a new UUID for anonymousId and let mongosh, Compass, and Atlas CLI read and use this if it exists already. Seems like hoping all tools can reliably com up with the same device identifier is more unreliable anyhow.

gagik: haha just realized the "big projects which use this" includes https://www.npmjs.com/package/realm?activeTab=dependencies though it seems like it's just used in an example project 😅

addaleax: > I'm wondering if we could instead do a shared mongodb devtools like `~/.mongodb-tools/config.yaml` directory or something where we write a new UUID for anonymousId and let mongosh, Compass, and Atlas CLI read and use this if it exists already.

If that is an option, we should definitely go with it. I had assumed we'd want a true "device" ID, not something that would be per-user, but in today's world the latter is probably at least as good as the former. (It also has the advantage of the user being able to control it, if necessary.)

> Seems like hoping all tools can reliably come up with the same device identifier is more unreliable anyhow.

Agreed 👍

.depalignrc.json
```diff
@@ -20,6 +20,10 @@
     "@typescript-eslint/parser": [
       "^5.59.0",
       "^4.33.0"
+    ],
+    "node-fetch": [
```
gagik, line 24: I think `depalign` was silently failing for a while; I ended up fixing it but the actual issue was the lack of sinon devdependency with `depcheck`

gagik:
@addaleax I ended up sticking with device ID for reasons I mentioned in **Alternative Considerations** in the PR description, mainly easier adaptability from existing Atlas CLI data. 

I also re-organized the logic so now we have 2 parallel buffers for bus events in general as well as telemetry events in particular until the device ID is resolved. Let me know if you have any thoughts about either of that.

packages/logging/src/logging-and-telemetry.ts
```diff
@@ -82,30 +103,72 @@ class LoggingAndTelemetry implements MongoshLoggingAndTelemetry {
...
+
+    this.resolveDeviceId('unknown');
+  }
+
+  private async setupTelemetry(): Promise<void> {
+    if (!this.deviceId) {
+      try {
+        this.deviceId ??= await Promise.race([
+          getDeviceId(),
+          new Promise<string>((resolve) => {
```
gagik, line 161: This may not be necessary I am just not sure if there's an issue with leaving the `getDeviceId()` "running" while we're flushing the events; does this get killed? Don't want to end up delaying the shell exit because of this.

packages/cli-repl/src/cli-repl.ts
```diff
@@ -1206,11 +1207,14 @@ export class CliRepl implements MongoshIOProvider {
    * @param code The user-provided exit code, if any.
    */
   async exit(code?: number): Promise<never> {
+    this.loggingAndTelemetry?.flush();
```
gagik, line 1210: I think we should be doing this generally in any case to make sure all telemetry (including the error afterwards?) get reported even before this change

packages/logging/src/logging-and-telemetry.ts
```diff
@@ -82,30 +103,72 @@ class LoggingAndTelemetry implements MongoshLoggingAndTelemetry {
...
+            this.resolveDeviceId = resolve;
+          }),
+        ]);
+      } catch (error) {
+        this.bus.emit('mongosh:error', error as Error, 'telemetry');
```
gagik, line 166: do we want to actually log this error? Seems like something we'd like to track ourselves but not necessarily show to users as it's not going to affect anything other than telemetry

addaleax (commented):
So ... I don't want to slow down work that I know we want to just get done¹, but I do still feel pretty uneasy about using node-machine-id, the fact that we're spawning multiple extra child processes on each startup of mongosh just feels fairly wrong, from a performance¹ and security perspective, even if we don't see an immediate critical issue with it.

Can we try to align even more closely with the Atlas CLI and use the same approach, which is to essentially perform the lookups in native/compiled code? I know that that comes with a bit of overhead, but it's far from infeasible (we've done that for other smaller things as well, like https://github.com/mongodb-js/glibc-version) and I'm happy to help with getting that off the ground.

¹ Telemetry is still turned off for mongosh, so it's probably not _super_ time-sensitive?
² Startup performance has been a big pain point in the past for us and even though the perf tests in CI seem okay here, this change feels a bit like we'd be pushing it 😞

packages/logging/src/logging-and-telemetry.ts
gagik: ah yeah right, I was under the false impression that logging in this case would also be `stdout`-ing the output, nevermind then, agreed

gagik:
@addaleax Sounds good, I can look into ways we could put this into the larger plan and see what we could come up with. Sorry, might have been meaningful to have more discussion about this earlier, I just figured after talking with Atlas CLI folk that we'd likely end up in some form of machine ID-powered setup anyways unless we'd like to push them to adopt the alternative instead. And potential of having to deal with directory permissions or whatnot seemed like a good enough argument against the shared directory idea.

But yeah about `node-machine-id` and performance concerns, that makes a lot of sense. The native lookup does sound worthwhile and hopefully not much effort from our end (honestly surprised this is something that doesn't exist already). I'll follow-up about that and regarding expected timeline.

addaleax:
> But yeah about `node-machine-id` and performance concerns, that makes a lot of sense. The native lookup does sound worthwhile and hopefully not much effort from our end (honestly surprised this is something that doesn't exist already). I'll follow-up about that and regarding expected timeline.

Yeah, I'm also happy to support this in any way I can, overall I'd expect it to be somewhat straightforward to put together

packages/logging/src/logging-and-telemetry.ts
lerouxb: I think it makes sense. Like you said probably not strictly necessary, but nice to have. Personally I'd leave a comment in this code with this reasoning for our future selves, but consider that a nit.

packages/logging/src/logging-and-telemetry.ts
```diff
@@ -284,10 +395,8 @@ class LoggingAndTelemetry implements MongoshLoggingAndTelemetry {
         }
         this.busEventState.telemetryAnonymousId =
           newTelemetryUserIdentity.anonymousId;
-        this.analytics.identify({
-          anonymousId: newTelemetryUserIdentity.anonymousId,
-          traits: getUserTraits(),
-        });
+
+        void identify();
```
lerouxb, line 399: Total nit, but there's no promise you're not awaiting, so the `void` pattern is probably not necessary?

packages/logging/package.json
```diff
@@ -22,15 +22,17 @@
     "@mongosh/history": "2.4.6",
     "@mongosh/types": "3.6.0",
     "mongodb-log-writer": "^2.3.1",
-    "mongodb-redact": "^1.1.5"
+    "mongodb-redact": "^1.1.5",
+    "native-machine-id": "^0.1.1"
```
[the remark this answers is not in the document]
gagik: this is actually our native addon, I ended up using a more generic name to make it feel less internal in case others want to adopt it. This seems to have caused more confusion that needed generally so I'll probaby stick to `@mongodb-js` for the future.

packages/build/src/compile/signable-compiler.ts
```diff
@@ -132,6 +132,10 @@ export class SignableCompiler {
       path: await findModulePath('cli-repl', 'glibc-version'),
       requireRegexp: /\bglibc_version\.node$/,
     };
+    const nativeMachineIdAddon = {
+      path: await findModulePath('logging', 'native-machine-id'),
+      requireRegexp: /\bnative_machine_id\.node$/,
+    };
```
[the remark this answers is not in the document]
gagik: the compilation of the artifacts were failing before so it does work on that front but I'll also add a test to make sure some value is resolved 👍

packages/logging/src/logging-and-telemetry.ts
```diff
@@ -52,6 +53,40 @@ import type {
   MongoshLoggingAndTelemetryArguments,
   MongoshTrackingProperties,
 } from './types';
+import { createHmac } from 'crypto';
+
+/**
+ * @returns A hashed, unique identifier for the running device or `"unknown"` if not known.
+ */
+export async function getDeviceId({
```
gagik, line 61: This could go inside the class but seems clearer to make it independent of it and avoid tying it to it to potentially even expose it on its own if we have some need for that in the future

addaleax (approved):

PR #2416, opened 2025-03-25
feat(shell-api): add Mongo bulkWrite API MONGOSH-1100

gagik:
New `bulkWrite(...)` command for Mongo instances.

[As stated for the driver](https://github.com/mongodb/node-mongodb-native/releases/tag/v6.10.0), this API is meant to replace the existing bulk write API on the Collection as it supports a bulk write across multiple databases and collections in a single call.
The command uses the active mongosh client and replicates the API from the driver in form of 
```ts
Mongo().bulkWrite([{
  namespace: string | Collection /** '<db>.<collection>' */,
  name: ‘<operation>’,
		  /* Other operation-specific props */
}, ...]);
```


packages/shell-api/src/mongo.ts
```diff
@@ -365,6 +368,18 @@ export default class Mongo extends ShellApiClass {
     return await this._listDatabases(options);
   }
 
+  @returnsPromise
+  @serverVersions(['8.0.0', ServerVersions.latest])
+  @apiVersions([1])
+  bulkWrite(
+    models: AnyClientBulkWriteModel<Document>[],
+    options: ClientBulkWriteOptions = {}
+  ): Promise<ClientBulkWriteResult> {
```
addaleax, line 377: We've typically wrapped result types in our own types so that we're a bit more independent when it comes to server changes, see `result.ts` – I'd consider doing that here as well if we can (this mostly already matches our existing` BulkWriteResult` as well?)

packages/shell-api/src/mongo.ts
```diff
@@ -62,6 +64,7 @@ import { ShellApiErrors } from './error-codes';
 import type { LogEntry } from './log-entry';
 import { parseAnyLogEntry } from './log-entry';
 import type { ShellBson } from './shell-bson';
+import type { ClientBulkWriteOptions } from 'mongodb';
```
addaleax, line 67: No imports from `mongodb` in `shell-api`, the dependency on that package is only for use in the integration tests of shell-api

If you need driver types, they should be exported from `service-provider-core`. The fact that we're just re-exporting the driver types there is honestly not _great_, but it's just pretty practical for now

addaleax (approved):

addaleax (approved):
Looks great! :shipit:

PR #2421, opened 2025-03-28
chore(release): write to new release s3 bucket MONGOSH-2124

drichmdb:
This commit adjusts the mongosh release process to dual write to our new
release S3 bucket. Instead of using long-lived static credentials, we have
moved over to an IAM role and are assuming it via Evergreen's ec2.assume_role
and in GitHub Actions via configure-aws-credentials.


lerouxb (approved):
Looks good. I guess we should remember to check after the next release that the assets are also in the correct new location?

PR #2422, opened 2025-03-31
feat(shell-api): add automerge information to sh.status() MONGOSH-1649

gagik:
We expose a status for the automerge being enabled/disabled so we'd want it to be shown in `sh.status()`

This does that in form:
```
automerge: 
    Currently enabled: yes
```

### DIFF
--- a/packages/shell-api/src/helpers.ts
+++ b/packages/shell-api/src/helpers.ts
@@ -328,6 +328,14 @@ export async function getPrintableShardStatus(
       };
     })(),
     (async (): Promise<void> => {
+      // Is automerge currently enabled, available since >= 7.0
+      const automerge = await settingsColl.findOne({ _id: 'automerge' });
+      result.automerge = {
+        'Currently enabled':
+          automerge === null || automerge.enabled ? 'yes' : 'no',
+      };
+    })(),
+    (async (): Promise<void> => {
       // Is the balancer currently enabled
       const balancerEnabled = await settingsColl.findOne({ _id: 'balancer' });
       balancerRes['Currently enabled'] =
@@ -711,6 +719,10 @@ export type ShardingStatusResult = {
             };
       }[];
   autosplit: {
+    'Currently enabled': 'yes' | 'no';
+  };
+  /** Available from 7.0.0 */
+  automerge: {
     'Currently enabled': 'yes' | 'no';
   };
   balancer: {
--- a/packages/shell-api/src/shard.spec.ts
+++ b/packages/shell-api/src/shard.spec.ts
@@ -2419,6 +2419,7 @@ describe('Shard', function () {
           'shardingVersion',
           'shards',
           'autosplit',
+          'automerge',
           'balancer',
           'databases',
         ]);
@@ -2463,6 +2464,7 @@ describe('Shard', function () {
             'shardingVersion',
             'shards',
             'autosplit',
+            'automerge',
             'balancer',
             'databases',
           ]);
--- a/packages/shell-api/src/helpers.spec.ts
+++ b/packages/shell-api/src/helpers.spec.ts
@@ -224,6 +224,7 @@ describe('getPrintableShardStatus', function () {
     );
     expect(status['most recently active mongoses']).to.have.lengthOf(1);
     expect(status.autosplit['Currently enabled']).to.equal('yes');
+    expect(status.automerge['Currently enabled']).to.equal('yes');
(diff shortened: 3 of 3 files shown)


packages/shell-api/src/helpers.ts
```diff
@@ -327,6 +327,14 @@ export async function getPrintableShardStatus(
           autosplit === null || autosplit.enabled ? 'yes' : 'no',
       };
     })(),
+    (async (): Promise<void> => {
+      // Is automerge currently enabled, available since >= 7.0
```
gagik, line 331: So this is available since 7.0 which means in older versions it'd also similarly say `'yes'` which is weird.
The solution to this is to check for the server version and compare with semver but I don't see this being used anywhere
1. Do we want to introduce this kind of differentiation and hide this field < 7.0?
2. What does a server version mean for automerge in sharded cluster? If I understand right it's possible for shards to have i.e. mix of 6.0 and 7.0 Mongo versions, does one get it from i.e. adminDB or something of sort?

packages/shell-api/src/helpers.ts
```diff
@@ -713,6 +721,10 @@ export type ShardingStatusResult = {
   autosplit: {
     'Currently enabled': 'yes' | 'no';
   };
+  /** Available from 7.0.0 */
+  automerge: {
+    'Currently enabled': 'yes' | 'no';
```
gagik, line 726: Related: is there a documented way of disabling this? Closest is https://www.mongodb.com/docs/manual/reference/method/sh.disableAutoMerger/ but this is collection-specific so not relevant for this.

packages/shell-api/src/helpers.ts
addaleax: Just wondering – what does `automerge === null` correspond to? Is that the case in which we would want to not print this information?

packages/shell-api/src/helpers.ts
addaleax: Does https://www.mongodb.com/docs/manual/reference/method/sh.stopAutoMerger/#mongodb-method-sh.stopAutoMerger do the right thing?

packages/shell-api/src/helpers.ts
```diff
@@ -327,6 +327,14 @@ export async function getPrintableShardStatus(
           autosplit === null || autosplit.enabled ? 'yes' : 'no',
       };
     })(),
+    (async (): Promise<void> => {
+      // Is automerge currently enabled, available since >= 7.0
+      const automerge = await settingsColl.findOne({ _id: 'automerge' });
```
addaleax, line 332: Not a problem to be solved in this PR, but the fact that we're doing 4 parallel `fineOne()`s on the same collection ... that's very how-our-users-use-MongoDB but not very how-users-should-use-MongoDB 😂

packages/shell-api/src/helpers.ts
gagik: Yeah just confirmed this more or less with `start/`stop` `autoMerger` is when that'd appear.

We could omit it as information which isn't explicitly set and show only if it has been configured; users looking for it would probably end up starting/stopping so seems reasonable.

packages/shell-api/src/helpers.ts
gagik: oh yes! reasonable naming was just a bit of confusing naming when I was looking at it with the context of `disableAutoSplit` which does change things globally

packages/shell-api/src/helpers.ts
gagik: truly committing to 'think like your user' 😄

packages/shell-api/src/shard.spec.ts
```diff
@@ -2516,6 +2518,26 @@ describe('Shard', function () {
         );
       });
     });
+    describe('automerge', function () {
+      it('not shown if sh.status() if not explicitly enabled', async function () {
+        expect((await sh.status()).value.automerge).is.undefined;
```
gagik, line 2523: this works but does kind of mess with the `sh.status` integrity a little as we don't really hide fields beyond that. 
I'm unsure how problematic would splitting by version be but that seems to be the only other options (or setting the status to `'unknown'` but that is also a bit weird and would be shown to a lot of users by default

addaleax: > If we went that route, the version would probably be derived from the admin DB right?

The server version? The only reliable version that clients are supposed to make decisions on is the wire version reported by `hello`

addaleax (approved):

PR #2424, opened 2025-04-01
feat: initial feature-flagged, new autocompleter MONGOSH-2036

lerouxb:
I filed a few TODOs as follow-up tickets.

```
~/mongo/mongosh % USE_NEW_AUTOCOMPLETE=true npm run start -- 'your connection string here'
```


packages/shell-api/src/collection.ts
```diff
@@ -2497,6 +2498,29 @@ export default class Collection extends ShellApiWithMongoClass {
       definition
     );
   }
+
+  async _getSampleDocs(): Promise<Document[]> {
+    this._cachedSampleDocs = await (await this.aggregate([])).toArray();
```
lerouxb, line 2503: yes I have to actually set a limit :P

packages/e2e-tests/test/e2e-snippet.spec.ts
```diff
@@ -72,6 +72,7 @@ describe('snippet integration tests', function () {
     shell.assertNoErrors();
   });
 
+  // TODO: port to the new autocomplete
```
lerouxb, line 75: I'll either do this in this PR or I'll file a follow-up ticket

packages/browser-runtime-core/src/open-context-runtime.ts
```diff
@@ -29,6 +29,7 @@ export interface InterpreterEnvironment {
  */
 export class OpenContextRuntime implements Runtime {
   private interpreterEnvironment: InterpreterEnvironment;
+  // TODO: we have to also port this to the new autocomplete
```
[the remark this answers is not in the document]
lerouxb: (I might even do it in this PR)

packages/cli-repl/src/mongosh-repl.spec.ts
```diff
@@ -352,6 +363,14 @@ describe('MongoshNodeRepl', function () {
     };
     const tabtab = async () => {
       await tab();
+      if (process.env.USE_NEW_AUTOCOMPLETE) {
+        // TODO: This is because autocomplete() is async and will either list
```
lerouxb, line 367: Less of a TODO than a NOTE.

packages/cli-repl/src/mongosh-repl.ts
```diff
@@ -131,6 +133,53 @@ type Mutable<T> = {
...
+     still autocomplete global variables and functions if you type part of the
+     name.
+  2. Don't filter out exact matches (where filter === name) so that we match the
+     behaviour of the node completer.
+  3. Make sure the name starts with the trigger, otherwise it will return every
+     possible property/name that's available at that level. ie. all the "peer"
+     properties of the things that match.
+  */
+  //console.log(name, kind);
+  // TODO: This can be improved further if we first see if there are any
```
lerouxb, line 161: Filed a ticket to just improve the filtering: https://jira.mongodb.org/browse/MONGOSH-2207

packages/cli-repl/src/mongosh-repl.spec.ts
addaleax: Yeah, I feel like setting a fixed timeout of 210 is either almost guaranteed flakiness or something that at least looks like it would be flaky ... can we somehow wait for the right event (e.g. some autocompletion data fteched event) at this point?

packages/shell-api/src/shell-instance-state.ts
```diff
@@ -135,6 +140,14 @@ export interface ShellPlugin {
 // eslint-disable-next-line no-control-regex
 const CONTROL_CHAR_REGEXP = /[\x00-\x1F\x7F-\x9F]/g;
 
+function connectionIdFromURI(uri: string): string {
+  // turn the uri into something we can safely use as part of a "filename"
+  // inside autocomplete
+  const hash = crypto.createHash('sha256');
+  hash.update(uri);
+  return hash.digest('hex');
+}
```
addaleax, line 149: This feels like it miiiight be worth caching on the `Mongo` instance, not sure if that would be premature optimization but while hashing is cheap, it might not be do-a-few-times-per-autocomplete cheap?

packages/shell-api/src/shell-instance-state.ts
```diff
@@ -402,6 +415,80 @@ export default class ShellInstanceState {
...
+      }
+    }
+    throw new Error(`mongo with connection id ${connectionId} not found`);
+  }
+
+  public getAutocompletionContext(): AutocompletionContext {
+    return {
+      currentDatabaseAndConnection: () => {
+        return {
+          connectionId: connectionIdFromURI(this.currentDb.getMongo().getURI()),
```
addaleax, line 431: Is this intended to be the ID that the autocompleter passes to things like `connectionId`? Would it be simpler to just assign an increasing counter id to each `Mongo` instance?

packages/cli-repl/src/mongosh-repl.spec.ts
lerouxb: I updated the existing autocomplete event and added a new helper around that so we can wait until the event fires with results.

packages/cli-repl/src/mongosh-repl.ts
```diff
@@ -423,17 +436,96 @@ class MongoshNodeRepl implements EvaluationListener {
...
+        connectionId: string
+      ): Promise<string[]> => {
+        const mongo = instanceState.getMongoByConnectionId(connectionId);
+        try {
+          const dbNames = await mongo._getDatabaseNamesForCompletion();
+          return dbNames.filter(
+            (name: string) => !CONTROL_CHAR_REGEXP.test(name)
+          );
+        } catch (err: any) {
+          // TODO: move this code to a method in the shell instance so we don't
```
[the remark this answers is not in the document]
addaleax: (Except probably for the `analyzeDocuments()` call, which should maybe live inside the autocompleter?)

packages/shell-api/src/collection.ts
```diff
@@ -2528,6 +2528,31 @@ export class Collection<
       definition
     );
   }
+
+  async _getSampleDocs(): Promise<Document[]> {
+    this._cachedSampleDocs = await (
+      await this.aggregate([{ $sample: { size: 10 } }])
```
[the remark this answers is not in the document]
lerouxb: Yeah sorry I meant to leave myself some TODO there to check up on the latest "best practise" from compass' schema tab then forgot about it. I definitely mean to still do that.

packages/cli-repl/src/mongosh-repl.ts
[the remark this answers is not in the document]
lerouxb: (We can obviously still extract this code that takes a sample and turns it into a schema elsewhere.)

packages/shell-api/src/collection.ts
```diff
@@ -2389,7 +2389,7 @@ export class Collection<
   ): Promise<RunCommandCursor> {
     this._emitCollectionApiCall('checkMetadataConsistency', { options });
 
-    return this._database._runCursorCommand({
+    return await this._database._runCursorCommand({
```
lerouxb, line 2392: drive-by to silence eslint

PR #2425, opened 2025-04-01
feat(service-provider-core): support new BSON vector types in mongosh MONGOSH-1873

gagik:
Add friendlier support for displaying BSON binary vector types in mongosh by displaying them as reproducable Binary constructor calls. 
Screenshot 2025-04-03 at 1 19 53 AM

### DIFF
--- a/packages/service-provider-core/src/printable-bson.spec.ts
+++ b/packages/service-provider-core/src/printable-bson.spec.ts
@@ -95,6 +95,34 @@ describe('BSON printers', function () {
     ).to.equal("MD5('0123456789abcdef0123456789abcdef')");
   });
 
+  describe('with Vector types', function () {
+    it('formats Int8Array correctly', function () {
+      expect(
+        inspect(bson.Binary.fromInt8Array(new Int8Array([1, 2, 3])))
+      ).to.equal('Binary.fromInt8Array(new Int8Array([ 1, 2, 3 ]))');
+    });
+
+    it('formats PackedBits correctly', function () {
+      expect(
+        inspect(bson.Binary.fromPackedBits(new Uint8Array([1, 2, 3])))
+      ).to.equal('Binary.fromPackedBits(new Uint8Array([ 1, 2, 3 ]))');
+    });
+
+    it('formats Float32Array correctly', function () {
+      expect(
+        inspect(
+          bson.Binary.fromFloat32Array(
+            new Float32Array([1.1111, 2.2222, 3.3333])
+          ),
+          { compact: true }
+        )
+      ).matches(
+        // Precision is lost because of float handling, so we use regex to match
+        /Binary.fromFloat32Array\(new Float32Array\(\[ 1\.1\d*, 2\.2\d*, 3\.3\d* \]\)\)/
+      );
+    });
+  });
+
   it('formats any other value with the new format using createfromBase64', function () {
     expect(
       inspect(bson.Binary.createFromBase64('SGVsbG8sIFdvcmxkIQo='))
--- a/packages/service-provider-core/src/printable-bson.ts
+++ b/packages/service-provider-core/src/printable-bson.ts
@@ -1,4 +1,5 @@
 import { bson as BSON } from './bson-export';
+import { inspect as utilInspect } from 'util';
 const inspectCustom = Symbol.for('nodejs.util.inspect.custom');
 type BSONClassKey = (typeof BSON)[Exclude<
   keyof typeof BSON,
@@ -19,7 +20,7 @@ function makeClasslessInspect<K extends BSONClassKey>(className: K) {
     this: (typeof BSON)[typeof className]['prototype'],
     ...args: any
   ) {
-    return removeNewFromInspectResult(originalInspect.apply(this, args as any));
+    return removeNewFromInspectResult(originalInspect.apply(this, args));
   };
 }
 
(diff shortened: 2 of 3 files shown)


packages/service-provider-core/src/printable-bson.ts
```diff
@@ -43,10 +44,38 @@ export const bsonStringifiers: Record<
...
+              ...args
+            )})`;
+          case BSON.Binary.VECTOR_TYPE.Float32:
+            return `Float32(${utilInspect(
+              Array.from(this.toFloat32Array()),
+              // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
+              ...args
+            )})`;
+          case BSON.Binary.VECTOR_TYPE.PackedBit:
+            return `PackedBit(${utilInspect(
```
gagik, line 67: Not too familiar with the usecase of PackedBit so not sure what representation bits best here.

packages/service-provider-core/src/printable-bson.ts
```diff
@@ -43,10 +44,38 @@ export const bsonStringifiers: Record<
...
+        switch (this.buffer[0]) {
+          case BSON.Binary.VECTOR_TYPE.Int8:
+            return `Int8(${utilInspect(
+              Array.from(this.toInt8Array()),
+              // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
+              ...args
+            )})`;
+          case BSON.Binary.VECTOR_TYPE.Float32:
+            return `Float32(${utilInspect(
+              Array.from(this.toFloat32Array()),
```
gagik, line 62: not sure if this is a problematically heavy operation here, `utilInspect`-ing directly causes it to display like a `{'0': 1... }` so not really what we want.

addaleax: This is what I get without the `Array.from()` bit:

```js
test> db.bsonvec.findOne()
{
  _id: ObjectId('67ed08d64a05814b92a9ff62'),
  x: Float32(Float32Array(1) [ 1.539989614439558e-36 ])
}
```

packages/service-provider-core/src/printable-bson.ts
```diff
@@ -43,10 +44,38 @@ export const bsonStringifiers: Record<
...
+              ...args
+            )})`;
+          case BSON.Binary.VECTOR_TYPE.PackedBit:
+            return `PackedBit(${utilInspect(
+              hexString,
+              // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
+              ...args
+            )})`;
+          default:
+            return `Vector(${utilInspect(
```
addaleax, line 73: In general, the representation we go for in our inspect function is one that allows re-creating the object by pasting the output back into input, unless we have a reason not to do that here, let's keep that intact

packages/service-provider-core/src/printable-bson.ts
gagik: Ah, I was testing in it in a playground which was using the DOM implementation for util... that makes more sense

packages/service-provider-core/src/printable-bson.ts
```diff
@@ -7,26 +9,73 @@ type BSONClassKey = (typeof BSON)[Exclude<
...
+          maxArrayLength: utilInspect.defaultOptions.maxArrayLength,
+        })
+      )})`;
+    case BSON.Binary.VECTOR_TYPE.Float32:
+      return `Binary.fromFloat32Array(new Float32Array(${removeTypedArrayPrefixFromInspectResult(
+        utilInspect(this.toFloat32Array(), {
+          depth,
+          ...options,
+          // These arrays can be very large, so would prefer to use the default options instead.
+          maxArrayLength: utilInspect.defaultOptions.maxArrayLength,
```
gagik, line 59: not sure if we want this, could remove. it just seems possible that these arrays can be very large I'd imagine so the `{maxArrayLength: Infinity}` that we're getting from https://github.com/mongodb-js/mongosh/blob/d93774d30aa5d64d5cdf52e58915a0fc1e2a031f/packages/cli-repl/src/format-output.ts#L68 can be problematic

packages/service-provider-core/src/printable-bson.spec.ts
```diff
@@ -95,6 +95,34 @@ describe('BSON printers', function () {
...
+      expect(
+        inspect(
+          bson.Binary.fromFloat32Array(
+            new Float32Array([1.1111, 2.2222, 3.3333])
+          ),
+          { compact: true }
+        )
+      ).matches(
+        // Precision is lost because of float handling, so we use regex to match
+        /Binary.fromFloat32Array\(new Float32Array\(\[ 1\.1\d*, 2\.2\d*, 3\.3\d* \]\)\)/
```
gagik, line 121: not sure if we're meant to be dealing with float precision weirdness in JS in this case. can definitely see how it'd be weird to see imprecise floats while trying to inspect data. we could do something like https://github.com/mongodb-js/compass/pull/6834/files#diff-3d4b298fbb2cfe42467edc59277483eb9bcfdc46b07ea658ab3dc4f66903da88R174

packages/service-provider-core/src/printable-bson.ts
```diff
@@ -64,7 +116,7 @@ export const bsonStringifiers: Record<
       default:
         return binaryInspect.apply(this, args);
     }
-  },
+  } satisfies CustomInspectFunction,
```
addaleax, line 119: 💙

addaleax (approved):
This looks great! 🚀

PR #2427, opened 2025-04-04
chore(tests): add tests for substrings, suffixes, prefixes with field level encryption COMPASS-8816

gagik:
It has previously been not possible to use $lookup stage of a pipeline with automatic encryption attached to it. This adds a test to verify it is possible along with another dependency bump.


packages/service-provider-node-driver/package.json
```diff
@@ -59,7 +59,7 @@
   },
   "optionalDependencies": {
     "kerberos": "2.1.0",
-    "mongodb-client-encryption": "^6.1.1"
+    "mongodb-client-encryption": "^6.3.0"
```
gagik, line 62: Seems like `mongodb-client-encryption` was previously bumped but this package was left at the old version; was this intentional?

addaleax:
Just quoting myself from the ticket:

> We will still want to add at least a mongosh e2e test on top of those, and we'll need to bump the crypt_shared library to 8.1 in both Compass and mongosh (for arm64/x64).

This should probably be an e2e test since we don't have the infrastructure for integration tests with the crypt_shared library in place in the shell-api package, and all similar tests are also e2e tests.

And, well, we need the 8.1 crypt_shared library, and that's not released yet.

packages/service-provider-node-driver/package.json
addaleax: No, I think that's just an oversight (or maybe a bad merge conflict resolve?) 👍

.evergreen/constants.js
```diff
@@ -37,8 +37,8 @@ exports.MONGODB_VERSIONS = [
   { shortName: '70xe', versionSpec: '7.0.x-enterprise' },
   { shortName: '80xc', versionSpec: '8.0.x' },
   { shortName: '80xe', versionSpec: '8.0.x-enterprise' },
-  { shortName: '82rc', versionSpec: '8.2.0-rc4' },
-  { shortName: '82rce', versionSpec: '8.2.0-rc4-enterprise' },
+  { shortName: '82rc', versionSpec: '8.2.0' },
```
gagik: I guess we should remove the `rc` suffix I added too from shortname 
(in hindsight, this choice ends up messing with task history which in some way makes sense i.e. rc vs non-rc but for our purposes should probably have followed the typical naming scheme from getgo)

packages/e2e-tests/test/e2e-fle.spec.ts
```diff
@@ -914,6 +913,311 @@ describe('FLE tests', function () {
...
+          textOptions: { caseSensitive: false, diacriticSensitive: false, ...details }
+        });
+        substringOpts = explicitOpts({ substring: { ...substringOptions, strMaxLength: 60 } });
+        prefixSuffixOpts = explicitOpts({ prefix: substringOptions, suffix: substringOptions });
+        methods = {
+          substring: { queryType: 'substringPreview', keyId: keyId1, options: substringOpts },
+          prefix: { queryType: 'prefixPreview', keyId: keyId2, options: prefixSuffixOpts },
+          suffix: { queryType: 'suffixPreview', keyId: keyId2, options: prefixSuffixOpts }
+        }
+        explicitEncrypt = (method, data) => {
```
gagik, line 1107: nice idea with defining the methods, opts variables. a function for explicit, make it much cleaner to follow. I think untangling the different formats was the biggest problem for me.

packages/shell-api/src/field-level-encryption.spec.ts
```diff
@@ -962,7 +962,7 @@ srDVjIT3LsvTqw==`,
               return;
             }
             expect.fail('missed exception');
-            break;
+          // eslint-disable-next-line no-fallthrough
```
gagik: might be coming from a merge conflict or my own changes but is this an intentional choice? `.fail` will essentially throw so no use in the break but break seems preferable to a eslint override

testing/integration-testing-hooks.ts
```diff
@@ -259,6 +259,7 @@ export function startTestCluster(id: string, ...argLists: Partial<MongoClusterOp
 }
 
 function skipIfVersion(test: any, testServerVersion: string, semverCondition: string): void {
+  //console.log({testServerVersion, semverCondition});
```
gagik:
```suggestion
```

packages/shell-api/src/field-level-encryption.spec.ts
addaleax: Yeah, this is a leftover from the original contents of the PR – removed!

.evergreen/constants.js
addaleax: Yup, done!

gagik:
Looks good, the check failure is confusing, going to try to run it again 🤷

PR #2430, opened 2025-04-07
chore(ci): include purls in the node-runtime-worker-thread package MONGOSH-1856

gagik:


.evergreen/compile-artifact.sh
```diff
@@ -96,10 +96,10 @@ if uname -a | grep -q 'Linux.*x86_64'; then
...
-
-cat dist/.purls.txt
-
 npm run create-dependency-sbom-lists
+
+ls -lhA .sbom
+for dir in .sbom/*/; do
+  cp ${dir}purls.txt dist/${dir}purls.txt
+  cat dist/${dir}purls.txt
+done
```
addaleax, line 105: How does this work when publishing packages through GHA?

The ticket was created before that was a thing, but now I think we'll need to account for that – maybe it could be encapsulated so that it only really affects the compile step of the node-runtime-worker-thread package? That would mean missing out on Silk/Kondukto integration for it, but I think that's okay given that it's not a standalone supported product

gagik: Good point; will look into potentially moving this to the build step instead

packages/node-runtime-worker-thread/package.json
```diff
@@ -29,9 +29,11 @@
     "webpack-build": "npm run compile && webpack --mode production",
     "webpack-build-dev": "npm run compile && webpack --mode development",
     "compile": "tsc -p tsconfig.json",
-    "prepublish": "npm run webpack-build",
+    "prepublish": "npm run webpack-build && npm run create-purls-file",
```
gagik, line 32: @addaleax I simplified the whole setup to minimize changes and just going to append the purls file into `dist/purls.txt` of the package since that seems to be all we need. I assume we don't need to append `.mongosh_crypt_...` as I don't see the files for it there.

addaleax: > I assume we don't need to append `.mongosh_crypt_...` as I don't see the files for it there.

Right, we only include the crypt_shared library in the distributed Compass and mongosh packages but not the individual npm packages in the monorepo 👍

addaleax (approved):

PR #2432, opened 2025-04-08
chore(ci): skip chromium download for most tasks

gagik:
Out of experiencing the pain of seeing what seemed to be potentially a 10 minute download and install of puppeteer each time during release, I became curious whether we're actually skipping things correctly. 

`PUPPETEER_SKIP_DOWNLOAD` only applies at a post-install script but in many places we set it after all dependencies have already been installed. This sets it at the install stage instead. Not sure whatever the issues were with https://github.com/mongodb-js/mongosh/pull/2311/ would come up here, seems all good so far.

Not super scientific but this leads to quicker execution on hosts which tend to not have the Chromium in their cache like RHEL:

### Before
billede

### After
billede

### DIFF
--- a/.evergreen.yml
+++ b/.evergreen.yml
@@ -121,6 +121,7 @@ functions:
           NODE_JS_VERSION: ${node_js_version}
           DISTRO_ID: ${distro_id}
           MONOGDB_DRIVER_VERSION_OVERRIDE: ${mongodb_driver_version_override}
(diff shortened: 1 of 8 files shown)


nirinchev (approved):
That 30+ minute speedup on RHEL 😱

addaleax (approved):

PR #2434, opened 2025-04-17
feat: Shell API autocomplete type definitions MONGOSH-2031 MONGOSH-2032 MONGOSH-2173

lerouxb:
This is a continuation of Anna's work [here](https://github.com/mongodb-js/mongosh/pull/1802).

Usage:

```
npm run api-generate
```

Implements:
* https://jira.mongodb.org/browse/MONGOSH-2031
* https://jira.mongodb.org/browse/MONGOSH-2032
* https://jira.mongodb.org/browse/MONGOSH-2173

Changes:
* merged in main, resolving the conflicts
* added generics to the rs, sh and sp objects
* added a separate api.ts entrypoint, generating api-processed.d.ts (although this needs work - trying to figure out how to reference it in package.json)
* bundled more deps so that we don't import those types in the api-processed.d.ts file. The node.js packages are still being imported, though. Will have to test if that works with the language server
* fixed the tests (at least the ones I've noticed were broken so far)
* added an api-globals.d.ts which declares the ShellAPI methods and the ShellBson functions as globals. This then gets appended to the end of api-processed.d.ts.
* now exporting `@mongosh/shell-api/api` which is just api-processed exported as one giant string.

**Notes:**

* We'll have to do https://jira.mongodb.org/browse/MONGOSH-2170 as a follow-up once https://github.com/mongodb-js/devtools-shared/pull/521 is in


packages/autocomplete/src/index.spec.ts
```diff
@@ -294,7 +294,7 @@ describe('completer.completer', function () {
 
     it('returns all suggestions', async function () {
       const i = 'db.';
-      const attr = shellSignatures.Database.attributes as any;
+      const attr = shellSignatures.DatabaseImpl.attributes as any;
```
lerouxb, line 297: The fact that Database and Collection are now types and the classes got replaced with DatabaseImpl and CollectionImpl is having a lot of knock-on effects and kinda ended up being responsible for most of this diff.

I'm wondering if we should keep them as Database and Collection but then rather name the new types to be something like DatabaseWithSchema and CollectionWithSchema. Then maybe the change will be smaller?

Not sure which way around would work out better ergonomically overall.

packages/shell-api/package.json
```diff
@@ -4,6 +4,15 @@
   "description": "MongoDB Shell API Classes Package",
   "main": "lib/index.js",
   "types": "lib/index.d.ts",
+  "exports": {
+    ".": {
+      "default": "./lib/index.js",
+      "types": "./lib/index.d.ts"
+    },
+    "./api": {
+      "types": "./lib/api-processed.d.ts"
```
lerouxb, line 13: I'm not sure what would be the best way to export/publish this and then reference it from the outside.

I think `import type ShellApi from '@monogdb-js/shell-api/api'` works with this and I tried it manually from a file, but I'll see shortly when I plug this into mongodb-ts-autocomplete

packages/shell-api/src/collection.spec.ts
```diff
@@ -33,24 +33,26 @@ import {
...
+    const apiClass = new CollectionImpl({} as any, {} as any, 'name');
     it('calls help function', async function () {
       expect((await toShellResult(apiClass.help())).type).to.equal('Help');
       expect((await toShellResult(apiClass.help)).type).to.equal('Help');
     });
   });
   describe('signatures', function () {
     it('type', function () {
-      expect(signatures.Collection.type).to.equal('Collection');
+      // TODO: do we want the signatures to be CollectionImpl or Collection?
```
lerouxb, line 51: Just a reminder for myself to remove this TODO comment before merging and after bringing it up with the team.

packages/shell-api/src/api.ts
```diff
@@ -0,0 +1,102 @@
+// TODO: does it make sense to have all this stuff? Don't we just need enough for the top-level API, ie. the globals?
```
lerouxb, line 1: Just a reminder for myself to remove this TODO comment before merging and after bringing it up with the team.

packages/build/src/packaging/package/helpers.ts
```diff
@@ -13,7 +13,9 @@ export async function execFile(
 ): Promise<ReturnType<typeof execFileWithoutLogging>> {
   const joinedCommand = [args[0], ...(args[1] ?? [])].join(' ');
   console.info(
-    'Running "' + joinedCommand + '" in ' + args[2]?.cwd ?? process.cwd()
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore TS2869 Right operand of ?? is unreachable because the left operand is never nullish.
+    `Running "${joinedCommand}" in ${args[2]?.cwd ?? process.cwd()}`
```
lerouxb, line 18: drive-by. For some reason I got a compile error on this line in CI.

packages/shell-api/api-extractor.json
```diff
@@ -0,0 +1,57 @@
...
+  },
+  "bundledPackages": [
+    "@mongodb-js/devtools-connect",
+    "@mongodb-js/devtools-proxy-support",
+    "@mongodb-js/oidc-plugin",
+    "@mongosh/arg-parser",
+    "@mongosh/service-provider-core",
+    "@mongosh/types",
+    "mongodb",
+    "bson"
```
lerouxb, line 18: We have bson set up in mongodb-ts-autocomplete, so I might remove it from here.

packages/shell-api/src/api-globals.d.ts
```diff
@@ -0,0 +1,44 @@
+declare global {
```
lerouxb, line 1: This file is basically a snippet that gets appended at the end of api-globals.d.ts after we api-extract that. See bin/api-postprocess.ts.

I first tried to write a babel plugin to walk the ShellAPI class's methods and the ShellBson interface's methods, but that turned out to be really complicated.

There is one benefit to having them all mapped here explicitly: We don't end up accidentally leaking everything on ShellAPI into the global scope whether we want to or not. And it is arguably handy seeing them all enumerated here.

The only other variables added as globals would be db, sh, sp and rs which all get specified dynamically by mongodb-ts-autocomplete so they reference the server/database/collection schema generics.

packages/shell-api/src/api.ts
lerouxb: This would all become one big ShellAPI import in monogdb-ts-autocomplete where we set up the db, sh, rs, sp globals so we can reference things like ShellAPI.Database. But I suppose we can hide it by importing it as a type.

packages/shell-api/src/api.ts
```diff
@@ -0,0 +1,102 @@
...
+  Mongo,
+  Shard,
+  ReplicaSet,
+  ShellApi,
+  ServerVersions,
+  Topologies,
+  InterruptFlag,
+};
+
+// TODO: do we really want all these?
```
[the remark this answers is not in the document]
lerouxb: I started looking into all the ae-forgotten-export errors, but in the end I made those warnings for now. It is kinda a lot and as you start adding them one by one _their_ dependencies end up becoming errors..

I don't think we need this as the shell api, though - that warning/error is probably more useful for something that's used as a library.

packages/shell-api/bin/api-postprocess.ts
```diff
@@ -0,0 +1,166 @@
...
+  );
+  const result = babel.transformSync(apiRaw, {
+    code: true,
+    ast: false,
+    configFile: false,
+    babelrc: false,
+    browserslistConfigFile: false,
+    compact: false,
+    sourceType: 'module',
+    plugins: [applyAsyncRewriterChanges()],
```
lerouxb, line 108: We end up with these errors:

```
~/mongo/mongosh/packages/shell-api % npx tsc --lib es2019 --target es2018 --moduleResolution node --types node lib/api-processed.d.ts
lib/api-processed.d.ts:4412:3 - error TS2416: Property 'hasNext' in type 'Cursor' is not assignable to the same property in base type 'AggregateOrFindCursor<ServiceProviderFindCursor<Document_2>>'.
  Type '() => boolean' is not assignable to type '() => Promise<boolean>'.
    Type 'boolean' is not assignable to type 'Promise<boolean>'.

4412   hasNext(): boolean;
       ~~~~~~~

lib/api-processed.d.ts:4418:3 - error TS2416: Property 'next' in type 'Cursor' is not assignable to the same property in base type 'AggregateOrFindCursor<ServiceProviderFindCursor<Document_2>>'.
  Type '() => Document_2' is not assignable to type '() => Promise<Document_2>'.
    Type 'Document_2' is missing the following properties from type 'Promise<Document_2>': then, catch, finally, [Symbol.toStringTag]

4418   next(): Document_2 | null;
       ~~~~


Found 2 errors in the same file, starting at: lib/api-processed.d.ts:4412
```

Weirdly these seem to be the only two cases - if I revert those two return types manually then the file is all green.

The ts language service works around it as far as I can tell, so it doesn't impact autocomplete. If we did want to make the generated .ts valid we'll have to do something like modify the base type's methods' return types or somehow flatten the return type so that it doesn't inherit anymore?

Maybe there are other ideas.

lerouxb: Wild idea: we can just add the comment

```
  // @ts-expect-error The type of this function is not compatible with the type of the function in the base class
```

as part of the babel processing...

packages/shell-api/bin/api-postprocess.ts
```diff
@@ -3,6 +3,18 @@ import type * as BabelTypes from '@babel/types';
 import { promises as fs } from 'fs';
 import path from 'path';
 import { signatures } from '../';
+import enUs from '../../i18n/src/locales/en_US';
+
+function getHelpClassName(className: string) {
+  switch (className) {
+    case 'CollectionImpl':
```
lerouxb, line 10: I pointed this out below as well, but this is an example of the kind of fallout from renaming Collection to CollectionImp and reusing the name for a type. And doing the same for Database.

Maybe this is OK but we could also consider leaving Collection and Database unchanged and then just having the types named something else like CollectionWithSchema and DatabaseWithSchema. Or whatever might be a better name.

packages/shell-api/package.json
lerouxb: In the end I have to fs.readFile() to get the file as text so we can bundle it as a giant string, so this is kinda moot.

packages/shell-api/src/api.ts
lerouxb: I added some regression tests to make sure we don't leak ShellAPI or anything on it as a varaible name, so I think that's at least safe.

packages/shell-api/bin/api-postprocess.ts
lerouxb: I filed https://jira.mongodb.org/browse/MONGOSH-2209

Had a quick look and decided to split that into its own ticket rather.

packages/shell-api/package.json
lerouxb: Right now I generate that giant string in mongodb-ts-autocomplete, but I suppose we can also export it here. I don't know which way is best. I don't want to accidentally bundle a giant string..

packages/shell-api/bin/api-postprocess.ts
```diff
@@ -0,0 +1,197 @@
...
+              !signatures[className].attributes?.[methodName].inherited
+            ) {
+              if (
+                !this.processedMethods.find(
+                  ([cls, method]) => cls === className && method === methodName
+                )
+              ) {
+                // TODO: fix this
+                if (className === 'Mongo') {
+                  return;
```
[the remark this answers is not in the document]
lerouxb: I have no idea why that was there. I removed the comment and the if and early return 🤷

packages/shell-api/package.json
addaleax: So ... the two options seem to be:

- Make shell-api a dev dependency of mongodb-ts-autocomplete, run the extract-types script there as a compile/prepublish step and then include the generated "big string" file in the mongodb-ts-autocomplete exports
- Make shell-api a peer dependency of mongodb-ts-autocomplete, move the "big string" generation script here a compile/prepublish and export it from shell-api, then consume it from mongodb-ts-autocomplete

Obviously both of these work, but I don't really see disadvantages to the latter; the former does have at least one notable disadvantage (having to update mongodb-ts-autocomplete to incorporate shell-api changes).

> I don't want to accidentally bundle a giant string..

I think we'll end up bundling this string in all our products anyway. But for now, this should actually just work fine, right? webpack wouldn't include it if it isn't being used.

packages/shell-api/api-extractor.json
lerouxb: I removed it from here and then mongodb-ts-autocomplete in devtools-shared replaces the bson import to use its one. That way we can use the same bson shared everywhere (shell-api and mql) and we don't end up extracting all the types again and again. This might premature optimisation, I don't know. But it isn't hard to undo either.

packages/shell-api/package.json
lerouxb: Here's what I do now: https://github.com/mongodb-js/mongosh/pull/2434/commits/0bd794f51ddf347109295d6db7d065c5a98e8777

Then in devtools-shared I can:

```
import ShellApiText from '@mongosh/shell-api/api';
```

Without having to require.resolve() and read the file and all that. Which is.. marginally better, I guess? I still read mql  lib for now (until we do the same thing in the mql types package) and I still do it for reading the bson package.

We can only really get rid of the bson package if we reverthttps://github.com/mongodb-js/mongosh/pull/2434/commits/9b3779d0d96c2b285405a9a903ca418828193198 and therefore bundle the bson types we use inside the extracted API. But then we'd have to do the same in mql types too.

packages/shell-api/src/database.ts
```diff
@@ -68,23 +74,37 @@ type AuthDoc = {
...
+  _mongo: Mongo<M>;
+  _name: StringKey<M>;
+  _collections: Record<string, Collection<M, D>>;
   _session: Session | undefined;
-  _cachedCollectionNames: string[] = [];
+  _cachedCollectionNames: StringKey<D>[] = [];
   _cachedHello: Document | null = null;
 
-  constructor(mongo: Mongo, name: string, session?: Session) {
+  _typeLaunder(): Database<M, D> {
```
nirinchev, line 96: Is this an idiomatic way for this kind of pattern? At least for me, a clearer naming would be `asDatabaseWithSchema` or something similar. `_typeLaunder` doesn't really convey what's happening here, but similarly to the generics - if it's used often enough elsewhere, happy with keeping it.

packages/shell-api/src/collection.ts
```diff
@@ -94,13 +97,37 @@ import PlanCache from './plan-cache';
...
+> extends ShellApiWithMongoClass {
+  _mongo: Mongo<M>;
+  _database: Database<M, D>;
+  _name: N;
+
+  _typeLaunder(): Collection<M, D> {
+    return this as Collection<M, D>;
+  }
+
+  constructor(mongo: Mongo<M>, database: Database<M, D>, name: N) {
```
nirinchev, line 130: Should we accept `Database | DatabaseImpl` here? I see a lot of places where we need to call `database._typeLaunder` which is a bit smelly when done outside. We're not really getting rid of the smell by accepting an -Impl here, but at least concentrating it in one place.

packages/shell-api/src/helpers.spec.ts
```diff
@@ -131,7 +131,7 @@ describe('getPrintableShardStatus', function () {
   const testServer = startSharedTestServer();
 
   let mongo: Mongo;
-  let configDatabase: Database;
+  let configDatabase: DatabaseImpl;
```
nirinchev, line 134: Why not `Database`?

packages/shell-api/src/index.ts
```diff
@@ -42,7 +42,9 @@ export {
   Cursor,
   CursorIterationResult,
   Database,
+  DatabaseImpl,
```
nirinchev, line 45: Do we want to be exporting the `Impl`s here? I'd expect those to be used only in this package and then we use the `Collection`/`Database` outside.

packages/shell-api/src/mongo-errors.spec.ts
```diff
@@ -78,10 +78,10 @@ describe('mongo-errors', function () {
   describe('intercepts shell API calls', function () {
     let mongo: Mongo;
     let serviceProvider: StubbedInstance<ServiceProvider>;
-    let database: Database;
+    let database: DatabaseImpl;
```
nirinchev, line 81:
```suggestion
    let database: Database;
```

packages/shell-api/src/mongo-errors.spec.ts
```diff
@@ -78,10 +78,10 @@ describe('mongo-errors', function () {
...
     let bus: StubbedInstance<EventEmitter>;
     let instanceState: ShellInstanceState;
-    let collection: Collection;
+    let collection: CollectionImpl;
```
nirinchev, line 84:
```suggestion
    let collection: Collection;
```

packages/shell-api/src/mongo.spec.ts
```diff
@@ -99,7 +99,7 @@ describe('Mongo', function () {
     const driverSession = { driverSession: 1 };
     let mongo: Mongo;
     let serviceProvider: StubbedInstance<ServiceProvider>;
-    let database: StubbedInstance<Database>;
+    let database: StubbedInstance<DatabaseImpl>;
```
nirinchev, line 102: Should we have `DatabaseImpl` here vs just stubbing `Database`?

nirinchev (approved):
Looks good - I mostly have naming/nitty suggestions, but the core logic makes sense to me.

packages/shell-api/src/database.ts
```diff
@@ -68,23 +74,37 @@ type AuthDoc = {
...
+};
+
 @shellApiClassDefault
-export default class Database extends ShellApiWithMongoClass {
-  _mongo: Mongo;
-  _name: string;
-  _collections: Record<string, Collection>;
+export class DatabaseImpl<
+  M extends GenericServerSideSchema = GenericServerSideSchema,
+  D extends GenericDatabaseSchema = GenericDatabaseSchema
```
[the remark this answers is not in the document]
lerouxb: I admittedly just left them as I found them and haven't given it much thought. Any suggestions?

packages/shell-api/src/bulk.spec.ts
```diff
@@ -57,7 +57,7 @@ describe('Bulk API', function () {
     });
     describe('Metadata', function () {
       describe('toShellResult', function () {
-        const collection = stubInterface<Collection>();
+        const collection = stubInterface<CollectionImpl>();
```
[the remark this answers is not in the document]
lerouxb: Collection (like Database) is now just a type, not a class. This is what I called out in a review comment [here](https://github.com/mongodb-js/mongosh/pull/2434#discussion_r2049155104). We can keep Collection and Database as classes and then make types named something like CollectionWithSchema and DatabaseWithSchema.

I'm in two minds about which way around is better. I just left it this way because it is what @addaleax did in the PoC.

packages/shell-api/src/database.ts
lerouxb: I honestly don't know - I just kept it because that's how it was in the PoC. In general I just tried to "productionize" the PoC which I took as "try and get the tests to pass" 😆 @addaleax what do you think?

packages/shell-api/src/bulk.spec.ts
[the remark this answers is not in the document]
lerouxb: I suppose there is probably also a third way where we just extend the class and end up with another Class rather than building a type 🤔

packages/shell-api/src/helpers.spec.ts
lerouxb: These are all fallout from the Database vs DatabaseImpl and Collection vs CollectionImpl thing above. But also as it stands Database & Collection are now generic which means every unit test in the codebase that uses those have to specify the generic types which.. would be a very large PR. And unit tests generally don't care about the schema provided by the generics - they just test the base Collection and Database classes which are just called DatabaseImpl and CollectionImpl now.

I think I can probably "solve" it by renaming CollectionImpl back to Collection and DatabaseImpl back to Database then making either derived classes or types called CollectionWithSchema and DatabaseWithSchema.

lerouxb: DatabaseImpl is gone now.

packages/shell-api/src/mongo.spec.ts
```diff
@@ -940,14 +940,16 @@ describe('Mongo', function () {
         expect(coll._database._name).to.equal('db1');
       });
 
-      it('throws if name is not a valid connection string', function () {
+      it('throws if name is not a valid collnection string', function () {
```
addaleax, line 943: a what? 😄 

```suggestion
      it('throws if name is not a valid collection string', function () {
```

packages/shell-api/src/database.ts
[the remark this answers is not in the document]
addaleax: I felt like the default values make it clear enough what these are respectively, and it doesn't take a ton of getting used to, but if there are better names for these, then I don't think there's any reason not to adopt them

packages/shell-api/package.json
```diff
@@ -13,7 +23,8 @@
   },
   "scripts": {
     "compile": "tsc -p tsconfig.json",
-    "pretest": "npm run compile",
+    "api-generate": "api-extractor run && ts-node bin/api-postprocess.ts",
+    "pretest": "npm run compile && npm run api-generate",
```
addaleax, line 27: Any reason not to make this part of `compile`? You'd also need this when building the mongosh executable, for example, not just for testing or publishing

packages/shell-api/bin/api-postprocess.ts
```diff
@@ -0,0 +1,156 @@
...
+`;
+  await fs.writeFile(
+    path.resolve(__dirname, '..', 'lib', 'api-export.js'),
+    exportCode
+  );
+
+  await fs.writeFile(
+    path.resolve(__dirname, '..', 'lib', 'api-export.d.ts'),
+    'declare const _default: string; export default _default;'
+  );
```
addaleax, line 149: Just gonna say, my gut feeling is that it would be easiest to just avoid default exports (good practice anyway) and stick to a named export à la

```suggestion
  const exportCode = `"use strict";
module.exports = { api: ${JSON.stringify(code)} };
`;
  await fs.writeFile(
    path.resolve(__dirname, '..', 'lib', 'api-export.js'),
    exportCode
  );

  await fs.writeFile(
    path.resolve(__dirname, '..', 'lib', 'api-export.d.ts'),
    'export declare const api;'
  );
```

instead of jumping through hooks to get a default export to work (definitely non-blocking though)

addaleax (approved):
LGTM – only real blocker is the fact that we need to make sure to return `DatabaseWithSchema` and `CollectionWithSchema` in public APIs now

packages/shell-api/src/mongo.spec.ts
lerouxb: I have no idea how I do these things.

packages/shell-api/src/database.ts
```diff
@@ -473,17 +496,19 @@ export default class Database extends ShellApiWithMongoClass {
   }
 
   @returnType('Database')
-  getSiblingDB(db: string): Database {
+  getSiblingDB<K extends StringKey<M>>(db: K): Database<M, M[K]> {
```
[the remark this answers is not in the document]
lerouxb: Oof sorry I was so sure I double checked this. I'll go over it again.

packages/shell-api/package.json
lerouxb: No good reason. I'll change it.

packages/shell-api/bin/api-postprocess.ts
lerouxb: Sure I don't mind. It isn't exactly a public API anyway 😆

packages/cli-repl/src/cli-repl.spec.ts
```diff
@@ -2466,8 +2466,8 @@ describe('CliRepl', function () {
       });
 
       afterEach(async function () {
-        expect(output).not.to.include('Tab completion error');
-        expect(output).not.to.include(
+        expect(output, output).not.to.include('Tab completion error');
+        expect(output, output).not.to.include(
```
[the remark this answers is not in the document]
lerouxb: This is kinda a hacky but very easy way to make sure it print the full output if it ever fails which is the very first thing anyone debugging it would do. And it should keep quiet when the expect succeeds so is kinda zero cost.

PR #2446, opened 2025-05-05
chore: switch to `@mongodb-js/device-id`

gagik:
This package is meant to minimize code replication and to make sure the approach we're taking with hashing is consistent and timeouts are automatically handled.

See https://github.com/mongodb-js/devtools-shared/pull/532

### DIFF
--- a/packages/logging/src/logging-and-telemetry.ts
+++ b/packages/logging/src/logging-and-telemetry.ts
@@ -53,40 +53,8 @@ import type {
   MongoshLoggingAndTelemetryArguments,
   MongoshTrackingProperties,
 } from './types';
-import { createHmac } from 'crypto';
-
-/**
- * @returns A hashed, unique identifier for the running device or `"unknown"` if not known.
- */
-export async function getDeviceId({
-  onError,
-}: {
-  onError?: (error: Error) => void;
-} = {}): Promise<string | 'unknown'> {
-  try {
-    // Create a hashed format from the all uppercase version of the machine ID
-    // to match it exactly with the denisbrodbeck/machineid library that Atlas CLI uses.
-    const originalId: string =
-      // eslint-disable-next-line @typescript-eslint/no-var-requires
-      await require('native-machine-id').getMachineId({
-        raw: true,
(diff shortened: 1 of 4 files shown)


packages/logging/src/logging-and-telemetry.ts
```diff
@@ -125,11 +92,11 @@ export class LoggingAndTelemetry implements MongoshLoggingAndTelemetry {
...
 
   private deviceId: string | undefined;
-  /** @internal */
+
+  /** @internal Used for awaiting the telemetry setup in tests. */
   public setupTelemetryPromise: Promise<void> = Promise.resolve();
 
-  // eslint-disable-next-line @typescript-eslint/no-empty-function
-  private resolveDeviceId: (value: string) => void = () => {};
+  private readonly telemetrySetup: AbortController = new AbortController();
```
gagik: not a huge fan of this name, maybe `deviceIdResolution` though I was thinking this would be shared by the entire setup. 
I could also call it `abortController` but seems like a implementation detail.

addaleax: I do think it's odd to refer to an abort mechanism as `...Setup` 🙂 No strong feelings here but maybe just `telemetrySetupAbort`?

packages/logging/src/logging-and-telemetry.ts
```diff
@@ -160,26 +127,29 @@ export class LoggingAndTelemetry implements MongoshLoggingAndTelemetry {
...
-        getDeviceId({
-          onError: (error) =>
-            this.bus.emit('mongosh:error', error, 'telemetry'),
-        }),
-        new Promise<string>((resolve) => {
-          this.resolveDeviceId = resolve;
-        }),
-      ]);
+      // eslint-disable-next-line @typescript-eslint/no-var-requires
+      const getMachineId = require('native-machine-id').getMachineId;
```
addaleax: This was previously in a try/catch (to avoid crashing mongosh in situations in which we cannot load this), should it continue to be?

PR #2452, opened 2025-05-12
chore: fix missing fields in package-lock.json

yuyuyureka:
Due to a [bug in npm](https://github.com/npm/cli/issues/6301) the integrity fields of some dependencies are missing when regenerating package-lock.json in presence of an existing node_modules

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -125,6 +125,8 @@
     },
     "node_modules/@ampproject/remapping": {
       "version": "2.2.0",
+      "resolved": "https://registry.npmjs.org/@ampproject/remapping/-/remapping-2.2.0.tgz",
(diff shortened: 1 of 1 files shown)


addaleax:
@yuyuyureka I think this PR is large enough that it would require you to sign our CLA, see https://www.mongodb.com/legal/contributor-agreement. Alternatively, can you let us know which commands you ran to create this change?

addaleax (approved):

addaleax:
> I don't think the change constitutes a creative work, as it is only generated changes. In case it does: I hereby release any changes made within this PR into the public domain.
> 
> I used the [npm-lockfile-fix](https://github.com/jeslie0/npm-lockfile-fix) tool to add the missing fields without changing any dependency versions, and then ran `npm install --package-lock-only` once more to ensure the formatting doesn't change on consecutive `npm install` runs.

Cool yeah, that makes sense, and I've verified that I'm getting the same results through these steps. Just leaving this extra comment as a reminder to include this information in the commit message when merging. Thanks for the PR!

addaleax:
@yuyuyureka Thanks, merged! 🚀

PR #2455, opened 2025-05-15
fix: unsafe shell command constructed from library input

odaysec:
https://github.com/mongodb-js/mongosh/blob/e9bfb2b3cc9575da07ec0ea2ab4b54867a1901ce/packages/editor/src/editor.ts#L243-L246

fix the issue should avoid using `shell: true` and instead pass the command and its arguments as an array to the `spawn` function. This ensures that the arguments are not interpreted by the shell, mitigating the risk of shell injection. Specifically:
1. Replace the `spawn` call with a version that does not use `shell: true`.
2. Pass the full path to the editor and the temporary file as separate arguments to `spawn`.

This approach ensures that the command is executed directly without shell interpretation, making it safe from injection attacks.

### DIFF
--- a/packages/editor/src/editor.ts
+++ b/packages/editor/src/editor.ts
@@ -240,10 +240,9 @@ export class Editor {
     const { spawn } =
       // eslint-disable-next-line
       require('child_process') as typeof import('child_process');
-    const proc = spawn(editor, [path.basename(tmpDoc)], {
+    const proc = spawn(editor, [tmpDoc], {
       stdio: 'inherit',
       cwd: path.dirname(tmpDoc),
-      shell: true,
     });
 
     // Pause the parent readable stream to stop emitting data events


addaleax:
@odaysec It seems you are creating pull requests across multiple GitHub repositories, potentially based on detecting specific patterns in source code automatically. Unfortunately, it also seems that you are not taking into account the circumstances under which the code in question runs, and what its intended behavior would be. I will close this PR and #2454 as they do not appear to be targeting legitimate vulnerabilities in this software. 

For the future, please be encouraged to bring up genuine security issues, but be advised that opening PRs across a broad set of repositories automatically without understanding the target source code will generally not help your reputation as a security researcher, and rather will put your PRs at risk of being marked as spam and reported to GitHub.

jacopotediosi:
Things like this make legitimate security researchers look bad. It's a shame.

PR #2458, opened 2025-05-22
chore(e2e-test): add regression test for dynamic import() MONGOSH-1062

addaleax:


packages/e2e-tests/test/e2e.spec.ts
```diff
@@ -1170,22 +1170,39 @@ describe('e2e', function () {
...
+      expect(result).to.match(/^B-ESM$/m);
       result = await shell.executeLine('require("c")');
       expect(result).to.match(/^C$/m);
+      result = await shell.executeLine('import("b").then(m => m.default)');
+      expect(result).to.match(/^B$/m);
+      result = await shell.executeLine('import("b-esm").then(m => m.value)');
+      expect(result).to.match(/^B-ESM$/m);
+    });
+
+    it('import() works when interleaved with GC', async function () {
```
gagik, line 1199: what is the relationship between the garbage collector and import that would cause issues?

addaleax: @gagik Sorry, completely missed this comment! The original ticket is https://github.com/nodejs/node/issues/38695, I'll link it here in the comments as well – basically, Node.js internals had a memory management bug where a C++ object describing modules could get deleted during garbage collection while the "public" JS module instance was still alive, and then accessing that C++ object from JS would crash

```suggestion
    // Regression test for https://github.com/nodejs/node/issues/38695
    it('import() works when interleaved with GC', async function () {
```

Obviously this is really more of a Node.js issue and one could argue that we don't need a specific regression test here in mongosh, but I'd feel more comfortable having it in place for when we start having a "proper" ESM story in mongosh (i.e. mongosh scripts with import/export).

gagik: thank you! the link in comments is helpful.

gagik (approved):

PR #2460, opened 2025-05-28
fix: set `chunkFormat: false` in webpack config MONGOSH-2225

addaleax:
The mongosh CLI cannot currently handled chunked input, as the process for building the single-executable application only uses a single file as its input.


addaleax:
(We'll want to check the benchmark CI task before merging)

PR #2464, opened 2025-06-03
feat: port autocomplete direct commands and browser-runtime-core to the new autocomplete MONGOSH-2205 MONGOSH-2035 MONGOSH-2206

lerouxb:
This essentially reuses (as in makes a copy of) the existing direct shell command autocompleter and then ports it to use the new AutocompleteContext vs the old AutocompleteParams. Unfortunately it meant more duplicate types and props between the old and new than I'd really like, but this way at least means we can have the feature flag and keep the old way around for now. But I'm open to suggestion.

I also fixed up the code that initialises the autocompleter and uses it, moving to packages/autocomplete in the process. This way the common code can be reused between the cli and browser repls. It lives in packages/autocomplete. The idea is we'd remove the old autocomplete code from there and keep this new wrapper.

Then to validate that the above design works I ported usage of autocomplete in browser-runtime-core to do the same thing we do in cli-repl. I ported its tests too, so you can do:

```
~/mongo/mongosh/packages/browser-runtime-core % USE_NEW_AUTOCOMPLETE=true npm test
```

I also narrowed some types along the way just to make porting tests a bit easier without requiring all of ShellInstanceState, for example.

### DIFF
--- a/packages/autocomplete/src/index.ts
+++ b/packages/autocomplete/src/index.ts
@@ -1,5 +1,6 @@
 import type { Topologies, TypeSignature } from '@mongosh/shell-api';
 import { signatures as shellSignatures } from '@mongosh/shell-api';
+import type { AutocompletionContext } from '@mongodb-js/mongodb-ts-autocomplete';
 import semver from 'semver';
 import {
   CONVERSION_OPERATORS,
@@ -13,6 +14,7 @@ import {
   ON_PREM,
   DATABASE,
 } from '@mongodb-js/mongodb-constants';
+import { directCommandCompleter } from './direct-command-completer';
 
 type TypeSignatureAttributes = { [key: string]: TypeSignature };
 
@@ -78,7 +80,7 @@ const GROUP = '$group';
  *
  * @returns {array} Matching Completions, Current User Input.
  */
-async function completer(
+export async function completer(
   params: AutocompleteParameters,
   line: string
 ): Promise<[string[], string, 'exclusive'] | [string[], string]> {
@@ -398,4 +400,51 @@ function filterShellAPI(
   return hits;
 }
 
-export default completer;
+type AutocompleteShellInstanceState = {
+  getAutocompleteParameters: () => AutocompleteParameters;
+  getAutocompletionContext: () => AutocompletionContext;
+};
+
+function transformAutocompleteResults(
+  line: string,
+  results: { result: string }[]
+): [string[], string] {
+  return [results.map((result) => result.result), line];
+}
+
+export type CompletionResults =
+  | [string[], string, 'exclusive']
+  | [string[], string];
+
+export async function initNewAutocompleter(
+  instanceState: Pick<
+    AutocompleteShellInstanceState,
+    'getAutocompletionContext'
+  >
+): Promise<(text: string) => Promise<CompletionResults>> {
+  // only import the autocompleter code the first time we need it to
+  // hide the time it takes from the initial startup time
+  const { MongoDBAutocompleter } = await import(
+    '@mongodb-js/mongodb-ts-autocomplete'
+  );
+
+  const autocompletionContext = instanceState.getAutocompletionContext();
+  const mongoDBCompleter = new MongoDBAutocompleter({
+    context: autocompletionContext,
+  });
+
+  return async (text: string): Promise<CompletionResults> => {
+    const directResults = await directCommandCompleter(
+      autocompletionContext,
+      text
+    );
+
+    if (directResults.length) {
+      return [directResults, text, 'exclusive'];
+    }
+
+    const results = await mongoDBCompleter.autocomplete(text);
+    const transformed = transformAutocompleteResults(text, results);
+    return transformed;
+  };
+}
--- a/packages/shell-api/src/decorators.ts
+++ b/packages/shell-api/src/decorators.ts
@@ -1,5 +1,6 @@
 import { MongoshInternalError } from '@mongosh/errors';
 import type { ReplPlatform } from '@mongosh/service-provider-core';
+import type { AutocompletionContext } from '@mongodb-js/mongodb-ts-autocomplete';
 import type { Mongo, ShellInstanceState } from '.';
 import type { Topologies } from './enums';
 import {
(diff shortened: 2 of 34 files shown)


packages/autocomplete/src/index.ts
```diff
@@ -78,7 +80,7 @@ const GROUP = '$group';
  *
  * @returns {array} Matching Completions, Current User Input.
  */
-async function completer(
+export async function completer(
```
lerouxb, line 83: I removed the default export, so you now either use `completer` (the old one) or `initNewAutocompleter` (the new one).

I know having "new" in there and having them asymmetrical like that is a bit ugly. Open to suggestions. I can go through these things and rename everything to old and new or we can live with the fact that this is temporary and clean it up when we remove the old one? Probably various other options.

packages/autocomplete/src/index.ts
```diff
@@ -398,4 +400,51 @@ function filterShellAPI(
...
+  results: { result: string }[]
+): [string[], string] {
+  return [results.map((result) => result.result), line];
+}
+
+export type CompletionResults =
+  | [string[], string, 'exclusive']
+  | [string[], string];
+
+export async function initNewAutocompleter(
```
lerouxb, line 419: I factored this out here because we have to do it for cli-repl and the browser one. The function we return is going to have to merge together the direct command completer results and the mongodb-ts-autocomplete results and I figured it is best to just do that once. Hence this thin wrapper that closes over the lazily imported & initialised state.

packages/shell-api/src/decorators.ts
```diff
@@ -350,12 +351,14 @@ function getShellInstanceState(apiObject: any): ShellInstanceState | undefined {
  * as needed.
  */
 export interface ShellCommandAutocompleteParameters {
-  getCollectionCompletionsForCurrentDb: (
-    collName: string
-  ) => string[] | Promise<string[]>;
   getDatabaseCompletions: (dbName: string) => string[] | Promise<string[]>;
 }
 
+export type NewShellCommandAutocompleteParameters = Pick<
```
lerouxb, line 357: Don't know how obvious it is from the diff, but this is just new types everywhere ShellCommand* was with that now the legacy type. We'll have to just clean this up when we remove the old autocompleter.

Again: open to suggestions for how to rather do this.

packages/shell-api/src/shell-api.ts
```diff
@@ -135,16 +136,38 @@ async function useCompleter(
...
 }
 
-/**
- * Complete a `show` subcommand.
- */
-// eslint-disable-next-line @typescript-eslint/require-await
-async function showCompleter(
-  params: ShellCommandAutocompleteParameters,
+// eslint-disable-next-line no-control-regex
+const CONTROL_CHAR_REGEXP = /[\x00-\x1F\x7F-\x9F]/g;
```
lerouxb, line 140: This is now the third copy of this regex..

packages/shell-api/src/shell-api.ts
```diff
@@ -135,16 +136,38 @@ async function useCompleter(
...
+    throw err;
+  }
+
+  const dbNames = await context.databasesForConnection(connectionId);
+
+  const prefix = (args[1] ?? '').toLowerCase();
+
+  return dbNames.filter(
+    (name) =>
+      name.toLowerCase().startsWith(prefix) && !CONTROL_CHAR_REGEXP.test(name)
```
addaleax, line 160: For use we'd probably want to be a bit stricter and only allow alphanumeric names, no?

packages/shell-api/src/shell-api.ts
```diff
@@ -135,16 +136,38 @@ async function useCompleter(
...
 ): Promise<string[] | undefined> {
   if (args.length > 2) return undefined;
-  if (args[1] === 'd') {
+
+  let connectionId: string;
+
+  try {
+    ({ connectionId } = context.currentDatabaseAndConnection());
+  } catch (err: any) {
+    if (err.name === 'MongoshInvalidInputError') {
```
addaleax: I still really think this interface should not be try/catch based and instead allow for returning `undefined`, the fact that we now have to do this in multiple places does make me want to push for this a bit harder

packages/autocomplete/src/index.ts
addaleax: This is fine, removing default exports is a good idea anyway

packages/snippet-manager/src/snippet-manager.ts
```diff
@@ -83,6 +83,36 @@ async function packBSON(data: any): Promise<Buffer> {
   return await brotliCompress(bson.serialize(data));
 }
 
+function completeSnippetsCommand(
+  args: string[],
+  snippets: SnippetManager['snippets']
```
addaleax:
```suggestion
  snippets: SnippetDescription[]
```

No need to refer to the internals of the class, right? (And if there was, I'd probably make this a static method on it)

packages/shell-api/src/shell-api.ts
lerouxb: I'll change it. I"m 50/50. I dislike special return values you have to check, but I agree that's better than this.

packages/snippet-manager/src/snippet-manager.ts
lerouxb: You're right. Not sure how I ended up with this. Probably copied from elsewhere when I factored it out or lazily went with autocomplete.

packages/shell-api/src/shell-api.ts
lerouxb: I just copied this from what we had, but I can change it.

packages/shell-api/src/shell-api.ts
addaleax: If you _do_ want to keep an exception-based approach, I'd definitely turn this into a specific exception that can be tested for instead of relying on the error name (that we also use in other places in mongosh). But ultimately, what we're saying _is_ that there is no current database and connection, so returning a null value for those cases does seem to match those semantics best

lerouxb: https://github.com/mongodb-js/devtools-shared/pull/543

addaleax (approved):

PR #2466, opened 2025-06-10
fix: provide license override for cli-table package MONGOSH-2248

addaleax:
This package has a license file but no `license` specified in its package.json file.

Broken since c8941c12a9457e8029.

### DIFF
--- a/licenses.json
+++ b/licenses.json
@@ -1,5 +1,7 @@
 {
   "ignoredOrgs": ["@mongodb-js", "@leafygreen-ui"],
   "doNotValidatePackages": ["argparse@2.0.1"],
-  "licenseOverrides": {}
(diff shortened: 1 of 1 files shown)


gagik (approved):
thanks for looking into this!

lerouxb (approved):
Thanks!

licenses.json
```diff
@@ -1,5 +1,7 @@
 {
   "ignoredOrgs": ["@mongodb-js", "@leafygreen-ui"],
   "doNotValidatePackages": ["argparse@2.0.1"],
-  "licenseOverrides": {}
+  "licenseOverrides": {
+    "cli-table@0.3.11": "MIT"
```
lerouxb, line 5: Good to know about this.

PR #2467, opened 2025-06-10
feat(shell-api): add collection.getShardLocation helper method MONGOSH-1996

nirinchev:
Adds the `collection.getShardLocation` helper that returns `{ shards: string[], sharded: boolean }` as a response. Under the hood calls 

```js
db.aggregate([$listClusterCatalog: {
    shards: true,
}]);
```

and filters the result to extract the information about the current collection.

### DIFF
--- a/.github/workflows/smoke-tests.yml
+++ b/.github/workflows/smoke-tests.yml
@@ -17,7 +17,7 @@ jobs:
         node: [20.x, 22.x, 23.x]
       fail-fast: false
     runs-on: ${{ matrix.runner }}-latest
-    timeout-minutes: 15
+    timeout-minutes: 20 # Installing dependencies on windows can take a while
     env:
       npm_config_loglevel: verbose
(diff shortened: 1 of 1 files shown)


packages/shell-api/src/collection.ts
```diff
@@ -2282,6 +2282,38 @@ export class Collection<
...
+    const result = await this._database.aggregate([
+      {
+        $listClusterCatalog: {
+          shards: true,
+        },
+      },
+    ]);
+
+    for await (const doc of result) {
+      if (doc.ns === this.getFullName()) {
```
addaleax: This could be a $match stage of the aggregation pipeline, right?

nirinchev: It could 🤦‍♂️ Clearly haven't moved on from Realm where everything was in-memory 😬

gagik (approved):

addaleax (approved):

PR #2475, opened 2025-06-16
feat(shell-api): add `.finish()` method to `ExplainableCursor` MONGOSH-1493

addaleax:


packages/shell-api/src/cursor.ts
```diff
@@ -62,7 +62,7 @@ export default class Cursor extends AggregateOrFindCursor<ServiceProviderFindCur
 
   @returnType('Cursor')
   @serverVersions([ServerVersions.earliest, '3.2.0'])
-  addOption(optionFlagNumber: number): Cursor {
+  addOption(optionFlagNumber: number): this {
```
addaleax, line 65: No real impact from these but it a) makes the types more correct and b) avoids having to cast in the integration test

Anemy (approved):

PR #2478, opened 2025-06-18
[WIP] use mql

lerouxb:


packages/build/src/download-center/config.ts
```diff
@@ -6,7 +6,7 @@ import { major as majorVersion } from 'semver';
 import type {
   DownloadCenterConfig,
   PlatformWithPackages,
-} from '@mongodb-js/dl-center/dist/download-center-config';
+} from '@mongodb-js/dl-center';
```
lerouxb, line 9: The new module resolution compiler option is not liking the /dist/ hack so I changed `@mongodb-js/dl-center` to just export the types.

packages/shell-api/src/collection.ts
```diff
@@ -96,6 +96,7 @@ import { HIDDEN_COMMANDS } from '@mongosh/history';
 import PlanCache from './plan-cache';
 import ChangeStreamCursor from './change-stream-cursor';
 import { ShellApiErrors } from './error-codes';
+import type * as mql from '@mongodb-js/mql-typescript/schema';
```
lerouxb, line 98: This is why I'm changing the moduleResolution ts compiler option. I want to be able to import a separate export from the default one.

packages/shell-api/src/database.ts
```diff
@@ -87,7 +87,7 @@ export class Database<
 > extends ShellApiWithMongoClass<M> {
   _mongo: Mongo<M>;
   _name: StringKey<M>;
-  _collections: Record<StringKey<D>, CollectionWithSchema<M, D>>;
+  _collections: { [k in StringKey<D>]: CollectionWithSchema<M, D, D[k], k> };
```
lerouxb, line 90: aah this was what I was looking for!

packages/shell-api/src/shard.ts
```diff
@@ -205,21 +205,21 @@ export default class Shard<
     }
   }
 
-  async _getConfigDB(): Promise<DatabaseWithSchema<M, M[keyof M]>> {
+  async _getConfigDB(): Promise<DatabaseWithSchema<M, M['config']>> {
```
lerouxb, line 208: I went back and forth on exactly this change and I seemed to just trade one set of errors for another. Although I agree that M['config'] is more correct - it is what I tried first.

packages/shell-api/src/helpers.ts
```diff
@@ -851,7 +849,7 @@ export function addHiddenDataProperty<T = any>(
 
 export async function iterate(
   results: CursorIterationResult,
-  cursor: AbstractCursor<any, any> | ChangeStreamCursor,
+  cursor: { isClosed(): boolean; tryNext(): Promise<Document | null> },
```
lerouxb, line 852: Good idea to specify smaller types. Didn't consider that.

PR #2481, opened 2025-06-19
fix: use on create with tags

gagik:
That was not it. 

Though I am not sure why pushing isn't covered (I suppose it'd need to be a push into an existing tag?), the fun long thread of issues people have been running into with https://github.com/actions/runner/issues/1007 seems relevant here.

Incorporated https://github.com/actions/runner/issues/1007#issuecomment-808904408 into our setup 🤞


addaleax (approved):
🤞

PR #2482, opened 2025-06-20
fix(cli-repl): account for Node.js 24 multi-line REPL handling update MONGOSH-2233

addaleax:
Still needs tests, but I think I got close to the bottom of this and this patch _should_ resolve this issue. Opening a PR already to see what CI says.

### DIFF
--- a/packages/cli-repl/src/mongosh-repl.ts
+++ b/packages/cli-repl/src/mongosh-repl.ts
@@ -47,7 +47,10 @@ import type { FormatOptions } from './format-output';
 import { markTime } from './startup-timing';
 import type { Context } from 'vm';
 import { Script, createContext, runInContext } from 'vm';
-import { installPasteSupport } from './repl-paste-support';
+import {
+  addReplEventForEvalReady,
+  installPasteSupport,
+} from './repl-paste-support';
 import util from 'util';
 
 declare const __non_webpack_require__: any;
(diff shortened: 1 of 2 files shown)


packages/cli-repl/src/smoke-tests.ts
```diff
@@ -120,6 +121,19 @@ export async function runSmokeTests({
       exitCode: 0,
       perfTestIterations: 0,
     },
+    {
+      // Regression test for MONGOSH-2233, included here because multiline support is a bit
+      // more fragile when it comes to newer Node.js releases and these are the only tests
+      // that run as part of the homebrew setup.
```
addaleax, line 127: Worth mentioning that our regular test suite does catch this (when run against Node.js 24, so we should add that to our CI soon too), we just don't run the relevant tests in homebrew.

packages/cli-repl/src/mongosh-repl.ts
```diff
@@ -260,7 +260,7 @@ class MongoshNodeRepl implements EvaluationListener {
       repl = asyncRepl.start({
         // 'repl' is not supported in startup snapshots yet.
         // eslint-disable-next-line @typescript-eslint/no-var-requires
-        start: require('pretty-repl').start,
+        //start: require('pretty-repl').start,
```
lerouxb: Do you want to leave this commented?

addaleax: Nope, thanks for catching this!

lerouxb (approved):

PR #2484, opened 2025-06-24
test: Re-apply streams e2e test and add retries to commands MONGOSH-2300

malexandert:


packages/e2e-tests/test/e2e-streams.spec.ts
```diff
@@ -0,0 +1,384 @@
...
+import type { TestShell } from './test-shell';
+import { sleep } from './util-helpers';
+import { eventually } from '../../../testing/eventually';
+
+const {
+  STREAMS_E2E_SPI_CONNECTION_STRING = '',
+  STREAMS_E2E_DB_USER = '',
+  STREAMS_E2E_DB_PASSWORD = '',
+  STREAMS_E2E_CLUSTER_CONNECTION_STRING = '',
+} = process.env;
```
malexandert, line 14: Something that confuses me is that between the OG commit and the revert, the `.evergreen.yml` got updated by a bot which [removed my env vars from `.evergreen.yml`](https://github.com/mongodb-js/mongosh/commit/b58ecda8550599e7948e9e23ff47f768aa1a200d), so reapplying the commit didn't put them back. However, the tests [still seem to run](https://parsley.mongodb.com/evergreen/mongosh_e2e_tests_amazon2023.0_arm64_small_e2e_tests_linux_arm64_patch_62609f535c9476f883a35776b2bcf2c46c256f79_685ae570fa9a410007526f30_25_06_24_17_50_42/0/task?bookmarks=0,9968&selectedLineRange=L9161-L9170), even though [they should skip and log that](https://github.com/mongodb-js/mongosh/pull/2484/files#diff-28e4a61eefaede0b794f731b96ddca4e2d0c4018f41bb1bf7e268769b99b6587R20-R48) if the right variables aren't present. So I guess having the right values in project settings without specifying them in the `.evergreen.yml` is sufficient?

addaleax: You'll need to add them to `.evergreen/evergreen.yml.in`, which the `.evergreen.yml` file is generated from

Having them in the project settings may currently work as a way of passing these env vars, but I'd definitely prefer to use an explicit way of passing them if we can

malexandert:
Quick bump on this @addaleax @himanshusinghs, would appreciate a review when you get a chance 🙏🏾 thanks in advance!

gagik (approved):
Not very familiar with streams but code looks reasonable to me.

PR #2485, opened 2025-06-25
feat: use placeholder mql types MONGOSH-2333

lerouxb:
The idea is that these would be easy to find/replace in the autocompleter while being semantically better than what we have and not adding any complexity for TypeScript.

### DIFF
--- a/packages/shell-api/src/collection.ts
+++ b/packages/shell-api/src/collection.ts
@@ -564,7 +564,7 @@ export class Collection<
     query: MQLQuery = {},
     projection?: Document,
     options: FindOptions = {}
-  ): Promise<C['schema'] | null> {
+  ): Promise<MQLDocument | null> {
     if (projection) {
       options.projection = projection;
     }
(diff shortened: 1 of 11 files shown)


packages/shell-api/src/collection.ts
```diff
@@ -560,7 +561,7 @@ export class Collection<
   @returnType('Document')
   @apiVersions([1])
   async findOne(
-    query: Document = {},
+    query: MQLQuery = {},
     projection?: Document,
     options: FindOptions = {}
   ): Promise<C['schema'] | null> {
```
addaleax: ^ should we also do something for the return types?

lerouxb: oh wow I totally missed that. Yes I think so.

lerouxb: I think this was actually the only usage of `C['schema']` and I changed it to `MQLDocument` now. But I can't easily think of other places that should change too 🤔

addaleax (approved):

PR #2489, opened 2025-06-26
chore: also bundle mongodb-schema and json-schema

lerouxb:
This came up once I bundled `@mongodb-js/mongodb-ts-autocomplete`

### DIFF
--- a/packages/shell-api/api-extractor.json
+++ b/packages/shell-api/api-extractor.json
@@ -15,7 +15,9 @@
     "@mongosh/arg-parser",
     "@mongosh/service-provider-core",
     "@mongosh/types",
-    "mongodb"
+    "mongodb",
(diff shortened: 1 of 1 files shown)


packages/shell-api/api-extractor.json
```diff
@@ -15,7 +15,9 @@
     "@mongosh/arg-parser",
     "@mongosh/service-provider-core",
     "@mongosh/types",
-    "mongodb"
+    "mongodb",
+    "mongodb-schema",
+    "@types/json-schema"
```
lerouxb, line 20: Not at all obvious, but this is actually a types-only package and json-schema is a "historical package" https://github.com/kriszyp/json-schema. Yet we import json-schema..

addaleax (approved):

PR #2501, opened 2025-07-07
feat: add skipStartupWarnings flag to skip warnings on demand MONGOSH-2371

dlombard:
Added a flag (`skipStartupWarning`) to skip the server's startup warnings

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ variable. For detailed instructions for each of our supported platforms, please 
         --eval [arg]                           Evaluate javascript
         --json[=canonical|relaxed]             Print result of --eval as Extended JSON, including errors
         --retryWrites[=true|false]             Automatically retry write operations upon transient network errors (Default: true)
-        --skipStartupWarnings                  Hides the startup warnings
+        --skipStartupWarnings                  Do not display server startup warnings
 
   Authentication Options:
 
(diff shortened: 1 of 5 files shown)


packages/cli-repl/src/mongosh-repl.spec.ts
```diff
@@ -1345,6 +1345,26 @@ describe('MongoshNodeRepl', function () {
...
+          sp.runCommandWithCheck
+            .withArgs(
+              ADMIN_DB,
+              {
+                getLog: 'startupWarnings',
+              },
+              {}
+            )
+            .resolves({ ok: 1, log: logLines });
+          // Make sure the connection info indicates a local Atlas server
```
addaleax: This comment is probably copied over and doesn't apply here?

README.md
```diff
@@ -44,6 +44,7 @@ variable. For detailed instructions for each of our supported platforms, please
         --eval [arg]                           Evaluate javascript
         --json[=canonical|relaxed]             Print result of --eval as Extended JSON, including errors
         --retryWrites[=true|false]             Automatically retry write operations upon transient network errors (Default: true)
+        --skipStartupWarnings                  Hides the startup warnings
```
addaleax: Let's maybe make it a bit more specific that this is about server warnings, not mongosh's own startup:

```suggestion
        --skipStartupWarnings                  Do not display server startup warnings
```

packages/cli-repl/src/mongosh-repl.spec.ts
dlombard: Correct. That's my bad. Removing it now

PR #2503, opened 2025-07-08
feat: allow opting out of schema sampling with config disableSchemaSampling=true MONGOSH-2370

lerouxb:
This is only relevant for the new autocompleter which is currently feature-flagged and only accessible if you run mongosh with the environment variable USE_NEW_AUTOCOMPLETE=true

Without it set we'll still read database and collection names like we've always done. With this set we'll just assume an empty collection/schema, so you won't see field names autocomplete.

### DIFF
--- a/packages/autocomplete/src/index.ts
+++ b/packages/autocomplete/src/index.ts
@@ -293,7 +293,9 @@ export async function completer(
 // from https://github.com/mongodb-js/devtools-shared/commit/e4a5b00a83b19a76bdf380799a421511230168db
 function satisfiesVersion(v1: string, v2: string): boolean {
   const isGTECheck = /^\d+?\.\d+?\.\d+?$/.test(v2);
-  return semver.satisfies(v1, isGTECheck ? `>=${v2}` : v2);
+  return semver.satisfies(v1, isGTECheck ? `>=${v2}` : v2, {
+    includePrerelease: true,
+  });
 }
 
 function isAcceptable(
--- a/packages/autocomplete/src/index.ts
+++ b/packages/autocomplete/src/index.ts
@@ -332,11 +332,6 @@ function isAcceptable(
       ? entry.env.includes(ATLAS)
       : entry.env.includes(ON_PREM));
 
-  console.log({
-    isAcceptableVersion,
-    isAcceptableEnvironment,
-    serverVersion: connectionInfo?.server_version,
-  });
   return isAcceptableVersion && isAcceptableEnvironment;
 }
 
--- a/packages/autocomplete/src/index.ts
+++ b/packages/autocomplete/src/index.ts
@@ -412,7 +412,7 @@ function filterShellAPI(
 type AutocompleteShellInstanceState = {
   getAutocompleteParameters: () => AutocompleteParameters;
   getAutocompletionContext: (options: {
-    disableSchemaSampling?: boolean;
+    disableSchemaSampling?: () => Promise<boolean>;
   }) => AutocompletionContext;
 };
 
@@ -433,9 +433,9 @@ export async function initNewAutocompleter(
     'getAutocompletionContext'
   >,
   {
-    disableSchemaSampling = false,
+    disableSchemaSampling,
   }: {
-    disableSchemaSampling?: boolean;
+    disableSchemaSampling?: () => Promise<boolean>;
   }
 ): Promise<(text: string) => Promise<CompletionResults>> {
   // only import the autocompleter code the first time we need it to
--- a/packages/autocomplete/src/index.ts
+++ b/packages/autocomplete/src/index.ts
@@ -411,9 +411,7 @@ function filterShellAPI(
 
 type AutocompleteShellInstanceState = {
   getAutocompleteParameters: () => AutocompleteParameters;
-  getAutocompletionContext: (options: {
-    disableSchemaSampling?: () => Promise<boolean>;
-  }) => AutocompletionContext;
+  getAutocompletionContext: () => AutocompletionContext;
 };
 
 function transformAutocompleteResults(
@@ -431,12 +429,7 @@ export async function initNewAutocompleter(
   instanceState: Pick<
     AutocompleteShellInstanceState,
     'getAutocompletionContext'
-  >,
-  {
-    disableSchemaSampling,
-  }: {
-    disableSchemaSampling?: () => Promise<boolean>;
-  }
+  >
 ): Promise<(text: string) => Promise<CompletionResults>> {
   // only import the autocompleter code the first time we need it to
   // hide the time it takes from the initial startup time
@@ -444,9 +437,7 @@ export async function initNewAutocompleter(
     '@mongodb-js/mongodb-ts-autocomplete'
   );
 
-  const autocompletionContext = instanceState.getAutocompletionContext({
-    disableSchemaSampling,
-  });
+  const autocompletionContext = instanceState.getAutocompletionContext();
   const mongoDBCompleter = new MongoDBAutocompleter({
     context: autocompletionContext,
   });
--- a/packages/autocomplete/src/index.ts
+++ b/packages/autocomplete/src/index.ts
@@ -411,7 +411,9 @@ function filterShellAPI(
 
 type AutocompleteShellInstanceState = {
   getAutocompleteParameters: () => AutocompleteParameters;
-  getAutocompletionContext: () => AutocompletionContext;
+  getAutocompletionContext: (options: {
+    disableSchemaSampling?: boolean;
+  }) => AutocompletionContext;
 };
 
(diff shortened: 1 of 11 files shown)


packages/cli-repl/src/mongosh-repl.ts
```diff
@@ -457,7 +457,11 @@ class MongoshNodeRepl implements EvaluationListener {
         (async () => {
           if (process.env.USE_NEW_AUTOCOMPLETE) {
             if (!newMongoshCompleter) {
-              newMongoshCompleter = await initNewAutocompleter(instanceState);
+              newMongoshCompleter = await initNewAutocompleter(instanceState, {
+                disableSchemaSampling: await this.getConfig(
+                  'disableSchemaSampling'
+                ),
```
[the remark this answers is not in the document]
addaleax: Yeah, Copilot isn't wrong here, this could be a `() => Promise<boolean>` instead of a `boolean` and everything would still work, only changes to the option take an effect more immediately, right?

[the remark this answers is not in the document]
lerouxb: How would you change the setting at runtime?

packages/cli-repl/src/mongosh-repl.spec.ts
```diff
@@ -481,14 +482,34 @@ describe('MongoshNodeRepl', function () {
           await tick();
           expect(output, output).to.include('db.coll.updateOne');
         });
-        // this will eventually be supported in the new autocomplete
-        it.skip('autocompletes collection schema fields', async function () {
-          input.write('db.coll.find({');
+        it('autocompletes collection schema fields', async function () {
+          if (!process.env.USE_NEW_AUTOCOMPLETE) {
+            // auto-completing collection field names only supported by new autocomplete
+            this.skip();
```
addaleax:
```suggestion
            this.skip();
            return;
```

packages/cli-repl/src/mongosh-repl.spec.ts
```diff
@@ -481,14 +482,34 @@ describe('MongoshNodeRepl', function () {
...
           await tabtab();
           await waitCompletion(bus);
           await tick();
           expect(output, output).to.include('db.coll.find({foo');
         });
+
+        it('does not autocomplete collection schema fields if disableSchemaSampling=true', async function () {
+          if (!process.env.USE_NEW_AUTOCOMPLETE) {
+            // auto-completing collection field names only supported by new autocomplete
+            this.skip();
```
addaleax:
```suggestion
            this.skip();
            return;
```

packages/cli-repl/src/mongosh-repl.ts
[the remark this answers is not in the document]
lerouxb: I'm not very keen on passing a function across 3 or 4 levels. Feels ugly.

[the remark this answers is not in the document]
addaleax: `config.set()`, for example

packages/shell-api/src/shell-instance-state.ts
```diff
@@ -483,17 +487,19 @@ export class ShellInstanceState {
...
-            ._getDb(databaseName)
-            .getCollection(collectionName)
-            ._getSampleDocsForCompletion();
-        } catch (err: any) {
-          if (
-            err?.code !== ShellApiErrors.NotConnected &&
-            err?.codeName !== 'Unauthorized'
-          ) {
-            throw err;
+        if (!disableSchemaSampling) {
```
addaleax:
```suggestion
        if ((await this.evaluationListener.getConfig?.('disableSchemaSampling')) === false) {
```

This does require moving the key from `CliUserConfig` to `ShellUserConfig` but that's fine

packages/cli-repl/src/mongosh-repl.ts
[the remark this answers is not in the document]
addaleax: See the suggestion below for a more direct way to solve this that doesn't require passing a function around 🙂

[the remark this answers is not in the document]
lerouxb: changed it now

packages/cli-repl/src/mongosh-repl.spec.ts
lerouxb: That return would be unreachable code.

packages/cli-repl/src/mongosh-repl.spec.ts
lerouxb: same

packages/shell-api/src/shell-instance-state.ts
```diff
@@ -483,17 +487,19 @@ export class ShellInstanceState {
...
-            ._getDb(databaseName)
-            .getCollection(collectionName)
-            ._getSampleDocsForCompletion();
-        } catch (err: any) {
-          if (
-            err?.code !== ShellApiErrors.NotConnected &&
-            err?.codeName !== 'Unauthorized'
-          ) {
-            throw err;
+        if (!(disableSchemaSampling ? await disableSchemaSampling() : false)) {
```
addaleax: If you'd prefer to keep the callback solution instead of directly reading the config value, I'd maybe phrase this as

```suggestion
        if (!(await disableSchemaSampling?.() ?? false)) {
```

that's probably just personal style, but I find this a bit easier to follow (i.e. it reads as "false" being the default value for `disableSchemaSampling`)

packages/cli-repl/src/cli-repl.spec.ts
```diff
@@ -2479,10 +2479,6 @@ describe('CliRepl', function () {
         // be listed
         wantWatch = true;
         wantShardDistribution = true;
-
-        // TODO: we need MQL support in mongodb-ts-autocomplete in order for it
-        // to autocomplete collection field names
-        wantQueryOperators = false;
```
lerouxb, line 2484: drive-by but I realised this test should actually be passing by now (with the env var set, obviously

packages/cli-repl/src/cli-repl.spec.ts
```diff
@@ -2509,9 +2510,25 @@ describe('CliRepl', function () {
         if (!(testServer as any)?._opts.args?.includes('--auth')) {
           // make sure there are some collections we can autocomplete on below
           input.write('db.coll.insertOne({})\n');
-          input.write('db.movies.insertOne({})\n');
+          input.write('db.movies.insertOne({ year: 1 })\n');
```
lerouxb, line 2512: Something I haven't really given much thought before, but the mql library won't autocomplete under fields that it doesn't know about:

```javascript
  const query: Query<{/* empty schema */}> = {
    year: {
      $g // you can't auto-complete here
    }
  }
```

packages/cli-repl/test/repl-helpers.ts
```diff
@@ -71,10 +71,7 @@ async function waitEval(bus: MongoshBus) {
 }
 
 async function waitCompletion(bus: MongoshBus) {
-  await Promise.race([
```
lerouxb, line 74: This timeout was just always going to be flaky and pressing tab twice doesn't necessarily mean it completed a potentially slow operation anyway. tabtab() always relied on 2 * 250ms being enough.

packages/cli-repl/src/mongosh-repl.ts
```diff
@@ -510,6 +505,8 @@ class MongoshNodeRepl implements EvaluationListener {
           results = results.filter(
             (result) => !CONTROL_CHAR_REGEXP.test(result)
           );
+          // emit here so that on nextTick the results should be output
```
lerouxb, line 511: OMG. autocomplete became async and there's another await.

packages/autocomplete/src/index.ts
```diff
@@ -293,7 +293,9 @@ export async function completer(
 // from https://github.com/mongodb-js/devtools-shared/commit/e4a5b00a83b19a76bdf380799a421511230168db
 function satisfiesVersion(v1: string, v2: string): boolean {
   const isGTECheck = /^\d+?\.\d+?\.\d+?$/.test(v2);
-  return semver.satisfies(v1, isGTECheck ? `>=${v2}` : v2);
+  return semver.satisfies(v1, isGTECheck ? `>=${v2}` : v2, {
+    includePrerelease: true,
```
lerouxb, line 297: 🤦

lerouxb:
```
> semver.satisfies('8.2.0-alpha-2686-g3770008', '>=8.0', {
...     includePrerelease: true,
...   });
true
>
> semver.satisfies('8.2.0-alpha-2686-g3770008', '>=8.0', {
...     includePrerelease: false, // the default
...   });
false
```

Didn't realise this was only failing on latest and that latest means latest alpha, not latest production release.

PR #2508, opened 2025-07-21
chore(ci): push mongosh tag separately during release

gagik:
Since GitHub Actions only runs create tag workflows when no more than 3 tags are created (https://docs.github.com/en/actions/reference/events-that-trigger-workflows#create), we will push the mongosh tag separately to ensure it triggers the merge tag workflow


addaleax (approved):
LGTM but Copilot's comment does seem to make sense :)

PR #2510, opened 2025-07-24
fix(ci): replace perf.send command with new API end point for performance data submission DEVPROD-19866

MAhmadShah:
The perf.send command is no longer being maintained and will be decommissioned soon. Users can now submit performance data using the "raw_perf_results" end point provided by the performance-monitoring-api. This PR migrates the mongosh evergreen project over to that API.


MAhmadShah:
Test patch can be found here: https://spruce.mongodb.com/version/687ff1bac119cd0007b2471d/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

PR #2511, opened 2025-07-29
chore: bump node driver to 6.18.0 MONGOSH-2413

gagik:
Bumps the node driver to the latest version.

Pending CI confirmation that no additional adjustments are needed.

PR #2520, opened 2025-08-20
chore(shell-api): add test and type updates for sort

gagik:
Followed up from the driver update, tests to ensure the new sort option works.

### DIFF
--- a/packages/service-provider-core/src/writable.ts
+++ b/packages/service-provider-core/src/writable.ts
@@ -319,7 +319,7 @@ export default interface Writable {
     collection: string,
     filter: Document,
     update: Document,
-    options?: UpdateOptions,
+    options?: UpdateOptions & { sort?: Document },
     dbOptions?: DbOptions
   ): Promise<UpdateResult>;
 
--- a/packages/shell-api/src/collection.ts
+++ b/packages/shell-api/src/collection.ts
@@ -930,7 +930,7 @@ export class Collection<
    * @param {Object} filter - The filter.
    * @param {Object} replacement - The replacement document for matches.
    * @param {Object} options - The replace options.
-   *    <upsert, writeConcern, collation, hint>
+   *    <upsert, writeConcern, collation, hint, sort>
    *
    * @returns {UpdateResult} The promise of the result.
    */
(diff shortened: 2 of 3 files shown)


packages/service-provider-core/src/writable.ts
```diff
@@ -319,7 +319,7 @@ export default interface Writable {
     collection: string,
     filter: Document,
     update: Document,
-    options?: UpdateOptions,
+    options?: UpdateOptions & { sort?: Document },
```
lerouxb, line 322: I was about to say "shouldn't that just be added to UpdateOptions?" but then realised it is also used for updateMany. Wondering if we should have a type next to UpdateOptions wherever it is defined that's `type UpdateOneOptions = UpdateOptions & { sort?: Document }` rather.

packages/shell-api/src/collection.ts
```diff
@@ -930,7 +930,7 @@ export class Collection<
    * @param {Object} filter - The filter.
    * @param {Object} replacement - The replacement document for matches.
    * @param {Object} options - The replace options.
-   *    <upsert, writeConcern, collation, hint>
+   *    <upsert, writeConcern, collation, hint, sort>
```
lerouxb, line 933: Shouldn't sort be added to the options that replaceOne takes? Or is it already on there? I suppose there's a test so somehow it has to be working.

packages/service-provider-core/src/writable.ts
lerouxb: But also shouldn't sort already be on an underlying driver type?

gagik: this is some weirdness from the driver: https://github.com/mongodb-js/mongosh/pull/2511#discussion_r2248144897

lerouxb: I'm wondering if we should at least make an UpdateOneOptions on our end so this is contained once, but that's not blocking.

lerouxb (approved):

packages/service-provider-core/src/writable.ts
gagik: for now I'd match what the driver does so we don't end up with our type system too deeply, also have some precedent for this (admittedly that was also done by me I believe..) https://github.com/mongodb-js/mongosh/blob/a845602e5961b86cf6c1ed7dd6fc04e65499428f/packages/shell-api/src/database.ts#L452

PR #2521, opened 2025-08-20
chore(e2e_tests): add 8.2.0-rc4 to e2e tests MONGOSH-2505

gagik:
`crypt_shared` is only available as an rc so not going to bump it, just adding tests to make sure there's nothing inherently problematic with publishing with the new PPA links.

### DIFF
--- a/.evergreen.yml
+++ b/.evergreen.yml
@@ -969,8 +969,8 @@ functions:
       params:
         aws_key: ${aws_key}
         aws_secret: ${aws_secret}
-        local_file: src/nyc-output-tests_darwin-m82xc_n20-cli_repl.tgz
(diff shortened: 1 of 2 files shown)


packages/build/src/packaging/download-crypt-library.ts
```diff
@@ -36,7 +36,7 @@ export async function downloadCryptLibrary(
   // (for the platforms that they exist for, i.e. for ppc64le/s390x only pick stable releases).
   let versionSpec = 'stable'; // TODO(MONGOSH-2192): Switch back to 'continuous' and deal with affected platform support.
   if (/ppc64|s390x/.test(opts.arch || process.arch)) {
-    versionSpec = 'stable';
+    versionSpec = '8.2.0';
```
lerouxb: I'm not sure I understand this, but ok.

lerouxb (approved):

PR #2525, opened 2025-08-28
chore(release) fix iam role used for dry run releases DEVPROD-21408

drichmdb:
This commit adjusts our release dry run task to use the correct IAM role for its operations. The non-dry-run release task was using the expected role, but dry-run wasn't assuming the role we need to use.


gagik (changes requested):
blocking the merge before release, LGTM generally though

PR #2530, opened 2025-09-05
fix(cli-repl): work around buggy Node.js autocomplete MONGOSH-2635

addaleax:
Work around the fact that, in some versions of the Node.js REPL,
tab completion may cause side effects.

(First commit is #2529)

### DIFF
--- a/packages/cli-repl/src/node-repl-fix-completer-side-effects.ts
+++ b/packages/cli-repl/src/node-repl-fix-completer-side-effects.ts
@@ -19,6 +19,8 @@ async function detectReplNodeBug59774(): Promise<boolean> {
   let ranFunction = false;
   repl.context.causeSideEffect = () => (ranFunction = true);
   try {
+    // Try to autocomplete a random property access on the result of calling
+    // causeSideEffect(). If the function is actually called, then the bug is present.
     await promisify(repl.completer)('causeSideEffect().x');
   } finally {
     repl.close();


packages/cli-repl/src/node-repl-fix-completer-side-effects.ts
```diff
@@ -0,0 +1,73 @@
...
+  const repl = replStart({
+    input: new PassThrough(),
+    output: new PassThrough(),
+    terminal: true,
+    useGlobal: false,
+  });
+  let ranFunction = false;
+  repl.context.causeSideEffect = () => (ranFunction = true);
+  try {
+    await promisify(repl.completer)('causeSideEffect().x');
```
gagik, line 24: I'm a bit confused by this, why is it trying to access `.x` in particular? Where is that coming from?

addaleax: I've added a comment here – it doesn't matter which property we try to autocomplete, the important bit is that is is trying to autocomplete _a_ property of the function call result (which it shouldn't evaluate)

gagik (approved):

packages/cli-repl/src/node-repl-fix-completer-side-effects.ts
gagik: got it, thank you! the comment is helpful too

PR #2532, opened 2025-09-10
feat: add a download script to facilitate easy use in CI MONGOSH-2687

FGasper:


download_latest.sh
```diff
@@ -0,0 +1,74 @@
...
+
+case "$arch" in
+    amd64|x86_64)
+        arch=x64
+        ;;
+    aarch64)
+        arch=arm64
+esac
+
+urls=$(curl -fsSL https://api.github.com/repos/mongodb-js/mongosh/releases/latest | jq -r '.assets[] | .browser_download_url' | grep -v -e \.sig -e shared -e openssl)
```
addaleax: Let's use https://downloads.mongodb.com/compass/mongosh.json for this rather than GH

download_latest.sh
```diff
@@ -0,0 +1,74 @@
...
+        echo >&2 "This script does not support this OS ($os). Download mongosh manually."
+        exit 1
+esac
+
+case "$arch" in
+    amd64|x86_64)
+        arch=x64
+        ;;
+    aarch64)
+        arch=arm64
```
addaleax, line 53: Just for completeness, we also support ppc64le and s390x

FGasper: Indeed … I just didn’t have access to such machines. Do you know what `uname -m` reports for those?

addaleax: Those are the uname strings (i.e. literal `s390x` and `ppc64le`) 🙂

FGasper: `uname` on some distros gets creative. `aarch64` versus `arm64`, etc.

Anyhow, I’ll update.

download_latest.sh
FGasper: Definitely better, yes. Adjusted jq query accordingly.

download_latest.sh
FGasper: Actually, _does_ this need any explicit support for those architectures, if the uname-reported arch is the same as what goes into the jq query?

README.md
```diff
@@ -22,6 +22,12 @@ Once downloaded, you will have to extract the binary and add it to your PATH
 variable. For detailed instructions for each of our supported platforms, please visit
 [installation documentation](https://www.mongodb.com/docs/mongodb-shell/install#mdb-shell-install).
 
+You can also run `download_latest.sh` to download a `mongosh` binary. You can use
+this script without cloning the repository thus:
+```
+curl -sSL https://raw.githubusercontent.com/mongodb-js/mongosh/refs/heads/main/download_latest.sh | sh
```
addaleax:
```suggestion
curl -fsSL https://raw.githubusercontent.com/mongodb-js/mongosh/refs/heads/main/download_latest.sh | sh
```
?

download_latest.sh
addaleax: Nope, you're right about that – Sorry! 🙂

download_latest.sh
```diff
@@ -0,0 +1,85 @@
...
+    echo >&2 "No download found for $os on $arch; download manually."
+    exit 1
+fi
+
+case "$ext" in
+    zip)
+        file=$(mktemp)
+
+        echo "Downloading $url to $file …"
+        trap 'rm -f $file' EXIT
```
addaleax:
```suggestion
        trap 'rm -f "$file"' EXIT
```

download_latest.sh
```diff
@@ -0,0 +1,85 @@
...
+
+        curl -fsSL "$url" > "$file"
+        echo "Downloaded $ext file; extracting mongosh …"
+
+        unzip -j "$file" '*/mongosh'
+        ;;
+    tgz)
+        echo "Downloading & extracting from $url …"
+
+        curl -fsSL "$url" | tar -xzf - \
```
addaleax, line 103:
```suggestion
        curl -fsSL "$url" | tar -xz \
```

download_latest.sh
```diff
@@ -0,0 +1,85 @@
...
+            --transform "s/.*\///" \
+            --wildcards "**/mongosh"
```
addaleax: We'd probably want `mongosh_crypt_v1.so` to be extracted next to mongosh

download_latest.sh
FGasper: Hm … macOS `man tar` says `-f`’s default varies across systems.

OK to leave this in?

download_latest.sh
FGasper: I should maybe know this, but: why so? What breaks if it’s not there? (It worked for me without.)

addaleax: Automatic encryption for Queryable Encryption & CSFLE don't – not everybody cares about those, but I don't think we'd like to ship "partially broken" versions of mongosh if we don't need to

download_latest.sh
addaleax: Sure!

download_latest.sh
FGasper: OK, the script now downloads both the executable and the shared library.

download_latest.sh
```diff
@@ -0,0 +1,96 @@
...
+        ;;
+    aarch64)
+        arch=arm64
+        ;;
+    *)
+        # Use uname’s reported architecture in the jq query.
+esac
+
+jq_query=$(cat <<EOF
+.versions[0].downloads[] |
```
addaleax, line 70: We can't really do a semver comparison here, right? Otherwise I think that would be highly preferable.

If we leave this as-is, I'd leave a comment in https://github.com/mongodb-js/mongosh/blob/e9dfc410785c173253788d3653af76fb5095029c/packages/build/src/download-center/config.ts#L205 about this script depending on that specific ordering

download_latest.sh
```diff
@@ -0,0 +1,96 @@
...
+select(
+    .distro == "$os" and
+    .arch == "$arch" and
+    (has("sharedOpenssl") | not)
```
addaleax: Generally the sharedOpenssl versions are mildly preferable – we could test for the presence on the current system through e.g. `ldd $(which curl) | grep -q libssl.so.3`?

download_latest.sh
```diff
@@ -0,0 +1,96 @@
...
+    echo >&2 "❓ No download found for $os on $arch."
+    show_download_link
+    exit 1
+fi
+
+case "$ext" in
+    zip)
+        file=$(mktemp)
+
+        echo "Downloading $url to $file …"
```
addaleax: Would keep using `>&2` like above for all diagnostic information

download_latest.sh
```diff
@@ -0,0 +1,96 @@
...
+        | sed -E 's/^.*[/]//'
+
+        ;;
+    *)
+        echo >&2 "Bad file extension: $ext"
+        show_download_link
+        exit 1
+esac
+
+echo "✅ Success! 'mongosh' and its crypto library are now saved in this directory."
```
addaleax: (including this line)

README.md
```diff
@@ -22,6 +22,12 @@ Once downloaded, you will have to extract the binary and add it to your PATH
 variable. For detailed instructions for each of our supported platforms, please visit
 [installation documentation](https://www.mongodb.com/docs/mongodb-shell/install#mdb-shell-install).
 
+You can also run `download_latest.sh` to download a `mongosh` binary. You can use
+this script without cloning the repository thus:
```
addaleax: I'd mention `npx mongosh` here, since that's probably easier for most people who don't truly need a local installation

download_latest.sh
FGasper: Are they preferable because we want updates from system library maintenance?

I get antsy testing for anything OpenSSL-related. macOS’s default `curl` actually uses SecureTransport, so I’m not sure testing `curl` is the best approach.

Does mongosh always require OpenSSL?

download_latest.sh
FGasper: fixed

download_latest.sh
FGasper: fixed

README.md
FGasper: I pushed something … it may not be the best verbiage.

download_latest.sh
FGasper: Would it maybe suffice merely to output a sufficiently-loud warning in the console that the downloaded `mongosh` uses a static OpenSSL & thus is only recommended for short-term usage?

addaleax: I've pushed a change that does this – only for OpenSSL 3, but I think that's fine, since most new systems would have that anyway.

We do care mostly because of FIPS support, but yes, system library maintenance upgrades also a big plus here.

On macOS and Windows, mongosh always uses a bundled OpenSSL library, so that's a non-concern, we only care on Linux.

FGasper: sgtm. Thank you!

download_latest.sh
FGasper: (@addaleax added the comment since jQ is a funky—if poetic, heh—place to do things like semver sorting.)

README.md
```diff
@@ -22,6 +22,15 @@ Once downloaded, you will have to extract the binary and add it to your PATH
 variable. For detailed instructions for each of our supported platforms, please visit
 [installation documentation](https://www.mongodb.com/docs/mongodb-shell/install#mdb-shell-install).
 
+Alternatively:
+- Run `npx mongosh` to run mongosh without a full installation. This is
+  easiest if you already have npm installed.
+- Run `download_latest.sh` to download a `mongosh` binary. You can use
+  the following script:
+```sh
+curl -fsSL https://raw.githubusercontent.com/mongodb-js/mongosh/refs/heads/main/download_latest.sh | sh
```
lerouxb, line 31: Do you have an example of what you're using to test this? I tried this locally by pointing at your branch:

```
% curl -fsSL https://raw.githubusercontent.com/FGasper/mongosh/refs/heads/add_download_script/download_latest.sh | sh
jq: error: syntax error, unexpected ')' (Unix shell quoting issues?) at <top-level>, line 6:
) |
jq: 1 compile error
curl: (56) Failure writing output to destination, passed 8192 returned 0
```

lerouxb: Works now with Anna's macOS fix 🎉

.github/workflows/test-download-latest.yml
```diff
@@ -0,0 +1,36 @@
+name: "Test download_latest script"
+on:
+  push:
+    branches:
+      - main
+  pull_request:
```
gagik, line 8: would it make sense to have this run every day etc. as a health check?

addaleax: @gagik Sure: 2a71d4959934cb471aa747cb688add9218dfd855

gagik (approved):

PR #2534, opened 2025-09-12
WIP chore(e2e-tests): add queryable encryption suffix/prefix/substring tests MONGOSH-1351

gagik:
Very WIP - Will probably need a few iterations

### DIFF
--- a/packages/build/src/packaging/download-crypt-library.ts
+++ b/packages/build/src/packaging/download-crypt-library.ts
@@ -35,10 +35,13 @@ export async function downloadCryptLibrary(
   // Download mongodb for latest server version, including rapid releases
   // (for the platforms that they exist for, i.e. for ppc64le/s390x only pick stable releases).
   let versionSpec = '8.0.12'; // TODO(MONGOSH-2192): Switch back to 'continuous' and deal with affected platform support.
-  if (/ppc64|s390x/.test(opts.arch || process.arch)) {
+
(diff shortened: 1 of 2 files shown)


packages/build/src/packaging/download-crypt-library.ts
```diff
@@ -35,10 +35,13 @@ export async function downloadCryptLibrary(
   // Download mongodb for latest server version, including rapid releases
   // (for the platforms that they exist for, i.e. for ppc64le/s390x only pick stable releases).
   let versionSpec = '8.0.12'; // TODO(MONGOSH-2192): Switch back to 'continuous' and deal with affected platform support.
-  if (/ppc64|s390x/.test(opts.arch || process.arch)) {
+
+  // For 8.2.0-rc4, we use the equivalent crypt shared library version for testing.
+  if (process.env.MONGOSH_SERVER_TEST_VERSION === '8.2.0-rc4-enterprise') {
+    versionSpec = '8.2.0-rc4';
```
addaleax, line 41: Fwiw, I'd have a mild preference for testing this in a way that would be reflective of real user behavior, i.e. adding the ability to pass through a version of the crypt_shared library to download through this function, then downloading that file and using it for only the QE text search tests.

If not that, then I'd maybe spent a minute to think about whether there's an easy way to make sure that this branch here _only_ is taken in tests, because setting this environment variable could be done quite easily on accident

addaleax:
Taken care of as part of https://github.com/mongodb-js/mongosh/pull/2427

PR #2537, opened 2025-09-17
chore(shell-bson): split shell BSON logic into separate package

addaleax:
This is admittedly a large refactor, but one that I would believe is generally worth it:

- Moving the BSON print handling out of `service-provider-core` makes sense, as `service-provider-core` is generally intended to be a library providing an *interface*, not partial implementations.
- Moving the BSON classes out of `shell-api` is less architecturally important, but it does provide a little bit of additional encapsulation. It does, for example, force us to separate mongosh-specific `.help` property from core BSON constructor logic.
- A number of files in other packages were using `@mongosh/service-provider-core` only as a convenient way of getting the current `bson` package, which works but is really more of a hack and there's no good reason not to just use the `bson` package that the driver team ships directly.

This is also convenient on a broader basis, though, because it makes mongosh's BSON constructors generally available for other projects with similar needs to use. In Compass, we are avoiding this by using a Babel-based "manual JavaScript evaluator", but not all projects have the same technical requirements as Compass, and some may prefer the flexibility of a full-fledged JS execution environment like mongosh does.


packages/service-provider-core/src/printable-bson.ts
addaleax, line 1: This file was moved, but the modifications were a little too extensive for the review display to follow

packages/shell-api/src/helpers.ts
```diff
@@ -1226,75 +1222,6 @@ export function shallowClone<T>(input: T): T {
...
-  return Object.create(
-    Object.getPrototypeOf(o),
-    Object.fromEntries(
-      Object.entries(Object.getOwnPropertyDescriptors(o)).filter(([k]) =>
-        (keys as string[]).includes(k)
-      )
-    )
-  );
-}
-
```
addaleax, line 1297: Extracted into a separate file

packages/shell-bson/README.md
addaleax, line 1: Probably not a bad idea to start a review by looking at this file and trying the example code here 🙂

packages/shell-bson/src/helpers.ts
```diff
@@ -0,0 +1,123 @@
...
+    )
+  );
+}
+
+function getAssertCaller(caller?: string): string {
+  return caller ? ` (${caller})` : '';
+}
+
+// NB: Duplicate of the one in packages/shell-api/src/helpers.ts
+export function assertArgsDefinedType(
```
addaleax, line 82: ^ Worth calling out the duplication here, I think

packages/shell-bson/src/shell-bson.spec.ts
```diff
@@ -1,26 +1,48 @@
...
 
+interface TestHelp {
+  testHelpName: string;
+  type: 'Help';
+}
+const ALL_SERVER_VERSIONS = ['1.0'];
+
 describe('Shell BSON', function () {
-  let shellBson: any;
+  let shellBson: ShellBson<BSON, TestHelp>;
```
addaleax, line 25: A lot of the changes in this file are because I naively thought this doesn't need to be `any`, it's quite noisy and introduces `as any` calls in other places so I honestly also wouldn't mind going back to the previous variant

packages/shell-bson/src/helpers.ts
```diff
@@ -0,0 +1,123 @@
...
+  T extends Function & { new (...args: any): any }
+>(
+  ClassCtor: T
+): {
+  new (...args: ConstructorParameters<T>): InstanceType<T>;
+  (...args: ConstructorParameters<T>): InstanceType<T>;
+} {
+  function fnCtor(...args: any[]) {
+    if (new.target) {
+      return Reflect.construct(ClassCtor, args, new.target);
```
gagik, line 17: today I learnt 😮

packages/shell-bson/src/shell-bson.spec.ts
gagik: either is fine with me, since this is slightly more typed we can just stick with this 🙂

packages/shell-bson/src/shell-bson.ts
```diff
@@ -99,37 +117,28 @@ export default function constructShellBson(
...
-  (bson.BSONSymbol as any).prototype.serverVersions = [
-    ServerVersions.earliest,
-    '1.6.0',
-  ];
-  (bson.BSONSymbol as any).prototype.deprecated = true;
+    if (!('prototype' in bson[className])) continue;
+    assignMetadata?.(bson[className].prototype, {
+      help: help,
+      // Symbol is deprecated
+      ...(className === 'BSONSymbol'
```
gagik, line 135: not needed for this PR but makes me think for future if we should aim to have a single point of reference for different types of deprectations across mongosh. I know we're also ideally looking for this information to come from driver/bson(?) in the future so it'd probably make transition easier.

gagik (approved):
This is a nice improvement 🙂 Many changes but all seems good to me

packages/shell-bson/src/shell-bson.ts
addaleax: Yeah, it would be awesome to have something standardized ... might require a bit of brainstorming together with the driver team about the approach 🙂

PR #2538, opened 2025-09-17
chore(ci): partially fix vscode test

addaleax:
Right now, the VSCode test is failing in CI because the versions of `@mongodb-js/devtools-connect` mismatch. We can align them the same way we aligned the driver versions.

This does not turn the task green yet, but it is a necessary step (the next one would be adjusting the VSCode source code for the removal of the argument to `serviceProvider.close()` from c8e2697e).


gagik (approved):
Thanks for looking into this! I can try to do the VSCode adjustments

PR #2544, opened 2025-10-03
fix(shell-api,shell-bson): increase runtime independence

addaleax:
Make the `shell-bson` and `shell-api` packages more runtime-independent (as they are supposed to be) by using `util` and `crypto` as progressive enhancement enablers rather than required packages.

This specifically makes `shell-bson` fully usable in a bare-bones JS environment.


lerouxb (approved):
Looks good, although it does feel like ideally these should be pluggable so someone could still make something like inspect work the way they want if they care about it. Rather than it be either hardcoded as the common-js-imported module OR the hardcoded fallback.

kmruiz (approved):

PR #2547, opened 2025-10-06
fix(logging): fix log flushing on exit MONGOSH-2883

addaleax:
Depends on https://github.com/mongodb-js/devtools-shared/pull/578

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -25890,9 +25890,9 @@
       }
     },
     "node_modules/mongodb-log-writer": {
-      "version": "2.4.1",
-      "resolved": "https://registry.npmjs.org/mongodb-log-writer/-/mongodb-log-writer-2.4.1.tgz",
-      "integrity": "sha512-kTVWtiUbayr2S54WeOeHpXvR80ASwlmoMsA3LIxH+PVZle8ddq7cXJXM3O5kkuT+Uni9+YNOTBwoRYVQlIAEUQ==",
+      "version": "2.4.3",
+      "resolved": "https://registry.npmjs.org/mongodb-log-writer/-/mongodb-log-writer-2.4.3.tgz",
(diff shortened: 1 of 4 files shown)


packages/e2e-tests/test/e2e.spec.ts
```diff
@@ -2182,6 +2177,46 @@ describe('e2e', function () {
...
+            args: [
+              '--nodb',
+              '--eval',
+              'for(let i=0; i < 10; i++) log.info("logging", {i}); print(log.getPath()); throw new Error("uh oh")',
+            ],
+          });
+          expect(await shell.waitForAnyExit()).to.equal(1);
+          const log = await readReplLogFile(
+            shell.output.replace(/Error: uh oh/g, '').trim()
+          );
```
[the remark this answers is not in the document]
addaleax: I mean, I had similar thoughts here as well, although I was ultimately feeling like this is just fine for a test

packages/cli-repl/src/cli-repl.ts
```diff
@@ -659,7 +671,7 @@ export class CliRepl implements MongoshIOProvider {
     if (enabled) {
       await this.startLogging();
     } else {
-      this.loggingAndTelemetry?.detachLogger();
+      await this.loggingAndTelemetry?.detachLogger();
```
gagik, line 674: for future, we should probably figure out why we don't have a lint rule for this

addaleax: @gagik We do 🙂 This is because I changed the type of `.detachLogger()` to return a `Promise` in this PR, too (in fact, the lint rule helped with catching all of these cases)

gagik: ah yes

PR #2709, opened 2026-04-08
fix(node-runtime-worker): do not relax long and other non int or double numbers COMPASS-6714

Anemy:
COMPASS-6714

| before | after |
| - | - |
| Screenshot 2026-04-10 at 07 01 06 | Screenshot 2026-04-10 at 07 01 13 |

### DIFF
--- a/packages/node-runtime-worker-thread/src/serializer.ts
+++ b/packages/node-runtime-worker-thread/src/serializer.ts
@@ -91,10 +91,38 @@ export function serializeEvaluationResult({
     type: 'SerializedShellApiResult',
     printable: {
       origType: type,
-      serializedValue: EJSON.serialize(printable),
+      serializedValue: EJSON.serialize(printable, {
+        relaxed: false,
+      }),
       constructionOptions,
     },
   };
+}
+
+/**
+ * After we EJSON.deserialize, Int32 and Double values come
+ * back as BSON objects. We want to show them as regular numbers,
+ * similar to the cli repl, so we relax them back to numbers.
+ * Things like Long values we want as BSON objects since they can exceed
+ * Number.MAX_SAFE_INTEGER and would lose precision as numbers.
+ */
+function relaxSafeNumericTypes(value: any): any {
+  if (value === null || typeof value !== 'object') return value;
+  if (value._bsontype === 'Int32' || value._bsontype === 'Double') {
+    return value.valueOf();
+  }
+  // Don't recurse into other BSON types (Long, ObjectId, Binary, etc.)
+  if (value._bsontype) return value;
+  if (Array.isArray(value)) {
+    for (let i = 0; i < value.length; i++) {
+      value[i] = relaxSafeNumericTypes(value[i]);
+    }
+    return value;
+  }
+  for (const key of Object.keys(value)) {
+    value[key] = relaxSafeNumericTypes(value[key]);
+  }
+  return value;
 }
 
 export function deserializeEvaluationResult({
@@ -115,7 +143,9 @@ export function deserializeEvaluationResult({
   if (type === 'SerializedShellApiResult') {
     return {
       type: printable.origType,
-      printable: EJSON.deserialize(printable.serializedValue),
+      printable: relaxSafeNumericTypes(
+        EJSON.deserialize(printable.serializedValue, { relaxed: false })
+      ),
       source,
       constructionOptions,
     };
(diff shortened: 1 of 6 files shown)


packages/node-runtime-worker-thread/src/serializer.ts
```diff
@@ -91,12 +91,40 @@ export function serializeEvaluationResult({
...
+/**
+ * After we EJSON.deserialize, Int32 and Double values come
+ * back as BSON objects. We want to show them as regular numbers,
+ * similar to the cli repl, so we relax them back to numbers.
+ * Things like Long values we want as BSON objects since they can exceed
+ * Number.MAX_SAFE_INTEGER and would lose precision as numbers.
+ */
+function relaxSafeNumericTypes(value: any): any {
+  if (value === null || typeof value !== 'object') return value;
+  if (value._bsontype === 'Int32' || value._bsontype === 'Double') {
```
gribnoysup: Take a look at getBsonType method in hadron-type-checker. I think accessing _bsontype is highly not recommended anymore (if it even works) in recent bson versions, at least I saw Anna replacing all those checks in compass

Anemy: Updated! Thanks for catching.

packages/node-runtime-worker-thread/src/serializer.ts
```diff
@@ -91,12 +92,41 @@ export function serializeEvaluationResult({
...
+ * similar to the cli repl, so we relax them back to numbers.
+ * Things like Long values we want as BSON objects since they can exceed
+ * Number.MAX_SAFE_INTEGER and would lose precision as numbers.
+ */
+function relaxSafeNumericTypes(value: any): any {
+  if (value === null || typeof value !== 'object') return value;
+  const bsonType = getBsonType(value);
+  if (bsonType === 'Int32' || bsonType === 'Double') {
+    return value.valueOf();
+  }
```
nbbeeken, line 115: https://github.com/mongodb/js-bson/blob/0712bb15653093766a80bbd4ba104353cd4581e3/src/double.ts#L90-L102

Perhaps overkill but should we replicate the way doubles are rendered in Ext json, could be useful to see a ".0" even though it is insignificant to the value.

packages/node-runtime-worker-thread/src/serializer.ts
```diff
@@ -91,12 +92,41 @@ export function serializeEvaluationResult({
...
+  // Don't recurse into other BSON types (Long, ObjectId, Binary, etc.)
+  if (bsonType) return value;
```
nbbeeken, line 117: Like here https://github.com/mongodb/js-bson/blob/0712bb15653093766a80bbd4ba104353cd4581e3/src/parser/deserializer.ts#L347-L350 we could only render `Long` when the value falls outside of the `MAX/MIN_SAFE_INTEGER` range. Thoughts?

Anemy: As far as I can tell regular mongosh cli repl isn't doing this, so I'm thinking it's better if we don't intentionally relax here. An aside, I'm in the show things in the canonical format camp when we have the opportunity to, so that folks see the type they inserted with and they don't end up changing a type or losing precision unexpectedly.

packages/node-runtime-worker-thread/src/serializer.ts
Anemy: I'm not sure I follow, are you suggesting we show the .0 for doubles? That sounds like it would be different from what mongosh's cli repl is doing. The .valueOf here should maintain the -0 typing which TIL javascript supports.

nbbeeken: > are you suggesting we show the .0 for doubles?

Yes I was but yea it would be different so never mind 🤔  but was it not already different that this was rendering Longs as plain numbers? normal Mongosh doesn't do that right? 

This is fine as is :)

packages/node-runtime-worker-thread/src/serializer.ts
nbbeeken: Yea, I agree with that, type information loss is a different form of precision loss. There's no way around it some of the time, in BSON we use Number.isInteger and its magnitude to decide whether the plain JS number is going to be an int, long, or double. So `0.7` is a double, `0.7 + 0.3` is an int32. `0x8000_0000` is a Long, `0x8000_0000 - 1` is an int32. 

Plain numbers don't have the associated type information so there's always going to be these hazards, but quite often, they do not matter, its just when they do it can be surprising. 

Anyway we def shouldn't be making the embedded shell special, so never mind again. :)

packages/node-runtime-worker-thread/src/serializer.ts
Anemy: Normal mongosh renders long values as strings, which is what this pr updates the browser-repl to do. Currently browser-repl was rendering them as numbers which made it lose precision with the big numbers.

lerouxb (approved):

PR #2720, opened 2026-04-20
chore(tests): adjust tests to be compatible with 9.0

gagik:


packages/service-provider-node-driver/src/node-driver-provider.integration.spec.ts
```diff
@@ -759,15 +759,21 @@ describe('NodeDriverServiceProvider [integration]', function () {
...
           v: 2,
           clustered: true,
           unique: true,
-        });
+        };
+        // Starting in MongoDB 9.0, listIndexes includes expireAfterSeconds
+        // for clustered indexes even when no TTL is configured (SERVER-83221).
+        if (await testServer.serverSatisfies('>= 9.0.0')) {
+          expectedClusteredIndex.expireAfterSeconds = 0;
+        }
```
addaleax: Based on https://github.com/mongodb-js/compass/pull/8000 I'd guess this should be

```suggestion
        if (await testServer.serverSatisfies('>= 9.0.0-alpha0')) {
          expectedClusteredIndex.expireAfterSeconds = 0;
        }
```

That being said, I'd be good with not gating this behind a version check and just using 

```suggestion
        expectedClusteredIndex.expireAfterSeconds?? = 0;
```

to keep it simple

gagik: yeah actually not 100% sure if this change is specific to this alpha

addaleax (approved):

PR #2721, opened 2026-04-20
chore(ci): update OpenSSL 1.1 download URL MONGOSH-3316 MONGOSH-3317 MONGOSH-3318

addaleax:
I'm not sure if this was intentional on the OpenSSL side, but the previous download link for the 1.1.1o source tarball was broken. These links are just redirects to Github's CDN anyway, so pointing it at the Github source should be fine.


gagik (approved):

gagik:
Thanks for looking into this!

PR #2725, opened 2026-04-28
chore: add nightly Node.js variant MONGOSH-2969

nbbeeken:
MONGOSH-2969

Introduce a new nightly Node.js build variant


.evergreen/setup-env.sh
```diff
@@ -4,6 +4,43 @@ set -x
...
+    NODE_JS_VERSION=$(nvm version node | sed 's/^v//')
+    echo "$NODE_JS_VERSION" > "$RESOLVED_NIGHTLY_FILE"
+  fi
+  nvm use "v$NODE_JS_VERSION"
+  set -x
+  export PATH="$NVM_BIN:$PATH"
+  export NODE_JS_VERSION
+  export NODE_JS_VERSION_OVERRIDE="$NODE_JS_VERSION"
+  export USE_NIGHTLY_NODE=1
+fi
```
addaleax, line 37: nvm feels a bit like overkill for downloading and extracting from https://nodejs.org/download/nightly/ for a single architecture and target OS

(aside: the ticket itself wasn't asking for nightlies, but I don't mind using them as long as the team is okay with that overall)

.evergreen/setup-env.sh
```diff
@@ -43,11 +80,22 @@ if [ "$OS" != "Windows_NT" ]; then
...
     export PYTHON="/opt/devtools/bin/python3"
 
+    # /opt/devtools/bin/gcc is pinned at 12.4 for reproducible stable builds.
+    # Node.js v26+ needs GCC >= 13.2 (C++20 in V8 turboshaft), so for the
+    # nightly variant prefer the host's system compiler — RHEL 10 ships
+    # gcc 14.2 / clang 19.1 stock, both new enough.
+    if [ -n "$USE_NIGHTLY_NODE" ] && [ -x /usr/bin/gcc ]; then
+      export CC=/usr/bin/gcc
+      export CXX=/usr/bin/g++
+    fi
```
addaleax: Any reason not to update the toolchain?

packages/e2e-tests/test/e2e.spec.ts
```diff
@@ -63,6 +63,15 @@ describe('e2e', function () {
...
     it('shows build info in JSON format', async function () {
+      // TODO(MONGOSH-2969): Under Node.js nightlies the glibc-version
+      // native addon's dlsym for `gnu_get_libc_version` returns null on
+      // some builders (e.g. ubuntu2004), so mongosh reports
+      // runtimeGlibcVersion = "N/A" while Node.js's process.report still
+      // sees the host's glibc. Investigate the addon's loader behavior
+      // under the nightly's prebuilt binary and re-enable.
+      if (process.version.includes('-nightly')) {
+        return this.skip();
+      }
```
addaleax, line 74: Do we create separate tickets for these?

packages/e2e-tests/test/e2e.spec.ts
```diff
@@ -1358,6 +1375,14 @@ describe('e2e', function () {
         });
 
         it('runs scripts in the right environment', async function () {
+          // TODO(MONGOSH-2969): Node.js nightlies (v26+) report
+          // executionAsyncId() === 0 inside the REPL context, where stable
+          // Node returns > 1. Investigate whether this is an intentional
+          // async_hooks change in V8/Node or whether the REPL setup needs
+          // adjustment, and re-enable once classified.
```
addaleax: This is from https://github.com/nodejs/node/pull/61227

We can probably get away with just adjusting the test here

.evergreen/setup-env.sh
nbbeeken: > Node.js latest or nightly builds as an early warning system for breakages

I guess I just went with the second half of the OR and didn't think too much about it 😅 latest node would def be easier, but I'm wondering if we're a good canary in the coal mine with how deeply we integrate with the platform that nightlies could be more interesting, could also be more noisy...

.evergreen/setup-env.sh
nbbeeken: This isn't needed anymore, we can still file a follow up for a toolchain update

packages/e2e-tests/test/e2e.spec.ts
nbbeeken: Yes, created!

.evergreen/setup-env.sh
addaleax: We are a good canary and I do like that the PR chooses this approach, but yeah, we need to be okay with the fact that we'll likely end up getting more alerts than we otherwise would

addaleax (approved):
:shipit:

PR #2726, opened 2026-04-28
chore(cli-repl): add more CTA matching logic MONGOSH-2967

nbbeeken:


packages/build/src/config/config.ts
```diff
@@ -10,6 +10,11 @@ interface ManPageConfig {
 
 // This needs to match the interface in cli-repl/update-notification-manager.ts
 export interface GreetingCTADetails {
+  // Optional filter on buildInfo() fields. When set, the CTA is only shown to
+  // shells whose buildInfo matches every key (string = exact, array = any-of,
+  // boolean for boolean fields). When omitted, the CTA targets the version key
+  // alone (current behavior).
+  match?: { [field: string]: string | string[] | boolean };
```
addaleax: This works, but a regular expression on a JSON or `env`-like representation of the `buildInfo` struct seems like it would a) be strictly more flexible and b) simpler in terms of logic

nbbeeken: Makes sense 🙂

packages/cli-repl/src/update-notification-manager.ts
```diff
@@ -2,14 +2,49 @@ import semver from 'semver';
...
+      for (const [nestedKey, nestedValue] of Object.entries(value)) {
+        if (nestedValue === null || nestedValue === undefined) continue;
+        lines.push(`${key}.${nestedKey}=${String(nestedValue)}`);
+      }
+    } else {
+      lines.push(`${key}=${String(value)}`);
+    }
+  }
+  return lines.join('\n');
+}
```
addaleax:
```suggestion
function serializeBuildInfo(buildInfo: Record<string, unknown>): string {
  const lines: string[] = [];
  for (const [key, value] of Object.entries(buildInfo)) {
    if (typeof value === 'object' && value !== null) {
      for (const line of serializeBuildInfo(value))
        lines.push(`${key}.${line}`);
    } else {
      lines.push(`${key}=${String(value)}`);
    }
  }
  return lines.join('\n');
}
```

Simplifies this a bit and makes it work for arbitrary nesting (and allows matching on `false`-y keys)

packages/cli-repl/src/update-notification-manager.ts
```diff
@@ -2,14 +2,49 @@ import semver from 'semver';
...
+function ctaMatchesBuildInfo(
+  match: string | undefined,
+  buildInfo: BuildInfo | undefined
+): boolean {
+  if (!match) return true;
+  if (!buildInfo) return false;
+  try {
+    return new RegExp(match, 'm').test(serializeBuildInfo(buildInfo));
+  } catch {
+    return false;
```
addaleax, line 45: I know it requires a bit of extra wiring, but I'd consider logging this somehow ... because it's really a "should never happen" type of error

packages/cli-repl/src/update-notification-manager.ts
```diff
@@ -76,7 +111,11 @@ export class UpdateNotificationManager {
       /* already handled in fetchUpdateMetadata() */
     }
 
-    return this.currentVersionGreetingCTA?.chunks;
+    const cta = this.currentVersionGreetingCTA;
+    if (!cta) return undefined;
+    if (cta.match && !ctaMatchesBuildInfo(cta.match, buildInfo))
```
addaleax: `ctaMatchesBuildInfo()` already handles false-y `cta.match`, I'd drop it in one place just to make sure we have a single consistent branch point that handles this flag

addaleax (approved):
Looks good! Only a few (but completely optional) suggestions

packages/cli-repl/src/update-notification-manager.ts
nbbeeken: wired, I also figure we should catch people when they write new CTAs? so I scanned them and check that if there's a `match` it can be made into a regex

addaleax (approved):

PR #2727, opened 2026-04-29
fix(cli-repl): send OS information in update notification request MONGOSH-3332

addaleax:
Example UA strings:

```
mongosh/2.8.2; Darwin; 25.4.0; arm64; ; ; 129a101ab7e5259325c347829751b5196742ead74b7ed6fcb961dea44d7f30c7
mongosh/2.8.2; Darwin; 25.4.0; arm64; ; ; unknown
mongosh/2.8.2; Linux; 6.8.0-106-generic; x64; ubuntu; 24.04; 390cf97309e176725a53c683b54830eed768644426a0f7965465c90f8ed83261
```

### DIFF
--- a/packages/cli-repl/src/cli-repl.ts
+++ b/packages/cli-repl/src/cli-repl.ts
@@ -143,7 +143,11 @@ export class CliRepl implements MongoshIOProvider {
     useEnvironmentVariableProxies: true,
   };
   agent: AgentWithInitialize | undefined;
-  fetch: (url: string, init?: RequestInit) => Promise<Response>;
+  fetch: ({
+    includeDeviceId,
+  }: {
+    includeDeviceId: boolean;
+  }) => (url: string, init?: RequestInit) => Promise<Response>;
   updateNotificationManager: UpdateNotificationManager;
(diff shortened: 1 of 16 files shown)


packages/cli-repl/src/cli-repl.ts
```diff
@@ -194,8 +207,52 @@ export class CliRepl implements MongoshIOProvider {
...
+            // TODO: There should be a single way used throughout this class
+            // to check whether telemetry is enabled or not, instead of directly checking
+            // config values in some places and passing an `enabled` flag to
+            // `setTelemetryEnabled` in others.
+            if (
+              !includeDeviceId ||
+              this.forceDisableTelemetry ||
+              !(await this.getConfig('enableTelemetry'))
+            ) {
+              return 'unknown';
```
Anemy: Should we differentiate the disabled case from the case where `getDeviceIdForMongosh` fails?

addaleax: Yeah, I like that idea – done in ae54cf3c66d229c55a5194dc371598a02c817102 👍

gagik (approved):

PR #2729, opened 2026-05-12
chore: update readme with latest node version MONGOSH-3292

johnjackweir:
MONGOSH-3292

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,7 +127,7 @@ curl -fsSL https://raw.githubusercontent.com/mongodb-js/mongosh/refs/heads/main/
 
 ### Requirements
 
-- Node.js v20.x
+- Node.js v24.x
 
 ### Install
 


Anemy (approved):

README.md
```diff
@@ -127,7 +127,7 @@ curl -fsSL https://raw.githubusercontent.com/mongodb-js/mongosh/refs/heads/main/
 
 ### Requirements
 
-- Node.js v20.x
+- Node.js v24.x
```
Anemy, line 130: Let's also update the package.json for this?
https://github.com/mongodb-js/mongosh/blob/21afacc8deb95ee809af65675e21e3fd2ec4befc/package.json#L100 
Looks like a lot of the packages also have `node` in their `engines` and it's gotten out of date. If you're up for a drive by we can likely remove those and just keep it at the top level.

addaleax: Let's maybe not? That would prevent people from using mongosh through `npx` without additional flags – we do show a deprecation warning to the user in that case, that should be good enough

addaleax (approved):

PR #2731, opened 2026-05-21
chore(cli-repl): move FIPS enabling into separate file

addaleax:
This ensures that even when mongosh is run as ESM instead of (as it currently is) being transpiled naïvely to CJS, enabling FIPS still happens as the first action when mongosh-owned code takes control of the process; in particular, it runs this code before other modules are loaded.

### DIFF
--- a/packages/cli-repl/src/enable-fips.ts
+++ b/packages/cli-repl/src/enable-fips.ts
@@ -1,0 +1,16 @@
+export let fipsError: Error | undefined;
+export function enableFipsIfRequested() {
+  if (process.argv.includes('--tlsFIPSMode')) {
+    // FIPS mode should be enabled before we run any other code, including any dependencies.
+    // We still wrap this into a function so we can also call it immediately after
+    // entering the snapshot main function.
+    // Similarly, this file is imported first at the very startup of mongosh, so that
+    // it runs even before any other imported files have a chance to do so.
+    try {
+      require('crypto').setFips(1);
+    } catch (err: any) {
+      fipsError ??= err;
+    }
+  }
+}
+enableFipsIfRequested();
--- a/packages/cli-repl/src/run.ts
+++ b/packages/cli-repl/src/run.ts
@@ -1,17 +1,9 @@
-let fipsError: Error | undefined;
-function enableFipsIfRequested() {
-  if (process.argv.includes('--tlsFIPSMode')) {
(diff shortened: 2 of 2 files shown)


packages/cli-repl/src/enable-fips.ts
```diff
@@ -0,0 +1,16 @@
...
+  if (process.argv.includes('--tlsFIPSMode')) {
+    // FIPS mode should be enabled before we run any other code, including any dependencies.
+    // We still wrap this into a function so we can also call it immediately after
+    // entering the snapshot main function.
+    // Similarly, this file is imported first at the very startup of mongosh, so that
+    // it runs even before any other imported files have a chance to do so.
+    try {
+      require('crypto').setFips(1);
+    } catch (err: any) {
+      fipsError ??= err;
```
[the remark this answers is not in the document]
addaleax: This is fine for now, it's just guarding against cases where `require('crypto')` itself fails

Anemy (approved):

packages/cli-repl/src/enable-fips.ts
```diff
@@ -0,0 +1,16 @@
+export let fipsError: Error | undefined;
+export function enableFipsIfRequested() {
+  if (process.argv.includes('--tlsFIPSMode')) {
+    // FIPS mode should be enabled before we run any other code, including any dependencies.
+    // We still wrap this into a function so we can also call it immediately after
```
lerouxb, line 5: > We still wrap this into a function so we can also call it immediately after entering the snapshot main function.

I'm not following this. What do you mean call it immediately after entering the snapshot main function? I can't see it imported or called anywhere. How does that work?

gagik (approved):

packages/cli-repl/src/enable-fips.ts
addaleax: It refers to this call site: https://github.com/mongodb-js/mongosh/blob/3740b7757f9d6d6e8be8207bf0fdc53aa4104632/packages/cli-repl/src/run.ts#L56

The idea being that it runs immediately once mongosh-owned code takes control of a, either when run directly without startup snapshot support, or if built with snapshot support, both when first building and when actually executing the snapshot

lerouxb: Aah I see now. I was confused by looking at the diff. Should have looked at the resulting file in full.

PR #2738, opened 2026-06-15
fix(browser-repl): show error violations COMPASS-10637

mabaasit:
We are not showing violations currently on `browser-repl` if server reports them. 

image

### DIFF
--- a/packages/browser-repl/src/components/types/error-output.tsx
+++ b/packages/browser-repl/src/components/types/error-output.tsx
@@ -101,6 +101,18 @@ export class ErrorOutput extends Component<ErrorOutputProps> {
     return undefined;
   }
 
+  formatErrorWriteErrors(): JSX.Element | undefined {
+    if (this.props.value.writeErrors) {
+      return (
+        <div>
+          Write Errors:
+          <SimpleTypeOutput value={this.props.value.writeErrors} />
+        </div>
+      );
+    }
+    return undefined;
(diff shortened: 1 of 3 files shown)


nbbeeken (approved):
LGTM

packages/browser-repl/src/components/types/error-output.tsx
```diff
@@ -109,6 +121,7 @@ export class ErrorOutput extends Component<ErrorOutputProps> {
           {this.formatErrorBugReportInfo()}
           {this.formatErrorInfo()}
           {this.formatErrorResult()}
+          {this.formatErrorViolations()}
```
addaleax, line 154: Can we sync this with the CLI just in general, instead of piecemeal additions?

https://github.com/mongodb-js/mongosh/blob/31f3e8ab5dad41e600c10957dd41900475538701/packages/cli-repl/src/format-output.ts#L382

`.cause` and `.writeErrors` in particular would be nice to have as well

addaleax (approved):

packages/browser-repl/src/components/types/error-output.tsx
mabaasit: Added in [16445cb](https://github.com/mongodb-js/mongosh/pull/2738/commits/16445cb13cf25c56c645fc3db1e1134e5e2002e7)

PR #2747, opened 2026-06-24
fix(browser-repl): show dates as ISODate MONGOSH-1781

Anemy:
MONGOSH-1781

| before | after |
| - | - |
| Screenshot 2026-06-24 at 12 38 48 | Screenshot 2026-06-24 at 12 37 30 |

### DIFF
--- a/packages/browser-repl/src/components/utils/inspect.ts
+++ b/packages/browser-repl/src/components/utils/inspect.ts
@@ -79,7 +79,10 @@ function attachInspectMethods(obj: any): void {
     tryAddInspect(obj, bsonStringifiers[bsontype]);
   } else if (isDate(obj)) {
     tryAddInspect(obj, function (this: Date): string {
-      return this.toISOString();
+      if (isNaN(this.valueOf())) {
+        return 'Invalid Date';
+      }
+      return `ISODate('${this.toISOString()}')`;
     });
   }
 }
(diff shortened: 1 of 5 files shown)


packages/browser-repl/src/components/utils/inspect.ts
```diff
@@ -79,7 +79,10 @@ function attachInspectMethods(obj: any): void {
     tryAddInspect(obj, bsonStringifiers[bsontype]);
   } else if (isDate(obj)) {
     tryAddInspect(obj, function (this: Date): string {
-      return this.toISOString();
+      if (isNaN(this.valueOf())) {
+        return 'Invalid Date';
+      }
+      return `ISODate('${this.toISOString()}')`;
```
Anemy, line 85: We want to do a bit more here actually. `ISODate('2019-07-11T02:26:45.758Z')` in mongosh cli-repl responds with ISODate. With these changes we're still showing `2019-07-11T02:26:45.758Z` as the response for a basic `new Date()` and it should be ISODate.

addaleax (approved):

nbbeeken (approved):

packages/browser-repl/src/components/utils/inspect.ts
Anemy: Done in https://github.com/mongodb-js/mongosh/pull/2747/commits/a3e7f532029fa863d99fc20a28b2605f841ad588 (the comment was improved in the following commit)

PR #2751, opened 2026-06-30
feat(logging): new telemetry design MONGOSH-3408, MONGOSH-3376, MONGOSH-3406

alenakhineika:
- Replace the Segment analytics integration with a self-hosted fire-and-forget HTTP transport (an intermediate step before integrating with CloudFront). Only `TelemetryClient` knows about the endpoint URL and HTTP format.
- The previous design sent one analytics event per user action (14 event types); this PR reduces that to three events per session: `Identify` (device and OS traits), `New Connection` (connection metadata including topology type, SRV, CSFLE), and `Session Ended` (aggregated session summary emitted once at close).
- Session data is accumulated incrementally in `SessionTelemetryState` as the user works, command counts split by `REPL` vs `rc/eval` context, error count, snippet count, startup timings, sequence of commands capped at 100, and flushed into a single end-of-session event. The event is only emitted for interactive sessions or identified agentic sessions. Containerized non-interactive non-agentic runs are excluded.
- The internal `get-os-info.ts` implementation is replaced with the shared `@mongodb-js/get-os-info` package.
- Use `device_id` as the sole telemetry user identifier. Remove `userId` and `telemetryAnonymousId` (config fields, validators, payload, and the `mongosh:new-user`/`mongosh:update-user` bus events).
- Key `ThrottledAnalytics` cross-session throttle state on `device_id` instead of `anonymousId`.
- Add `telemetryEndpoint` as a `CliUserConfig` option (with URL validator) carrying the production default. Allow resolving the endpoint as override via `MONGOSH_TELEMETRY_ENDPOINT` env variable.
- When no endpoint is configured, the telemetry is not sent over the network, but still can be enabled and events are still written to the local log.
- Remove the `IS_MONGOSH_EVERGREEN_CI` guard and the `analyticsOptions`/`alwaysEnable` constructor option (endpoint overrides now via env or config, consistent with `updateURL`).
- Read user/global config before setting up analytics/telemetry, so the configured telemetry endpoint is honored (safe now that the `new-user` flow is gone). Defer setup errors until logging is in place.
- Extract analytics wiring into `packages/cli-repl/src/setup-analytics.ts` (`setupTelemetryAnalytics()` + `AnalyticsOptions`), leaving `setupAnalytics()` a thin wrapper.
- Split the telemetry/logging tests out of `cli-repl.spec.ts` into new `cli-repl-telemetry.spec.ts` (fake-server integration), `cli-repl-logging.spec.ts` (logging config), and `setup-analytics.spec.ts` (pure-function unit tests); kept log-based unit tests in `cli-repl.spec.ts`.
- Close each `CliRepl` instance in a top-level `afterEach` so per-test resources (driver connections, log writers, process listeners) don't accumulate - the full cli-repl suite now runs within the default heap instead of OOMing.
- Update `logging`, `cli-repl`, `types`, and `e2e` specs and the cli-repl README for the new telemetry model.


packages/logging/src/telemetry-client.ts
```diff
@@ -0,0 +1,33 @@
...
+      method: 'POST',
+      headers: { 'Content-Type': 'application/json' },
+      body: JSON.stringify(event),
+    }).catch(() => {
+      // fire-and-forget: network errors are intentionally ignored
+    });
+  }
+
+  async flush(): Promise<void> {
+    // fire-and-forget: no in-flight tracking, nothing to await
```
addaleax: Adding some sort of timeout is probably still desirable, because otherwise there are events that will be consistently under-reported

packages/cli-repl/src/cli-repl.ts
```diff
@@ -44,7 +43,8 @@ import {
 } from '@mongosh/types';
 import { promises as fs } from 'fs';
 import path from 'path';
-import { getOsInfo, type OsInfo } from './get-os-info';
+import { getOsInfo } from '@mongodb-js/get-os-info';
+type OsInfo = Awaited<ReturnType<typeof getOsInfo>>;
```
nbbeeken: Should we get this type exported from the package? 

If not probably should appear below all the imports, but that's just organizational.

packages/logging/src/helpers.spec.ts
```diff
@@ -23,4 +23,55 @@ describe('logging helpers', function () {
       });
     }
   });
+
+  describe('getAiAgent', function () {
+    const cleanup = () => {
+      for (const v of Object.keys(KNOWN_AGENT_ENV_VARS)) {
```
nbbeeken, line 32: A minor thing, we don't ever expect these env vars when running our tests but they could be there, its worth taking extra care to restore the env to exactly as it was, avoids future surprises.

nbbeeken (changes requested):
small things, looks really good!

packages/cli-repl/src/cli-repl.ts
alenakhineika: Yeah, this is not exported, created [MONGOSH-3489](https://jira.mongodb.org/browse/MONGOSH-3489) to make it right. For now, moved below 👍

packages/logging/src/helpers.spec.ts
alenakhineika: Good idea! Updated here and in another test suite.

nbbeeken (approved):

packages/cli-repl/src/cli-repl.ts
```diff
@@ -61,21 +60,20 @@ import {
 import { fullDepthInspectOptions } from './format-output';
 import { getDeviceIdForMongosh } from './device-id';
 
+// TODO: OsInfo is not exported from @mongodb-js/get-os-info. Derive it until that is fixed MONGOSH-3489.
+// https://github.com/mongodb-js/devtools-shared/blob/main/packages/get-os-info/src/get-os-info.ts#L15C6-L15C12
+type OsInfo = Awaited<ReturnType<typeof getOsInfo>>;
```
nbbeeken: 0.7.0 is released now!

nbbeeken (approved):

packages/cli-repl/src/cli-repl-telemetry.spec.ts
```diff
@@ -0,0 +1,572 @@
...
+} from '../test/repl-helpers';
+import type { CliReplOptions } from './cli-repl';
+import { CliRepl } from './cli-repl';
+import { KNOWN_AGENT_ENV_VARS } from '@mongosh/logging';
+import type { DevtoolsConnectOptions } from '@mongosh/service-provider-node-driver';
+import type { AddressInfo } from 'net';
+import sinon from 'sinon';
+import type { MongoLogWriter } from 'mongodb-log-writer';
+
+const delay = promisify(setTimeout);
```
nbbeeken: minor but `timers.promises.setTimeout` exists so you don't need to make a promise sleep func

packages/logging/src/telemetry-client.ts
```diff
@@ -0,0 +1,58 @@
...
+    this.endpoint = endpoint;
+    this.fetch = fetch;
+    this.flushTimeoutMs = flushTimeoutMs;
+  }
+
+  track(event: TelemetryEvent): void {
+    const p = this.fetch(this.endpoint, {
+      method: 'POST',
+      headers: { 'Content-Type': 'application/json' },
+      body: JSON.stringify(event),
```
nbbeeken: Should we set a bound on the amount of time this should run for? If the request is held by server indefinitely there's nothing to clear up resources on our end, we could throw an `AbortSignal.timeout` with some ceiling here.

packages/logging/src/telemetry-client.ts
```diff
@@ -0,0 +1,58 @@
...
+const FLUSH_TIMEOUT_MS = 2_000;
+
+type FetchFn = (
+  url: string,
+  init?: { method?: string; headers?: Record<string, string>; body?: string }
+) => Promise<unknown>;
+
+/**
+ * Sends telemetry events to the MongoDB telemetry HTTP endpoint.
+ * Network errors are silently dropped. flush() waits up to 2 s for
```
nbbeeken: Where is the "up to 2" enforced?

nbbeeken (changes requested):

packages/logging/src/telemetry-client.ts
```diff
@@ -0,0 +1,58 @@
...
+    flushTimeoutMs: number = FLUSH_TIMEOUT_MS
+  ) {
+    this.endpoint = endpoint;
+    this.fetch = fetch;
+    this.flushTimeoutMs = flushTimeoutMs;
+  }
+
+  track(event: TelemetryEvent): void {
+    const p = this.fetch(this.endpoint, {
+      method: 'POST',
```
nbbeeken: From slack: Why is this POST and not HEAD? (just leaving this here so we have a reference to the discussion)

nbbeeken: It's a placeholder

packages/logging/src/telemetry-client.ts
```diff
@@ -32,18 +44,61 @@ export class TelemetryClient implements MongoshAnalytics {
...
+    )}`;
+
+    // A hung request (e.g. a stalled proxy) would otherwise never settle its
+    // promise, leaking a socket and an entry in inflight forever — the
+    // abort signal bounds every request so it always settles.
+    const controller = new AbortController();
+    const timeoutId = setTimeout(
+      () => controller.abort(),
+      REQUEST_TIMEOUT_MS
+    ).unref?.();
```
nbbeeken: Is there a reason not to use `AbortSignal.timeout`? IIRC it also is an unref-ed timer

packages/logging/src/telemetry-client.ts
```diff
@@ -32,18 +44,61 @@ export class TelemetryClient implements MongoshAnalytics {
...
-      headers: { 'Content-Type': 'application/json' },
-      body: JSON.stringify(event),
+    const payload: Record<string, unknown> = event.payload;
+    const query = new URLSearchParams({
+      deviceId: String(payload.device_id ?? ''),
+      sessionId: String(payload.session_id ?? ''),
+    });
+    const url = `${this.endpoint}${eventPath(event.name)}?${query.toString()}`;
+
+    // TODO: It might be worth using something like zstd
```
nbbeeken: Can we ticket this todo?

packages/logging/src/telemetry-client.ts
```diff
@@ -32,18 +44,61 @@ export class TelemetryClient implements MongoshAnalytics {
...
+    // and/or use a custom dictionary rather than plain gzip.
+    const cookie = `mge=${gzipSync(Buffer.from(JSON.stringify(event))).toString(
```
nbbeeken: Why synchronous zipping? Shouldn't we let the event loop turn here

packages/logging/src/telemetry-client.ts
```diff
@@ -32,18 +44,61 @@ export class TelemetryClient implements MongoshAnalytics {
...
+    // abort signal bounds every request so it always settles.
+    const controller = new AbortController();
+    const timeoutId = setTimeout(
+      () => controller.abort(),
+      REQUEST_TIMEOUT_MS
+    ).unref?.();
+
+    // User-Agent is intentionally not set here: the `fetch` passed in by the
+    // caller (see cli-repl's fetch wrapper) always attaches its own
+    // OS/version-derived User-Agent, which already covers the client identity.
```
nbbeeken: Sure, does this need to be a comment? or could we just have a unit test for this? seems specific to leave a note about

packages/logging/src/telemetry-client.ts
```diff
@@ -32,18 +44,61 @@ export class TelemetryClient implements MongoshAnalytics {
...
+
+    // TODO: It might be worth using something like zstd
+    // and/or use a custom dictionary rather than plain gzip.
+    const cookie = `mge=${gzipSync(Buffer.from(JSON.stringify(event))).toString(
+      'base64'
+    )}`;
+
+    // A hung request (e.g. a stalled proxy) would otherwise never settle its
+    // promise, leaking a socket and an entry in inflight forever — the
+    // abort signal bounds every request so it always settles.
```
nbbeeken: This feels like an overly specific description of something that should just be a unit test, we can verify a signal is always given to fetch in a unit test and the test's title will serve as a description as to why its there

packages/logging/src/telemetry-client.ts
```diff
@@ -1,13 +1,25 @@
+import { gzipSync } from 'zlib';
 import type { TelemetryEvent } from './telemetry-events';
 import type { MongoshAnalytics } from './analytics-helpers';
 
+export const REQUEST_TIMEOUT_MS = 5_000;
```
nbbeeken, line 9: Fine with it remaining this value, but we should say why, even if its picked out of thin air the next person will think there's a reason if we don't state there is none.

nbbeeken (changes requested):

packages/logging/src/telemetry-client.ts
alenakhineika: I actually didn't know about it, so placed this for myself for later reference :) I noticed it when was investigating why CloudFront receives 2 `User-Agent` ([MONGOSH-3503](https://jira.mongodb.org/browse/MONGOSH-3503)).

nbbeeken: Ah cool, yea it's just like not obvious that one would set a user-agent so then this comment about its absence to me was like "yea... wasn't planning on setting one" 😆 

We control the typescript for the headers input too, you could use `{ Cookie: string }` instead of `Record<string, string>` to block someone from adding it

packages/logging/src/telemetry-client.ts
alenakhineika: I have added test as well. I write such comments for myself, because i forget details later, but i will try to keep them short :)

packages/logging/src/telemetry-client.ts
alenakhineika: `{ Cookie: string }` doesn't help, it still displays 2 `user-agent` in a request on the AWS WAF Dashboard. Maybe the bug is there 😄 but we can look deeper into it in the scope of the MONGOSH-3503 work.

nbbeeken: No I meant that you could prevent another who comes along here from mistakenly adding it. "User-Agent is intentionally not set here" it could be a compiler error to set if if the types were more restrictive. 🤷🏻 optional, just could catch an accident

alenakhineika: Ah gotcha! Yes in this case it would be helpful.

packages/logging/src/telemetry-client.ts
alenakhineika: [MONGOSH-3504](https://jira.mongodb.org/browse/MONGOSH-3504)

packages/logging/src/telemetry-client.ts
alenakhineika: Converted to async.

nbbeeken (approved):

nbbeeken (approved):

PR #2753, opened 2026-06-30
chore(npm-packages): regenerate the shrinkwrap during release

nbbeeken:
run shrinkwrap regen after npm install when package bumper has bumped


addaleax (approved):
🤞

PR #2765, opened 2026-07-08
test(shell-api): timeseries namespace change in 9.0 MONGOSH-3451

nbbeeken:
MONGOSH-3451 — linux-mlatest
MONGOSH-3473 — win32-mlatest
MONGOSH-3422 — darwin-mlatest


nbbeeken:
better solution: https://github.com/mongodb-js/mongosh/pull/2767/

PR #2768, opened 2026-07-10
feat: remove support for preview text, suffix, and substring QE FLE APIs MONGOSH-3403

ivandevp:
MONGOSH-3336

The e2e-fle "Queryable Encryption Prefix/Suffix/Substring Support" test uses the `substringPreview`/`prefixPreview`/`suffixPreview` query types and the `TextPreview` algorithm, which were the **public-preview** QE text-search API introduced in 8.2. That API was removed at GA: `9.0+` servers reject these query types at collection creation (`... query type prefixPreview, as it is deprecated`, code `12341600`), so the test fails on the `tests_linux-mlatest_n24` variant once the nightly server rolls to 9.0.

This is not a mongosh regression — the failure was mis-attributed by foliage to an unrelated update-notification commit. The server-side removal is tracked in SERVER-116329 (rename to GA `prefix`/`suffix`/`substring`) and SERVER-123416 (block the deprecated names), both fixed in `9.0.0-rc0`.

Gate the preview block with `skipIfServerVersion(testServer, '>= 9.0')`, scoped to the nested Prefix/Suffix/Substring context so the sibling 8.2+ `$lookup` test still runs on 9.0+. Coverage is preserved on 8.2.x (the preview window). This mirrors the driver-team remedy in DRIVERS-3461, which skips these tests on 9.0+. Test-only change.

Porting the test to the GA API (query types `prefix`/`suffix`/`substring`, 9.0+ crypt_shared) is a follow-up, blocked on a libmongocrypt/driver bump that recognizes the GA names — tracked in MONGOSH-3336.

## Related tickets

This is a single failure fanned out by foliage across both test modes (automatic/explicit) and three platforms (linux/win32/darwin). The `skipIfServerVersion('>= 9.0')` gate sits on the shared context that runs all of them, so this PR resolves all six:

- MONGOSH-3336 — automatic, tests_linux-mlatest_n24
- MONGOSH-3337 — explicit, tests_linux-mlatest_n24
- MONGOSH-3338 — automatic, tests_win32-mlatest_n24
- MONGOSH-3339 — explicit, tests_win32-mlatest_n24
- MONGOSH-3340 — automatic, tests_darwin-mlatest_n24
- MONGOSH-3341 — explicit, tests_darwin-mlatest_n24

## Additional fix on this branch

While verifying on the 9.0-alpha `mlatest` server, the unrelated `FLE tests > 7.0+ > allows explicit enryption with bypassQueryAnalysis` test also failed there (`Location9188701: max contention 4 in find payload ... does not match collection's configured contention 8`). Same class of issue — 9.0 tightened validation the test relied on. Fixed by configuring the collection field's `contention` to match the payloads' `contentionFactor` (4).

### DIFF
--- a/package.json
+++ b/package.json
@@ -170,9 +170,5 @@
     "packages/node-runtime-worker-thread",
     "packages/browser-repl",
     "packages/connectivity-tests"
-  ],
-  "overrides": {
-    "mongodb": "^7.5.0",
-    "mongodb-client-encryption": "^7.2.0"
-  }
+  ]
 }
--- a/package.json
+++ b/package.json
@@ -116,7 +116,7 @@
     "husky": "^9.0.11",
     "knip": "^5.71.0",
     "mocha": "^11.7.5",
-    "mongodb-runner": "^6.5.3",
+    "mongodb-runner": "^6.8.2",
     "nyc": "^15.1.0",
     "pkg-up": "^3.1.0",
     "rimraf": "^3.0.2",
@@ -170,5 +170,9 @@
     "packages/node-runtime-worker-thread",
     "packages/browser-repl",
     "packages/connectivity-tests"
-  ]
+  ],
+  "overrides": {
+    "mongodb": "^7.5.0",
+    "mongodb-client-encryption": "^7.2.0"
+  }
 }
--- a/package.json
+++ b/package.json
@@ -116,7 +116,7 @@
     "husky": "^9.0.11",
     "knip": "^5.71.0",
     "mocha": "^11.7.5",
-    "mongodb-runner": "^6.8.2",
+    "mongodb-runner": "6.5.3",
     "nyc": "^15.1.0",
     "pkg-up": "^3.1.0",
     "rimraf": "^3.0.2",
@@ -170,5 +170,9 @@
     "packages/node-runtime-worker-thread",
     "packages/browser-repl",
     "packages/connectivity-tests"
-  ]
+  ],
+  "overrides": {
+    "mongodb": "^7.5.0",
+    "mongodb-client-encryption": "^7.2.0"
+  }
 }
--- a/packages/shell-api/src/integration.spec.ts
+++ b/packages/shell-api/src/integration.spec.ts
@@ -3050,6 +3050,16 @@ describe('Shell API (integration)', function () {
 
   describe('maxTimeMS support', function () {
     skipIfServerVersion(testServer, '< 4.2');
+    // These tests trigger maxTimeMS by running a `$where: while (true)` query,
+    // which relies on the server interrupting server-side JavaScript promptly.
+    // On 9.0+ server-side JS runs in a new WASM engine whose interruption is
+    // coarse/poll-based, so a maxTimeMS'd `$where` loop is no longer aborted
+    // near-instantly (and two back-to-back loops can saturate the engine). That
+    // is a server-side behavior detail unrelated to what these tests check
+    // (mongosh applying the maxTimeMS config option and explicit option to
+    // operations), so restrict them to servers that still interrupt `$where`
+    // promptly.
+    skipIfServerVersion(testServer, '>= 9.0');
 
     beforeEach(async function () {
       await collection.insertMany([...Array(10).keys()].map((i) => ({ i })));
--- a/packages/shell-api/src/integration.spec.ts
+++ b/packages/shell-api/src/integration.spec.ts
@@ -3050,16 +3050,6 @@ describe('Shell API (integration)', function () {
 
   describe('maxTimeMS support', function () {
     skipIfServerVersion(testServer, '< 4.2');
-    // These tests trigger maxTimeMS by running a `$where: while (true)` query,
-    // which relies on the server interrupting server-side JavaScript promptly.
-    // On 9.0+ server-side JS runs in a new WASM engine whose interruption is
-    // coarse/poll-based, so a maxTimeMS'd `$where` loop is no longer aborted
-    // near-instantly (and two back-to-back loops can saturate the engine). That
-    // is a server-side behavior detail unrelated to what these tests check
-    // (mongosh applying the maxTimeMS config option and explicit option to
-    // operations), so restrict them to servers that still interrupt `$where`
-    // promptly.
-    skipIfServerVersion(testServer, '>= 9.0');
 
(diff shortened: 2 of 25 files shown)


nbbeeken (changes requested):
https://github.com/mongodb/node-mongodb-native/releases/tag/v7.5.0

The driver's released GA support! We should update and get the tests passing again, so instead of skipping on 9 I think we need to duplicate these and verify the preview API continues to work on old versions and the ga API works on 9.0

nbbeeken:
From slack: we're going to drop preview as described in https://jira.mongodb.org/browse/MONGOSH-3403 and its linked tickets

package.json
```diff
@@ -170,5 +170,9 @@
     "packages/node-runtime-worker-thread",
     "packages/browser-repl",
     "packages/connectivity-tests"
-  ]
+  ],
+  "overrides": {
+    "mongodb": "^7.5.0",
+    "mongodb-client-encryption": "^7.2.0"
+  }
```
nbbeeken: We shouldn't have overrides in mongosh, fairly certain they don't play well with shrink-wrap 🤔 But I also don't think we have any incompatibilities in the dep tree that requires the use of override, we just need to bump them:

If you just want to upgrade the driver then 
```sh
env REPLACE_PACKAGE=mongodb:^7.5.0 npm run replace-package
```

might as well hit the supporting packages while we're here:
```sh
env REPLACE_PACKAGE=bson:^7.3.1 npm run replace-package
env REPLACE_PACKAGE=mongodb-client-encryption:^7.2.0 npm run replace-package
env REPLACE_PACKAGE=mongodb-connection-string-url:^7.0.1 npm run replace-package
```

Then just `npm i` to update the lock file

if mongodb-runner needs a fix, we own it:

https://github.com/mongodb-js/devtools-shared/tree/main/packages/mongodb-runner

packages/shell-api/src/integration.spec.ts
```diff
@@ -3050,6 +3050,16 @@ describe('Shell API (integration)', function () {
...
+    // These tests trigger maxTimeMS by running a `$where: while (true)` query,
+    // which relies on the server interrupting server-side JavaScript promptly.
+    // On 9.0+ server-side JS runs in a new WASM engine whose interruption is
+    // coarse/poll-based, so a maxTimeMS'd `$where` loop is no longer aborted
+    // near-instantly (and two back-to-back loops can saturate the engine). That
+    // is a server-side behavior detail unrelated to what these tests check
+    // (mongosh applying the maxTimeMS config option and explicit option to
+    // operations), so restrict them to servers that still interrupt `$where`
+    // promptly.
+    skipIfServerVersion(testServer, '>= 9.0');
```
nbbeeken: I'm not too sure about the truth of this comment but I believe the reason you may be seeing this test fail is because the fix is in a 9.0 release that isn't available on ubuntu 20.04 we need to ensure that we don't use "latest" on ubuntu 20 anymore, we have to pin that OS to 8.x and ensure "latest" is tested on 24+

Is that what we're doing and this is still observed?

package.json
ivandevp: Thanks @nbbeeken . I'll check this after confirming tests work

packages/shell-api/src/integration.spec.ts
ivandevp: with changes included in this PR, mlatest is running in ubuntu 22.04 - the issue happening for this in particular is that $where test is using a JS loop and the test times out before the op is actually killed. This does not happen in 8.0 because it looks like it kills immediately the op while 9.0 uses a poll-based system now. I'm trying to get tests stable first and then will double check your comments again before marking this PR as ready for review.

ivandevp:
Alright, so this was the best CI state I could get with changes related to bumping the driver, supporting 9.0 and bumping ubuntu to 2204 for mlatest tests.
https://spruce.corp.mongodb.com/version/6a5689830eaedb0007258455/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

Remaining failing tasks which seem unrelated to the changes in this PR are:

- [mongosh_e2e_tests_windows_vsCurrent_small_e2e_tests_win32_patch_879833575f366e722102298e72b88213a144b6ff_6a5689830eaedb0007258455_26_07_14_19_10_00](https://spruce.corp.mongodb.com/task/mongosh_e2e_tests_windows_vsCurrent_small_e2e_tests_win32_patch_879833575f366e722102298e72b88213a144b6ff_6a5689830eaedb0007258455_26_07_14_19_10_00/tests?execution=0&sorts=STATUS%3AASC)
- [mongosh_linux_other_test_vscode_patch_879833575f366e722102298e72b88213a144b6ff_6a5689830eaedb0007258455_26_07_14_19_10_00](https://spruce.corp.mongodb.com/task/mongosh_linux_other_test_vscode_patch_879833575f366e722102298e72b88213a144b6ff_6a5689830eaedb0007258455_26_07_14_19_10_00/logs?execution=0)

Given that said, I'd focus now on trying to get the same state by implementing the suggestions left in previous comments.

packages/shell-api/src/integration.spec.ts
leemaguire: @ivandevp what you observed was a real server bug, not the intended 9.0 design: in earlier alphas, a maxTimeMS expiry was never delivered into JS that was actively running, so a `$where: while(true)` ran until the 60s JS-function timeout and the test gave up first. That was SERVER-130767, fixed in alphas cut on/after July 10 (e.g. 9.0.0-alpha0-224742c7). So @nbbeeken's instinct is right: if the mlatest download predates that, you'll still see the old behavior regardless of OS.

On a current alpha the behavior is: kills are delivered by a ~1s poll (scriptingEngineInterruptIntervalMS, default 1000ms), so the op fails with MaxTimeMSExpired within roughly maxTimeMS + 1s. Slightly coarser than 8.0's near-instant interrupt, but should be fine for these tests, I'd re-run them on a fresh alpha before keeping the >= 9.0 skip. If an assertion truly needs sub-second kill latency, lower scriptingEngineInterruptIntervalMS on the test server instead.

ivandevp: Thanks @leemaguire and @nbbeeken. Looks like it succeeded after running in a fresh alpha 🎉

nbbeeken (approved):

PR #2770, opened 2026-07-13
fix(cli-repl): correctly return false from enableTelemetry setting when forceDisableTelemetry is set MONGOSH-1141

nbbeeken:
When `forceDisableTelemetry` is set don't consult the user preferences just return false. Fixes misleading state.

### DIFF
--- a/packages/cli-repl/src/cli-repl.spec.ts
+++ b/packages/cli-repl/src/cli-repl.spec.ts
@@ -1852,6 +1852,11 @@ describe('CliRepl', function () {
             "Cannot modify telemetry settings while 'forceDisableTelemetry' is set to true"
           );
 
+          output = '';
+          input.write('config.get("enableTelemetry")\n');
+          await waitEval(cliRepl.bus);
+          expect(output).to.include('false');
+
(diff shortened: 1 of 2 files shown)


alenakhineika:
Can we leverage your PR to clean up this? :) https://github.com/mongodb-js/mongosh/blob/main/packages/cli-repl/src/cli-repl.ts#L221-L224 or crate a follow-up PR/ticket to not forget again.

Would be great to introduce a single `isTelemetryEnabled()` helper that checks `forceDisableTelemetry` and `enableTelemetry` config value, and have all callers defer to it. We have these checks in multiple places already, e.g. the `fetch` closure, `setTelemetryEnabled,` `enableTelemetry`, maybe more. Combining both would give us the single source of truth.

alenakhineika (approved):

nbbeeken:
Yea that's fair, I'll investigate what that would take

nbbeeken:
Because of async-ness I had to change some functions return type, what do we think? @alenakhineika

alenakhineika (approved):
This is great 🙌 Thx!

PR #2774, opened 2026-07-20
chore(ci): fix windows e2e with server pin, upgrade OS for nightlies and add skips, skip vscode tests, fix evergreen warnings

nbbeeken:
- fix all the evergreen warnings: removed unused tasks, and ensured all variants have unique names
- Skipped vscode testing: MONGOSH-3496, VSCODE-788
- added a script that will test cached browsers on the host and if they are corrupted it deletes them MONGOSH-3502
- upgrade linux for node nightlies to ubuntu 26 and migrated them to arm (only hardware available for that linux version)
- Node.js nightly has reimplemented repl completer to be an async function, using util.promisfy emits a warning when you use it on an async function so basically wrote our own for this one case. We'll be able to clean it up when we upgrade.
- A number of nightly skips that have been ticketed: MONGOSH-3498 and put in the node 26 epic. This will help us estimate the size of that effort
- Replaces rimraf with the native nodejs rm method, it has built in retries for windows EBUSY errors
- Windows server 2019 (or "pre-2022") was fixed by pinning to server 8.0, 8.3 dropped support for vs2019

### DIFF
--- a/packages/cli-repl/src/node-repl-fix-completer-side-effects.ts
+++ b/packages/cli-repl/src/node-repl-fix-completer-side-effects.ts
@@ -4,7 +4,7 @@ import { PassThrough } from 'stream';
 function isThenable<T>(value: unknown): value is PromiseLike<T> {
   return !!(
     value &&
-    typeof value === 'object' &&
+    (typeof value === 'object' || typeof value === 'function') &&
     'then' in value &&
     typeof value.then === 'function'
   );
--- a/packages/cli-repl/src/node-repl-fix-completer-side-effects.ts
+++ b/packages/cli-repl/src/node-repl-fix-completer-side-effects.ts
@@ -1,6 +1,31 @@
 import type { CompleterResult } from 'readline';
 import { PassThrough } from 'stream';
-import { promisify } from 'util';
+
+function isThenable<T>(value: unknown): value is PromiseLike<T> {
+  return !!(
+    value &&
+    typeof value === 'object' &&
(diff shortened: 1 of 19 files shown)


packages/cli-repl/src/node-repl-fix-completer-side-effects.ts
```diff
@@ -1,6 +1,31 @@
...
+function isThenable<T>(value: unknown): value is PromiseLike<T> {
+  return !!(
+    value &&
+    typeof value === 'object' &&
+    'then' in value &&
+    typeof value.then === 'function'
+  );
+}
+
+export function promisifyCompleter(
```
nbbeeken, line 13: `util.promisify` now prints a warning because this method is async as of https://github.com/nodejs/node/pull/64034

packages/e2e-tests/test/e2e-direct.spec.ts
```diff
@@ -42,7 +42,10 @@ describe('e2e direct connection', function () {
       });
     });
 
-    context('after rs.initiate()', function () {
+    // TODO(MONGOSH-3498): after() cleanup uses exit which hangs on Node nightly
+    (process.version.includes('-nightly')
```
alenakhineika: We do such nightly skips quite a number of times, maybe we can introduce a hook, something like `skipOnNightly`. We have similar hooks in `packages/testing/src/integration-testing-hooks.ts`

alenakhineika (approved):

PR #2779, opened 2026-07-21
feat(ci): build shared-OpenSSL binaries for ppc64le and s390x MONGOSH-1681

ivandevp:
Adds build_linux_ppc64le/s390x_openssl11/openssl3 build variants and matching e2e test tasks, mirroring the existing x64/arm64 setup, so shared-OpenSSL mongosh binaries exist on these architectures too.

Also fixes two issues surfaced only when compiling on the ppc64le host: `LD_LIBRARY_PATH` leaking from the shared-OpenSSL build into the FLE addon's git-clone subprocess (breaking git's Kerberos auth), and OpenSSL 3.0.5's `make install` failing to build HTML docs due to a missing Perl module (fixed by using `install_sw` instead, which only installs the libs/headers we actually need).

Verified with Evergreen patches covering the new variants individually, then the full build matrix (all existing + new variants) to confirm no regressions: https://spruce.corp.mongodb.com/version/6a5f83268ce8f70007bf1346

### DIFF
--- a/.evergreen.yml
+++ b/.evergreen.yml
@@ -6355,6 +6355,42 @@ tasks:
           node_js_version: ${node_js_version}
           mongosh_server_test_version: ${mongosh_server_test_version}
           mongosh_test_e2e_force_fips: ${mongosh_test_e2e_force_fips}
+  - name: e2e_tests_linux_ppc64le_openssl11
+    tags: ["e2e-test", "assigned_to_jira_team_mongosh_mongosh"]
+    depends_on:
+      - name: compile_artifact
+        variant: build_linux_ppc64le_openssl11
+    commands:
+      - func: checkout
+      - func: install
+        vars:
+          node_js_version: ${node_js_version}
+      - func: download_compiled_artifact
+        vars:
+          executable_os_id: linux-ppc64le-openssl11
+      - func: run_e2e_tests
+        vars:
+          node_js_version: ${node_js_version}
+          mongosh_server_test_version: ${mongosh_server_test_version}
+          mongosh_test_e2e_force_fips: ${mongosh_test_e2e_force_fips}
+  - name: e2e_tests_linux_ppc64le_openssl3
+    tags: ["e2e-test", "assigned_to_jira_team_mongosh_mongosh"]
+    depends_on:
+      - name: compile_artifact
+        variant: build_linux_ppc64le_openssl3
+    commands:
+      - func: checkout
+      - func: install
+        vars:
+          node_js_version: ${node_js_version}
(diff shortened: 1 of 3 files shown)


packages/build/src/compile/signable-compiler.ts
```diff
@@ -15,6 +15,13 @@ async function preCompileHook(nodeSourceTree: string) {
     ),
     'package.json'
   )).version;
+  // LD_LIBRARY_PATH may point at a custom-built shared OpenSSL used for the
+  // Node.js compile step (see .evergreen/compile-artifact.sh); this script
+  // only clones a git repo and builds the FLE addon, and git subprocesses
+  // (git-remote-https) can fail to resolve symbols against that OpenSSL on
+  // some platforms, so we don't want it inherited here.
+  const envWithoutLdLibraryPath = { ...process.env };
+  delete envWithoutLdLibraryPath.LD_LIBRARY_PATH;
```
ivandevp, line 24: These happened here: https://spruce.corp.mongodb.com/version/6a5e28563d36740007859606/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

.evergreen/compile-artifact.sh
```diff
@@ -59,7 +59,10 @@ elif [ -n "$MONGOSH_SHARED_OPENSSL" ]; then
   pushd openssl-*
   ./config --prefix=/tmp/m/opt --libdir=lib shared
   make -j12
-  make -j12 install install_ssldirs
+  # install_sw (not install) skips building/installing HTML man pages via
+  # mkpod2html.pl, which can fail if the host's Perl lacks required modules;
+  # we only need the libs/headers here.
+  make -j12 install_sw install_ssldirs
```
ivandevp, line 65: This happened here: https://spruce.corp.mongodb.com/task/mongosh_build_linux_ppc64le_openssl3_compile_artifact_patch_d14626f8f0b94b46b2b8dfab1bdcff1a317059dd_6a5e6745382dbc00077ac266_26_07_20_18_22_00/logs?execution=1

paula-stacho (approved):
LGTM

.evergreen.yml
addaleax, line 1: @ivandevp This file is automatically generated, so any changes to are also automatically undone unless they are accompanied by changes to the original source files. This PR also doesn't actually take care of publishing the generated packages, which would be a core part of the ticket in question.

I think for both of these you'll need to modify `config/release-package-matrix.js` and probably at least tests for the `build` package. I'll re-open the JIRA ticket

PR #2783, opened 2026-07-22
chore(ci): refresh MongoDB version support matrix MONGOSH-3283

ivandevp:
MONGOSH-3283

Refreshes the MongoDB server version support matrix used for testing:

- Add **9.0** (`9.0.0-rc0`) community + enterprise to the matrix, mirroring how 8.3 was covered during its pre-GA window. Since 9.0 does not publish an Ubuntu 20.04 build, its linux variants run on `ubuntu2204-small` — the same handling already used for the `latest` variant.
- Promote **8.3** from `8.3.0-rc3` to the GA `8.3.x` series, and reconcile the e2e spec (was pinned to `8.3.0-rc5`) to `8.3.x`.
- Drop **4.2** — it is >3 years past its 2023-04 EOL per the platform support policy. Removed the now-dead 4.2-specific `continue` guards in the unit-test build-variant generator.
- Regenerated `.evergreen.yml`.

Validated with a [full unit-test](https://spruce.corp.mongodb.com/version/6a6120c96cbcbc00075721bd/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) patch across the new/changed variants — all green after pinning 9.0 linux to Ubuntu 22.04.

Follow-up (not in this PR): bump `90xc`/`90xe` from `9.0.0-rc0` to `9.0.x` once 9.0 GAs.


nbbeeken (approved):
Changes lgtm!! ✅  let's wait on the CI to finish all things and restart any setup failures if you see them, even if they're unrelated-ly flaky (which we should fix) we'll want to make sure we get a green run on the server upgrades.

ivandevp:
CI is still green 🚀

PR #2819, opened 2026-08-17
feat(shell-api): allow empty pipeline in stream processor creation functions STREAMS-2592

mdb-dgil:
for materialized views, an empty pipeline is a valid use case- it means materializing a source onto a sink without any transformations. we should allow this use case and let the server do the validation


nbbeeken (approved):
lgtm with green ci, I'm subscribed to the patch, I'll let you know if there's anything that's just a flake!

PR #2824, opened 2026-08-21
chore: snyk MONGOSH-3592

alenakhineika:
Add an overrides for `browserslist`.


nbbeeken (changes requested):
Seems like `npm upgrade browserslist` does the trick then we don't keep around an override, right?

nbbeeken (approved):

PR #2828, opened 2026-08-31
feat(browser-repl): update background color on dark mode

Anemy:
Match the rest of editors in Compass.

| before | after |
| - | - |
| Screenshot 2026-08-31 at 12 49 39 | Screenshot 2026-08-31 at 12 49 32 |

We updated the background styles in Compass, so with the current mongosh we're seeing some issues without this:
image

### DIFF
--- a/packages/browser-repl/src/components/shell.tsx
+++ b/packages/browser-repl/src/components/shell.tsx
@@ -63,7 +63,7 @@ const shellContainerLightModeStyles = css({
 });
 
 const shellContainerDarkModeStyles = css({
-  backgroundColor: palette.gray.dark4,
+  backgroundColor: palette.black,
   color: palette.gray.light3,
 });
 


mabaasit (approved):

packages/browser-repl/src/components/shell.tsx
```diff
@@ -63,7 +63,7 @@ const shellContainerLightModeStyles = css({
 });
 
 const shellContainerDarkModeStyles = css({
-  backgroundColor: palette.gray.dark4,
+  backgroundColor: palette.black,
```
gribnoysup, line 66: Maybe worth using the same `codePalette.light / codePalette.dark` we use in the editor itself just for consistency?

gribnoysup: Ah, it's already merged, nevermind

PR #2831, opened 2026-09-02
build: CI fixes

alenakhineika:
History of experiments to see whether CI goes green.

- Shrinkwrap generation (split to https://github.com/mongodb-js/mongosh/pull/2836)
- Avoid loading the deprecated Node domain module (split to https://github.com/mongodb-js/mongosh/pull/2837)
- Missing 7zip package on `windows-2022-latest` (split to https://github.com/mongodb-js/mongosh/pull/2838)
- Flaky GC test  (split to https://github.com/mongodb-js/mongosh/pull/2835)
- Connectivity tests fail on KDC (split to https://github.com/mongodb-js/mongosh/pull/2839)

PR #2835, opened 2026-09-06
test(cli-repl): fix flaky CliRepl GC test MONGOSH-3505

alenakhineika:
- Re-take the heap snapshot a second time before failing, so the timing artifact doesn't fail the run. A genuine leak (a strong reference) stays visible on every attempt.
- Log the retainer path when the object is still present, so CI failures are actually debuggable.
- Add `the retry loop detects a genuine leak`, which plants a deliberate strong reference and asserts the object is still reported. Without it, the retry could mask real leaks indefinitely.

CI on this PR will still show unrelated failures. See #2831 that combines all fixes to demonstrate that CI actually reaches green.

### DIFF
--- a/packages/cli-repl/src/cli-repl-gc.spec.ts
+++ b/packages/cli-repl/src/cli-repl-gc.spec.ts
@@ -61,11 +61,11 @@ describe('CliRepl GC', function () {
 
   function* listNodesInHeapSnapshot(
     snapshotString: string
-  ): Iterable<{ type: string; name: string }> {
+  ): Iterable<{ type: string; name: string; index: number }> {
     const { snapshot, nodes, strings } = JSON.parse(snapshotString);
     const { node_fields, node_types } = snapshot.meta;
     for (let i = 0; i < nodes.length; i += node_fields.length) {
-      const description: any = {};
+      const description: any = { index: i / node_fields.length };
       for (let j = 0; j < node_fields.length; j++) {
         const type = node_types[j];
         if (Array.isArray(type)) {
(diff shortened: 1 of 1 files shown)


packages/cli-repl/src/cli-repl-gc.spec.ts
```diff
@@ -80,12 +80,88 @@ describe('CliRepl GC', function () {
     }
   }
 
+  function retainerPath(snapshotString: string, targetIndex: number): string {
```
kmruiz: I'm not sure we need all of this just for reporting. At the end, we are testing Node.js GC and we can use their heap profiler, GC traces and so on if we detect that something weird is going on here.

alenakhineika: Fair! Dropped it 👍

packages/cli-repl/src/cli-repl-gc.spec.ts
```diff
@@ -107,11 +107,40 @@ describe('CliRepl GC', function () {
...
+    objHolder.obj = null;
+
+    let leaked: { index: number }[] = [];
+    for (let attempt = 0; attempt < 2; attempt++) {
+      await new Promise(setImmediate);
+      leaked = [...taggedObjectsInHeap(await takeHeapSnapshot())];
+    }
+    expect(leaked).to.have.lengthOf(1);
+    expect(leakSink).to.have.lengthOf(1); // keep leakSink alive until here
+  });
```
addaleax, line 141: I've been trying to understand what this actually tests, but I don't really get it. What is `objHolder` expected to do here? Its contents are immediately moved to another object. What kind of failure condition would be indicated if this test fails?

alenakhineika: Good catch! objHolder is left over i copied from the test above. Cleaned up.

kmruiz (approved):

packages/cli-repl/src/cli-repl-gc.spec.ts
addaleax: That doesn't really answer the bigger question about what this test does

PR #2840, opened 2026-09-07
test(cli-repl): remove the redundant GC test MONGOSH-3505

alenakhineika:
Follow-up to https://github.com/mongodb-js/mongosh/pull/2835.

The intent was to make sure the retry loop couldn't swallow a real leak. But a real leak stays visible on every attempt, the loop only breaks when nothing is found, so the assertion fails anyway.


alenakhineika:
Thanks @addaleax for pushing on it!

addaleax (approved):

PR #2841, opened 2026-09-07
test: fix flaky requests password when user is redirecting output MONGOSH-3663

alenakhineika:
Pass a callback to write() so the error is reported there instead of being thrown.

### DIFF
--- a/packages/cli-repl/src/run.spec.ts
+++ b/packages/cli-repl/src/run.spec.ts
@@ -109,11 +109,13 @@ describe('CLI entry point', function () {
       });
       let stdout = '';
       let promptedForPassword = false;
+      // The write can race the child exiting, so handle EPIPE in
+      // the callback rather than letting it be thrown asynchronously.
       proc.stdout?.setEncoding('utf8').on('data', (chunk) => {
         stdout += chunk;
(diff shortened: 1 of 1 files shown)


packages/cli-repl/src/run.spec.ts
```diff
@@ -109,11 +109,13 @@ describe('CLI entry point', function () {
       });
       let stdout = '';
       let promptedForPassword = false;
+      // The write can race the child exiting, so handle EPIPE in
+      // the callback rather than letting it be thrown asynchronously.
```
addaleax: ... but if the child exits while waiting for an answer, that would be a bug, so why are we masking that bug?

I think what's way more likely the issue here is that we're writing multiple newlines if multiple chunks arrive after `Enter password`, and we should only write a single newline

addaleax (changes requested):

packages/cli-repl/src/run.spec.ts
alenakhineika: You're right. I treated it as an unavoidable race and just gave the error somewhere to go. But i just kept the same problem and masked it. Updated.

addaleax (approved):

PR #2845, opened 2026-09-11
chore: update node.js to 24.21.0 and bump shared OpenSSL 3 to 3.0.12 MONGOSH-3676

alenakhineika:
Doing the Node.js bump manually because the automation PR https://github.com/mongodb-js/mongosh/pull/2814 can't succeed on its own, because Node 24.21.0 fails to compile against the OpenSSL 3.0.5 we pin for the shared-openssl builds.

```
ncrypto.cc:1906: error: 'OPENSSL_DH_CHECK_MAX_MODULUS_BITS' was not declared in this scope
ncrypto.cc:3276: error: 'EVP_PKEY_PRIVATE_KEY' was not declared in this scope
make: *** [node] Error 2
```

### DIFF
--- a/.evergreen/compile-artifact.sh
+++ b/.evergreen/compile-artifact.sh
@@ -42,7 +42,10 @@ elif [ -n "$MONGOSH_SHARED_OPENSSL" ]; then
     curl -sSfLO https://github.com/openssl/openssl/releases/download/OpenSSL_1_1_1o/openssl-1.1.1o.tar.gz
     MONGOSH_OPENSSL_LIBNAME=:libcrypto.so.1.1,:libssl.so.1.1
   elif [ "$MONGOSH_SHARED_OPENSSL" == "openssl3" ]; then
-    curl -sSfLO https://www.openssl.org/source/old/3.0/openssl-3.0.5.tar.gz
+    # Node.js 24.21 needs EVP_PKEY_PRIVATE_KEY and
+    # OPENSSL_DH_CHECK_MAX_MODULUS_BITS, which were backported to the 3.0
+    # branch in 3.0.13, so anything older fails to compile deps/ncrypto.
+    curl -sSfLO https://www.openssl.org/source/old/3.0/openssl-3.0.22.tar.gz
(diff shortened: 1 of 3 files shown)


.evergreen/compile-artifact.sh
```diff
@@ -42,7 +42,10 @@ elif [ -n "$MONGOSH_SHARED_OPENSSL" ]; then
     curl -sSfLO https://github.com/openssl/openssl/releases/download/OpenSSL_1_1_1o/openssl-1.1.1o.tar.gz
     MONGOSH_OPENSSL_LIBNAME=:libcrypto.so.1.1,:libssl.so.1.1
   elif [ "$MONGOSH_SHARED_OPENSSL" == "openssl3" ]; then
-    curl -sSfLO https://www.openssl.org/source/old/3.0/openssl-3.0.5.tar.gz
+    # Node.js 24.21 needs EVP_PKEY_PRIVATE_KEY and
+    # OPENSSL_DH_CHECK_MAX_MODULUS_BITS, which were backported to the 3.0
+    # branch in 3.0.13, so anything older fails to compile deps/ncrypto.
+    curl -sSfLO https://www.openssl.org/source/old/3.0/openssl-3.0.22.tar.gz
```
addaleax: Library dependencies in general, including this one specifically, should use the _oldest_ version we can build against, not the newest, since that will limit what our binaries can be used with

alenakhineika: Good point. Also figured out by the current CI failure that 3.0.22 was to high and started requiring `Time::Piece`. Looks like `PENSSL_DH_CHECK_MAX_MODULUS_BITS` was added in [3.0.10](https://github.com/openssl/openssl/blob/openssl-3.0.10/include/openssl/dh.h#L96) and `EVP_PKEY_PRIVATE_KEY` is absent in [3.0.11](https://github.com/openssl/openssl/blob/openssl-3.0.11/include/openssl/evp.h), added in [3.0.12](https://github.com/openssl/openssl/blob/openssl-3.0.12/include/openssl/evp.h#L88). Will try to build against 3.0.12 and see if this works.

dudaschar (approved):

addaleax (approved):

PR #2846, opened 2026-09-14
feat: integrate with prod CloudFront MONGOSH-3406

alenakhineika:
Sets the production telemetry endpoint, and keeps the test suite from sending to it.


kmruiz (approved):

LuciaHarcekova (approved):
The 20-minute timeout looks good for Evergreen/Linux and prevents the Kerberos job from hanging indefinitely. Can we confirm this script is not expected to run directly on macOS/Windows, where GNU timeout may not be available?

alenakhineika:
@LuciaHarcekova Every variant running these tests is Linux, so GNU timeout is always available. Local runs on macOS/Windows are possible in principle, but running the whole connectivity suite locally isn't the usual workflow, and if you do run an individual test, you can add an OS-specific timeout.

PR #2847, opened 2026-09-15
fix: count this session log file towards log retention limits MONGOSH-3680

alenakhineika:
- `cleanupOldLogFiles()` has to run after this session's log file exists: with `logMaxFileCount: 4` and 4 existing logs, cleanup deleted nothing and the new log made 5 every session. The same applied to `logRetentionGB` that we have seen in flaky CI (mongosh kicked `cleanupOldLogFiles()` and `createLogWriter` nearly simultaneously so it was sometimes failing and sometimes succeeding).
- Log file names encoded a timestamp with one-second granularity, and retention deleted oldest-first, making the assertions flaky. Now we space the batches 60s apart so the ordering is deterministic.


addaleax (approved):

packages/cli-repl/src/cli-repl-logging.spec.ts
```diff
@@ -121,6 +122,29 @@ describe('CliRepl logging', function () {
...
+          .stub(MongoLogManager.prototype, 'cleanupOldLogFiles')
+          .callsFake(async () => {
+            const logPath = cliRepl.getLogPath();
+            logFileExistedDuringCleanup =
+              !!logPath &&
+              (await fs.access(logPath).then(
+                () => true,
+                () => false
+              ));
+          });
```
addaleax: This is fine but I don't think we really need to stub the method here, we could just emit an event on the bus (in cli-repl.ts) and then listen to that ... it's a bit odd to stub out a method that we aren't actually trying to modify

addaleax (approved):

packages/cli-repl/src/cli-repl-logging.spec.ts
alenakhineika: Wasn't sure about adding an event just for testing, but with the event the test becomes actually better and covers the user-visible scenario. Switched to a bus event as suggested!

addaleax (approved):

PR #2849, opened 2026-09-16
test(cli-repl): stop retaining REPL instances across the suite MONGOSH-3683

alenakhineika:
`test_cli_repl` aborts on Windows with `ATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory`:

https://spruce.corp.mongodb.com/task/mongosh_tests_win32_m60xe_n24_test_cli_repl_patch_844825cc2f4dcee23a255dfeb9efa0a8ef57de55_6aaaa033a5228300071c587e_26_09_16_13_57_11/html-log?execution=0&origin=task

The suite retains 138 of 274 MongoshNodeRepl objects, each holding a MongoDBAutocompleter with its own TypeScript language service (~3.3 GB against a ~4.1 GB limit).

Reasons: mongosh-repl.spec.ts never called `sinon.restore()`, and `verifyAutocompletion()` closed only the REPL, leaving the MongoClient open.


packages/cli-repl/src/cli-repl.spec.ts
```diff
@@ -2087,7 +2087,12 @@ describe('CliRepl', function () {
         expect(output, output).not.to.include(
           'listCollections requires authentication'
         );
-        await cliRepl.mongoshRepl.close();
+        // close() the whole CliRepl, not just the REPL: these suites connect to
```
kmruiz: The comment is not right, cliRepl.mongoshRepl is who holds the "MongoClient" through the ServiceProvider. Autocomplete is also in the cliRepl.mongoshRepl; cliRepl.close will only close (in addition to the mongoshRepl) logs and telemetry, which should not be an issue. It is fine to close at the root level, I just want the comment to be removed or fixed because it is misleading.

packages/cli-repl/src/mongosh-repl.spec.ts
```diff
@@ -130,7 +130,22 @@ describe('MongoshNodeRepl', function () {
...
+    // Release the Node.js REPL instance and the shell instance state of this
+    // test, so that the autocompleter they reference - which keeps a
+    // TypeScript language service around - can be collected.
+    try {
+      await mongoshRepl?.close();
+    } catch {
+      /* not initialized or already closed */
+    }
+
+    // ts-sinon's stubInterface() and bare sinon.stub() register with sinon's
```
kmruiz: We can keep the sinon.restore for hygiene, but the comment doesn't seem right. I don't see sinon referencing the REPL or depending on it and I'm not aware of any circular dependency across the REPL and it's dependencies. Can you confirm that is correct?

kmruiz (changes requested):

kmruiz (approved):

packages/cli-repl/src/mongosh-repl.spec.ts
```diff
@@ -130,7 +130,16 @@ describe('MongoshNodeRepl', function () {
...
+  afterEach(async function () {
+    // Release the Node.js REPL instance and the shell instance state of this
+    // test, so that the autocompleter they reference - which keeps a
+    // TypeScript language service around - can be collected.
+    try {
+      await mongoshRepl?.close();
+    } catch {
+      /* not initialized or already closed */
+    }
+    sinon.restore();
```
addaleax, line 142: Obviously don't mind, but just for my own understanding, does this call actually do anything? We don't stub out methods or properties of persisted objects

addaleax (approved):

packages/cli-repl/src/mongosh-repl.spec.ts
alenakhineika: The case here is not about restoring original values, it's about memory. Turned out `restore()` has a side effect, it also [empties the sandbox's array](https://github.com/sinonjs/sinon/blob/v19.0.5/lib/sinon/sandbox.js#L215), and without it every stub stays registered for the process, holding the args and this values of every call. Those are `stubInterface()` objects which create a registered stub for every method the REPL touches on them, plus the bare `sinon.stub()`.

alenakhineika: Here, [sandbox.stub()](https://github.com/sinonjs/sinon/blob/v19.0.5/lib/sinon/sandbox.js#L454) calls `commonPostInitSetup`, which [puts the stub into collection array](https://github.com/sinonjs/sinon/blob/v19.0.5/lib/sinon/sandbox.js#L442). The `stubInterface()` methods take the same path.

PR #2851, opened 2026-09-16
test(e2e): poll db.currentOp() instead of racing a fixed sleep MONGOSH-3685

alenakhineika:
e2e has the currentOp timing race:
https://evergreen.mongodb.com/task_log_raw/mongosh_e2e_tests_rhel9_zseries_small_e2e_tests_linux_s390x_patch_844825cc2f4dcee23a255dfeb9efa0a8ef57de55_6aaaa033a5228300071c587e_26_09_16_13_57_11/1?type=T&text=true
```
1) e2e
     currentOp
       with 2 shells
         should return the current operation and clear when it is complete:
   AssertionError: expected '{\n  inprog: [\n    {\n      type: \'…' to include 'testProjection'
    at Context.<anonymous> (test/e2e.spec.ts:2696:34)
```

Widen the window to 5 s and poll with eventually.


packages/e2e-tests/test/e2e.spec.ts
```diff
@@ -2679,25 +2704,35 @@ describe('e2e', function () {
         await helperShell.waitForPrompt();
         await currentOpShell.waitForPrompt();
 
-        // Insert a dummy object so find commands will actually run with the delay.
-        await helperShell.executeLine('db.coll.insertOne({})');
+        // Start from a known state with a single dummy object, so that find
+        // commands run with the delay exactly once.
+        await helperShell.executeLine(`db.${COLLECTION}.drop()`);
```
kmruiz, line 2701: We don't need the drop and insertOne. We only need one document _as least_ in the collection, having more don't affect the test outcome.

packages/e2e-tests/test/e2e.spec.ts
```diff
@@ -2666,8 +2665,34 @@ describe('e2e', function () {
       let helperShell: TestShell;
       let currentOpShell: TestShell;
 
-      const CURRENT_OP_WAIT_TIME = 400;
-      const OPERATION_TIME = CURRENT_OP_WAIT_TIME * 2;
+      // How long the server-side operation stays in progress. This is the
+      // window in which db.currentOp() has to observe it, so it needs to be
+      // comfortably longer than one shell round trip - on the emulated
+      // variants (s390x, ppc64le) a single executeLine() can take a second or more.
+      const OPERATION_TIME = 5000;
```
kmruiz: Wouldn't this make the op run for 5s? That seem too much.

packages/e2e-tests/test/e2e.spec.ts
```diff
@@ -2666,8 +2665,34 @@ describe('e2e', function () {
...
+
+      // The shell that runs the operation only gets its prompt back once the
+      // operation completes, so it has to wait for longer than
+      // waitForPrompt() allows by default. Be generous: the server may
+      // evaluate $where more than once per document, and on MongoDB 9.0 the
+      // default 10s was not enough even for a single document.
+      const SLOW_COMMAND_OPTIONS = { timeout: 60_000 };
```
kmruiz: Independently of how many times $where is executed, it should not affect negatively because the more the merrier (it will just wait more, so more chances to get the currentOp). The timeout likely has to be OPERATION_TIME because we have the eventually doing active polling.

packages/e2e-tests/test/e2e.spec.ts
```diff
@@ -2666,8 +2665,34 @@ describe('e2e', function () {
...
+      const COLLECTION = 'currentOpColl';
+
+      // eventually() adds attempts while the sleeps between them still fit in
+      // `timeout`, so that allows 1500 / 250 = 6 attempts.
+      //
+      // The time each attempt itself takes is not part of that,
+      // so the total time has to fit in OPERATION_TIME.
+      const CURRENT_OP_POLL_OPTIONS = { initialInterval: 250, timeout: 1500 };
+
+      before(function () {
```
kmruiz:
```suggestion
      beforeEach(function () {
```

IIRC, if we do this.timeout in a `before` it applies to the whole suite, and we likely want to apply it to each test independently.

kmruiz (changes requested):

packages/e2e-tests/test/e2e.spec.ts
alenakhineika: It should cover the find and a full db.currentOp() round trip in a second shell, which is what kept failing on s390x. Happy to lower it to e.g. 3000? I was just generous enough to fix the race.

packages/e2e-tests/test/e2e.spec.ts
alenakhineika: It does affect negatively, just not the polling. Every extra $where evaluation makes the find take longer, and the helper shell can't print a prompt until it returns, so await currentCommand awaits for all of them. SLOW_COMMAND_OPTIONS is that prompt wait, not the poll timeout. The polling side is as you said "the more the merrier".

packages/e2e-tests/test/e2e.spec.ts
alenakhineika: Same issue as with $where per document. The assertion only needs one document agreed, but the count controls how long the test blocks.

packages/e2e-tests/test/e2e.spec.ts
kmruiz: Are we sure of this behaviour? It surprises me because let's say that we wait for this line:
```
const currentCommand = helperShell.executeLine(
          `db.${COLLECTION}.find({$where: function() { sleep(${OPERATION_TIME}) }}).projection({testProjection: 1})`,
          SLOW_COMMAND_OPTIONS
        );
```
We will only have an answer from the REPL when the find is done, so we won't be able to tail a currentOp later.  What you say would be true if we do in fact await, like we do later in currentOp:

```
const currentOpCall = await currentOpShell.executeLine(
            `db.currentOp()`
          );
```
as it would wait for the shell to finish evaluatio, but it's not the case, we are running both shells concurrently. That's why we don't await in helperShell but we do await in currentOpShell.

packages/e2e-tests/test/e2e.spec.ts
kmruiz: 3s is fine, if we get the current op earlier we close the connection so it should be aborted. It's just that we don't want to wait for too long if the test is broken.

packages/e2e-tests/test/e2e.spec.ts
kmruiz: Not exactly the same, because it's gated by the timeout you added earlier. So even if we have 1K documents, it would take as long as 5s (3s from your later suggestion).

If we feel strongly about this, for me it's fine to keep the drop and insertOne, it's just that it is not necessary for now.

packages/e2e-tests/test/e2e.spec.ts
alenakhineika: It actually belongs on the context better. Updated.

packages/e2e-tests/test/e2e.spec.ts
alenakhineika: But isn't the moment we call helperShell.executeLine it types the line and enters waitForPrompt, which starts eventually's retry loop with its 10s budget? So that time is ticking while the other shell polls, and if the find outlives it the promise rejects on its own. Or I understand this wrong..

kmruiz: Yes, that's right, if we look at the implementation of executeLine from the TestShell, it looks like this:

```ts
async executeLine(
    line: string,
    opts: TestShellWaitForPromptOptions & TestShellInputOptions = {}
  ): Promise<string> {
    const previousOutputLength = this._output.length;
    this.writeInputLine(line, opts);
    await this.waitForPrompt(previousOutputLength, opts); // <-- this is what you mention
    return this._output.slice(previousOutputLength);
  }
```

However, because we do not await in helperShell.executeLine, it doesn't block the test. So let's say (not totally accurate) that you have 2 threads:

1. A blocked thread in waitForPrompt, that will wait until the find is finished (we have a prompt back). It does not block the test because we do not await executeLine.
2. A running thread with the currentOpShell that we in fact await in the test, and we do the eventually loop.

If you put an await in helperShell.executeLine, you'll see the behaviour I mention: the test will wait until the server evaluates the find($where) and then the test will timeout because either the find($where) takes too long or there will be never a currentOp, because when we reach the eventually loop, the helperShell is already done.

packages/e2e-tests/test/e2e.spec.ts
alenakhineika: I tried to measure it so we have real number to discuss :)
```js
const unlimitedFind = `db.${COLLECTION}.find({$where: function() { sleep(${OPERATION_TIME}) }})`;
const seed = async (n: number) => {
  await helperShell.executeLine(`db.${COLLECTION}.drop()`);
  await helperShell.executeLine(
    `db.${COLLECTION}.insertMany(Array.from({length: ${n}}, () => ({})))`
  );
};

for (const n of [1, 2, 3, 4]) {
  await seed(n);
  const started = Date.now();
  let error: Error | undefined;
  try {
    await helperShell.executeLine(unlimitedFind);
  } catch (err: any) {
    error = err;
  }
  const elapsed = Date.now() - started;
  console.log(`${n} document(s): ${elapsed}ms ${error ? `- ${error.message.split('\n')[0]}` : ''}`);
```
The output:
```
1 document(s): 3154ms
2 document(s): 6109ms
3 document(s): 9044ms
4 document(s): 10159ms - [Timed out 10000ms] expected prompt
```
OPERATION_TIME isn't a ceiling on the operation - $where is evaluated once per scanned document, so the sleep runs per document and the total is n × OPERATION_TIME. That's what the numbers above show: 3s, 6s, 9s for 1, 2, 3 documents, then a failure at 4 once it outgrows the 10s that waitForPrompt allows by default. Nothing caps it.

But `$where: function() { sleep(...) }` was returning `undefined`, so nothing matched and the server scaned the whole collection. Changed it to return `true` and added `.limit(1)` so the scan stops at the first document - with that, 1 document and 50 documents both take ~3s, and the `drop()` is no longer needed, so I removed it.

Maybe i should rename OPERATION_TIME to SLEEP_PER_DOCUMENT so it is less confusing.

packages/e2e-tests/test/e2e.spec.ts
alenakhineika: Removed SLOW_COMMAND_OPTIONS. One note: we don't await executeLine at the call site, but we do await the promise after the polling, then the eventually loop, then await currentCommand. That keeps the concurrency you describe (the poll still runs while the find is in flight) while still observing the result.

kmruiz (approved):

PR #2852, opened 2026-09-17
feat(browser-repl)!: replace enzyme with @testing-library/react and move browser-repl to react 18 COMPASS-11110

dudaschar:
Follow up for React 18 upgrade in Compass.

We are removing enzyme and replacing with RTL (which needed to be updated to support React 18).

browser-repl package is also upgraded to React 18, which caused peer dependency to move up and this will be a breaking change to clients.

https://jira.mongodb.org/browse/COMPASS-11110


packages/browser-repl/src/components/types/error-output.spec.tsx
```diff
@@ -38,21 +39,29 @@ describe('ErrorOutput', function () {
     errorInfo: 'More details about the error',
   });
 
+  // The collapsed view renders the error name as a link that toggles the
+  // Expandable it lives in.
+  async function expand(): Promise<void> {
+    await userEvent.click(screen.getAllByRole('link')[0]);
```
addaleax: Can we give the target element something that uniquely identifies it, e.g. a `data-testid` attribute? Not blocking but would be nice

package.json
```diff
@@ -139,7 +139,9 @@
     "lerna": "^9.0.3"
   },
   "overrides": {
-    "js-yaml@^4.0.0": "^4.3.0"
+    "js-yaml@^4.0.0": "^4.3.0",
+    "react": "^18.3.1",
+    "react-dom": "^18.3.1"
```
addaleax, line 144: What are these overrides solving? We generally try to avoid them, it's fine to use them if they are genuinely necessary, but there should be some kind of context in the PR description/commit message

addaleax (commented):
No such thing as `chore!` in conventional-commits, breaking changes are always `feat!` 🙂

package.json
dudaschar: I was getting 2 react versions (17 and 18) because @mongodb-js/diagramming is pinning react 17 as dependency, and we get it through compass-components.

I didn't find this initially because it is not causing an issue in compass, so updating this other package (probably migrating to devtools shared I guess) was not part of this scope.

dudaschar: I can add a comment with a warning to remove it once react 17 leaves the dependency three, so we won't have the 2 copies being resolved.

packages/browser-repl/src/components/types/error-output.spec.tsx
dudaschar: updated

package.json
addaleax: Okay no, that totally makes sense, we're just going to have to remind ourselves to remove these overrides when the diagramming package is updated 🙂

addaleax (approved):

PR #2853, opened 2026-09-21
fix: telemetry disabled by a lockfile mtime MONGOSH-3687

alenakhineika:
`fs.utimes` takes seconds, but the code refreshing the lockfile timestamp passed milliseconds, which pushed its mtime to the year 2262. A lock dated in the future never counts as stale, so it blocks every later session and telemetry is silently dropped. Fixed by passing Date objects, plus treating a future mtime as stale so already-affected machines recover.


kmruiz (approved):

packages/logging/src/analytics-helpers.ts
```diff
@@ -137,7 +137,9 @@ async function lockfile(
     // created by long running process (longer than staleDuration) we make sure
     // that another process doesn't consider lockfile stale
     intervalId = setInterval(() => {
-      const now = Date.now();
+      // Do not use Date.now() because utimes() reads numbers as seconds,
+      // so milliseconds would date the lockfile to the year 2262.
```
gribnoysup: This comment seems out of place unless you know what code was here before, there's probably a better way to document this

gribnoysup (approved):

lerouxb (approved):

packages/logging/src/analytics-helpers.ts
```diff
@@ -137,7 +137,10 @@ async function lockfile(
     // created by long running process (longer than staleDuration) we make sure
     // that another process doesn't consider lockfile stale
     intervalId = setInterval(() => {
-      const now = Date.now();
+      // Use Dates so that utimes() cannot interpret the time wrong: it reads
```
gribnoysup, line 140: Yup, that's way better, thanks!

